PR #129, opened 2016-08-09
Change format of output so we can associate each file with a target.

peterebden:


csdigi:
Not a listed reviewer but YOLO, LGTM.

dimpavloff:
Obviously this will print a bit more than currently. Would this be annoying when using plz via the CLI?

peterebden:
Only a little. I thought I might suppress the target name when you only build one, but could also control it by a flag. Thoughts?

dimpavloff:
If only a little, then great, +1

Re suppressing the only target, I think that while it makes it friendlier it will be inconsistent compared to when building multiple targets which might be confusing. A flag feels like an overkill

peterebden:
Okay thanks, I will leave it as is for now, we can easily revisit if it's a problem.

PR #130, opened 2016-08-15
More robust system tools

### LINKED ISSUE #9: System level dependencies part 2
Extension to #8 , add ability to identify system-level tools and add them as dependencies so changes in them invalidate other outputs. For example protoc, if someone installs a new version we'd know it was different and regenerate all generated code correctly.

peterebden:
Allow attaching arbitrary system-level files to a rule as tools which get hashed into the rule's inputs. Will mean that we don't need to specify versions of things in .plzconfig any more, changing e.g. the Go version will automatically rebuild all the Go rules.

Go rules and proto rules change a bit to take advantage of this, the others are mostly handled more generically already.

Fixes #9


src/build/incrementality.go
```diff
@@ -336,10 +344,6 @@ func ruleHash(target *core.BuildTarget, runtime bool) []byte {
 	// Obviously we don't include the code pointer because it's a pointer.
 	h.Write(target.PreBuildHash)
 	h.Write(target.PostBuildHash)
-	// The Go version affects the hash for Go targets only.
```
csdigi: NIce to get rid of this special casing

PR #131, opened 2016-08-19
Support for optional outputs

peterebden:
As discussed, they get moved to the output dir, but other rules can't consume them afterwards.

Have allowed them to be glob patterns, which was a bit more work, but I'm pretty sure I'm going to want it in one or two places.

### DIFF
--- a/src/cache/server/cache.go
+++ b/src/cache/server/cache.go
@@ -141,8 +141,8 @@ func (cache *Cache) removeAndDeleteFile(p string, file *cachedFile) {
 // multiple files to be returned.
 func (cache *Cache) RetrieveArtifact(artPath string) (map[string][]byte, error) {
 	ret := map[string][]byte{}
-	if strings.Contains(artPath, "*") {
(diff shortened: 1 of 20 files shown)


src/cache/server/cache.go
```diff
@@ -140,6 +140,23 @@ func (cache *Cache) removeAndDeleteFile(p string, file *cachedFile) {
 // return whatever's been stored there, which might be a directory and therefore contain
 // multiple files to be returned.
 func (cache *Cache) RetrieveArtifact(artPath string) (map[string][]byte, error) {
+	ret := map[string][]byte{}
+	if strings.Contains(artPath, "*") {
```
csdigi: Curious, is this how we detect we are globbing? Globs don't always need to contain that wildcard right? (It could equally be something like "?" etc right?

peterebden: good point, yeah. Have made it more robust - there are a pretty small set of special characters. I do think it's a good idea to do the check rather than globbing every time though.

PR #132, opened 2016-08-19
Add default labels to proto rules

fs75:


src/parse/rules/proto_rules.py
```diff
@@ -20,6 +20,10 @@
 }
 
 
+DEFAULT_PROTO_LABELS = ['protobuf']
```
peterebden: Can you make these variables private please?

peterebden: We already have this, but it's called "proto" - try e.g. `plz query alltargets --include proto --hidden`. Adding grpc separately is probably useful though.

fs75: ok, removed protobuf
didn't know about proto

peterebden:
Looks good, thanks Fabian!

PR #133, opened 2016-08-20
Support for pushing metrics to Prometheus

peterebden:
I imagine we might want other metrics later, but this should be enough to get started with.

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -11,10 +11,12 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"gopkg.in/op/go-logging.v1"
 
 	"core"
+	"metrics"
 	"parse"
 )
 
@@ -24,6 +26,7 @@ var log = logging.MustGetLogger("build")
 var stopTarget = fmt.Errorf("stopping build")
 
 func Build(tid int, state *core.BuildState, label core.BuildLabel) {
+	start := time.Now()
 	target := state.Graph.TargetOrDie(label)
 	target.SetState(core.Building)
 	if err := buildTarget(tid, state, target); err != nil {
@@ -39,6 +42,7 @@ func Build(tid int, state *core.BuildState, label core.BuildLabel) {
 		target.SetState(core.Failed)
 		return
 	}
+	metrics.Record(target, time.Since(start))
 
 	// Add any of the reverse deps that are now fully built to the queue.
 	for _, reverseDep := range state.Graph.ReverseDependencies(target) {
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -126,7 +126,7 @@ func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget) (err
 				RemoveOutputs(target)
 				return false
 			} else if outputHashErr != nil || !bytes.Equal(oldOutputHash, newOutputHash) {
-				target.SetState(core.Built)
+				target.SetState(core.Cached)
 				state.LogBuildResult(tid, target.Label, core.TargetCached, "Cached")
 			} else {
 				target.SetState(core.Unchanged)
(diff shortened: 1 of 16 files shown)


src/build/build_step.go
```diff
@@ -39,6 +42,7 @@ func Build(tid int, state *core.BuildState, label core.BuildLabel) {
 		target.SetState(core.Failed)
 		return
 	}
+	metrics.Record(target, time.Since(start))
```
csdigi: Should we record this even if there was an error above?

src/core/config.go
```diff
@@ -108,6 +108,7 @@ func DefaultConfiguration() *Configuration {
 	config.Cache.Dir = ".plz-cache"
 	config.Cache.DirCacheHighWaterMark = "10G"
 	config.Cache.DirCacheLowWaterMark = "8G"
+	config.Metrics.PushFrequency = 100 // push every 100ms
```
csdigi: Seems like a good value, do we have the notion of pushing after the whole build has failed too (so if I set this high, it will still try to flush unpushed metrics before please quits?)

src/metrics/prometheus.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/push"
+	"gopkg.in/op/go-logging.v1"
+
+	"core"
+)
+
+var log = logging.MustGetLogger("metrics")
+
+const maxErrors = 3
```
csdigi: nit: A little documentation that says please will stop trying to push metrics for the build after it has failed this many times?

csdigi:
Thanks pebers LGTM, thanks for the test too, looks like it was a bit of a fiddle!

src/build/build_step.go
peterebden: The Record function takes that into account; I thought it might be interesting to record which targets fail the most often.

src/metrics/prometheus.go
peterebden: done

src/metrics/prometheus.go
```diff
@@ -0,0 +1,207 @@
...
+
+// initMetrics initialises a new metrics instance.
+// This is deliberately not exposed but is useful for testing.
+func initMetrics(url string, frequency int) *metrics {
+	u, err := user.Current()
+	if err != nil {
+		log.Warning("Can't determine current user name for metrics")
+		u = &user.User{Username: "unknown"}
+	}
+	constLabels := prometheus.Labels{
```
dimpavloff: Hey, apologies for the late comment but I think it would be good to also have a label about the branch on which this has been ran, if that information is easy provide here. However, the Prometheus docs explicitly [advice against](https://prometheus.io/docs/practices/naming/#labels) labels with large cardinality so perhaps a boolean about whether we're on the master branch or not would be good enough. Because of the same warning, the user label might be a problem, although we could just wait and see because I think it's definitely a useful dimension

peterebden: No problem, it's a good idea; my only concern is that plz is agnostic to the version control system so at the moment we never know the branch. I suppose we could quickly run a cheeky git command to find it.

I think I prefer a boolean for master or not; we've already got quite a large set of build targets so it'd make sense not to get the cross product of both, and I think in most cases all we'd care about analysing was master vs. non-master builds.

dimpavloff: > plz is agnostic to the version control system so at the moment we never know the branch. I suppose we could quickly run a cheeky git command to find it.
Hm, I thought this might be the case :/ I think a better solution is to have a section in the plzconfig about the additional dimensions/labels to apply to each metric, although that feels like a more substantial piece of work.

Re master/non-master -- definitely agree, that was my thinking, too.

PR #135, opened 2016-08-23
Slightly experimental support for C++ coverage using gcov

peterebden:
Not sure this will account for files outside the test package, and it doesn't deal so well with declaration-only headers, but it's better than nothing :)


src/parse/rules/cc_rules.py
```diff
@@ -71,6 +71,7 @@ def cc_library(name, srcs=None, hdrs=None, private_hdrs=None, deps=None, visibil
     cmds = {
         'dbg': cmd_template % (CONFIG.CC_TOOL, dbg_flags),
         'opt': cmd_template % (CONFIG.CC_TOOL, opt_flags),
+        'cover': cmd_template % (CONFIG.CC_TOOL, dbg_flags) + ' -ftest-coverage -fprofile-arcs -fprofile-dir=.',
```
alamaison: Might be good to pull the cover flags out and reuse them below

alamaison:
LGTM

src/parse/rules/cc_rules.py
peterebden: Good idea thanks, done.

PR #137, opened 2016-08-26
Add flag to display test output always, even on success.

### LINKED ISSUE #136: Please allow me to see logging output when running tests
When running tests with `plz test //pkg:my_test -p` I don't get any output the test/code might have put out, unless the test failed.

It would be handy to have a flag that you could toggle to always get the output.

peterebden:
Fixes #136 

Output gets printed after each test result at the end. If using `-p` it will be printed as each individual test completes too.


janhancic:
amazing, thanks!

PR #138, opened 2016-08-28
Allow attaching arbitrary labels to metrics

peterebden:
As we discussed in #133, it'd be nice to be able to attach the git branch or similar, but I don't think plz should have too much knowledge about the specifics of the version control system. This seems like a pretty generic middle ground such that the users can customise it to their workflow.

### DIFF
--- a/docs/config.html
+++ b/docs/config.html
@@ -191,6 +191,33 @@
         <code>.pb.go</code>, <code>_pb2.py</code>, etc.</li>
     </ul>
 
+    <h3>[Metrics]</h3>
+
+    <p>Options related to metric collection. Currently the only metrics implementation pushes
+      to a <a href="https://prometheus.io/">Prometheus</a>
+      <a href="https://github.com/prometheus/pushgateway">pushgateway</a> for collection.</p>
+
+    <ul>
+      <li><b>PushGatewayURL</b><br/>
+	The URL of the pushgateway to send metrics to.</li>
+
+      <li><b>PushFrequency</b> (integer)<br/>
+	The frequency, in milliseconds, to push statistics at. Defaults to 100.</li>
+    </ul>
+
+    <h3>[CustomMetricLabels]</h3>
+
+    <p>Describes optional labels to attach to metrics.<br/>
+      For technical reasons this is a separate section to the main metrics options.</p>
+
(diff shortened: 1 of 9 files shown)


src/metrics/prometheus.go
```diff
@@ -205,3 +211,21 @@ func (m *metrics) pushMetrics() int {
 	log.Debug("Push #%d of metrics in %0.3fs", m.pushes, time.Since(start).Seconds())
 	return 0
 }
+
+// deriveLabelValue runs a command and returns its output.
+// It returns the empty string on error; we assume it's better to keep the set of labels constant on failure.
```
peterebden: I'm not 100% sure on this; we might be better attaching no label at all in that case. Thoughts?

docs/config.html
```diff
@@ -191,6 +191,33 @@
...
+    </ul>
+
+    <h3>[CustomMetricLabels]</h3>
+
+    <p>Describes optional labels to attach to metrics.<br/>
+      For technical reasons this is a separate section to the main metrics options.</p>
+
+    <p>These are defined as key-value pairs, where the key is the label attached to the metric
+      and the value is a command to run on the system that defines the value to attach.<br/>
+      Hence these labels are constant across all build targets.</p>
```
dimpavloff: It might be good to point out the label values are evaluated once per `plz` execution.

src/metrics/prometheus.go
dimpavloff: I think it makes sense to keep a constant number. 

Alternatively, could we fail hard since this is done once in the `initMetrics`? And if we do, what do you think about checking that the command produced only one value (or one line) and failing otherwise? I don't know how the Prometheus API/UI will handle newlines, and besides, anything that is not 1 value feels like an error to me

dimpavloff:
Looks great, thanks!

src/metrics/prometheus.go
peterebden: Agreed, it's probably better to be stricter now, we can always loosen it later if it becomes a problem.

dimpavloff:
That was fast! Lgtm and thanks! :)

peterebden:
No prob, thanks for the review!

PR #140, opened 2016-08-30
make the license matching case insensitive

csdigi:


peterebden:
LGTM, thanks Will!

Not sure what's up with Travis. I had a similar thing from it the other day, it went away the next time though so never figured out what it was :(

PR #143, opened 2016-09-12
Build linter

peterebden:
Lints for a bunch of things:
 - Syntax errors
 - Use of dict.iteritems, dict.iterkeys, dict.itervalues
 - Unsorted iteration of sets or dicts (a little unreliably)
 - Calling builtin rules through non-keyword arguments
 - Deprecation warnings
 - Incorrect arguments to builtin functions
 - Duplicates in argument lists

Also a small fix for an issue around pexes sometimes not working properly when they don't have .pex in the name.


csdigi:
Thanks Pebers, LGTM. I assume this is going to be used in conjunction with buildifier for ordering / formatting in practise?

src/lint/linter.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def _lint(contents):
+    try:
+        tree = ast.parse(contents)
+    except SyntaxError as err:
+        yield err.lineno, SYNTAX_ERROR
+        return
+
+    for n in ast.walk(tree):
```
dimpavloff: nit: it might be more readable if the below conditions are split into separate functions which will also make testing them much easier and not requiring file fixtures

dimpavloff:
Awesome, thanks Pebers

src/lint/linter.py
peterebden: I quite liked the file fixtures, dumping them all into a Python file as constants seemed kind of messy. Also the interface to those functions would be lower level than I want - something above it would have to parse the file into an AST and pass that in, and I'd rather keep the `ast` classes within this module.

peterebden:
@csdigi yes, but this deliberately doesn't do any formatting, just "correctness" style checks for things I can't rule out in the parser itself.

src/lint/linter.py
dimpavloff: Definitely keep the fixture files -- they are effectively an acceptance test -- and have the lint checks inside the same module. My thinking was just splitting them in separate functions. Something along the lines of the following:

```
def _check_unsorted_iteration(n):
   if isinstance(n, ast.For):
       if isinstance(n.iter, ast.Call) and isinstance(n.iter.func, ast.Name):
           if n.iter.func.id in UNSORTED_CALLS:
               return n.lineno, UNSORTED_CALLS[n.iter.func.id]
       elif isinstance(n.iter, (ast.Dict, ast.DictComp)):
           return n.lineno, UNSORTED_DICT_ITERATION
       elif isinstance(n.iter, (ast.Set, ast.SetComp)):
           return n.lineno, UNSORTED_SET_ITERATION
  return None

def _check_builtin_args(n):
   ...

def _check_deprecated_functions(n):
  ...

...
LINTER_FNS = [...]
def _lint(contents):
  ...
  lint_errors = (fn(n) for n in ast.walk(tree) for fn in LINTER_FNS)  # or just spell out the functions explicitly
  return (e for e in lint_errors if e)  # avoiding using filter vs itertools.ifilter :P
```

peterebden: Ah I see what you mean. Done.

dimpavloff:
\o/ Thanks :)

PR #144, opened 2016-09-19
Sh binary fix

dimpavloff:
- fix sh_binary def by ensuring the output script header is right
- make sure sh_binary can work when the target does not specify any dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ The core contributors so far are:
  * [Will Montgomery](https://github.com/csdigi)
  * [Fabian Siddiqi](https://github.com/FS89)
  * [Diana-Maria Costea](https://github.com/dianacostea)
+ * [Dimitar Pavlov](https://github.com/dimpavloff)
 
 Progress has been slightly hindered by our fearless leader Paul
 who continually insists that we should "get on with our work" and
--- a/src/parse/rules/sh_rules.build_defs
+++ b/src/parse/rules/sh_rules.build_defs
@@ -8,8 +8,8 @@ implement something more advanced (ala .sar files or whatever).
 _TARBALL_BOUNDARY = '__TARFILE_FOLLOWS__'
 
 # NB: This string is formatted strangely due to bash variable expansion being a massive pain.
-_SH_BINARY_TMPL_PREFIX = """\
-#!%(shell_binary)s\n\
+_SH_EXEC="#!%s\n"
+_SH_BINARY_TMPL_PREFIX = """
 SKIP=\\\$(awk '/^%(tarball_boundary)s/ { print NR + 1; exit 0; }' \\\$0)\n\
 THIS=\\\$0\n\
 tail -n +\\\$SKIP \\\$THIS | tar -xz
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -71,6 +71,7 @@ The core contributors so far are:
  * [Will Montgomery](https://github.com/csdigi)
  * [Fabian Siddiqi](https://github.com/FS89)
  * [Diana-Maria Costea](https://github.com/dianacostea)
+ * [Dimitar Pavlov](https://github.com/dimpavloff)
```
csdigi: We did just talk about removing this list (as not many people remember to update it / are unsure when they should). Let's see what pebers thinks..

src/parse/rules/sh_rules.build_defs
```diff
@@ -55,21 +55,29 @@ def sh_binary(name, main, deps=None, visibility=None):
...
+            'echo "%s" > $OUT' % (_SH_BINARY_TMPL_PREFIX % tmpl_vars),
+            # Inject the user defined script.
+            'cat $SRCS >> $OUT',
+            # Inject the tar ball boundary.
+            'echo "%s" >> $OUT' % (_SH_BINARY_TMPL_SUFFIX % tmpl_vars),
+            # Compress the dependent files and dump out into the bash script.
+            'find . -type f | grep -v $SRCS | grep -v $(echo $OUT | sed "s|$TMP_DIR||g") | sed "s/^\.\///g" | tar zc -f - --files-from - >> $OUT',
+        ])
+    else:
+        cmds = ' && '.join([
```
peterebden: This branch is short enough that it might be easier to do it as one line. Up to you.

src/parse/rules/sh_rules.build_defs
```diff
@@ -55,21 +55,29 @@ def sh_binary(name, main, deps=None, visibility=None):
...
-        # Inject bash code to untar the compressed files.
-        'echo "%s" > $OUT' % (_SH_BINARY_TMPL_PREFIX % tmpl_vars),
-        # Inject the user defined script.
-        'cat $SRCS >> $OUT',
-        # Inject the tar ball boundary.
-        'echo "%s" >> $OUT' % (_SH_BINARY_TMPL_SUFFIX % tmpl_vars),
-        # Compress the dependent files and dump out into the bash script.
-        'find . -type f | grep -v $SRCS | grep -v $(echo $OUT | sed "s|$TMP_DIR||g") | sed "s/^\.\///g" | tar zc -f - --files-from - >> $OUT',
-    ])
+    if deps:
```
peterebden: Might be worth adding a comment here about why you need to branch on deps; it's not a usual thing in build rules.

dimpavloff:
Thanks. Btw, I realised that I can avoid passing the `shell` kwarg and instead figure it out from the script. That is much better because there will be only 1 place to set -- the script itself, which is the right place. I'll send an update in a bit

README.md
peterebden: There's a slightly more complete one on the website, don't mind Dimitar adding this for now as long as he doesn't mind if it goes away soon too :)

src/parse/rules/sh_rules.build_defs
dimpavloff: simplified, thanks

PR #145, opened 2016-09-20
Fix race condition with concurrently building filegroup rules

peterebden:
Basically builds can sometimes fail with mysterious errors when there are multiple filegroups outputting the same file. That's normally banned but we make a special case for filegroups because they're often implicitly created by other rules and since they're special-cased we know they're OK.

But as we saw once or twice yesterday that can still fail. There are various implicit guards like checking if the hash is the same, but turns out that's not always enough; the different rules race attempting to link the same file, one fails and then tries to copy it, then others can do a partial read on it and get bad hash values.

Have fixed by removing the fallback to copy behaviour and adding a check that if the error was that the file already existed and the two have the same inode, then it's effectively been linked anyway so we can just carry on.

### DIFF
--- a/src/core/utils_test.go
+++ b/src/core/utils_test.go
@@ -3,6 +3,7 @@ package core
 import (
 	"crypto/sha1"
 	"encoding/base64"
+	"io/ioutil"
 	"os"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


src/core/utils_test.go
```diff
@@ -130,6 +131,18 @@ func TestLookPathColons(t *testing.T) {
 	assert.Equal(t, "bash", info.Name())
 }
 
+func TestIsSameFile(t *testing.T) {
+	err := ioutil.WriteFile("issamefile1.txt", []byte("hello"), 0644)
```
dimpavloff: shouldn't these be cleaned at the end of the test?

peterebden: plz will clean them for us :) the test runs in a temporary directory that it cleans up after.

PR #148, opened 2016-10-05
Add a --unique flag on query deps to only output each target once

peterebden:
The current implementation prints a complete tree, which is not always what you want and can produce a lot of output.


csdigi:
Thanks Pebers, LGTM. Was a little strange at first to see print targets tracking the unique stuff, but no the option due to the recursion.

fs75:
LGTM thanks

PR #150, opened 2016-10-10
Fix for case where one filegroup could remove another's file

peterebden:
aka. mysterious "no such file" errors.

At least I hope this is the fix, it is quite hard to trigger reliably & repeatably.

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -334,19 +334,30 @@ func moveOutput(target *core.BuildTarget, tmpOutput, realOutput string, filegrou
 	if err != nil {
 		return true, err
 	}
-	// The tmp output can be a symlink back to the real one; this is allowed for rules like
-	// filegroups that attempt to link outputs of other rules. In that case we can't
-	// remove the original because that'd break the link, but by definition we don't need
-	// to actually do anything more.
-	// TODO(pebers): The logic here is quite tortured, consider a (very careful) rewrite.
-	dereferencedOutput, _ := filepath.EvalSymlinks(realOutput)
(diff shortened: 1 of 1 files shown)


src/build/build_step.go
```diff
@@ -334,19 +334,30 @@ func moveOutput(target *core.BuildTarget, tmpOutput, realOutput string, filegrou
...
-	// TODO(pebers): The logic here is quite tortured, consider a (very careful) rewrite.
-	dereferencedOutput, _ := filepath.EvalSymlinks(realOutput)
-	if absOutput, _ := filepath.Abs(realOutput); tmpOutput == absOutput || realOutput == tmpOutput || dereferencedOutput == tmpOutput {
-		log.Debug("%s and %s are the same file, nothing to do", tmpOutput, realOutput)
+	realOutputExists := core.PathExists(realOutput)
+	// If this is a filegroup we hardlink the outputs over and so the two files may actually be
+	// the same file. If so don't do anything else and especially don't delete & recreate the
+	// file because other things might be using it already (because more than one filegroup can
+	// own the same file).
+	if filegroup && realOutputExists && core.IsSameFile(tmpOutput, realOutput) {
```
peterebden: The diff isn't brilliant, but basically bailing out early here when the two are the same file is the real fix. The second block below is a mild optimisation so we don't start dereferencing symlinks & comparing paths if the real output doesn't exist anyway since we now check that sooner. Not that I'm even sure that code is still needed...

fs75:
Thanks pebers. Hopefully this will fix the issue.

PR #151, opened 2016-10-13
Add a timeout to metric pushes.

peterebden:
Quite a lot of this is fiddling about with the config so you can write `500ms` etc.


samueljamesbell:
LGTM

PR #155, opened 2016-11-14
Add support for an experimental dir.

peterebden:
The rules are:
- Targets within that directory can depend on any target outside, even if they normally couldn't see it.
- Targets outside can never depend on any target inside it.
- When scanning the repo at the top level it is not recursed into.

### DIFF
--- a/src/core/build_label.go
+++ b/src/core/build_label.go
@@ -199,7 +199,7 @@ func (label BuildLabel) IsAllTargets() bool {
 	return label.Name == "all"
 }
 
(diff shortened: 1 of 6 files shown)


fs75:
Just to confirm, this goes in the plzconfig?

peterebden:
Yes correct, the change in config.go adds that field to the struct that defines what we read from .plzconfig.

fs75:
OK, makes sense. I wonder if there is a better name then. Not sure `experimental` fully conveys that idea...

Can't think of a better name myself tbh, so just a thought.

PR #156, opened 2016-11-14
HA / cluster support for RPC cache server

peterebden:
Servers gossip amongst themselves to identify one another, and use RPC for 'real' communication.

It's a pretty simple scheme, the cluster size is fixed at startup, there's no online hash redistribution, a fixed replication factor of 2 and the client takes care of routing requests to the correct node. It's backwards compatible for both client & server, except in the case where one ran a clustered server with > 2 nodes and connected to it with an old client - that would still work but the client would get some cache misses because it'd not know to send them to the right place.

We briefly discussed putting something else in front of it last week, I thought about that but don't think it'll work because it would have to understand gRPC, and would essentially end up looking pretty much the same as the current server. Plus there will be a benefit to not having all the artifacts on one service since it can be slow to start since it needs to walk them all.


src/cache/cluster/cluster.go
```diff
@@ -0,0 +1,298 @@
...
+			cluster.nodes = resp.Nodes
+			cluster.size = int(resp.Size)
+			return
+		}
+	}
+	log.Fatalf("Unable to contact any other cluster members")
+}
+
+// InitCluster seeds a new plz cache cluster.
+func (cluster *Cluster) Init(size int) {
```
peterebden: `newNode` does that, I'll add a comment pointing that out

src/cache/cluster/cluster.go
```diff
@@ -0,0 +1,298 @@
...
+			if n.Name == node.Name {
+				log.Warning("Node %s / %s is taking over slot %d", node.Name, node.Addr, i)
+			} else {
+				log.Notice("Populating node %d: %s / %s", i, node.Name, node.Addr)
+			}
+			cluster.nodes[i] = newNode(i)
+			// Remove any client that might exist for this node so we force a reconnection.
+			cluster.clientMutex.Lock()
+			defer cluster.clientMutex.Unlock()
+			delete(cluster.clients, cluster.nodes[i].Address)
```
peterebden: Fair point on both cases. I have changed `clients` to be keyed by name as well.

src/cache/cluster/cluster.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+	for i, n := range cluster.nodes {
+		if n.Name == "" || n.Name == node.Name {
+			// Available slot. Or, if they identified as an existing node, they can take that space over.
+			if n.Name == node.Name {
+				log.Warning("Node %s / %s is taking over slot %d", node.Name, node.Addr, i)
+			} else {
+				log.Notice("Populating node %d: %s / %s", i, node.Name, node.Addr)
+			}
+			cluster.nodes[i] = newNode(i)
```
peterebden: Nope, it can't, but two concurrent requests might race and various things could then go wrong. I've added a `RWMutex` around `nodes` which should fix it.

src/cache/server/rpc_server_main.go
```diff
@@ -44,11 +54,27 @@ func main() {
 	} else if opts.TLSFlags.KeyFile == "" && (opts.TLSFlags.WritableCerts != "" || opts.TLSFlags.ReadonlyCerts != "") {
 		log.Fatalf("You can only use --writable_certs / --readonly_certs with https (--key_file and --cert_file)")
 	}
+
 	log.Notice("Scanning existing cache directory %s...", opts.Dir)
 	cache := server.NewCache(opts.Dir, time.Duration(opts.CleanFlags.CleanFrequency),
 		time.Duration(opts.CleanFlags.MaxArtifactAge),
 		uint64(opts.CleanFlags.LowWaterMark), uint64(opts.CleanFlags.HighWaterMark))
+
+	var clusta *cluster.Cluster
```
peterebden: Haha I couldn't think of a good name for it, `cluster` would shadow the package name!

peterebden:
Thanks for the review, Dimitar!

src/cache/cluster/cluster.go
```diff
@@ -120,7 +120,7 @@ func (cluster *Cluster) GetMembers() []*pb.Node {
 	for _, m := range cluster.list.Members() {
 		cluster.newNode(m)
 	}
-	return cluster.nodes
+	return cluster.nodes[:]
```
dimpavloff: good spot

PR #161, opened 2016-11-25
Adding some real locking around building filegroups to prevent awkward races.

peterebden:
This is the latest attempt to fix that dreaded case where multiple filegroups build the same file at once and things can mysteriously break (aka. "the cloudformation thing"...).


fs75:
LGTM hopefully this will get rid of our CFN issues?

peterebden:
Theoretically yes :)

PR #162, opened 2016-11-29
Adding a gc command to identify unused targets / sources

peterebden:
It's possible to specify things to keep in .plzconfig for cases where it's not obvious to the algorithm what is a "gc root" of the repo.

I think this is definitely better to have in plz itself, as much as it's theoretically possible to use `plz query graph` to get enough info to do this outside it's quite slow on large repos with the JSON serialisation and it's a pain to reinvent the helper functions that are available internally.


fs75:
Thanks lgtm

PR #163, opened 2016-12-05
Adds support for remote workers

### LINKED ISSUE #159: Persistent workers
Basically the ability to persist a process for compilers to reduce per-rule overhead. Bazel can do it (http://www.bazel.io/blog/2015/12/10/java-workers.html) and so can Buck (https://buckbuild.com/rule/worker_tool.html).

This will be most relevant to Java, but I think we should implement as a generic API - Buck's approach seems a better fit for us, although I'm not sure we'll mimic the `$(worker :my_worker)` syntax.

peterebden:
Initial implementation for Java as a persistent wrapper around javac. Speedup of 4-5x on Java trees of reasonable size (consistent with https://bazel.build/blog/2015/12/10/java-workers.html).

I've tried a couple of different approaches, first a standalone client & server which is transparent to plz, but really awkward for the client (because it has to start the server and it's hard to know when it should), and second one using gRPC which is again awkward because of timing issues around opening ports etc. This one has plz spawn the workers so it knows it only does one at a time, and it communicates with them over stdin/stdout using a proto-based scheme. It's quite similar to Buck's approach in a lot of ways.

Resolves #159


src/build/java/build/please/compile/FileFinder.java
```diff
@@ -0,0 +1,48 @@
...
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.nio.file.FileVisitResult.CONTINUE;
+
+class FileFinder extends SimpleFileVisitor<Path> {
+
+  private final String extension;
+  private final List<String> files = new ArrayList<>();
```
csdigi: Not sure what people do with this list once it is returned by getFiles, but technically a LinkedList would save you any internal array resizing costs if adding many files here. (Practically though it is probably not worth it).

install.sh
```diff
@@ -35,6 +35,8 @@ cp -f plz-out/bin/src/build/go/please_go_test ${DEST}/please_go_test
 chmod 0775 ${DEST}/please_go_test
 cp -f plz-out/bin/src/lint/please_build_linter ${DEST}/please_build_linter
 chmod 0775 ${DEST}/please_build_linter
+cp -f plz-out/bin/src/build/java/javac_worker ${DEST}/javac_worker
+chmod 0775 ${DEST}/please_build_linter
```
peterebden: thx

src/build/java/build/please/compile/FileFinder.java
peterebden: Generally it just gets passed into getJavaFileObjectsFromStrings, so any iterable will do. Might as well change it...

peterebden:
Thanks Will!

PR #165, opened 2016-12-07
Download a portable PyPy if nothing else is available

### LINKED ISSUE #158: Consider using portable PyPy
See https://github.com/squeaky-pl/portable-pypy; basically rather than having to provide instructions about how to install PyPy on Linux we could automatically fetch a version for plz and use that.

Probably won't affect other platforms but at least it's pretty easy to install with Brew.

peterebden:
Should make it easy to get going on Linux without needing any system deps.

Fixes #158


peterebden:
Thanks for the review!

PR #166, opened 2016-12-09
Add the ability to horizontally scroll  in pre tags

kileras:
When the screen is small enough (for example is you have the devtools open all the time :D) the pre blocks are too narrow and the content gets cut.

This adds horizontal scroll to all the pre elements so it doesn’t happen. It sets the vertical scroll to hidden since the presence of the horizontal scrollbar can produce little unnecessary vertical scroll in some browsers and I don’t think there will be any pre with vertical scroll.

### DIFF
--- a/docs/styles.css
+++ b/docs/styles.css
@@ -55,6 +55,9 @@ pre {
     border-radius: 3px;
     padding: 2px;
     color: white;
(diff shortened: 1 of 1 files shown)


peterebden:
Thanks!

PR #168, opened 2016-12-22
Add labels to sh binary

fs75:


src/parse/rules/sh_rules.build_defs
```diff
@@ -19,29 +19,32 @@ exit 0\n\
...
-      src (str): Source file for the rule
-      deps (list): Dependencies of this rule
+      name (str): Name of the rule.
+      src (str): Source file for the rule.
+      deps (list): Dependencies of this rule.
       visibility (list): Visibility declaration of the rule.
       link (bool): Deprecated, has no effect.
+      labels (list): List of labels.
     """
+    labels = labels or []
```
peterebden: nit: this line is unnecessary (passing labels=None should be perfectly acceptable)

fs75: thanks, removed.

PR #169, opened 2017-01-01
plz help

peterebden:
Implements `plz help go_binary` and so forth, an instant reference for builtin rules & whatnot.

I've added stuff for .plzconfig too, although I'm not super happy; would rather have one central source of truth for that (maybe struct tags?). Will leave that for later though. I've also repurposed some existing levenshtein-based suggestion code to suggest sufficiently similar help topics.

What else does one do on a cold & rainy New Years Day...


src/help/help.go
```diff
@@ -0,0 +1,121 @@
...
+	topic = strings.ToLower(topic)
+	for i, filename := range allHelpFiles {
+		if message, found := halp(topic, filename, allHelpPreambles[i]); found {
+			return message
+		}
+	}
+	return ""
+}
+
+func halp(topic, filename, preamble string) (string, bool) {
```
fs75: nit: rename to `loadHelpFile`

src/utils/suggest.go
```diff
@@ -0,0 +1,57 @@
...
+	ret := make([]string, len(options))
+	for i, o := range options {
+		ret[i] = o.s
+	}
+	return ret
+}
+
+// Suggest implements levenshtein-based suggestions on a sequence of items and
+// produces a single message from them.
+func SuggestMessage(needle string, haystack []string, maxSuggestionDistance int) string {
```
fs75: PrettyPrintSuggestion or smth like that? Had to read through the code of both functions to understand the difference.

src/utils/suggest.go
```diff
@@ -0,0 +1,57 @@
...
+		if i > 0 {
+			if i < len(options)-1 {
+				msg += ", "
+			} else {
+				msg += " or "
+			}
+		}
+		msg += o
+	}
+	return msg + " ?" // Leave a space so you can select them without getting the question mark
```
fs75: so nice of you, hate it when this happens with CLIs

src/help/help.go
```diff
@@ -0,0 +1,121 @@
...
+`
+const miscPreamble = `
+%s is a general Please concept.
+
+`
+const defaultHelpMessage = `
+Please is a high-performance language-agnostic build system.
+
+Try plz help <topic> for help on a specific topic;
+plz --help if you want information on flags / options/ commands that it accepts;
```
dimpavloff: nit: space before the `/` ?

peterebden:
Thanks Fabian, have made those changes

src/help/help.go
```diff
@@ -0,0 +1,121 @@
...
+	return false
+}
+
+func help(topic string) string {
+	if topic == "" {
+		return defaultHelpMessage
+	} else if topic == "topics" {
+		return fmt.Sprintf(topicsHelpMessage, strings.Join(allTopics(), "\n"))
+	}
+	topic = strings.ToLower(topic)
```
dimpavloff: are you deliberately not interpreting `plz help TOPICS` as `plz help topics`? if not, obviously move this line at the start

src/help/help.go
```diff
@@ -0,0 +1,121 @@
...
+Or see the website (https://please.build) for more information.
+`
+const topicsHelpMessage = `
+The following help topics are available:
+
+%s
+`
+
+var allHelpFiles = []string{"rule_defs.json", "config.json", "misc.json"}
+var allHelpPreambles = []string{rulesPreamble, configPreamble, miscPreamble}
```
dimpavloff: Instead of having this files-preambles order-dependent mapping here, you could include the preamble in the file itself and nest the help one level deeper. However, I don't know if this makes things any more cohesive, since you probably are still going to have some string constants in this .go file

peterebden: Done, I think that is better overall (deserialising is a bit more of a faff but it's probably worth it for more flexibility later).

src/help/help.go
peterebden: Nope, just a random oversight!

peterebden:
Thanks Dimitar!

PR #171, opened 2017-01-06
Pre css border was a bit weird

fr1n63:
You could see the border radius hanging down, now it's sharp, shouldn't have annoyed me but it did. 

Also I don't think the overflow/scrolling does what you wanted it too.


peterebden:
Thanks!

PR #174, opened 2017-01-18
Initial cross-compiling support.

peterebden:
Essentially we identify cases when we require a different architecture to the host and create duplicates of targets for that arch. They live in different directories under plz-out etc. There are two ways that can happen; either on the command line there's a flag for the arch to target, or targets can specify that they always target a specific arch (for example, a target building a rkt container might always be Linux).

The actual support is fairly lightweight; BUILD files are not reparsed so swapping on CONFIG.OS / CONFIG.ARCH in them is a little dangerous. Those are exposed as variables when the target builds. That's sufficient for relatively simple cases like Go (where we just need to set GOOS / GOARCH to cross-compile), Java (where it's really just maven_jar rules with native dependencies to worry about) and possibly Python (pip_library for libraries with C bits being the only issue).
It's unlikely to be very satisfying for C, but I'm not sure I want to build complete support there right now.


src/core/build_label.go
```diff
@@ -303,6 +314,40 @@ func (label BuildLabel) IsEmpty() bool {
...
+
+// noArch returns a new build label with no architecture.
+func (label BuildLabel) noArch() BuildLabel {
+	label.Arch = ""
+	return label
+}
+
+// FullArch returns the full architecture, even if Arch is empty, in which case it will return
+// the host architecture.
+func (label BuildLabel) FullArch() string {
```
dimpavloff: nit: could be a ptr receiver method

src/core/build_label.go
```diff
@@ -303,6 +314,40 @@ func (label BuildLabel) IsEmpty() bool {
...
+		return HostArch
+	}
+	return label.Arch
+}
+
+// OsArch attempts to break the label's architecture into a Go-style os/arch pair,
+// for example linux_amd64 -> (linux, amd64).
+// If that can't be determined (since Please's architectures are freeform) the returned OS
+// will be the empty string.
+func (label BuildLabel) OsArch() (string, string) {
```
dimpavloff: same here

src/core/build_label_test.go
```diff
@@ -50,3 +50,38 @@ func TestUnmarshalText(t *testing.T) {
...
+	label := BuildLabel{PackageName: "src/core", Name: "core"}
+	assert.Equal(t, "//src/core:core", label.String())
+	label = BuildLabel{PackageName: "src/core", Name: "core", Arch: "test_x86"}
+	assert.Equal(t, "//src/core:core [test_x86]", label.String())
+}
+
+func TestToArch(t *testing.T) {
+	label1 := BuildLabel{PackageName: "src/core", Name: "core"}
+	label2 := label1.toArch("test_x86")
+	assert.Equal(t, "", label1.Arch)
```
dimpavloff: +1 for confirming the current label hasn't been mutated

src/core/build_target.go
```diff
@@ -286,6 +289,21 @@ func (target *BuildTarget) DeclaredDependencies() []BuildLabel {
 	return ret
 }
 
+// DeclaredArchDependencies returns the original declaration of this target's dependencies,
+// adjusted to be the appropriate architecture.
+func (target *BuildTarget) DeclaredArchDependencies() []BuildLabel {
```
dimpavloff: It would be nice to have unit tests for this one

src/core/graph.go
```diff
@@ -184,6 +216,66 @@ func (graph *BuildGraph) linkDependencies(fromTarget, toTarget *BuildTarget) {
...
+
+// cloneTargetForArch returns a build target for the given architecture. It's otherwise
+// identical to the original one.
+func (graph *BuildGraph) cloneTargetForArch(target *BuildTarget, arch string) *BuildTarget {
+	archLabel := target.Label.toArch(arch)
+	t, present := graph.targets[archLabel]
+	if present {
+		return t
+	}
+	graph.targets[archLabel] = target // Sentinel, so we don't recurse into here incorrectly.
```
dimpavloff: should there be a lock when interacting with `graph.targets`? Presumably it's fine without because even if different goroutines overwrote each other, the contents would still be equivalent?

src/core/build_label.go
peterebden: Done, this type is pretty old and for mostly bad reasons generally doesn't use pointer receivers (in fact I generally pass them by value). Sometime I'll change it all over although it's quite a wide-ranging change so I'll leave it out of this PR.

src/core/build_target.go
peterebden: Fair call :) Done.

src/core/graph.go
peterebden: There is, `graph.mutex` is already acquired at this point, and Go mutexes aren't re-entrant so we can't lock again. 
Go doesn't have any official or unofficial way of annotating that as a requirement though that I know of - Google used to have some stuff internally for C++ that was similar to [Clang's](http://clang.llvm.org/docs/ThreadSafetyAnalysis.html), it'd be sweet if they had something similar.

I'll add a comment to that effect, anyway.

peterebden:
> Just to confirm, if the command line flag specifies one arch and builds a target and that target or some of its dependencies has arch in the build rule which differs, that would cause a panic with an error message, right? +1 for that

Yes, exactly, if a target specifies an arch it should only be able to be built for that arch.
(Assuming I've got it all right, this one is quite involved...)

Might still need some more thought anyway, Go cross-compiling isn't quite working because it needs to rebuild the standard library for the target arch. Not quite sure how best to solve it though.

src/core/graph.go
dimpavloff: Cool, thanks. That link looks interesting... :)

PR #176, opened 2017-01-19
Add proto.jsdep config option that gets used when building JS protos

janhancic:
I didn't add it to `config.json` since we don't really support JS yet. But I'm not sure I missed anything else? I did test it against our repo and it seems to work fine when I set the appropriate dependency in our .plzconfig


peterebden:
Looks good, thanks!

The only other thing we might want as well is to add a `js_deps` argument to `proto_library`. That's mostly used for `grpc_library` though which would also need an extra config setting to match the ones for other languages. Up to you if you want to do that now or leave it for later.

janhancic:
I don't think we need it for now. I can have another look once we add proper JS support, if it's ok with you.

peterebden:
Yup all good, thanks!

PR #181, opened 2017-01-26
Kill any remaining workers on exit

peterebden:


fs75:
Thanks for looking into this!

PR #182, opened 2017-01-30
Rework Prometheus implementation a bit to avoid potential deadlock at close.

peterebden:
Basically it could sometimes deadlock when the metrics got shut down during the last push, and that push tipped it over the limit for errors; it would then exit the loop and never read the channel that was being pushed to.

Instead I'm erring on the side of possibly doing two pushes at the end, which should still be deadlined so shouldn't take too long.


fs75:
LGTM Thanks

PR #188, opened 2017-03-25
Fixate tarball timestamps

dimpavloff:
Thus the hash of the outputs does not change. It changing became apparent
with docker images where we use the plz stamp to derive the docker image
tag.


src/parse/rules/misc_rules.build_defs
```diff
@@ -537,15 +540,27 @@ def tarball(name, srcs, out=None, deps=None, subdir=None,
...
+        archive_cmd = 'tar -c * | gzip -n > $OUT'
+    else:
+        archive_cmd = 'tar %s -cf $OUT *' % compression
+
     build_rule(
         name=name,
         cmd=' && '.join([
             'mkdir -p _tmp/' + subdir,
-            'cp -r $(locations :_%s#files) _tmp/%s' % (name, subdir),
+            'cp -pr $(locations :_%s#files) _tmp/%s' % (name, subdir),
```
dimpavloff: I added the `-p` just to be explicit about something which was already doing. However, I'm not sure if preserving the files ownership desirable, given that those could be build by another user and shared via the remote cache. Let me know if you think we should just preserve the mode

ChangeLog
```diff
@@ -1,3 +1,6 @@
+Version 7.8.1
```
peterebden: FYI I've not done anything with 7.8.0 yet, so you could just roll it into that version.

src/parse/rules/misc_rules.build_defs
peterebden: I'm not sure that we should preserve mode bits. I'm a little worried that it might not be consistent between outputs that are retrieved from the cache and those that aren't.

We could try it and see though?

ChangeLog
dimpavloff: I see, thanks; I think it could be confusing, though, if there are changes included in a version which were committed after the commit with the corresponding version bump.

src/parse/rules/misc_rules.build_defs
dimpavloff: Hm, the cache having different mode bits does not sound right; I'll keep an eye on it next week. It shouldn't break, though, because `cp` preserves them by default, the above change just made it explicit. 
In general, I think the mode bits are useful to preserve for executables and probably something that should affect the output hash.

dimpavloff: p.s Agreed, let's release it on Monday and see what happens :)

peterebden: I'm not 100% sure that they are different any more; at some point we got a bit stricter about chmod'ing things in plz-out/gen. I just recall there being a codepath where I had to write a file and I didn't know what the mode should be, because the cache has lost it by then.

Hopefully it is not a problem, just mentioning it as a potential point of indeterminacy :)

PR #190, opened 2017-03-28
pass pkg_config_libs to cc_library in cc_binary

sfirmery:
When building a binary with cc_binary, pkg_config_libs must be passed to cc_library to have pkg-config added to build command.


peterebden:
Thanks for the fix, looks good!

PR #192, opened 2017-04-01
Add third_party top level folder to default GOPATH

dimpavloff:
The current default only works with deps that are declared under
third_party/go/ . However, for a Go-only repo the subfolder is
unnecessary and it shouldn't be required, I think


peterebden:
LGTM, thanks

PR #193, opened 2017-04-01
Fix go_get with dependencies when in declared in repo root

dimpavloff:
When in the root, building a go_get with a dependency, also in the root,
fails because with PKG="" , PKG_DIR == "." , and
GOPATH=$TMP_DIR/TMP_DIR/$PKG both the deps and the buildable get built
in the same folder. Thus, the usual mechanism of filtering out the dep
files (find $PKG_DIR and delete the equivalent in the $TMP_DIR) does not
work because we cannot distinguish these anymore. Furthermore, using
"." in the sed command gets interpreted as a wildcard so the sed .. | rm
currently does not delete anything. The end result is that the buildable
produces .a outputs which are also produced by its dependencies and plz
detects that multiple targets output the same file.

The PR saves the existing outputs in the PKG_DIR before performing go
install and uses this for the clean up. It also special cases the "." to
make the sed command work as desired.

### DIFF
--- a/src/parse/rules/go_rules.build_defs
+++ b/src/parse/rules/go_rules.build_defs
@@ -456,8 +456,16 @@ def go_get(name, get=None, outs=None, deps=None, exported_deps=None, visibility=
         if not binary:
             post_build = _extra_outs(get)
     cmd = [
-        'export GOPATH=$TMP_DIR:$TMP_DIR/$PKG',
+        # When PKG_DIR == ".", the install location likely coincides with the
+        # dependencies of this go_get (i.e they are all in the root's BUILD
+        # file). Thus we need to note which compiled .a files are already
+        # provided by the deps and filter these out from the
+        # output
+        'if [ -d $PKG_DIR ]; then '
+        'export EXISTING="$(find $PKG_DIR -name *.a)"; else export EXISTING=""; fi',
+
         'rm -rf pkg src',
+        'export GOPATH=$TMP_DIR:$TMP_DIR/$PKG',
         'go get -d ' + get,
     ]
     subdir = 'src/' + (get[:-4] if get.endswith('/...') else get)
@@ -468,7 +476,9 @@ def go_get(name, get=None, outs=None, deps=None, exported_deps=None, visibility=
         cmd.append('patch -s -d %s -p1 < ${TMP_DIR}/$(location %s)' % (subdir, patch))
     cmd.append('go install -gcflags "-trimpath $TMP_DIR" ' + ' '.join([get] + (install or [])))
     # Remove anything that was there already.
-    cmd.append('if [ -d $PKG_DIR ]; then find $PKG_DIR -name "*.a" | sed -e "s|$PKG_DIR/||g" | xargs rm -f; fi')
+    cmd.append('if [ -d $PKG_DIR ]; then '
+               'if [ $PKG_DIR = "." ]; then echo $EXISTING | xargs rm -f; '
+               'else echo $EXISTING | sed -e "s|$PKG_DIR/||g" | xargs rm -f; fi; fi')
     if not binary:
         cmd.extend([
             'find . -name .git | xargs rm -rf',
--- a/src/parse/rules/go_rules.build_defs
+++ b/src/parse/rules/go_rules.build_defs
@@ -456,16 +456,23 @@ def go_get(name, get=None, outs=None, deps=None, exported_deps=None, visibility=
         if not binary:
             post_build = _extra_outs(get)
     cmd = [
-        # When PKG_DIR == ".", the install location likely coincides with the
-        # dependencies of this go_get (i.e they are all in the root's BUILD
-        # file). Thus we need to note which compiled .a files are already
-        # provided by the deps and filter these out from the
-        # output
-        'if [ -d $PKG_DIR ]; then '
-        'export EXISTING="$(find $PKG_DIR -name *.a)"; else export EXISTING=""; fi',
+        # the TMP_DIR contains all the compiled dependencies already. Later when
+        # we go install, it is difficult to seperate its outputs from the
+        # deps outputs, hence why we capture them here in advance
+        'export existing="$(find $TMP_DIR -name \"*.a\")"',
 
-        'rm -rf pkg src',
-        'export GOPATH=$TMP_DIR:$TMP_DIR/$PKG',
+        # go_get dependencies could be in different paths, make sure they are
+        # all available in the GOPATH by detecting their pkg/ folder and using its parent
+        'export GOPATH="$TMP_DIR"',
+        # loop done like this because the GOPATH export should not be in a SubShell
+        'while IFS= read -r dep_dir; do '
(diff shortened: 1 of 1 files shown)


dimpavloff:
p.s. Admittedly, special-casing the root directory isn't pretty. Also, I think the above description of the dependency exclusion is what's happening in the current code but I could be wrong :) It took me a while to understand what's going on there.

My use case was having the prometheus client and protobuf as `go_get` targets in the root BUILD file, and the prometheus one depending on the protobuf one. When building the prometheus one, it would conflict with the protobuf one because it was still outputing `proto.a`:


```ENVIRONMENT:
PKG=
PKG_DIR=.
ARCH=amd64
OS=linux
LANG=en_GB.UTF-8
PATH=/usr/local/bin:/usr/bin:/bin
TMP_DIR=/home/dim/code/golang/src/github.com/dimpavloff/yap/plz-out/tmp/prometheus._build
SRCS=
OUTS=src/github.com/prometheus/client_golang/prometheus
NAME=prometheus
TOOLS=/usr/bin/go
OUT=/home/dim/code/golang/src/github.com/dimpavloff/yap/plz-out/tmp/prometheus._build/src/github.com/prometheus/client_golang/prometheus
TOOL=/usr/bin/go
export GOPATH=$TMP_DIR:$TMP_DIR/$PKG && rm -rf pkg src && go get -d github.com/prometheus/client_golang/prometheus && (cd src/github.com/prometheus/client_golang/prometheus && git checkout -q 738ed6c0b9b30ec936d4b74c1600ae76324efff5) && go install -gcflags "-trimpath $TMP_DIR" github.com/prometheus/client_golang/prometheus github.com/prometheus/client_golang/prometheus github.com/prometheus/client_golang/prometheus/promhttp && if [ -d $PKG_DIR ]; then find $PKG_DIR -name "*.a" | sed -e "s|$PKG_DIR/||g" | xargs rm -f; fi && find . -name .git | xargs rm -rf && find pkg -name "*.a" | sort && rm -rf src/github.com/prometheus/client_golang/prometheus/procfs/fixtures && rm -rf src/github.com/prometheus/client_golang/prometheus/golang/protobuf
19:53:14.102   DEBUG: Running post-build function for //:prometheus. Build output:
pkg/linux_amd64/github.com/beorn7/perks/quantile.a
pkg/linux_amd64/github.com/golang/protobuf/proto.a
pkg/linux_amd64/github.com/matttproud/golang_protobuf_extensions/pbutil.a
pkg/linux_amd64/github.com/prometheus/client_golang/prometheus.a
pkg/linux_amd64/github.com/prometheus/client_golang/prometheus/promhttp.a
pkg/linux_amd64/github.com/prometheus/client_model/go.a
pkg/linux_amd64/github.com/prometheus/common/expfmt.a
pkg/linux_amd64/github.com/prometheus/common/internal/bitbucket.org/ww/goautoneg.a
pkg/linux_amd64/github.com/prometheus/common/model.a
pkg/linux_amd64/github.com/prometheus/procfs.a
pkg/linux_amd64/github.com/prometheus/procfs/xfs.a
19:53:14.102   ERROR: ///BUILD: Looks like you have two go_get rules trying to output the same file; most likely you need to add some dependencies so that shared file is built independently.
19:53:14.102    INFO: //:prometheus: Running post-build function for //:prometheus
19:53:14.102   ERROR: //:prometheus: Failed post-build function for //:prometheus: Failed to run post-build function for target //:prometheus: Rules //:protobuf and //:prometheus in BUILD both attempt to output the same file: pkg/linux_amd64/github.com/golang/protobuf/proto.a

19:53:14.102   ERROR: //:prometheus failed:
Failed to run post-build function for target //:prometheus: Rules //:protobuf and //:prometheus in BUILD both attempt to output the same file: pkg/linux_amd64/github.com/golang/protobuf/proto.a

19:53:14.102   ERROR: //:prometheus: Build failed: Failed to run post-build function for target //:prometheus: Rules //:protobuf and //:prometheus in BUILD both attempt to output the same file: pkg/linux_amd64/github.com/golang/protobuf/proto.a
: Failed to run post-build function for target //:prometheus: Rules //:protobuf and //:prometheus in BUILD both attempt to output the same file: pkg/linux_amd64/github.com/golang/protobuf/proto.a

19:53:14.102   ERROR: //:prometheus failed:
Failed to run post-build function for target //:prometheus: Rules //:protobuf and //:prometheus in BUILD both attempt to output the same file: pkg/linux_amd64/github.com/golang/protobuf/proto.a

Build stopped after 13.29s. 1 target failed:
    //:prometheus
Failed to run post-build function for target //:prometheus: Rules //:protobuf and //:prometheus in BUILD both attempt to output the same file: pkg/linux_amd64/github.com/golang/protobuf/proto.a

    //:prometheus
Failed to run post-build function for target //:prometheus: Rules //:protobuf and //:prometheus in BUILD both attempt to output the same file: pkg/linux_amd64/github.com/golang/protobuf/proto.a

```

peterebden:
Sounds about right... the `go_get` stuff is not beautiful since we have to fit both our file model and Go's one :(

I'm not wowed about making it more complex but it's better that it works, so LGTM. FYI I am plotting a rewrite / alternate rule to `go_get` so we don't have everything happening under one `go` command.

src/parse/rules/go_rules.build_defs
```diff
@@ -466,9 +481,19 @@ def go_get(name, get=None, outs=None, deps=None, exported_deps=None, visibility=
         cmd.append('(cd %s && git checkout -q %s)' % (subdir, revision))
     if patch:
         cmd.append('patch -s -d %s -p1 < ${TMP_DIR}/$(location %s)' % (subdir, patch))
-    cmd.append('go install -gcflags "-trimpath $TMP_DIR" ' + ' '.join([get] + (install or [])))
+    cmd.append('go install -gcflags "-trimpath $INSTALL_DIR" ' + ' '.join([get] + (install or [])))
```
dimpavloff: Could you confirm whether `-trimpath $INSTALL_DIR` is correct?

dimpavloff:
@pebers I've updated the PR following our chat, please take another look. I think this should work in all cases with go_get dependencies of another go_get spread across the repo but I'll test it a bit more before merging. Let me know if you think this has got too complex

src/parse/rules/go_rules.build_defs
peterebden: Yes, that does look correct to me - you should be trimming everything above the directory it was installed into, otherwise those directories turn up in stacktraces.

peterebden:
I am a bit worried about the complexity of it, but LGTM anyway - it's not supposed to require any specific install location.
Although the repo root is a bit fiddly for various reasons, I wouldn't be surprised if there were other rules that don't work there.

dimpavloff:
Hey, sorry I keep forgetting to test this on our repo before merging. Would you rather we don't merge it or did you close it because it was inactive?

peterebden:
Whoops sorry! I was cleaning up a bunch of old branches and forgot about this (or that it would close the PR).

dimpavloff:
Tested on our repo and all Go targets built fine. Merging

PR #195, opened 2017-04-04
Add fixed modification dates to files produced by jarcat

llucherini:
Currently the early access java9 build does not support zeroed modification dates


peterebden:
Looks good, thanks Lorenzo!

PR #203, opened 2017-04-20
Maven classifier

chpeer:
Add support for maven classifier to the build definition maven_jar

from https://maven.apache.org/pom.html
> classifier:
> The classifier allows to distinguish artifacts that were built from the same POM but differ in their content. It is some optional and arbitrary string that - if present - is appended to the artifact name just after the version number.

### DIFF
--- a/src/parse/rules/java_rules.build_defs
+++ b/src/parse/rules/java_rules.build_defs
@@ -379,7 +379,7 @@ def maven_jars(name, id, repository=None, exclude=None, hashes=None, combine=Fal
 def maven_jar(name, id=None, repository=None, hash=None, hashes=None, deps=None,
               visibility=None, filename=None, sources=True, licences=None,
               exclude_paths=None, native=False, artifact_type=None, test_only=False,
-              binary=False, classifier=""):
+              binary=False, classifier="", classifier_sources_override=""):
     """Fetches a single Java dependency from Maven.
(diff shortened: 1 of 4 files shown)


src/parse/rules/java_rules.build_defs
```diff
@@ -379,7 +379,7 @@ def maven_jars(name, id, repository=None, exclude=None, hashes=None, combine=Fal
 def maven_jar(name, id=None, repository=None, hash=None, hashes=None, deps=None,
               visibility=None, filename=None, sources=True, licences=None,
               exclude_paths=None, native=False, artifact_type=None, test_only=False,
-              binary=False):
+              binary=False, classifier="", classifier_sources_override=""):
```
peterebden: nit: these should be single quotes (just for consistency, the other string literals use ' when they're not triple-quoted)

test/java_rules/BUILD
```diff
@@ -12,3 +12,24 @@ java_test(
...
+
+
+
+# Test for maven classifier
+maven_jar(
+    name = 'logback-core-tests',
+    id = 'ch.qos.logback:logback-core:1.1.3',
+    classifier = 'tests',
+    classifier_sources_override = 'test',
+    hash = '',
```
peterebden: this should be set

PR #206, opened 2017-04-24
Add a script for auto-uploading new Please versions

peterebden:
So it's a bit less manual than me doing it by hand. This will imply that changing the VERSION file always publishes a new release immediately, but maybe that is good discipline to be in (we can mark them as beta etc to avoid it going too public).


src/misc/upload_release.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+# Script to automatically upload new Please versions.
+# Will not overwrite existing versions, only adds new ones.
+# Should be run from the root of the repo, and only by a CI system.
+
+set -eu
```
dimpavloff: nit: add `-o pipefail`

src/misc/upload_release.sh
```diff
@@ -0,0 +1,45 @@
...
+RED="$(tput setaf 1)"
+GREEN="$(tput setaf 2)"
+YELLOW="$(tput setaf 3)"
+RESET="$(tput sgr0)"
+BUCKET="s3://get.please.build"
+
+VERSION="`cat VERSION`"
```
dimpavloff: nit: replace backticks with `$()` here and further down

src/misc/upload_release.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
dimpavloff: shouldn't the script be outside the `src/` folder?

peterebden: Yeah I guess, I've moved it to `tools/misc` for now.

src/misc/upload_release.sh
peterebden: done

src/misc/upload_release.sh
peterebden: done

peterebden:
> The script atm does not prevent the above

It sort of does, it doesn't overwrite old releases, so it won't automatically add newer commits to one. It doesn't enforce it that much but then also doesn't integrate with Github releases, for example. I'm idly thinking that a followup to draft releases might be nice though...

PR #210, opened 2017-04-29
Implementing several new linter features

peterebden:
Checking for deprecated arguments, missing required arguments and duplication of third party packages.

Also tied it into `plz help`.

### DIFF
--- a/tools/linter/linter.py
+++ b/tools/linter/linter.py
@@ -34,6 +34,12 @@ current things searched for are:
  - Detection of duplicates in argument lists - most usually this is useful for deps,
    but it applies to anything since there's no reason to have a duplicate in any
    argument list to any builtin Please function.
+ - Some specific checks on third-party library rules (maven_jar, pip_library and
+   go_get) that warn on duplicated artifacts.
+   This is deliberately version-agnostic, since in most cases having two different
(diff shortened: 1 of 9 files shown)


tools/linter/linter.py
```diff
@@ -122,10 +162,16 @@ def _lint_builtin_functions(n):
...
         if n.args or n.starargs:
             yield n.lineno, NON_KEYWORD_CALL
-        args = {arg['name'] for arg in BUILTIN_FUNCTIONS[n.func.id]['args']}
+        args = BUILTIN_FUNCTIONS[n.func.id]['args']
         for kwd in n.keywords or []:
             if kwd.arg not in args and not kwd.arg.startswith('_'):
                 yield kwd.value.lineno, INCORRECT_ARGUMENT
+        # Don't check kwargs if the caller is doing a **kwargs into it, assume that'll take care of it.
+        if not n.kwargs:
+            kwds = {kwd.arg for kwd in n.keywords or []}
```
dimpavloff: I don't think you need the `or []` ?

peterebden: if `n.keywords` is `None` then it's not iterable and we'd get an exception.

(Note that the if statement above is `kwargs` not `keywords`...)

PR #212, opened 2017-05-02
Enhancement on pkg_config_libs in cc_rules

sfirmery:
- Add `pkg_config_libs` from transitive labels to `cc_library` build commands
- Add config option to set PKG_CONFIG_PATH


src/core/config.go
```diff
@@ -147,6 +147,7 @@ func DefaultConfiguration() *Configuration {
 	config.Cpp.DefaultDbgCflags = "--std=c99 -g3 -pipe -DDEBUG -Wall -Werror"
 	config.Cpp.DefaultOptCppflags = "--std=c++11 -O3 -pipe -DNDEBUG -Wall -Werror"
 	config.Cpp.DefaultDbgCppflags = "--std=c++11 -g3 -pipe -DDEBUG -Wall -Werror"
+	config.Cpp.PkgConfigPath = ""
```
peterebden: FYI this probably isn't necessary, it'll default to an empty string anyway.

sfirmery: It works well without it. Thanks.

PR #214, opened 2017-05-04
Add a patch to make PEX-INFO files deterministic

peterebden:
i.e. sort their keys so they come out consistently.

I'm sending this upstream too (https://github.com/pantsbuild/pex/pull/384) but don't want to wait for that to merge & release before testing this ourselves.


peterebden:
Yeah I wonder... there's a set that runs in CI which basically asserts that we don't rebuild things, but that won't catch a thing like this. It'd be a lot slower but maybe we could force a rebuild of everything and assert that all the hashes match as you say.

PR #222, opened 2017-05-18
fix rpc_server which can't load certificates from a directory

sfirmery:


peterebden:
Whoops... Thanks for the fix!

PR #223, opened 2017-05-24
Creating tarballs in jarcat

### LINKED ISSUE #219: tarball altering file mtimes can be observed by other rules
In a heroic effort to make the output of `tarball` rules deterministic, Dimitar had it alter the mtimes of its sources. However because sources are hardlinked into the tmp directory this change can be observed by other rules - notably `sh_binary` sometimes complains about "file changed as we read it".

Consider extending `jarcat` to do tar for us as well so we have precise control over the output file; Go has support in its stdlib and it would avoid a lot of the cross-platform issues we've struggled with as well.

peterebden:
Means we can write our own fixated mtimes without having to struggle with OS-specific quirks or modify incoming files.

Is a breaking change though because it now only does gzip compression. We could pull in some third party packages to get bzip2 / xz writers but I'm not sure it's worth it.

Fixes #219

### DIFF
--- a/src/parse/rules/misc_rules.build_defs
+++ b/src/parse/rules/misc_rules.build_defs
@@ -509,8 +509,8 @@ def _deb_deps(cmd, current_label):
     return _find_deb_deps
 
 
-def tarball(name, srcs, out=None, deps=None, subdir=None,
-            compression='gzip', visibility=None, labels=None):
+def tarball(name, srcs, out=None, deps=None, subdir=None, gzip=True,
+            visibility=None, labels=None, compression=None):
     """Defines a rule to create a tarball containing outputs of other rules.
 
     File mode and ownership are preserved. However, the atime and mtime of all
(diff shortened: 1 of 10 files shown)


tools/jarcat/tar/tar_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"io"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const testDataDir = "tools/jarcat/tar/test_data"
```
dimpavloff: could the outputs be written in-memory instead, e.g with `bytes.Buffer` ? Or is that an issue because of filepath.Walk? If so, you could use `ioutil.Temp*`, which would already be in-memory for modern OSes

tools/jarcat/tar/tar_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+func TestNoCompression(t *testing.T) {
+	filename := "test_no_compression.tar"
+	err := Write(filename, testDataDir, "", false)
```
dimpavloff: should this and the other unit tests defer-call a best-effort deletion of the output file?

tools/jarcat/tar/tar_test.go
```diff
@@ -0,0 +1,105 @@
...
+	m := map[*tar.Header]string{}
+	for {
+		hdr, err := tr.Next()
+		if err == io.EOF {
+			break
+		}
+		assert.NoError(t, err)
+		var buf bytes.Buffer
+		_, err = io.Copy(&buf, tr)
+		assert.NoError(t, err)
```
dimpavloff: nit: these `assert`s should probably be `require` -- when they break, there's going to be less noise that way

src/parse/rules/misc_rules.build_defs
```diff
@@ -520,65 +520,39 @@ def tarball(name, srcs, out=None, deps=None, subdir=None,
       name (str): Rule name
       srcs (list): Source files to include in the tarball
       out (str): Name of output tarball (defaults to `name`.tar.gz, but see below re compression)
-      subdir (str): Subdirectory to create in (defaults to 'name')
-      compression (str): Kind of compression to use. Either one of {gzip, bzip2, xz, lzma}
-                         to filter through known tar methods, an explicit flag, or None for
-                         no compression.
+      subdir (str): Subdirectory to create in.
```
dimpavloff: might be worth documenting the flattening part here

tools/jarcat/jarcat.go
```diff
@@ -128,6 +129,10 @@ var opts = struct {
 	NoDirEntries            bool              `short:"n" long:"nodir_entries" description:"Don't add directory entries to zip"`
 	RenameDirs              map[string]string `short:"r" long:"rename_dir" description:"Rename directories within zip file"`
 	StripBytecodeTimestamps bool              `short:"b" long:"strip_bytecode_timestamps" description:"Strips timestamps from any .pyc / .pyo files encountered."`
+
+	Tar    bool   `long:"tar" description:"Write a tarball instead of a zipfile. Note that most other flags are not honoured if this is given."`
+	Gzip   bool   `short:"z" long:"gzip" description:"Apply gzip compression to the tar file. Only has an effect if --tar is passed."`
+	Prefix string `long:"prefix" description:"Flatten all files into a directory with this name within the tarball."`
```
dimpavloff: what's the flattening use case? wouldn't it be more common/expected to only prefix the usual tree with the given prefix? You can't run into file collisions then, too

tools/jarcat/tar/tar_test.go
peterebden: Yup, it would be an issue because of `filepath.Walk`. 

This path relates to the input test data not the temporary test files; I'd prefer to keep the outputs in the test data dir rather them in /tmp though so plz will clean them up afterwards.
Are you sure about it being in-memory? `ioutil.Temp*` calls `os.TempDir` which returns `/tmp` for me, that's not in-memory on Ubuntu AFAICT?

tools/jarcat/tar/tar_test.go
peterebden: plz cleans it up after the test exits

tools/jarcat/jarcat.go
peterebden: We do this at the moment, no? For Docker rules among other things?

tools/jarcat/tar/tar_test.go
dimpavloff: On ubuntu 16+ `/tmp` should be mounted as `tmpfs` by default; this is also the case for the Fedora-family and ArchLinux but I don't know about others. 
I forgot that plz would do this for you, so it's fine as is;

tools/jarcat/tar/tar_test.go
peterebden: huh, didn't know about that actually. thx!

src/parse/rules/misc_rules.build_defs
```diff
@@ -520,65 +520,39 @@ def tarball(name, srcs, out=None, deps=None, subdir=None,
...
+      gzip (bool): If True, the output will be gzipped. If False, it will just be a tarball.
       deps (list): Dependencies
       visibility (list): Visibility specification.
       labels (list): Labels associated with this rule.
+      compression (str): Deprecated. No longer has any effect.
```
peterebden: OK, removed.

src/parse/rules/misc_rules.build_defs
```diff
@@ -520,65 +520,39 @@ def tarball(name, srcs, out=None, deps=None, subdir=None,
...
-
-    out = out or name + '.tar' + extension
-    # Unlike the rest of the compressions, gzip preserves the file's (the tar)
-    # timestamp by default, making the output unreproducible.
-    # Hence the compression is done with the -n flag in a separate step
-    if extension == '.gz' or out.endswith('.gz'):
-        archive_cmd = 'tar -c * | gzip -n > $OUT'
-    else:
-        archive_cmd = 'tar %s -cf $OUT *' % compression
+    out = out or name + ('.tar.gz' if compress else '.tar')
```
peterebden: yup, good spot thanks

peterebden:
> IMO let's tag it as a major version release because of the breaking change.
Was planning to :) About time we had 8.0 too, the minor version number is getting too big...

PR #224, opened 2017-05-24
Support for not specifying Java test package in config.

peterebden:
Instead we default to anything named `*Test.java`. Turned out not to be that hard since we've already got a list of the files we compiled in the .jar in case we need them for coverage.

### DIFF
--- a/tools/junit_runner/src/build/please/test/TestMain.java
+++ b/tools/junit_runner/src/build/please/test/TestMain.java
@@ -43,16 +43,14 @@ public class TestMain {
 
   public static void main(String[] args) throws Exception {
     String testPackage = System.getProperty("build.please.testpackage");
-    if (testPackage == null || testPackage.equals("")) {
-      throw new RuntimeException("Test package not provided (define with -Dbuild.please.testpackage)");
-    }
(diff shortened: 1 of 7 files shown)


tools/junit_runner/src/build/please/test/TestMain.java
```diff
@@ -77,6 +75,26 @@ public static void main(String[] args) throws Exception {
...
   }
 
+  /**
+   * Loads all the available test classes.
+   * This is a little complex because we want to try to avoid scanning every single class on our classpath.
+   * @param testPackage the test package to load from. If empty we'll look for them by filename.
+   */
+  private static Set<Class> findClasses(String testPackage) throws Exception {
+    ClassLoader loader = Thread.currentThread().getContextClassLoader();
+    if (testPackage != null && !testPackage.equals("")) {
```
dimpavloff: nit: `testPackage.equals("")` could be `testPackage.isEmpty()`

peterebden: I was proud that I at least remembered not to do `== ""` this time... :(

PR #225, opened 2017-05-26
Support for running multiple targets at once

peterebden:
Either sequentially or in parallel.

Required a little hacking of the flags library; it's a bit nonobvious how it should handle optional subcommands versus positional arguments. Would send them a PR but the fix for this breaks some other tests in their repo (not things we care about though).

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -133,6 +133,11 @@ var opts struct {
 				Targets []core.BuildLabel `positional-arg-name:"target" description:"Targets to run"`
 			} `positional-args:"true" required:"true"`
 		} `command:"parallel" description:"Runs a sequence of targets in parallel"`
+		Sequential struct {
+			Args struct {
+				Targets []core.BuildLabel `positional-arg-name:"target" description:"Targets to run"`
+			} `positional-args:"true" required:"true"`
+		} `command:"sequential" description:"Runs a sequence of targets sequentially."`
 		Args struct {
 			Target core.BuildLabel `positional-arg-name:"target" required:"true" description:"Target to run"`
 			Args   []string        `positional-arg-name:"arguments" description:"Arguments to pass to target when running (to pass flags to the target, put -- before them)"`
@@ -321,6 +326,12 @@ var buildFunctions = map[string]func() bool{
 	"parallel": func() bool {
 		if success, state := runBuild(opts.Run.Parallel.Args.Targets, true, false); success {
 			run.Parallel(state.Graph, opts.Run.Parallel.Args.Targets)
+		}
+		return true // We do return from run.Parallel once it's all over.
+	},
+	"sequential": func() bool {
+		if success, state := runBuild(opts.Run.Sequential.Args.Targets, true, false); success {
+			run.Sequential(state.Graph, opts.Run.Sequential.Args.Targets)
 		}
 		return true // We do return from run.Parallel once it's all over.
 	},
--- a/src/please.go
+++ b/src/please.go
@@ -327,15 +327,15 @@ var buildFunctions = map[string]func() bool{
 	},
 	"parallel": func() bool {
 		if success, state := runBuild(opts.Run.Parallel.PositionalArgs.Targets, true, false); success {
-			run.Parallel(state.Graph, opts.Run.Parallel.PositionalArgs.Targets, opts.Run.Parallel.Args)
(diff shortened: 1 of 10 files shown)


src/please.go
```diff
@@ -313,6 +325,18 @@ var buildFunctions = map[string]func() bool{
...
 	},
+	"parallel": func() bool {
+		if success, state := runBuild(opts.Run.Parallel.PositionalArgs.Targets, true, false); success {
+			run.Parallel(state.Graph, opts.Run.Parallel.PositionalArgs.Targets, opts.Run.Parallel.Args)
+		}
+		return true // We do return from run.Parallel once it's all over.
+	},
+	"sequential": func() bool {
+		if success, state := runBuild(opts.Run.Sequential.PositionalArgs.Targets, true, false); success {
+			run.Sequential(state.Graph, opts.Run.Sequential.PositionalArgs.Targets, opts.Run.Sequential.Args)
```
dimpavloff: same here, should run.Sequential return true or false?

src/run/run_step.go
```diff
@@ -20,6 +21,37 @@ var log = logging.MustGetLogger("run")
...
+func Parallel(graph *core.BuildGraph, labels []core.BuildLabel, args []string) {
+	var g errgroup.Group
+	for _, label := range labels {
+		label := label // capture locally
+		g.Go(func() error {
+			return run(graph, label, args, true).Wait()
+		})
+	}
+	if err := g.Wait(); err != nil {
+		log.Fatalf("Command failed: %s", err)
```
dimpavloff: would this log which of the commands failed?

src/run/run_step.go
```diff
@@ -40,7 +72,21 @@ func Run(graph *core.BuildGraph, label core.BuildLabel, args []string) {
...
 	log.Info("Running target %s...", strings.Join(args, " "))
 	output.SetWindowTitle("plz run: " + strings.Join(args, " "))
-	if err := syscall.Exec(splitCmd[0], args, os.Environ()); err != nil {
-		log.Fatalf("Error running command %s: %s", strings.Join(args, " "), err)
+	if fork {
+		cmd := exec.Command(splitCmd[0], args[1:]...) // args here don't include argv[0]
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		// Note that we don't connect stdin. It doesn't make sense for multiple processes.
+		must(cmd.Start(), args)
```
dimpavloff: same here, consider propagating the exit code

src/run/run_step.go
```diff
@@ -20,6 +21,37 @@ var log = logging.MustGetLogger("run")
...
+	if err := g.Wait(); err != nil {
+		log.Fatalf("Command failed: %s", err)
+	}
+}
+
+// Sequential runs a series of targets sequentially.
+func Sequential(graph *core.BuildGraph, labels []core.BuildLabel, args []string) {
+	for _, label := range labels {
+		log.Notice("Running %s", label)
+		cmd := run(graph, label, args, true)
```
dimpavloff: might be nicer to not panic and propagate the exit code instead

src/please.go
```diff
@@ -313,6 +325,18 @@ var buildFunctions = map[string]func() bool{
 		}
 		return false // We should never return from run.Run so if we make it here something's wrong.
 	},
+	"parallel": func() bool {
+		if success, state := runBuild(opts.Run.Parallel.PositionalArgs.Targets, true, false); success {
+			run.Parallel(state.Graph, opts.Run.Parallel.PositionalArgs.Targets, opts.Run.Parallel.Args)
+		}
+		return true // We do return from run.Parallel once it's all over.
```
peterebden: Yeah sorry did some updates, ended up needing to pass arguments to commands and refactored some other bits at the same time.

Whether these things return is all a bit arbitrary. Originally my thought was that this was easier if it just died since on success `Run` doesn't ever return, so it's kind of weird to return a bool that's only really got one possible value.

src/run/run_step.go
peterebden: nope, we have lost it at this point. I see `plz run parallel` as pretty fire-and-forget though...

src/please.go
dimpavloff: True, makes sense to just exec into the command.

peterebden:
Okay, have rewritten somewhat to propagate the exit codes, PTAL.

PR #226, opened 2017-05-30
Secrets

peterebden:
As (briefly) discussed last week, seems useful to be able to track "secrets" that are used in a build but not copied about the place.

The behaviour should be that:
 - They're required to build a target that uses them
 - But you can retrieve a target from the remote cache without knowing them
 - Changing them locally forces a rebuild

Hopefully this will make it easier to track things like AWS credentials and so forth (possibly .apk signing keys too, although got a bit more thinking to do about how we make that practical).

### DIFF
--- a/.plzconfig
+++ b/.plzconfig
@@ -39,7 +39,6 @@ javacworker = //tools/javac_worker
 jarcattool = //tools/jarcat
 junitrunner = //tools/junit_runner
 pleasemaventool = //tools/please_maven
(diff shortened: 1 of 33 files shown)


peterebden:
Yeah having names normally ends up being more flexible, although since they're all absolute paths it's probably not so hard to just chuck them in the command. In several cases (e.g. AWS credentials) they're used implicitly anyway tho...

PR #230, opened 2017-07-04
Implement cleaning entire remote cache on plz clean

peterebden:
It's gated by a `--remote` flag since it seems a little expensive to be doing by default.

Refactored things a bit so the logic in `clean` is limited to cleaning outputs, and cache cleaning is all done by the implementations in `cache`.
Fortunately the server side actually had this logic already, we just weren't calling it...

Should make the remote cache a little easier to administer.

### DIFF
--- a/src/cache/rpc_cache.go
+++ b/src/cache/rpc_cache.go
@@ -233,7 +233,11 @@ func (cache *rpcCache) Clean(target *core.BuildTarget) {
 }
 
 func (cache *rpcCache) CleanAll() {
-	if cache.isConnected() && cache.Writeable {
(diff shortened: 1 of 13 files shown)


src/cache/rpc_cache.go
```diff
@@ -231,6 +231,20 @@ func (cache *rpcCache) Clean(target *core.BuildTarget) {
 		})
 	}
 }
+
+func (cache *rpcCache) CleanAll() {
+	if cache.isConnected() && cache.Writeable {
```
fs75: should this log an error message if it's not writeable?

PR #233, opened 2017-08-14
Lightweight test sandboxing

peterebden:
Introduces a small binary that can do test sandboxing in place of Docker, i.e. network / IPC segregation without a full-blown container & new filesystem, akin to Chrome's sandbox. It's not too complex since it's just using namespaces & cgroups.

Anecdotally this seems to help responsiveness of my machine under load of a lot of tests with `container = True`. I'm not sure that things are necessary faster, but it should also make it easier to analyse where time is going since containerised tests won't all be hidden behind a Docker client/server any more.

Also contains a small tweak to allow passing a list as the `cmd` to `genrule`, which is equivalent to `'&&'.join(cmd)` but slightly nicer to write.


src/core/config.go
```diff
@@ -218,8 +225,9 @@ type Configuration struct {
...
 	Test               struct {
-		Timeout          cli.Duration            `help:"Default timeout applied to all tests. Can be overridden on a per-rule basis."`
-		DefaultContainer ContainerImplementation `help:"Sets the default type of containerisation to use for tests that are given container = True.\nCurrently the only option is 'docker' but we intend to add rkt support at some point."`
+		Timeout           cli.Duration `help:"Default timeout applied to all tests. Can be overridden on a per-rule basis."`
+		DefaultContainer  string       `help:"Sets the default type of containerisation to use for tests that are given container = True.\nCurrently the available options are 'docker' and 'plz', the latter is an experimental custom sandbox that currently only restricts the process and network namespaces whereas the former is a full-blown container engine. We expect to add support for more engines in future."`
```
csdigi: #rkt

src/test/container.go
```diff
@@ -117,3 +119,19 @@ func retrieveFile(target *core.BuildTarget, cid []byte, filename string, warn bo
 		}
 	}
 }
+
+// runPlzSandboxedTest runs a test using please_sandbox, which simply restricts network & process
+// namespaces. It's only available on Linux and must be installed separately, since it has to be
+// suid root in order to use CLONE_NEWNET.
+func runPlzSandboxedTest(state *core.BuildState, target *core.BuildTarget) ([]byte, error) {
```
dimpavloff: As discussed, ideally it would be good to be able to use the rest of the namespaces, especially the PID, UID, and filesystem to make the builds even more hermetic. Regarding the FS namespace, that would just obfuscate mounts so it shouldn't cause issues (i.e I think both of us were thinking of this + chroot earlier). I don't know if we'll need to let the user control these via the build rule, hopefully there aren't such use cases

src/core/build_env.go
```diff
@@ -54,8 +54,10 @@ func BuildEnvironment(state *BuildState, target *BuildTarget, test bool) []strin
 	env := GeneralBuildEnvironment(state.Config)
 	env = append(env, "PKG="+target.Label.PackageName, "PKG_DIR="+target.Label.PackageDir())
 	if !test {
+		tmpDir := path.Join(RepoRoot, target.TmpDir())
 		env = append(env,
-			"TMP_DIR="+path.Join(RepoRoot, target.TmpDir()),
+			"TMP_DIR="+tmpDir,
+			"TMPDIR="+tmpDir,
```
peterebden: TMPDIR is the POSIX version that things honour for making temp files etc. Since we're not running in a container filesystem any more they need to not all use /tmp otherwise tests can fight with one another.

Indeed the ones that did fail were mostly Java, but that's just coincidence really.

src/core/config.go
```diff
@@ -34,8 +34,11 @@ const LocalConfigFileName string = ".plzconfig.local"
 // for a particular machine (eg. build machine with different caching behaviour).
 const MachineConfigFileName = "/etc/plzconfig"
 
-const TestContainerDocker = "docker"
-const TestContainerNone = "none"
+const (
+	ContainerImplementationNone   = "none"
+	ContainerImplementationDocker = "docker"
+	ContainerImplementationPlz    = "plz"
+)
```
peterebden: Yeah so I'm starting to think this is structured wrong; rather than using `container` for this maybe there should be a separate property called `sandbox` or something that uses this, with `container` effectively overriding it to use a full-blown container implementation.

src/test/container.go
peterebden: This comment is actually slightly out of date now (because it also restricts PID, IPC and UTS as well, for example).
Can look at the options around FS; I think we can also increase the level of sandboxing over time.
Unsure about UID. Docker caused many issues with that by creating files that were owned by root and couldn't be subsequently deleted. Would make sense, but only if that can be avoided.

peterebden:
Okay, have re-engineered a bit; it's a separate concept to `container` now and can be applied to build actions too. Individual rules can opt in/out where needed.

src/core/config.go
```diff
@@ -187,10 +193,12 @@ type Configuration struct {
...
-		FallbackConfig string       `help:"The build config to use when one is chosen and a required target does not have one by the same name. Also defaults to opt." example:"opt | dbg"`
+		Timeout           cli.Duration `help:"Default timeout for Dockerised tests, in seconds. Default is twenty minutes."`
+		Path              []string     `help:"The PATH variable that will be passed to the build processes.\nDefaults to /usr/local/bin:/usr/bin:/bin but of course can be modified if you need to get binaries from other locations." example:"/usr/local/bin:/usr/bin:/bin"`
+		Config            string       `help:"The build config to use when one is not chosen on the command line. Defaults to opt." example:"opt | dbg"`
+		FallbackConfig    string       `help:"The build config to use when one is chosen and a required target does not have one by the same name. Also defaults to opt." example:"opt | dbg"`
+		Sandbox           bool         `help:"True to sandbox individual build actions, which isolates them using cgroups. Somewhat experimental, only works on Linux and requires please_sandbox to be installed separately."`
```
dimpavloff: nit: `using namespaces`

peterebden: Thanks

https://s-media-cache-ak0.pinimg.com/736x/9f/84/3b/9f843bf305effdf18d6a187079b5eff8--computers-funny-stuff.jpg

PR #237, opened 2017-10-06
Fix Go builds with no deps.

alamaison:
If there are no deps, grep returns exit code 1, which we need to ignore.


alamaison:
@peterebden

peterebden:
LGTM, thanks Alex!

PR #238, opened 2017-10-09
Adding support for streaming build events to a remote client.

peterebden:
Builds can be configured to start a gRPC server, clients can connect to that and stream events back from it to get a facsimile of the normal live console plus some basic system stats.

Sorry for largeish PR; needed most of this stuff to get it working enough to actually try it.

Command is now `plz follow` at Will's suggestion


src/output/shell_output.go
```diff
@@ -110,6 +108,47 @@ func MonitorState(state *core.BuildState, numThreads int, plainOutput, keepGoing
...
+func PrintDisconnectionMessage(success, closed, disconnected bool) {
+	printf("${BOLD_WHITE}Disconnected from remote plz server.\nStatus: ")
+	if disconnected {
+		printf("${BOLD_YELLOW}Disconnected${RESET}\n")
+	} else if !closed {
+		printf("${BOLD_MAGENTA}Unknown${RESET}\n")
+	} else if success {
+		printf("${BOLD_GREEN}Success${RESET}\n")
+	} else {
+		printf("${BOLD_RED}Success${RESET}\n")
```
csdigi: Do we want the success message here?

peterebden: Thanks Will, good spot!

PR #242, opened 2017-11-06
Rewriting please_pex to not involve pex any more

peterebden:
This is really for optimisation; all the pex gubbins at startup adds 0.5-1 second of latency to every Python test, which makes them a bit disappointing for writing many small tests. Quite a lot of that time is pkg_resources which imports every imaginable piece of Python infrastructure, including `calendar` and `email.parser` 🙄 - unfortunately pex requires pkg_resources so that's not really negotiable :(

On the other hand we use very little of its functionality anyway and the useful bits (profiling and non-zip-safety) are not hard to write. Removing pex and doing those bits ourselves reduces the startup overhead of a test (or indeed any pex) to < 200ms or so which seems much more reasonable.

This is a breaking change, primarily for anything that assumed pkg_resources was importable; we didn't necessarily formally guarantee that but we use it often enough ourselves that it seems like it was assumed.
Anything that tried `import pex_main` or `import test_main` will be similarly disappointed, although that seems very much like an implementation detail that people really shouldn't have been touching.
I'm keeping the pex nomenclature since I'm sure many things will depend on the files being named .pex.

There are a few related cleanups and whatnot, feel free to ask if the rationale behind some of it is unclear.


agenticarus:
Interesting.

This seems like a decent approach, and I'm impressed by the speedups available. I don't think any users of this ought to have been depending on the implicit import of pkg_resources or pex_main/test_main so I'm happy for this to break anyone who was.

Do you think leaving the name 'pex' in the code will be confusing now we're effectively making a pex-lite file? Or is it sufficiently similar that it's not worth renaming?

peterebden:
I think it mimics things to a sufficient degree that it's reasonable to keep the name. I want to keep it in the filename anyway since I think it is reasonable for users to expect those to remain consistent, so there'll be some instances of it around still.

Plus we've got to call them *something* :-P

agenticarus:
Fair. Shipit.

PR #244, opened 2017-11-14
Add dynamic support for Python wheel naming schemes

natpen:
Hi! I've been checking out please the past day or two and it's super cool! I was hoping to get please talking to our internal wheel repository, but our naming format is slightly different than what please expects. This commit adds the ability for users to specify a custom wheel naming scheme via the `name_scheme` argument to python_wheel. 

Additionally, in the absence of a custom name_scheme, this commit adds search precedence to wheels adhering to the PEP wheel naming standard (as per [PEP 427](https://www.python.org/dev/peps/pep-0427/#file-name-convention) and [PEP 491](https://www.python.org/dev/peps/pep-0491/#file-name-convention)).

Please let me know if you have other ideas! :)


natpen:
Oh, interesting... the `ci/circleci: build-linux` is failing the same way my please was when trying to build against master, even before introducing my changes.

natpen:
Oh! I have the environment working now. I'll see if I can figure out what I broke.

peterebden:
Ah yeah that test is a bit flaky, I think; not your fault...

peterebden:
PR looks good, thanks! Before I merge, it seems like it might get tedious to add that format to a bunch of build rules - what would you think about adding it as a config parameter as well?

natpen:
Sounds good to me, I hope this last commit handles everything needed for that!

I made the precedence be build rule settings, then config settings, then the defaults.

LMK what you think :)

src/parse/rules/python_rules.build_defs
```diff
@@ -404,12 +409,32 @@ def python_wheel(name, version, hashes=None, package_name=None, outs=None,
...
+                                       version=version))
+    elif CONFIG.PYTHON_WHEEL_NAME_SCHEME:
+        urls.append(CONFIG.PYTHON_WHEEL_NAME_SCHEME.format(url_base=url_base,
+                                                           package_name=package_name,
+                                                           version=version))
+    else:
+        # Populate urls using a reasonable set of possible wheel naming schemes.
+        # Look for an arch-specific wheel first; in some cases there can be both (e.g. protobuf
+        # has optional arch-specific bits) and we prefer the one with the cool stuff.
+        urls.append('{url_base}/{package_name}-{version}-${{OS}}-${{ARCH}}.whl'.format(url_base=url_base,
```
peterebden: Do you still need this version as well as the one beneath?

natpen: Maybe not! I can definitely customize for our wheel naming schema using the functionality in this PR, I just thought it might make sense to leave this as one of the default schemas since it's the standard according to [wheel's](https://pypi.python.org/pypi/wheel) documentation.

It might have a speed cost for your use-case though, if it has to check that format first, not find it, then fallback to the thought-machine schema.

So... I'm okay with whatever you prefer, since it's firstly a thought-machine project :)

peterebden: Yeah, I was thinking of a speed cost but it's not very significant - by the design of the system we don't have to re-request those packages very often. Makes sense to leave as is.

PR #245, opened 2017-11-15
Ignore go-bindata-generated parser files for Python 3

natpen:
This PR adds python 3-specific lines to .gitignore that match the pre-existing python2 and pypy lines.


peterebden:
Thanks! (FYI those are generated by cffi rather than go-bindata - but same effect)

PR #249, opened 2017-11-25
Support for specifying absolute Go import paths

### LINKED ISSUE #247: golang: how to use  please with absolute golang import paths?
We are following in our Go applications the common approach to use absolute import paths, starting with a domainname.

For example we have in our git repository root the files:
`go/code/lib/log/log.go`
`go/code/myapp/cmd/myapp/main.go`

In the `main.go` file of myapp we import `example.com/go/code/lib/log`.
To make it build with `go build`, we check out the git repository into a directory called `example.com` and set the `$GOPATH` to the parent directory.

If I build myapp with please, it can't find the import `example.com/go/code/lib/log` because in please's temporarily build directory the package is located in `go/code/lib/log/` instead of in `example.com/go/code/lib/log`.

What is the common approach to solve it with please?
Is there a better way than putting the `.plzconfig` file into the parent directory of the checked out repository?

Bazel has an `importpath` directive for golang packages to specify an importpath that differs from the file location. 

**Update:**
With our layout Pyhthon builds are also broken.
`example.com` is used as part of the import path and python uses `.` as separator for directories in the import path.

```
|- .plzconfig
|- example.com/ (root of the cloned git repository)
    | - python/
        |- test/
            |- main.go
            |- BUILD
    |- go/
        | - vendor/
            |- BUILD
        |- code/
            |- lib/
                |- BUILD
                |- log/
                    |- log.go
            |- myapp/
                |- BUILD
                |- cmd/
                    |- myapp/
                        | cmd.go
```

peterebden:
Resolves #247

### DIFF
--- a/src/parse/rules/go_rules.build_defs
+++ b/src/parse/rules/go_rules.build_defs
@@ -637,11 +637,12 @@ def _go_library_cmds(complete=True, all_srcs=False):
     compile_cmd = '$TOOL tool %s -trimpath $TMP_DIR %s%s -pack -o $OUT ' % (go_compile_tool, complete_flag, _GOPATH)
     # Annotates files for coverage
     cover_cmd = 'for SRC in $SRCS; do mv -f $SRC _tmp.go; BN=$(basename $SRC); go tool cover -mode=set -var=GoCover_${BN//./_} _tmp.go > $SRC; done'
-    srcs = 'export SRCS="$PKG_DIR/*.go"; ' if all_srcs else ''
+    prefix = ('export SRCS="$PKG_DIR/*.go"; ' + _LINK_PKGS_CMD) if all_srcs else _LINK_PKGS_CMD
+    prefix += _go_import_path_cmd(CONFIG.GO_IMPORT_PATH)
     return {
-        'dbg': '%s%s; %s -N -l $SRCS' % (srcs, _LINK_PKGS_CMD, compile_cmd),
-        'opt': '%s%s; %s $SRCS' % (srcs, _LINK_PKGS_CMD, compile_cmd),
-        'cover': '%s%s; %s && %s $SRCS' % (srcs, _LINK_PKGS_CMD, cover_cmd, compile_cmd),
+        'dbg': '%s; %s -N -l $SRCS' % (prefix, compile_cmd),
+        'opt': '%s; %s $SRCS' % (prefix, compile_cmd),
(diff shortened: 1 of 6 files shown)


src/parse/rules/go_rules.build_defs
```diff
@@ -658,14 +660,24 @@ def _go_binary_cmds(static=False, ldflags=''):
...
         'go': [CONFIG.GO_TOOL],
         'ld': [CONFIG.LD_TOOL if CONFIG.LINK_WITH_LD_TOOL else CONFIG.CC_TOOL],
     }
 
 
+def _go_import_path_cmd(import_path):
+    """Returns a partial command which is used for setting up the Go import path."""
+    if not import_path:
+        return ''
+    elif '/' in import_path:
```
dimpavloff: do you need to guard against an import path starting with `/` ?

test/go_rules/importpath/importpath.go
```diff
@@ -0,0 +1,5 @@
+// Package importpath is used to implement a test on Go import paths.
+package importpath
+
+// Answer is the answer to Life, the Universe and Everything.
```
dimpavloff: ultra nit: missing Oxford comma :P

test/go_rules/importpath/importpath_test.go
```diff
@@ -0,0 +1,14 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/thought-machine/please/test/go_rules/importpath"
+)
+
+func TestTheAnswer(t *testing.T) {
+	// This is a pretty silly test, but it's really about whether this compiles or not.
+	assert.Equal(t, 42, importpath.Answer)
+}
```
dimpavloff: nice one adding a test for this, +1

src/parse/rules/go_rules.build_defs
peterebden: Probably should, yeah; it wouldn't make sense as a Go import path but someone might do it by accident.

test/go_rules/importpath/importpath.go
peterebden: Douglas Adams was a Cambridge graduate ;-)
https://en.wikipedia.org/wiki/Life,_the_Universe_and_Everything

peterebden:
Thanks for the fast review!

PR #251, opened 2017-12-03
python rules: switch test runner from unittest to pytest

uwedeportivo:
This is WIP. I'm having trouble even going past declaring pytest a python wheel dependency. So none of this works. I'm sending this to get comments and help. Obviously I would prefer if you guys take over and just support pytest :-).

But this is sort of what I was thinking for switching test runner to pytest. pytest can run unittest TestCases natively and it can output junit-like xml reports.

### DIFF
--- a/third_party/python/BUILD
+++ b/third_party/python/BUILD
@@ -13,10 +13,10 @@ REPOS = [
 ]
 
 python_wheel(
-    name = 'xmlrunner',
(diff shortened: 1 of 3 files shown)


peterebden:
Thanks for the PR! I've been working on this too - #252 makes it possible to swap to pytest as the test runner. Feel free to leave comments on that!

Issue #246, opened 2017-11-17
Support interactive debugging of Python test failures


natpen:
You can do this with some other build systems, [like Pants](https://www.pantsbuild.org/python_readme.html#interactive-debugging-on-test-failure) - upon a test failure, it will drop you into an interactive debugger (looks like pdb in the Pants example above).

I think it's a very nice feature, so I started noodling on would be necessary to support that workflow with Please (I'm by no means a Python expert, but I couldn't figure out a way to get comparable behavior out of Please+unittest).

You're already accepting flags, so my first thought is that switching the test runner to [pytest](https://docs.pytest.org/en/latest/) would give access to the `--pdb` flag out of the box. I'm happy to continue messing around with it, but I wanted to see if this was something you considered before, and what your thoughts are.

peterebden:
I've thought about it a little - some people here have asked for it. I agree it'd be great to have but it did seem tricky...

Off the top of my head, the tests would need to have stdin connected, and stdout/stderr would need to get redirected out to the terminal. Things could get a bit weird if there's more than one test that tries to drop into the debugger at the same time, but I guess you could just do it for the first failure and ignore the others.

Sounds like a great idea though - let me know if you have any questions about it all!

PR #252, opened 2017-12-03
Optional pytest runner for python_test

peterebden:
Allows swapping out unittest in favour of pytest. I'm not making pytest the default or only option since it's somewhat slower (imports a lot more, including pkg_resources, which we've been trying to cut down on) but it's more flexible & powerful than unittest.

First step towards #246 (but does not solve it completely).

Interesting side-feature: pytest is only happy when it has the test sources as data files. Turns out that is somewhat more efficient in some cases (when the test files are changed we don't have to rebuild the entire pex) so I've adopted that for unittest as well.


src/parse/rules/python_rules.build_defs
```diff
@@ -181,7 +181,8 @@ def python_test(name, srcs, data=None, resources=None, deps=None, labels=None, s
     """Generates a Python test target.
 
     This works very similarly to python_binary; it is also a single .pex file
-    which is run to execute the tests. The tests are run via unittest.
+    which is run to execute the tests. The tests are run via either unittest or pytest, depending
+    on which is set for the python.testrunner property in .plzconfig.
```
dimpavloff: nit: imo delete  `in .plzconfig.` because you can override it for a given folder which is likely to be a common use case as one starts writing new tests in pytest while still using unittest ones elsewhere or vice-versa

test/python_rules/pytest/inc.py
```diff
@@ -0,0 +1,3 @@
+def inc(x):
+    """boy this is one profound function"""
+    return x + 1
```
dimpavloff: I expect a separate PR with leftpad!

src/parse/rules/python_rules.build_defs
peterebden: I've expanded it a bit to better describe available options - I wanted to explain a bit how you'd change it. Agreed that it's useful to describe the per-folder case although I suspect a lot of potential users will be using one or the other already.

uwedeportivo:
👍

natpen:
Bravo! Can't wait to try this out during the week :smile_cat:

peterebden:
Thanks - will be good for it to have some testing, I'm not really familiar with pytest so maybe I've missed some use cases :)

Issue #250, opened 2017-11-29
Config file hierarchy


natpen:
:wave: hi again! I've been thinking about the config file hierarchy today. Copying from documentation, the current state is:

> Please looks in several locations for these files and builds it up bit by bit. In order (from lowest priority to highest):
> * `.plzconfig`
> * `.plzconfig_<arch>` (e.g. .plzconfig_linux_amd64)
> * `/etc/plzconfig`
> * `.plzconfig.local`

I get why `/etc/plzconfig` is higher-priority than the repo-specific configs, as you mention in documentation that this allows overriding repo-specific configuration in places like CI. Conceptually though, I was wondering if there would be a way to make increases in hierarchy correspond with increases in application breadth. Here was an idea I came up with, which I was curious what you thought of:

> Please looks in several locations for these files and builds it up bit by bit. In order (from lowest priority to highest):
> * `/etc/plzconfig`
> * `~/.please/plzconfig`
> * `.plzconfig`
> * `.plzconfig_<arch>` (e.g. .plzconfig_linux_amd64)
> * `.plzconfig.<env>` (where env is a string matching an env config set farther down in the config hierarchy)

The main ordering difference is that repo-specific configs would override lower-level configs, similar to how other major tools (e.g., [Git](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration)) do config file prioritization. Additionally, it adds a user-specific config file. Lastly, to cover the CI use-case, it adds a repo-specific, environment-specific config file at the highest priority (`.plzconfig.<env>`). In the absence of a set env, it could default to `local` to maintain existing functionality with respect to `.plzconfig.local`, but on, say, a CI machine, you could set the env to something like `ci` in `/etc/plzconfig`, then check in a `.plzconfig.ci` to the repository.

This wouldn't necessarily be backwards compatible with everyone's use cases, which would make it trickier, but still curious what your thoughts are, and if you've talked through this before :)

peterebden:
We've not really thought about this in a lot of depth - the current config setup appeared fairly early on to solve some immediate needs. I agree that the ordering is probably a bit unintuitive though since it doesn't match how other tools do it.

Our use has evolved a bit and I'm not sure we rely on the prioritisation so specifically any more, so it might be possible to simply reorder them.
I like the idea of adding `~/.please/plzconfig` although it interacts a little weirdly with the config file allowing you to define a different location for it. Again I'm not sure if there's really a lot of use to that any more though...

[...]

PR #256, opened 2017-12-11
Implement new config file hierarchy as discussed in #250

peterebden:
See #250 for details.

tl;dr `/etc/plzconfig` is now lower priority than `.plzconfig` in the repo, and adds `~/.please/.plzconfig`.

This is a breaking change, clearly. I am pretty sure it has relatively little effect in our case and can be fixed with some judicious config adjustments anyway.


natpen:
Cool! I like this change. One thing I'm curious about, after the discussion in #250, is how you're getting higher-priority configuration on your build machines. You mentioned that you weren't relying on `/etc/plzconfig` for that anymore.

In addition to the env-specific version-controlled option I described in #250, I see a few other options, and I'm curious what your thoughts are on them would be with the landing of this PR (or if you have other ideas from your internal workflow):
- use a test/build runner to copy in a `.plzconfig.local` after cloning a repo (I like that this is OS-agnostic, unlike the next option, but dislike that it doesn't gel well with checking the local config into the repo... unless you do something interesting like, check it in with a non-please-recognized filename, then rename it at remote build/test time, but that feels a bit odd as well)
- use a checked-in, arch-specific config, and make sure your build/test machines have a different arch than any development machines (I like that this gets you in-repo, version-controlled build/test machine configuration, but dislike that it's restrictive with respect to OSes)

peterebden:
Thanks!

> One thing I'm curious about, after the discussion in #250, is how you're getting higher-priority configuration on your build machines. You mentioned that you weren't relying on /etc/plzconfig for that anymore.

I realised that we don't need higher-priority configuration any more; we can remove a couple of things from `.plzconfig` where they match the default, and that should work out fine. Definitely prefer the idea of having a more intuitive config hierarchy, and we can solve higher-priority config in a different way if we need it.

> use a test/build runner to copy in a .plzconfig.local after cloning a repo (I like that this is OS-agnostic, unlike the next option, but dislike that it doesn't gel well with checking the local config into the repo... unless you do something interesting like, check it in with a non-please-recognized filename, then rename it at remote build/test time, but that feels a bit odd as well)

Indeed I think we do this somewhere internally - we certainly used to do it back when we used Travis for CI in this repo, but it's not necessary since we moved to Circle. Agreed that it's a bit rough-and-ready, I don't like the idea of it not being checked in although there is always the case where you want to do something that is only known to the CI system at runtime which prohibits checking everything in.

> use a checked-in, arch-specific config, and make sure your build/test machines have a different arch than any development machines (I like that this gets you in-repo, version-controlled build/test machine configuration, but dislike that it's restrictive with respect to OSes)

Fair point. Right now the arch-specific one is set up at compile time (it's just based on GOOS and GOARCH) so you can't really choose another one for different machines very readily. We use it for swapping in different behaviour on OSX / Linux but it's probably not suitable for much else.

I'm not against the idea of adding something else too, just a little unsure what is best - and I'd like to avoid it becoming too confusing about having too many locations. Another option is possibly to add a flag to specify arbitrary locations to load from (because all problems can be solved by adding more flags...).

PR #258, opened 2017-12-17
Add option to read a config profile by flag.

peterebden:
As discussed in #256 and slightly more recently #250 with @natpen , adding a flag allowing a choice of config profile which can be checked in and thus version-controlled, but also easily accessed.

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -52,7 +52,7 @@ var opts struct {
 		Include    []string        `short:"i" long:"include" description:"Label of targets to include in automatic detection."`
 		Exclude    []string        `short:"e" long:"exclude" description:"Label of targets to exclude from automatic detection."`
 		Option     ConfigOverrides `short:"o" long:"override" env:"PLZ_OVERRIDES" env-delim:";" description:"Options to override from .plzconfig (e.g. -o please.selfupdate:false)"`
-		Profile    string          `long:"profile" env:"PLZ_CONFIG_PROFILE" description:"Configuration profie to load; e.g. --profile=dev will load .plzconfig.dev if it exists."`
+		Profile    string          `long:"profile" env:"PLZ_CONFIG_PROFILE" description:"Configuration profile to load; e.g. --profile=dev will load .plzconfig.dev if it exists."`
 	} `group:"Options controlling what to build & how to build it"`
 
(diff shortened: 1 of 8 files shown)


src/please.go
```diff
@@ -52,6 +52,7 @@ var opts struct {
 		Include    []string        `short:"i" long:"include" description:"Label of targets to include in automatic detection."`
 		Exclude    []string        `short:"e" long:"exclude" description:"Label of targets to exclude from automatic detection."`
 		Option     ConfigOverrides `short:"o" long:"override" env:"PLZ_OVERRIDES" env-delim:";" description:"Options to override from .plzconfig (e.g. -o please.selfupdate:false)"`
+		Profile    string          `long:"profile" env:"PLZ_CONFIG_PROFILE" description:"Configuration profie to load; e.g. --profile=dev will load .plzconfig.dev if it exists."`
```
natpen: nit: profie --> profile

peterebden: Fixed, thanks!

natpen:
Hihi! Sorry to have just noticed this after you landed the PR, but I'm a little confused by the application of the profile name to the filename. In the `ReadConfigFiles` implementation it does some string concatenation like so `filename+"."+profile`. This makes sense to me, because then lexicographical ordering of your project's filetree would give you a cluster of plzconfigs, like:

* `.plzconfig`
* `.plzconfig.working`
* `.plzconfig.prod`
* etc.

But then the tests in this PR all seem to have the profiles set the other way around, like:

* `.plzconfig`
* `.working.plzconfig`
* `.prod.plzconfig`
* etc.

I think this may be a bug if I'm interpreting things correctly. I think it may have slipped through b/c all the test cases make config files, then pass the full config path, including profile, via the filename parameter, so none of the test cases are actually testing `ReadConfigFiles`'s implementation of the application of the profile to the filename.

peterebden:
I think that's just unclear naming of the test files; the relevant test does test the concatenation by loading `working.plzconfig` and `working.plzconfig.dev`. The real config file names are defined in config.go and all start with `.plzconfig`.

Sorry if that's unclear!

natpen:
Ahhhhh, okay, I follow! Totally missed the `dev` in there :)

PR #260, opened 2017-12-24
Remove use of github_file in pleasings documentation

natpen:
github_file itself was moved from Please core into pleasings, so the
documentation for how to set up pleasings initially should rely on the
core-included remote_file implementation.


peterebden:
Thanks for the correction!

(And sorry for slow response, have been away over Xmas / New Years...)

PR #262, opened 2018-01-05
Strip out all test dependencies completely

peterebden:
Fixes some output with -b where spurious dependencies were added.


tools/please_maven/maven/pom.go
```diff
@@ -274,6 +274,7 @@ func (pom *PomXML) Unmarshal(f *Fetch, response []byte) {
 	pom.Version = pom.replaceVariables(pom.Version)
 	// Arbitrarily, some pom files have this different structure with the extra "dependencyManagement" level.
 	pom.Dependencies.Dependency = append(pom.Dependencies.Dependency, pom.DependencyManagement.Dependencies.Dependency...)
```
dimpavloff: should you be stripping the recursive test dependencies at this point? otherwise you will pull some test dependency's compile dependency and then strip the former but not the latter

peterebden: hmmm no I don't think I need to? At this point we have only immediate dependencies, not transitive, so stripping all the ones with test scope should be fine?

dimpavloff: I didn't realise this was the case; yes, that should be fine then

PR #267, opened 2018-01-12
Initial support for java 9 and modules

llucherini:


src/core/config.go
```diff
@@ -117,6 +122,18 @@ func defaultPath(conf *string, dir, file string) {
 	}
 }
 
+// defaultPathIfExists sets a variable to a location in a directory if it's not already set and if the location exists.
+func defaultPathIfExists(conf *string, dir, file string) {
+	if *conf == "" {
+		location := path.Join(dir, file)
+		// check that the location is valid
+		_, err := os.Stat(location)
+		if err == nil {
```
peterebden: FYI you can do these two in one line:
```
if _, err := os.Stat(location); err == nil {
```

src/parse/rules/java_rules.build_defs
```diff
@@ -113,6 +114,91 @@ def java_library(name, srcs=None, src_dir=None, resources=None, resources_root=N
...
+    """
+    if not CONFIG.JAVA_HOME:
+        raise Exception('Java home needs to be set to link java runtime images against jmods.')
+    if not CONFIG.JLINK_TOOL:
+        raise Exception('A jlink tool is required to build java runtime images.')
+    if not modules:
+        raise Exception('You cannot assemble a java runtime image without specifying any modules.')
+    deps = deps or []
+    out = out or name
+    jlink_args = jlink_args or '--strip-debug --compress=2'
```
peterebden: If the default is a string, just make that the default argument value (we only avoid that for mutable things like lists which can get a bit weird)

src/parse/rules/java_rules.build_defs
```diff
@@ -113,6 +114,91 @@ def java_library(name, srcs=None, src_dir=None, resources=None, resources_root=N
...
+        data=data,
+        outs=[out],
+        cmd=cmd,
+        needs_transitive_deps=True,
+        output_is_complete=True,
+        binary=True,
+        building_description="Creating runtime image...",
+        requires=['java'],
+        visibility=visibility,
+        tools={'jlink': [CONFIG.JLINK_TOOL]}
```
peterebden: nit: if there's only one tool you can just give it as a list and access it as `$TOOL` which is (very) mildly more efficient.

PR #268, opened 2018-01-12
jarcat: handle akka reference.conf specially

dimpavloff:
When we package akka libs as a dependency, these will use the reference.conf each one of those packages with itself. However, since we are building a fatjar, those files need to be merged. Just concatenating them suffices.
See https://doc.akka.io/docs/akka/2.3.3/general/configuration.html#When_using_JarJar__OneJar__Assembly_or_any_jar-bundler


tools/jarcat/zip/writer_test.go
```diff
@@ -24,10 +24,64 @@ func TestAddZipFile(t *testing.T) {
...
+		if _, err := io.ReadFull(fReader, contents); err != nil {
+			return nil, err
+		}
+		res[file.FileInfo().Name()] = contents
+	}
+	return res, nil
+}
+
+func TestAddZipFileConcatenatesSpecialFiles(t *testing.T) {
+	t.Run("Scala Akka reference.conf", func(t *testing.T) {
```
dimpavloff: I haven't added the other special cases because I don't know what a good fixture looks like; but it is with those in mind why I've called my test like that and have a subtest

tools/jarcat/zip/writer_test.go
```diff
@@ -24,10 +24,64 @@ func TestAddZipFile(t *testing.T) {
 	assertExpected(t, "add_zip_file_test.zip", 0)
 }
 
+// getZipContents is a helper which returns a map of filename -> contents
+func getZipContents(zipfilePath string) (map[string][]byte, error) {
```
dimpavloff: please let me know if there's a more straightforward way

tools/jarcat/zip/writer_test.go
```diff
@@ -24,10 +24,64 @@ func TestAddZipFile(t *testing.T) {
...
+	})
+}
+
 func TestAddFiles(t *testing.T) {
 	f := NewFile("add_files_test.zip", false)
 	f.Suffix = []string{"zip"}
-	err := f.AddFiles("tools")
+	err := f.AddFiles("tools/jarcat/zip/test_data")
+	require.NoError(t, err)
+	err = f.AddFiles("tools/jarcat/zip/test_data_2")
```
dimpavloff: these had to be changed because the assertions we are making are for specific filenames in the zip files which the new fixture does not (and should not imo) have.

tools/jarcat/zip/writer_test.go
peterebden: Not that I can think of right now.

PR #274, opened 2018-02-08
Replace hyphen with underscore in [buildenv] plzconfig sections

clintharrison:
Please was previously leaving the `-` from the config file key in the name used of the environment variable. This is not technically valid, nor desired, but apparently permitted by my system!


peterebden:
Thanks, looks good!

I'm pretty sure the build failure is caused by https://github.com/golang/go/issues/23672 (fix is being tracked in https://github.com/golang/go/issues/23744). I'm trying to find a workaround - presumably there'll be a 1.9.5 that fixes it but it'd be nice not to be broken until then...

peterebden:
OK, it's fixed on master so will merge anyway despite the failure.

clintharrison:
Thanks for the quick merge!
Changing `--std=c99` to `-std=c99` will match the builtin allow list already; or you can set `export CGO_CFLAGS_ALLOW=--std=c99` to get it building as-is.

peterebden:
Thanks, it turned out in the end that we don't actually need it (I guess whatever c99 thing we originally needed it for is no longer around) so I just deleted it - but nice to know there were workarounds too :)

PR #276, opened 2018-02-15
Adding --run flag to plz watch

### LINKED ISSUE #272: watch & run functionality?
Hello again! :wave: 

I see there is `plz watch` functionality built into please as a result of #118  - I don't see this covered in the documentation, but from reading through the code it looks like it will watch the filesystem for changes, rebuild, and test if possible.

One thing we've been playing with the past few days is running a python/flask web app through please (which has been its own fun adventure in the interplay between gunicorn and pexes), and we were wondering if you had given any thought to the viability/usefulness of `watch&run` type functionality, as opposed to the current `watch[&test]`?

peterebden:
Fixes #272

### DIFF
--- a/src/watch/watch.go
+++ b/src/watch/watch.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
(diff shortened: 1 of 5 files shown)


src/watch/watch.go
```diff
@@ -151,3 +143,16 @@ func addSource(watcher *fsnotify.Watcher, state *core.BuildState, source core.Bu
 		}
 	}
 }
+
+// commands returns the plz commands that should be used for the given labels.
+func commands(state *core.BuildState, labels []core.BuildLabel, run bool) []string {
+	if run {
+		return []string{"run", "parallel"}
```
agenticarus: Why do you want to default to parallel?

peterebden: Mostly to allow for it to accept multiple targets. I was tossing up between `parallel` or `sequential` and settled on this - happy to change if you think sequential would make more sense.

agenticarus: Does it make sense to allow multiple targets to `run` at all?

peterebden: if someone passes multiple targets, why not? I'm not sure it will be common but we've had sufficient need for that to have added the `sequential` and `parallel` subcommands.

agenticarus: Fair enough.

Issue #277, opened 2018-02-18
Go 1.10 compatibility - missing Go dependencies


tomlanyon:
Go 1.0 [changes](https://golang.org/doc/go1.10#build) the behaviour of 'go install'; it now only installs packages you list on the command line, not any dependencies of those packages.

It turns out that building Please relied on the old behaviour, because there's quite a lot of dependencies which aren't listed explicitly in /third_party/go/BUILD.

A simple example of this is `github.com/Songmu/prompter` (//third_party/go:prompter) imports `github.com/mattn/go-isatty`, which isn't listed in the BUILD file as a dependency.

I have started fixing a bunch of these, but before going any further wanted to see if this was the desired fix.  The alternative would be updating the Go build rules to use 'go install -i', I guess, which returns the old behaviour.  To me, this seems counter to Please's style of being explicit about dependencies.

peterebden:
I hadn't looked at updating to 1.10 yet - seems like a good Monday morning task...

I'm not a huge fan of the way `go_get` works at the moment - as you say, being explicit about dependencies would be better. If 1.10 makes that easier I'm all for it, although I'd like to keep compatibility with older versions as well (in which the `-i` flag doesn't seem to exist...).

It's been a long time since I first wrote those rules, I can't remember if there was some reason why they work the way they do rather than installing only a single package each.

Anyway, I would be very happy to accept a fix that made all the dependencies explicit. I guess it will be a breaking change (although we may be able to think of ways of mitigating that). It definitely seems the way to go in future.

PR #279, opened 2018-02-18
go_get: support getting a/b/...

### LINKED ISSUE #278: go_get support for getting foo/bar/...
I was looking at updating a bunch of Go deps in //third_party/go for #277 - it would be useful to be able to specify:

    go_get(
        name = 'foo',
        get = 'github.com/foo/bar/...',
    )

to get a bunch of related packages in the same repo, but this doesn't currently seem to work, because the current rule tries to copy the '...' directory literally during the build.  You can't work around this by specifying your own `outs` either, because then the post-build scripts don't run to collect all the archives in pkg/.

tomlanyon:
Allows specifying paths ending in '...' as the 'get' parameter of go_get rules.

Fixes #278.

### DIFF
--- a/src/parse/rules/go_rules.build_defs
+++ b/src/parse/rules/go_rules.build_defs
@@ -529,10 +529,12 @@ def go_get(name:str, get:str, outs:list=None, deps:list=None, exported_deps:list
     post_build = None
     if binary and outs and len(outs) != 1:
         raise ValueError(name + ': Binary rules can only have a single output')
+    getroot = get[:-4] if get.endswith('/...') else get
+    subdir = 'src/' + getroot
     if not outs:
-        outs = [('bin/' + name) if binary else ('src/' + get)]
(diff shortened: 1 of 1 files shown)


tomlanyon:
I can't reproduce that CircleCI test failure on Darwin (and I can't see why it would fail); can I trigger CircleCI to re-run the test somehow?

peterebden:
It looks suspiciously related, although weird it is not broken for Linux. I've retriggered a build, let's see how that goes.

peterebden:
Ah, I guess this is go1.10 showing itself. The OSX build uses Brew to install things and is probably getting a new version of it (whereas the Linux one uses a Docker container that we've not updated yet).

We could pin it back temporarily, or just fix for 1.10 as in the other issue.

peterebden:
OK, gonna merge this as is since I seem to be running into the same issue while attempting to fix #277 :)

PR #280, opened 2018-02-19
Add missing //third_party/go dependencies

### LINKED ISSUE #277: Go 1.10 compatibility - missing Go dependencies

tomlanyon:
Be more explicit about go_get deps to ensure compatibility with Go 1.10 which now will not fetch dependencies by default during 'go install'.

Where many sub-packages of one repo were in use, the go_get now imports the whole repo (e.g. golang.org/x/net/...) and the name of the build rule has changed to the name of the repo.

Enables needs_transitive_deps for all go_get rules to maintain compatibility with Go 1.9 and below, which now needs to ensure all of the correct deps (and their deps, and so on) are available during the go_get build, otherwise those deps will get automatically installed and packaged in the wrong build rule.

Fixes #277 

Tested on Go 1.9.4 and Go 1.10 linux/amd64.

### DIFF
--- a/third_party/go/BUILD
+++ b/third_party/go/BUILD
@@ -19,10 +19,24 @@ go_get(
 )
 
 go_get(
+    name = 'warnings',
+    get = 'gopkg.in/warnings.v0',
+    revision = 'v0.1.2',
+)
+
+go_get(
     name = 'gcfg',
     get = 'gopkg.in/gcfg.v1',
+    install = [
+        'gopkg.in/gcfg.v1/scanner',
+        'gopkg.in/gcfg.v1/token',
+        'gopkg.in/gcfg.v1/types',
+    ],
     patch = 'gcfg_dynamic_fields.patch',
     revision = '27e4946190b4a327b539185f2b5b1f7c84730728',
+    deps = [
+        ':warnings',
+    ],
 )
 
 go_get(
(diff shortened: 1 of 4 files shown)


third_party/go/BUILD
```diff
@@ -61,24 +75,66 @@ go_get(
...
 )
 
 go_get(
     name = 'grpc',
-    exported_deps = [':context'],
+    exported_deps = [':net'],
     get = 'google.golang.org/grpc',
-    install = ['google.golang.org/grpc/health'],
+    install = [
+        'google.golang.org/grpc/balancer',
```
peterebden: This is a bit of a mouthful - would it be easier to use `get = 'google.golang.org/grpc/...'`?

tomlanyon: I tried that first. 😔 It fails, at least on Go 1.9, because one of the sub packages contains only tests; the way the go_get rule effectively calls 'go tool compile *.go' in that directory leads to a fatal error.

peterebden: Ah, right. Not much to be done then I guess :)

PR #281, opened 2018-02-19
Fix capture of loop variable.

tomlanyon:
Produces a golint warning and potentially other nastiness.


peterebden:
Good spot, thanks!

PR #282, opened 2018-02-21
Add please_go_filter command to filter Go source files against build constraints

### LINKED ISSUE #275: Go rules don't respect Go build constraints
As far as I can tell, the current Go build rules in Please don't support filtering source files using Go's [build constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints) because they directly call 'go tool compile', 'go tool link', etc.

Should be simple enough to filter the source files through [MatchFile](https://golang.org/pkg/go/build/#Context.MatchFile), but I couldn't see an appropriate way to hook this into the current rules.  Can you suggest the best place to hook in such a filter?

tomlanyon:
By default, Go source files will now be included/excluded from builds by applying
the build constraints as documented in the package docs for Go's `go/build` package.

For example, on GOOS=linux, source files containing a '// +build !linux' tag will not
be built.

* Refactor build constraints used during Please bootstrapping; the 'bootstrap' build
  tag is now applied during bootstrap.
* Use proper build constraints for src/core/exec*.go.
* Introduce the Go FilterTool / go_filter_tool config options.
* Introduce the filter_srcs parameter to go_binary and go_library rules, enabled
  by default.

Fixes #275


tomlanyon:
I'm fairly happy with this, but think it could do with some greater testing on more real life Please repos.

tools/please_go_filter/please_go_filter.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"flag"
+	"fmt"
+	"go/build"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+var taglist = flag.String("tags", "", "Comma-separated list of extra build tags to apply as build constraints.")
```
peterebden: The rest of the tools and whatnot don't use the standard `flag` package, although this one doesn't need any of the more advanced features of go-flags so I don't have very strong feelings on it.

tomlanyon: I started out using Please's `cli` package and, therefore, go-flags, but this was a dependency cycle because `cli` depends on please_go_filter for its own build.  It seemed simpler to only depend on built-in packages rather than work some way around this.

peterebden: Right, of course :)

peterebden:
Sorry for all the conflicts - I was moving a bunch of stuff about yesterday. I'll clean them up...

tomlanyon:
Nice, thanks for fixing the _ prefixed tests too.  I hadn't had a chance to clean them up after changing the filtering to _cgo_.

peterebden:
No problem!

PR #285, opened 2018-02-25
go_get rewrite

peterebden:
Revisiting it yet again...

Post #280 we were specifying all transitive dependencies anyway; this takes advantage of that to simplify the rule by having it output just the specific package in question. It then regained a bit of complexity with some Github-specific optimisations :)

`install` now automatically prefixes with the repo name if not given, and you have to explicitly pass an empty string if you want the root (this is a bit awkward, but `install` is not the common case, and it allows more flexibility for cases when the repo root is not a valid `go install` target itself).

I've also broken it into two rules internally (one for fetching and one for compiling) since the former should now be hash verifiable.

### DIFF
--- a/src/parse/rules/go_rules.build_defs
+++ b/src/parse/rules/go_rules.build_defs
@@ -512,7 +512,7 @@ def cgo_test(name:str, srcs:list, data:list=None, deps:list=None, visibility:lis
     )
 
 
-def go_get(name:str, get:str, deps:list=[], exported_deps:list=None,
+def go_get(name:str, get:str, repo:str='', deps:list=[], exported_deps:list=None,
            visibility:list=None, patch:str=None, binary:bool=False, test_only:bool&testonly=False,
            install:list=None, revision:str=None, strip:list=None, hashes:list=None):
     """Defines a dependency on a third-party Go library.
@@ -530,6 +530,7 @@ def go_get(name:str, get:str, deps:list=[], exported_deps:list=None,
     Args:
       name (str): Name of the rule
       get (str): Target to get (eg. "github.com/gorilla/mux")
+      repo (str): Location of a Git repository to fetch from.
       deps (list): Dependencies
       exported_deps (list): Dependencies to make available to anything using this rule.
       visibility (list): Visibility specification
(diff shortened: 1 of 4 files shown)


peterebden:
I have absolutely no understanding of how the builds that failed here ever managed to work. I believe the fix is correct, but the other changes would seem to be unrelated.

I'm sorely tempted to remove all the go_generate stuff from core though, I suspect it is a rich source of this kind of stuff (and is IMO not of an awful lot of value when you've got a build system supporting arbitrary actions...).

src/parse/rules/go_rules.build_defs
```diff
@@ -512,114 +514,140 @@ def cgo_test(name:str, srcs:list, data:list=None, deps:list=None, visibility:lis
...
-        # loop done like this because the GOPATH export should not be in a SubShell
-        'while IFS= read -r dep_dir; do '
-        'if [ -n "$dep_dir" ]; then export GOPATH="${dep_dir}:${GOPATH}" ; fi ; done '
-        '< <(echo "$existing" | sed -e \'s|pkg/.*||g\' | sort -u)',
-
-        'export INSTALL_DIR=$(mktemp -d ${TMP_DIR}/tmp.XXXXX)',
-        'export GOPATH="$INSTALL_DIR:$GOPATH"',
-
-        'cd $INSTALL_DIR',
-        '$TOOL get -d ' + get,
```
dimpavloff: good to see this madness gone :)

src/parse/rules/go_rules.build_defs
```diff
@@ -512,114 +514,140 @@ def cgo_test(name:str, srcs:list, data:list=None, deps:list=None, visibility:lis
...
+        test_only = test_only,
+        labels = labels,
+        hashes = hashes,
+        sandbox = False,
+    )
+
+    if install:
+        install = [i if i.startswith(get) else (get + '/' + i) for i in install]
+    else:
+        install = [get]
```
dimpavloff: should you be using `getroot` here instead of `get`?

third_party/go/BUILD
```diff
@@ -88,31 +84,44 @@ go_get(
...
+        'internal',
+        'keepalive',
+        'metadata',
+        'naming',
+        'peer',
+        'resolver',
+        'stats',
+        'status',
+        'tap',
+        'transport',
```
dimpavloff: nice!

src/parse/rules/go_rules.build_defs
peterebden: I thought maybe you'd miss it :)

src/parse/rules/go_rules.build_defs
peterebden: Not on line 606 for sure (it should install /... if specified). Maybe on 604; it doesn't really make sense to pass `install` with /... but `getroot` is probably better still.

dimpavloff: sorry, yes I meant mostly 604

peterebden:
> Have you tried it on a repo which has go_get targets in non-conventional places such as the project's root?

Yes, I've tried in third_party and in the root. The latter needs a teeny bit of special casing but it's only one if.

Issue #284, opened 2018-02-25
Try to run on Java 9 - Error


Beiri22:
I have installed JDK9 and tried to compile a small demo lib with the new compiler, resulting in the following error message and hanging plz process:

14:04:56.827   ERROR: Error from remote worker: Error: LinkageError occurred while loading main class build.please.compile.JavaCompiler
	java.lang.ExceptionInInitializerError: null
14:04:56.831   ERROR: Failed to read response: EOF

peterebden:
Thanks for the report - I don't have a java 9 setup handy to test on but will try to take a look when I can - or get some help if I can do that... :)

Beiri22:
The demo lib is just one file and a module-info; cleaning does not help. 
[Matrix.zip](https://github.com/thought-machine/please/files/1760561/Matrix.zip)

$ java -version
openjdk version "9.0.4"
OpenJDK Runtime Environment (build 9.0.4+11)
OpenJDK 64-Bit Server VM (build 9.0.4+11, mixed mode)

llucherini:
Unfortunately I can't replicate your issue, I seem to be able to build your example just fine.
Could you send over your `.plzconfig` file, and also the output of `plz --version`?
Thanks

peterebden:
Closing this unless there's more info; in general things work for us with Java 9 / 10 (as evidenced by CircleCI) so I assume it's not a fundamental issue.

PR #286, opened 2018-02-27
Worker hanging

peterebden:
Partial (but not complete) fix for #284; it resolves the issue where plz hangs forever when a worker dies unexpectedly, but does not answer anything about why the worker was dying in that case :(


llucherini:
Looks good!

PR #290, opened 2018-03-13
Example test using Numpy

### LINKED ISSUE #207: Working pex example with complex dependency (numpy)
Hi all,

I'm trying to get a pex file built that depends on a complex pypi library, numpy in this case, which includes C extensions.

Files are here:

https://gist.github.com/nimish/0c16fcfb5e9643f67440a87fd8d09c3f

`plz build` produces example.pex, but I get a "no module named numpy" when run.

If I drop `zip_safe=False` from the numpy declaration, nothing seems to change.

peterebden:
Resolves #207

### DIFF
--- a/.plzconfig
+++ b/.plzconfig
@@ -4,7 +4,6 @@
 pextool = //tools/please_pex
 defaultpiprepo = https://get.please.build/third_party/python/index.html
 moduledir = third_party.python
-defaultinterpreter = python3
 
 [parse]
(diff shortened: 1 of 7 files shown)


peterebden:
Thanks! I think the docker images need updating - they don't have pip3 installed.

Bit less sure about the OSX one...

peterebden:
Got it! That should make homebrewed pip a bit more working which is nice.

I am reminded of why I wrote python_wheel now though... :D

PR #291, opened 2018-03-17
Migrate filepath.Walk calls to godirwalk

### LINKED ISSUE #273: Investigate glob() speed
Suggestion from alandonovan:

> FYI: in my own experiments I have found that glob is crucial to performance, and that you can do much better by avoiding filepath.Walk and even os.Readdir and doing the recursion yourself directly over the Linux or Darwin getdents system call, which gives you the name and isdir bit of every entry in an entire directory each file in a single system call.

I've not looked at optimising it in any detail, but if we can make it faster that would be cool. The returned values probably still need to be deterministic though (since srcs preserve their original order - we could also change that but I'm sure we rely on that somewhere).

peterebden:
As mentioned in #273 and upstream, filepath.Walk could be faster. This swaps it out for a more optimal implementation which seems to preserve all the ordering that we require.

Resolves #273. I've still not done any real benchmarking of glob() but happy to assume that this is about as fast as it's going to get :)

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -28,6 +28,7 @@ go get github.com/texttheater/golang-levenshtein/levenshtein
 go get github.com/Workiva/go-datastructures/queue
 go get github.com/coreos/go-semver/semver
 go get github.com/djherbis/atime
+go get github.com/karrick/godirwalk
 
 # Clean out old artifacts.
 rm -rf plz-out src/parse/rules/builtin_rules.bindata.go src/parse/rules/builtin_data.bindata.go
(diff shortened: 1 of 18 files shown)


dimpavloff:
p.s. FYI, I tried some globs on 1k files in nested folders with and without the rewrite but there was no noticeable difference yet.

peterebden:
Any day now I'm sure... :) Maybe I should get Paulo to spin me up one of those AWS workspaces to actually try it.

Thanks for the test. I've not observed a case previously that's been limited by glob either, but it seemed worth doing since godirwalk is now a thing.

PR #293, opened 2018-03-21
Support for cleaning old versions

peterebden:
As discussed earlier, this will probably be useful to keep disk usage from growing uncontrollably.


fsiddiqia:
LGTM, thanks!

Issue #296, opened 2018-03-27
Slow python binaries with large library dependencies


matthen:
I am building python binaries with tensorflow as a dependency, and having some issues

## BUILD
```
pip_library(
  name="numpy",
  version="1.14.0",
  outs=["numpy"],
  zip_safe=False,
)

pip_library(
    name="protobuf",
    version="3.4.0",
    outs=["google"],
    deps=[
      ":six",
    ],
)

pip_library(
  name="six",
  version="1.11.0",
  outs=["six.py"],
)

pip_library(
    name="tensorflow",
    version="1.4.0",
    deps=[
      ":numpy",
      ":protobuf",
    ]
)

python_binary(
  name="test_main",
  main="test_main.py",
  deps=[":tensorflow"],
)
```

## test_main.py

```
import tensorflow


if __name__ == "__main__":
    print("tensorflow version: " + tensorflow.__version__)
```

```
$ please build :test_main
Build finished; total time 1m15.23s, incrementality 0.0%. Outputs:
//:test_main:
  plz-out/bin/test_main.pex
```

```
tensorflow version: 1.4.0

real	0m7.089s
user	0m3.786s
sys	0m2.554s
```

So there is a 7 second over-head.
Is that expected?

Also this results in 4059 objects in `.plz-cache`, including many small python files from tensorflow, making caching this directory between runs slow on travis.

Is there any way to make the binaries quicker, and to make the cache have fewer small files?

peterebden:
Hmmmm, a tricky one here :)

> So there is a 7 second over-head.
> Is that expected?
It's possible that it's taking a long time to import everything from within the zipfile. I guess I'd need to profile it although there's not much else that can be going on there...

In general there is some overhead for starting a `python_binary`, but we haven't found it significantly worse than the non-zipped equivalent. I've not tested Tensorflow though which I expect is quite a bit heavier than most other Python libraries.

> Is there any way to make the binaries quicker

Not much I can think of (but it could bear some more investigation for sure). TOne think you could try would be adding `zip_safe=False` (not that I'd hold out a lot of hope, but it's easy to try).

> and to make the cache have fewer small files?

[...]

PR #297, opened 2018-03-30
Python download rules output wheel files

peterebden:
As discussed in #296 , this more or less matches the suggestion there but uses `provides` to expose the `.whl` file. Hence when used as a dependency only the wheel gets output, but it's still possible to get all the .py files by building the target explicitly.


src/parse/rules/misc_rules.build_defs
```diff
@@ -307,8 +307,8 @@ def system_library(name:str, srcs:list, deps:list=None, hashes:list=None,
 
 
 def remote_file(name:str, url:str|list, hashes:list=None, out:str=None, binary:bool=False,
-                visibility:list=None, licences:list=None, test_only:bool&testonly=False, deps:list=None,
-                exported_deps:list=None, _tag=''):
+                visibility:list=None, licences:list=None, test_only:bool&testonly=False,
+                labels:list=[], deps:list=None, exported_deps:list=None, _tag=''):
```
csdigi: nit: labels:list=None?

peterebden: Thanks Will!

These guys are safe now, it's okay to default them to things like lists.

PR #298, opened 2018-04-04
Improve handling of Aliases

agenticarus:
Rather than doing a blind search-and-replace for aliases in the combined arguments string (which leads to it replacing after a -- and if the alias word appears at all) we scan the arguments and only replace if the exact argument matches.

Tested this with some problematic command lines that were failing before this change.


agenticarus:
I agree, that would seem more flexible.

PR #300, opened 2018-04-05
Set pipefail in plz_ci.sh

matthen:
Otherwise an error in plz query might not cause the entire script to fail.


peterebden:
Makes sense, but I don't think this works with the interpreter being `/bin/sh` - can you change that to `/bin/bash` please?

matthen:
Sure, changed to bash

peterebden:
Thanks!

PR #301, opened 2018-04-05
Make bytecode generation more consistent

peterebden:
Have tested locally on our troublesome case; previously `plz rebuild` produced different results frequently (noticeable via it saying `incrementality 0.0%`), now it is repeatably the same.


agenticarus:
LGTM. Thanks @peterebden

PR #302, opened 2018-04-11
Add config to pass in environment variables

### LINKED ISSUE #205: Proxy support in called commands
Unfortunately I have to live behind a corporate proxy, which means that the commands that have to traverse that proxy need to be aware of it.

Normall just setting HTTP_PROXY/http_proxy should be enough in the environment, but please doesn't pass it through (https://github.com/thought-machine/please/blob/master/src/core/build_env.go)

Pip supports a `--proxy` but some other tools don't.

jtdoepke:
This adds a config `Build.PassEnv` to copy environment variables from the user's current shell. Example:

```
[Build]
PassEnv = http_proxy
PassEnv = HTTP_PROXY
```

Should fix #205 in most cases, I think.

### DIFF
--- a/src/core/config.go
+++ b/src/core/config.go
@@ -83,6 +83,7 @@ func ReadConfigFiles(filenames []string, profile string) (*Configuration, error)
 		setDefault(&config.Parse.BuildFileName, []string{"BUILD"})
 	}
 	setDefault(&config.Build.Path, []string{"/usr/local/bin", "/usr/bin", "/bin"})
+	setDefault(&config.Build.PassEnv, []string{})
 	setDefault(&config.Cover.FileExtension, []string{".go", ".py", ".java", ".js", ".cc", ".h", ".c"})
 	setDefault(&config.Cover.ExcludeExtension, []string{".pb.go", "_pb2.py", ".pb.cc", ".pb.h", "_test.py", "_test.go", "_pb.go", "_bindata.go", "_test_main.cc"})
 	setDefault(&config.Proto.Language, []string{"cc", "py", "java", "go", "js"})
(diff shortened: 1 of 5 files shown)


src/core/config.go
```diff
@@ -435,6 +437,19 @@ func (config *Configuration) GetBuildEnv() []string {
 	return config.buildEnvStored
 }
 
+// GetPassEnv returns environment variables from the user's shell.
+func (config *Configuration) GetPassEnv() []string {
```
peterebden: Would be nice to add this to GetBuildEnv as well so it only has to build the slice once.

They should also form part of the config hash which that will take care of.

jtdoepke: Just combined into the same function, something like this?
```go
// GetBuildEnv returns the build environment configured for this config object.
func (config *Configuration) GetBuildEnv() []string {
	config.buildEnvOnce.Do(func() {
		config.buildEnvStored = []string{}

		// from the BuildEnv config keyword
		for k, v := range config.BuildEnv {
			pair := strings.Replace(strings.ToUpper(k), "-", "_", -1) + "=" + v
			config.buildEnvStored = append(config.buildEnvStored, pair)
		}

		// from the user's environment based on the PassEnv config keyword
		for _, k := range config.Build.PassEnv {
			v, isSet := os.LookupEnv(k)
			if isSet {
				config.buildEnvStored = append(config.buildEnvStored, k+"="+v)
			}
		}

		sort.Strings(config.buildEnvStored)
	})
	return config.buildEnvStored
}
```

peterebden: Yup, exactly!

peterebden:
Looks great, thanks!

PR #304, opened 2018-04-17
Fixes sh_binary to use zip/unzip instead of tar.

fsiddiqia:


src/parse/rules/sh_rules.build_defs
```diff
@@ -57,19 +52,20 @@ def sh_binary(name:str, main:str, deps:list=None, visibility:list=None, labels:l
...
-            'echo "%s" >> $OUT' % (_SH_BINARY_TMPL_PREFIX % _TARBALL_BOUNDARY),
+            'echo "%s" >> $OUT' % _SH_BINARY_TMPL_PREFIX,
             # Inject the user defined script.
             'cat $SRCS >> $OUT',
             # Inject the tar ball boundary.
-            'echo "%s" >> $OUT' % (_SH_BINARY_TMPL_SUFFIX % _TARBALL_BOUNDARY),
+            'echo "%s" >> $OUT' % _SH_BINARY_TMPL_SUFFIX,
             # Compress the dependent files and dump out into the bash script.
-            'find . -type f | grep -v $SRCS | grep -v $(echo $OUT | sed "s|$TMP_DIR||g") | sed "s/^\\.\\///g" | tar zc -f - --files-from - >> $OUT',
+            'find . -type f | grep -v $SRCS | grep -v $(echo $OUT | sed "s|$TMP_DIR||g") | sed "s/^\\.\\///g" | zip _%s#zipfile -@' % name,
```
peterebden: this should probably use jarcat rather than zip - the output of zip isn't very deterministic. I'm not sure that supports stuff from stdin though...

fsiddiqia: Thanks, this makes sense.

As discussed offline, since we're moving away from tar (which probably wasn't any better in this regard), it should be good enough. Will leave a TODO here to investigate using jarcat (or something with a deterministic output) in the future.

PR #305, opened 2018-04-18
Use Jarcat to construct sh_binary instead of zip

peterebden:
90% of this is just moving code around to keep dependencies clean. Also added a little wrapper type for the stdin thing.


fsiddiqia:
Thanks! LGTM. Will send a PR with the unzip fix soon.

peterebden:
Cool, let's wait for that before doing another version.

PR #311, opened 2018-04-23
Added Java support for the new `--release` flag

llucherini:
Also enhanced java rules to support changing the source, target and release level. This can be useful when, for example, part of your code base is targeting android.


src/parse/rules/java_rules.build_defs
```diff
@@ -2,6 +2,7 @@
 
 
 def java_library(name:str, srcs:list=None, src_dir:str=None, resources:list=None, resources_root:str=None,
+                 source_level:str=None, target_level:str=None, release_level:str=None,
```
peterebden: You can default these - e.g. `source_level:str=CONFIG.JAVA_SOURCE_LEVEL` which would simplify your ifs below

llucherini:
Oh nice, I didn't realise we could do that already. I agree, this is something that will most likely be just on the package level. How do the targets read it? Do they just see it as config?

peterebden:
Yup exactly, it just modifies values in the CONFIG object.

PR #312, opened 2018-05-01
Implements `plz query changes`

peterebden:
Replaces the previous multi-command dance with a single command (well, it's still a few to invoke the build properly, but it's not too hard). I don't think this was possible a year or two ago but things are a bit better factored now (especially the parser is a lot more cooperative).

The total runtime is rather faster; a large chunk of it now is the underlying git operations, which were previously dwarfed by all the other futzing about we were doing.

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -310,7 +310,7 @@ var opts struct {
 			} `positional-args:"true"`
 		} `command:"rules" description:"Prints built-in rules to stdout as JSON"`
 		Changes struct {
-			Since           string `short:"s" long:"since" description:"Revision to compare against"`
+			Since           string `short:"s" long:"since" default:"origin/master" description:"Revision to compare against"`
 			CheckoutCommand string `long:"checkout_command" default:"git checkout %s" description:"Command to run to check out the before/after revisions."`
 			CurrentCommand  string `long:"current_revision_command" default:"git rev-parse HEAD" description:"Command to run to get the current revision (which will be checked out again at the end)"`
 			Args            struct {
@@ -570,7 +570,6 @@ var buildFunctions = map[string]func() bool{
 		return true
 	},
 	"changes": func() bool {
-		opts.OutputFlags.PlainOutput = true
 		original := query.MustGetRevision(opts.Query.Changes.CurrentCommand)
 		files := opts.Query.Changes.Args.Files.Get()
 		query.MustCheckout(opts.Query.Changes.Since, opts.Query.Changes.CheckoutCommand)
@@ -600,8 +599,6 @@ func (overrides ConfigOverrides) Complete(match string) []flags.Completion {
 
 // Used above as a convenience wrapper for query functions.
 func runQuery(needFullParse bool, labels []core.BuildLabel, onSuccess func(state *core.BuildState)) bool {
-	opts.OutputFlags.PlainOutput = true // No point displaying this for one of these queries.
-	config.Cache.DirClean = false
 	if !needFullParse {
 		opts.ParsePackageOnly = true
 	}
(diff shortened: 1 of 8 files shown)


src/please.go
```diff
@@ -309,6 +309,14 @@ var opts struct {
 				Targets []core.BuildLabel `position-arg-name:"targets" description:"Additional targets to load rules from"`
 			} `positional-args:"true"`
 		} `command:"rules" description:"Prints built-in rules to stdout as JSON"`
+		Changes struct {
+			Before          string `short:"b" long:"before" description:"Revision to check out for the state before"`
+			After           string `short:"a" long:"after" required:"true" description:"Revision to check out for the state after"`
```
agenticarus: Why do we have both `before` and `after`? Naively I would expect this to work as follows:

`plz query changed` with no arguments compares the current working directory state to the last commit (i.e. HEAD, i.e. a noop when directly on a git commit).

`plz query changed --since [reflike]` compare the current working directory state to the given reflike.

So for CI I would expect to run something like `plz query changed --since @{upstream} (or origin/master I guess)`. And on merge I'd either run everything or just `plz query changed --since HEAD^` (given our workflow) - I don't think I'd ever want to do anything else.

Either way I think before and after are slightly confusing terms here.

peterebden: That is fair.

Either it's a copy-and-paste job from the old flow, or I was trying to answer the question of how I know what the current revision is - i.e. the logic for `plz query changes --since origin/master` looks like:
```
git checkout origin/master
plz do stuff
git checkout ????
plz do more stuff
```
I do agree it's a nicer interface - maybe another pseudo-flag to work out the current revision? I am trying not to insist on git here although obviously it's somewhat biased towards it - but I think that's okay as long as it's possible to choose something else via a sufficient supply of flags.

src/query/changes.go
```diff
@@ -0,0 +1,81 @@
...
+	for _, t2 := range targets {
+		if t1 := before.Graph.Target(t2.Label); t1 == nil || changed(before, after, t1, t2, files) || configChanged {
+			ret = append(ret, addRevdeps(after.Graph, t2, done)...)
+		}
+	}
+	sort.Sort(ret) // reverse dependencies aren't guaranteed to be consistent
+	return ret
+}
+
+// changed returns true if the given two targets are equivalent.
```
agenticarus: That seems a little confusing.

src/query/changes.go
```diff
@@ -0,0 +1,81 @@
...
+	return !bytes.Equal(h1, h2)
+}
+
+// addRevdeps walks back up the reverse dependencies of a target, marking them all changed.
+// It returns the list of any that have, plus the target itself.
+func addRevdeps(graph *core.BuildGraph, target *core.BuildTarget, done map[*core.BuildTarget]struct{}) []core.BuildLabel {
+	if _, present := done[target]; present {
+		return nil
+	}
+	done[target] = struct{}{}
```
agenticarus: We probably don't need both the map and the list - in the for loop you could check the map and if absent recurse, then finally return the map keys.

tools/please_diff_graphs/plz_diff_graphs_main.go
```diff
@@ -69,6 +73,7 @@ func readStdin() []string {
 func main() {
 	cli.ParseFlagsOrDie("Please graph differ", "9.1.2", &opts)
```
agenticarus: Is this a version number?

src/query/changes.go
peterebden: Indeed it does...

tools/please_diff_graphs/plz_diff_graphs_main.go
peterebden: Yeah. The various tools all got them at one point, but I later realised that was a bad idea (because upping the version number causes them all to change at each release which causes unnecessary rebuilds).

One day I should just get rid of them all again...

peterebden:
Yeah so at present it doesn't really know anything about the SCM. I was rather hoping to keep it that way...

PR #315, opened 2018-05-16
Better version of plz-out/go

peterebden:
Discovered that the old plz-out/go implementation didn't work for nontrivial cases (because sooner or later Go gets upset about having .cc files in the wrong place or similar).

This adds a dinky little protocol based on labels where targets can opt into getting their outputs linked somewhere, which ultimately creates a tree of only the Go files under `plz-out/go`. I prefer losing the explicit Go stuff from the core packages - would like to do the same for Python although not sure how best to indicate that they need `__init__.py` files.


agenticarus:
Oh that seems pretty cute and definitely a lot simpler than my approach. I'll patch it and have a play.

agenticarus:
This LGTM - will need to check about the `package main` / `package [something]` problems but that may just be fixed by restructuring the source itself.

agenticarus:
Where does Generated code (protos/grpc/magic etc) end up?

peterebden:
Generated code ends up in the expected places under plz-out/go - i.e. it's linked there in a similar way to normal sources. Anything is as long as it's a source of a `go_library` rule.

And agreed on the package main / other situation; if one wants to use normal Go tools with that then following Go's package setup seems like a good idea. Don't want to try to solve that here given how fiddly that'd be.

PR #320, opened 2018-05-26
fix MODULE_DIR for pandas

jeffols:
The pandas package does not work with the original method of allowing third_party.python to work.

Current method loads the module from third_party.python.pandas and then fixes sys.modules.
pandas refers to sys.modules expecting to find itself and it can not (see pandas.tools.plotting)

My proposed fix adds thrid_party/python to sys.path so it can be found as is.

I am not certain why I had to remove third_party.python from [proto].pythonpackage in .plzconfig, but the tests would not pass without it.

### DIFF
--- a/.plzconfig
+++ b/.plzconfig
@@ -47,7 +47,7 @@ targetlevel = 7
 
 [proto]
 protoctool = //third_party/proto:protoc
-pythonpackage = third_party.python.google.protobuf
+pythonpackage = google.protobuf
 grpcjavaplugin = //third_party/java:protoc-gen-grpc-java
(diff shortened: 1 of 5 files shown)


.plzconfig
```diff
@@ -47,7 +47,7 @@ targetlevel = 7
 
 [proto]
 protoctool = //third_party/proto:protoc
-pythonpackage = third_party.python.google.protobuf
+pythonpackage = google.protobuf
```
peterebden: It might be possible to totally remove this (which would be a nice cleanup). I might have a little look after it's in.

PR #323, opened 2018-05-30
Copy XML test reports to their own directory before collation occurs

agenticarus:
These files are directly usable by CIs such as Jenkins and I'd like to expose them as they contain rich information.

I'm a little unsure about the plumbing structure here - it looks like the Opts object is only available if copied into the state Config but I'd appreciate a headsup if that's not the case. Either way, it's just a case of passing the directory path all the way down to the code that determines if the XML file is JUnit/Surefire (etc) format or not, and if it is do a copy.

The output directory looks reasonable for the please repo, although SampleTest shows up which it shouldn't but trying to keep it out turned into an Inception-like exercise.

I'm not sure what the best option for CI is - possibly deleting the whole folder before beginning a test run? But it feels like that should be handled outside Please.

I'd appreciate your thoughts.

### DIFF
--- a/src/test/surefire.go
+++ b/src/test/surefire.go
@@ -3,7 +3,6 @@ package test
 import (
 	"io/ioutil"
 	"path/filepath"
-	"os"
 
 	"core"
 	"fs"
@@ -25,7 +24,7 @@ func CopySurefireXmlFilesToDir(graph *core.BuildGraph, surefireDir string) {
 					if looksLikeJUnitXMLTestResults(bytes) {
 			    		log.Infof("Found target with tests: %v", target)
 						surefireResult := filepath.Join(surefireDir, filepath.Base(path))
-						if err := fs.CopyOrLinkFile(path, surefireResult, os.ModePerm, true, true); err != nil {
+						if err := fs.CopyOrLinkFile(path, surefireResult, 0644, true, true); err != nil {
 							log.Errorf("Error linking %s to %s - %s", surefireResult, path, err)
 						}
 					}
--- a/src/test/surefire.go
+++ b/src/test/surefire.go
@@ -1,0 +1,37 @@
+package test
+
+import (
+	"io/ioutil"
+	"path/filepath"
+	"os"
+
+	"core"
+	"fs"
+)
+
+func CopySurefireXmlFilesToDir(graph *core.BuildGraph, surefireDir string) {
+	log.Infof("Copy files to %s", surefireDir)
+	for _, target := range graph.AllTargets() {
+		if target.IsTest {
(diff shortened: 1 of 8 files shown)


agenticarus:
screen shot 2018-05-30 at 22 11 59

peterebden:
On the general plumbing structure - the Opts object is for handling the CLI flags, Configuration is for reading the .plzconfig file and BuildState for holding all the stuff that relates to a build (so it's built up from both those two).

I'm not sure here that this should be both a configuration option and a flag. As it is the flag always overrides the config option, and there is a reasonable amount of precedent for things like this not being in the config (e.g. `--test_results_file`, `--log_file`, `--trace_file` etc).

agenticarus:
Ah yeah ok - do you have any suggestions on how to get the flag down to where  it needs to be? Just  pass  it as a parameter all the way down?

peterebden:
That is probably the most obvious way, yes. Conversely some stuff gets hooked onto the `BuildState` object where I couldn't be bothered doing that kind of plumbing.

In this case it feels to me like maybe it should be another top-level call (ala `test.WriteResultsToFileOrDie`) in which case it would be logical to pass the filename in there.

agenticarus:
That seems a better approach, and more easily testable. I'll rework this.

agenticarus:
That's far less invasive now - I've removed some of the cover/test duplication too. It's a little yucky reiterating over all the targets and then parsing all the files we see, when we've just done that in the test results code, but it doesn't seem to take too long. I guess we could just match on the filename (TEST-...xml) as that's just as standard and the python and java honour that too.

src/test/surefire.go
```diff
@@ -0,0 +1,38 @@
...
+	"path/filepath"
+
+	"core"
+	"fs"
+)
+
+func CopySurefireXmlFilesToDir(graph *core.BuildGraph, surefireDir string) {
+	outputDirs := make(map[string]struct{})
+	for _, target := range graph.AllTargets() {
+		if target.IsTest {
```
peterebden: probably worth checking `&& !target.NoTestOutput` here as well; that indicates that the target isn't expected to produce any test output.

agenticarus:
I think you reviewed an outdated copy (I had already added to cover etc). But thanks, I'll add the notestoutput check

agenticarus:
> I think at some point we may want to rewrite the outputs ourselves - that would let this work for
> languages that produce non-XML output (e.g. Go) - but this is fine for now.

Yes, there's definitely more that can be done with the test results object which would avoid all of this copying.

peterebden:
Thanks Mike!

PR #326, opened 2018-06-06
Close file manager when done.

peterebden:
AFAICT this is not required, but it may help to close them when they are no longer needed rather than waiting for a GC. I wonder if Java 10 simply has a different GC strategy which is less aggressive about reclaiming them.
Anecdotally this seems to improve T14586.

Have also allowed customising the threadpool executor - don't think this worker should make a policy decision on it, but this way it's easier to experiment with outside this repo.

### DIFF
--- a/tools/javac_worker/src/build/please/compile/JavaCompiler.java
+++ b/tools/javac_worker/src/build/please/compile/JavaCompiler.java
@@ -26,7 +26,7 @@ public class JavaCompiler {
      * run reads requests from stdin and sends them to stdout until they are closed.
      */
     public void run() {
-        ExecutorService executor = Executors.newFixedThreadPool(8);
+        ExecutorService executor = createExecutor();
         final Gson gson = new GsonBuilder()
             .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
             .create();
@@ -47,6 +47,13 @@ public class JavaCompiler {
                     }
                 });
         }
+    }
+
+    /**
+     * Overriding this allows customising the threadpool that is used.
+     */
+    protected ExecutorService createExecutor() {
+        return Executors.newFixedThreadPool(8);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tools/javac_worker/src/build/please/compile/JavaCompiler.java
```diff
@@ -125,6 +132,7 @@ private BuildResponse reallyBuild(BuildRequest request) throws IOException {
             }
             opts.add(finder.joinFiles(':'));
             response.success = compiler.getTask(writer, fileManager, reporter, opts, null, compilationUnits).call();
+            fileManager.close();
```
llucherini: This could potentially be left open should an exception happen. We should declare the `fileManager` variable in a `try-with-resources` statement like so:
`try (StandardJavaFileManager fileManager = compiler.getStandardFileManager(reporter, null, null)) {`

peterebden:
Yeah - I don't like the idea of it having a fixed size of 8 threads. That might be 7 more than needed, or it might be limiting to 1/4 of the possible parallelism if you've got some sweet threadripper thing. I think the decision there should be controlled by `plz --num_threads` instead - which this basically achieves because it'll expand to as many as needed.

If it proves undesirable for us in any case then we can always set it back using the new function.

tools/javac_worker/src/build/please/compile/JavaCompiler.java
```diff
@@ -49,6 +49,13 @@ public void run() {
         }
     }
 
+    /**
+     * Overriding this allows customising the threadpool that is used.
+     */
+    protected ExecutorService createExecutor() {
+        return Executors.newCachedThreadPool();
```
llucherini: A `cachedThreadPool` is a bit problematic under load as it doesn't have a task queue and will create threads as more tasks need to be submitted to it. `fixedThreadPool` does.

llucherini:
How about by default we scale the fixed thread pool based on the number of cpus? We can easily do that with `Runtime.getruntime().availableprocessors()`

tools/javac_worker/src/build/please/compile/JavaCompiler.java
peterebden: Thanks, I always forget those!

llucherini:
Looks good, thanks!

peterebden:
OK yup that makes sense. I still like the idea of it growing on-demand but if the cached pool has other issues let's steer clear.

PR #327, opened 2018-06-09
Adding support to optionally keep the tests namespace when pulling th…

msmall-amfam:
…ird party pip packages. theano, for example, requires its tests package during runtime.


msmall-amfam:
I like applying the consistency of `strip` elsewhere as well. Maybe accept a list with the default of `*.pyc`?

I'm happy to take stab at it if you like (might be a week or two before I get to it though).

msmall-amfam:
Looks great! Thx Peter.

PR #328, opened 2018-06-12
Shade jacoco deps using jarjar

peterebden:
Should avoid any clashes with anything downstream that uses the same dependencies.


peterebden:
I'm not sure that we can vendor the jUnit annotations - we need to compare
the actual class to identify tests.

Lorenzo suggested this, but he thought jUnit probably would be okay so I'm
leaving it for now :)

llucherini:
Looks great, thanks!

PR #337, opened 2018-06-20
Added worker arguments to python_test to allow python_test to accept worker scripts

BNMetrics:


src/parse/rules/python_rules.build_defs
```diff
@@ -248,6 +249,11 @@ def python_test(name:str, srcs:list, data:list=[], resources:list=[], deps:list=
         )
         deps = [pex_rule, lib_rule]
 
+    test_cmd = '$TEST ' + flags
+    if worker:
+        test_cmd += '$(worker %s) && ' % worker
```
agenticarus: Won't this put it on the end of the string instead of the start?

peterebden:
Nice, but I think we should do the same to the other `_test` rules as well for consistency. Do you mind having a look at those too please?

src/parse/rules/cc_rules.build_defs
```diff
@@ -477,11 +478,12 @@ def cc_test(name:str, srcs:list=[], hdrs:list=[], compiler_flags:list&cflags&cop
         )
         srcs.append(main_rule)
     cmds, tools = _binary_cmds(_c, linker_flags, pkg_config_libs)
-    test_cmd = None if not CONFIG.CPP_COVERAGE else {
-        'opt': '$TEST %s' % flags,
-        'dbg': '$TEST %s' % flags,
-        'cover': '$TEST %s; R=$?; cp $GCNO_DIR/*.gcno . && gcov *.gcda && cat *.gcov > test.coverage; exit $R' % flags,
-    }
+    # test_cmd = None if not CONFIG.CPP_COVERAGE else {
```
peterebden: delete these lines

src/parse/rules/cc_rules.build_defs
```diff
@@ -495,6 +497,17 @@ def cc_test(name:str, srcs:list=[], hdrs:list=[], compiler_flags:list&cflags&cop
             _c=_c,
         )
         deps.append(lib_rule)
+
+    test_cmd = '$TEST' % flags
```
peterebden: this is missing any symbol to interpolate?

src/parse/rules/cc_rules.build_defs
```diff
@@ -495,6 +497,17 @@ def cc_test(name:str, srcs:list=[], hdrs:list=[], compiler_flags:list&cflags&cop
...
+    if worker:
+        test_cmd = '$(worker %s) && %s ' % (worker, test_cmd)
+        deps += [worker]
+
+    test_cmds = {k: test_cmd for k in ['opt', 'dbg']}
```
peterebden: nit: move this inside the if statement like
```
if CONFIG.CPP_COVERAGE:
    test_cmd = {
        'opt': test_cmd,
        'dbg': test_cmd,
        'cover': test_cmd + '...',
    }
```

src/parse/rules/sh_rules.build_defs
```diff
@@ -110,6 +111,7 @@ def sh_test(name:str, src:str=None, labels:list&features&tags=None, data:list=No
         srcs=[src or test],
         data=data,
         deps=deps,
+        worker=worker,
```
peterebden: I don't think you can do this? `build_rule` doesn't take `worker` directly?

src/parse/rules/cc_rules.build_defs
BNMetrics: oops! Will fix this now!

src/parse/rules/sh_rules.build_defs
BNMetrics: More oops >.<

PR #341, opened 2018-06-22
Switching ~ for ${HOME} in .profile

edrogers:
This PR changes how the `$PATH` is updated in `~/.profile` by the install script. 

When trying to install on zsh, the tilde is not expanded in the same way as in bash. ([More info here.](https://unix.stackexchange.com/a/146697))  So far, the only way I can get the install script to work in zsh is to capture download of https://get.please.build and replace the `~` on line 38 with `${HOME}`. I think this change will simplify the install process.


katzdm:
@peterebden , it doesn't look like this change was ever published to https://get.please.build (which still builds the wrong `PATH` for `zsh`) - Would you mind?

PR #344, opened 2018-06-25
Make subrepo an explicit part of BuildLabel

peterebden:
This makes things a lot cleaner versus the previous setup of mashing it into the package name and gets rid of a bunch of hackorama functions as a result. OTOH lots of knock-on changes throughout since a few core APIs need to change.

Unfortunately I still haven't persuaded it that it wants to deal with subrepos away from the root :(

### DIFF
--- a/src/core/graph.go
+++ b/src/core/graph.go
@@ -6,7 +6,6 @@ package core
 
 import (
 	"sort"
-	"strings"
 	"sync"
 )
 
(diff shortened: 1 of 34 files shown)


src/core/graph.go
```diff
@@ -80,18 +79,27 @@ func (graph *BuildGraph) TargetOrDie(label BuildLabel) *BuildTarget {
...
-// PackageOrDie retrieves a package by name, and dies if it can't be found.
-func (graph *BuildGraph) PackageOrDie(name string) *Package {
-	pkg := graph.Package(name)
+// PackageOrDie retrieves a package by label, and dies if it can't be found.
+func (graph *BuildGraph) PackageOrDie(label BuildLabel) *Package {
+	pkg := graph.PackageByLabel(label)
 	if pkg == nil {
-		log.Fatalf("Package %s doesn't exist in graph", name)
+		if label.Subrepo != "" {
+			log.Fatalf("Package @%s//%s doesn't exist in graph", label.Subrepo, label.PackageName)
```
agenticarus: don't bother with the `if` and just one argument and use `label.String()` ?

peterebden: `label.String` isn't quite the same (I don't want to print the target name), but good idea, I can reuse the `packageKey` struct for the same and drop the if.

PR #345, opened 2018-06-25
added test case for custom defined userdata on behave

BNMetrics:
Testing to see if we can get userdata into our context object in behave test by passing in ```-D``` or ```--define``` flag


peterebden:
LGTM. 

That test failure is a new one - I wonder if CircleCI has changed something about the Mac agent that makes it parallelise more, and hence hit some fd limit. For now I'm just going to drop the size of the test down.

PR #346, opened 2018-06-26
Documentation updates

peterebden:
Also a few bugfixes:
 - handle multiline strings properly for built-in functions
 - die if URL is not passed to `plz follow`
 - Remove RunArgs config entry which isn't honoured any longer.

### DIFF
--- a/docs/config.html
+++ b/docs/config.html
@@ -8,16 +8,17 @@
     <p>Please looks in several locations for these files and builds it up bit by bit. In order
       (from lowest priority to highest):
       <ul>
-        <li>/etc/plzconfig</li>
-        <li>~/.please/plzconfig</li>
-        <li>.plzconfig</li>
-        <li>.plzconfig_&lt;arch&gt; (e.g. .plzconfig_linux_amd64)</li>
-        <li>.plzconfig.local</li>
+        <li><code>/etc/plzconfig</code></li>
+        <li><code>~/.please/plzconfig</code></li>
+        <li><code>.plzconfig</code></li>
+        <li><code>.plzconfig_&lt;arch&gt;</code> (e.g. .plzconfig_linux_amd64)</li>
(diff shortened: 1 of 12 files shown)


docs/config.html
```diff
@@ -8,16 +8,17 @@ <h1>Please config file reference</h1>
...
+        <li><code>/etc/plzconfig</code></li>
+        <li><code>~/.please/plzconfig</code></li>
+        <li><code>.plzconfig</code></li>
+        <li><code>.plzconfig_&lt;arch&gt;</code> (e.g. .plzconfig_linux_amd64)</li>
+        <li><code>.plzconfig.&lt;profile&gt;</code> (if the <code>--profile</code> flag is passed)</li>
+        <li><code>.plzconfig.local</code></li>
       </ul>
-      One would normally check in .plzconfig (and arch-specific equivalents if needed).<br/>
+      One would normally check in <code>.plzconfig</code> (and arch-specific equivalents if needed).<br/>
       /etc/plzconfig is there to facilitate per-machine config in cases where the repo is often
```
agenticarus: &lt;code&gt; here too

docs/config.html
```diff
@@ -313,6 +358,9 @@ <h3>[Python]</h3>
         specified on the rule itself. Defaults to <code>python</code> but you could of course set it to
         <code>pypy</code>.</li>
 
+      <li><b>TestRunner</b><br/>
+        The test runner used to discover & run Python tests; one of unittest, pytest or behave.</li>
```
agenticarus: &amp;amp;

agenticarus: &lt;code&gt;unittest&lt;/code&gt; etc.

peterebden:
Thanks, fixed those. Also cleaned up the flags stuff a bit (turns out go-flags was just fine, we were short-circuiting the error...).

PR #347, opened 2018-06-28
Support for f-strings

peterebden:
Very cut-down version of the Python one - only supports interpolating literal variable names.

Have rewritten builtin rules to use them where possible; it might be very slightly faster (since it effectively pushes some of the overhead of reading the format string to parse time)


peterebden:
Hmmmmm this requires Go 1.10 for `strings.Builder`

I am tempted to just say that we require that now, I don't want to mess about with `bytes.Buffer` instead...

PR #349, opened 2018-06-28
More sensible defaults for num_runs and flakiness

agenticarus:
As part of the work to improve reporting of flaky tests, it turned out to be useful to set the defaults for these values to `1` instead of `0`. It makes the calcNumRuns method a lot simpler too.

### DIFF
--- a/src/utils/utils.go
+++ b/src/utils/utils.go
@@ -60,3 +60,11 @@ func isABuildFile(name string, config *core.Configuration) bool {
 	}
 	return false
 }
+
+// Max is an incredibly complicated function, which is presumably why it didn't make it into the Go stdlib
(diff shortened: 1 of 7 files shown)


src/utils/utils.go
```diff
@@ -60,3 +60,11 @@ func isABuildFile(name string, config *core.Configuration) bool {
 	}
 	return false
 }
+
+// Max is an incredibly complicated function, which is presumably why it didn't make it into the Go stdlib
```
peterebden: There's one for floats but not for ints. The float implementation is non-trivial though which I guess is the argument there.

agenticarus:
I'm going to touch it a lot more!

PR #352, opened 2018-07-09
Re-enable pip tests on OSX

peterebden:
I *think* these might work again

This reverts commit 1dcc9f245ed5824f28baf9001e5300e3d361a2aa (PR #351)


peterebden:
OK looks like it needs a bump of numpy and pandas but there is still no TF wheel, so I'll leave this for a bit.

PR #355, opened 2018-07-10
Replace hash files with xattr tags

peterebden:
Turns out that it's possible to get incorrect builds due to stale hash files and (in)appropriate manipulations of BUILD files.

This resolves that by removing the separate file (in most cases) and instead recording hashes on output files using xattrs, which is less likely to get out of sync. I think this will work on most relevant systems (i.e. it should work by default on ext4 with no mount flags, and OSX 10.4+), don't *think* there will be Docker issues.

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -114,7 +114,10 @@ func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget) (err
 	if target.IsHashFilegroup {
 		updateHashFilegroupPaths(state, target)
 	}
-	if !needsBuilding(state, target, false) {
+	// We don't record rule hashes for filegroups since we know the implementation and the check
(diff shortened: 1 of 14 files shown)


src/build/build_step.go
```diff
@@ -456,20 +446,29 @@ func calculateAndCheckRuleHash(state *core.BuildState, target *core.BuildTarget)
...
 		}
 	}
-	if err := writeRuleHashFile(state, target); err != nil {
-		return nil, fmt.Errorf("Attempting to create hash file: %s", err)
+	if !target.IsFilegroup {
+		if err := writeRuleHash(state, target); err != nil {
+			return nil, fmt.Errorf("Attempting to record rule hash: %s", err)
+		}
+	}
+	// Set appropriate permissions on outputs
```
agenticarus: Doesn't feel like this belongs in `RemoveOutputs`?

Issue #361, opened 2018-07-16
Support pkg-config for cflags only


peterebden:
As discussed, `so_test` fails if compiled against a Python install that has static but not dynamic libaries available, as the ones distributed from python.org do.

Python extension modules don't actually need to be linked against libpython at all (because the symbols will already be provided by whatever loads them) and it is probably a better idea that they aren't. However this isn't easy to achieve right now - we need to use pkg-config to find the compiler flags (so we can `#include "Python.h"`) but then we use it at link time too.

Look at adding an option like `pkg_config_cflags` to `c_library` and `cc_library` which runs it just at compile time.

BNMetrics:
This issue has been resolved with [#363](https://github.com/thought-machine/please/pull/363)

PR #363, opened 2018-07-17
#361 Support pkg-config for cflags only 

BNMetrics:
- Added pkg_config_cflags option in cc related build rules in  `/src/parse/rules/cc_rules.build_defs` and `/src/parse/rules/c_rules.build_defs`

### DIFF
--- a/src/parse/rules/c_rules.build_defs
+++ b/src/parse/rules/c_rules.build_defs
@@ -10,7 +10,7 @@ you must pay attention to interoperability when needed (e.g. 'extern "C"' and so
 
 def c_library(name:str, srcs:list=[], hdrs:list=[], private_hdrs:list=[], deps:list=[],
               visibility:list=None, test_only:bool&testonly=False, compiler_flags:list&cflags&copts=[],
-              linker_flags:list&ldflags&linkopts=[], pkg_config_libs:list=[],
+              linker_flags:list&ldflags&linkopts=[], pkg_config_libs:list=[], pkg_config_cflags:list=[],
               includes:list=[], defines:list|dict=[], alwayslink:bool=False):
     """Generate a C library target.
 
@@ -27,8 +27,10 @@ def c_library(name:str, srcs:list=[], hdrs:list=[], private_hdrs:list=[], deps:l
       compiler_flags (list): Flags to pass to the compiler.
       linker_flags (list): Flags to pass to the linker; these will not be used here but will be
                            picked up by a c_binary or c_test rule.
-      pkg_config_libs (list): Libraries to declare a dependency on using pkg-config. Again, the ldflags
-                              will be picked up by c_binary or c_test rules.
+      pkg_config_libs (list): Libraries to declare a dependency on using `pkg-config --libs`. Again, the ldflags
+                              will be picked up by cc_binary or cc_test rules.
+      pkg_config_cflags (list): Libraries to declare a dependency on using `pkg-config --cflags`. Again, the ldflags
+                              will be picked up by cc_binary or cc_test rules.
       includes (list): List of include directories to be added to the compiler's path.
       defines (list | dict): List of tokens to define in the preprocessor.
                              Alternatively can be a dict of name -> value to define, in which case
@@ -48,6 +50,7 @@ def c_library(name:str, srcs:list=[], hdrs:list=[], private_hdrs:list=[], deps:l
         compiler_flags = compiler_flags,
         linker_flags = linker_flags,
         pkg_config_libs = pkg_config_libs,
+        pkg_config_cflags=pkg_config_cflags,
         includes = includes,
         defines = defines,
         alwayslink = alwayslink,
@@ -56,8 +59,9 @@ def c_library(name:str, srcs:list=[], hdrs:list=[], private_hdrs:list=[], deps:l
 
 
 def c_object(name:str, src:str, hdrs:list=[], private_hdrs:list=[], out:str=None, test_only:bool&testonly=False,
-             compiler_flags:list&cflags&copts=[], linker_flags:list&ldflags&linkopts=[], pkg_config_libs:list=[], includes:list=[],
-             defines:list|dict=[], alwayslink:bool=False, visibility:list=None, deps:list=[]):
+             compiler_flags:list&cflags&copts=[], linker_flags:list&ldflags&linkopts=[],
+             pkg_config_libs:list=[], pkg_config_cflags:list=[], includes:list=[], defines:list|dict=[],
+             alwayslink:bool=False, visibility:list=None, deps:list=[]):
     """Generate a C object file from a single source.
 
     N.B. This is fairly low-level; for most use cases c_library should be preferred.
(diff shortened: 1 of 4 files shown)


src/parse/rules/cc_rules.build_defs
```diff
@@ -755,18 +771,28 @@ def _default_cflags(c, dbg):
...
 
-def _build_flags(compiler_flags:list, pkg_config_libs:list, defines=None, c=False, dbg=False):
+def _build_flags(compiler_flags:list, pkg_config_libs:list, pkg_config_cflags:list, defines=None, c=False, dbg=False):
     """Builds flags that we'll pass to the compiler invocation."""
     compiler_flags = [_default_cflags(c, dbg), '-fPIC'] + compiler_flags  # N.B. order is important!
     if defines:
         compiler_flags += ['-D' + define for define in defines]
-    pkg_config_cmd = ' '.join([f'`pkg-config --cflags {x}`' for x in pkg_config_libs])
+
+    pkg_config_list = pkg_config_libs or pkg_config_cflags or []
```
peterebden: I'm not sure `or` is correct here. Shouldn't it be `pkg_config_libs + pkg_config_cflags`?

src/parse/rules/cc_rules.build_defs
```diff
@@ -755,18 +771,28 @@ def _default_cflags(c, dbg):
...
+    pkg_config_cmd = ' '.join([f'`pkg-config --cflags {x}`' for x in pkg_config_list])
+
     return ' '.join(compiler_flags) + ' ' + pkg_config_cmd
 
 
-def _binary_build_flags(linker_flags:list, pkg_config_libs:list, shared=False, alwayslink='', c=False, dbg=False):
+def _binary_build_flags(linker_flags:list, pkg_config_libs:list, pkg_config_cflags:list, shared=False, alwayslink='', c=False, dbg=False):
```
peterebden: I don't think it makes sense to pass pkg_config_cflags to this function. They're not going to have any impact on the linker.

src/parse/rules/cc_rules.build_defs
```diff
@@ -822,34 +850,43 @@ def _binary_cmds(c, linker_flags, pkg_config_libs, extra_flags='', shared=False,
...
     def apply_transitive_labels(name):
         labels = get_labels(name, 'cc:')
         flags = ['-isystem %s' % l[4:] for l in labels if l.startswith('inc:')]
         flags.extend(['-D' + l[4:] for l in labels if l.startswith('def:')])
-        pkg_config_libs.extend([l[3:] for l in labels if l.startswith('pc:') and l[3:] not in pkg_config_libs])
+
+        if pkg_config_libs:
+            pkg_config_libs.extend([l[3:] for l in labels if l.startswith('pc:') and l[3:] not in pkg_config_libs])
+        elif pkg_config_cflags:
+            pkg_config_libs.extend([l[3:] for l in labels if l.startswith('pcc:') and l[4:] not in pkg_config_cflags])
```
peterebden: should probably be `l[4:]` in both places

src/parse/rules/cc_rules.build_defs
```diff
@@ -822,34 +850,43 @@ def _binary_cmds(c, linker_flags, pkg_config_libs, extra_flags='', shared=False,
...
-def _library_transitive_labels(c, compiler_flags, pkg_config_libs, archive=True):
+def _library_transitive_labels(c, compiler_flags, pkg_config_libs, pkg_config_cflags, archive=True):
     """Applies commands from transitive labels to a cc_library rule."""
     def apply_transitive_labels(name):
         labels = get_labels(name, 'cc:')
         flags = ['-isystem %s' % l[4:] for l in labels if l.startswith('inc:')]
         flags.extend(['-D' + l[4:] for l in labels if l.startswith('def:')])
-        pkg_config_libs.extend([l[3:] for l in labels if l.startswith('pc:') and l[3:] not in pkg_config_libs])
+
+        if pkg_config_libs:
```
peterebden: I don't think these should be conditional, presumably it should extend `pkg_config_cflags` in the same way as the logic worked before

src/parse/rules/cc_rules.build_defs
```diff
@@ -822,34 +850,43 @@ def _binary_cmds(c, linker_flags, pkg_config_libs, extra_flags='', shared=False,
...
         if flags:  # Don't update if there aren't any relevant labels
-            cmds, _ = _library_cmds(c, compiler_flags, pkg_config_libs, ' '.join(flags), archive=archive)
+            cmds, _ = _library_cmds(c, compiler_flags, pkg_config_libs, pkg_config_cflags, ' '.join(flags), archive=archive)
             for k, v in cmds.items():
                 set_command(name, k, v)
     return apply_transitive_labels
 
 
-def _binary_transitive_labels(c, linker_flags, pkg_config_libs, shared=False):
+def _binary_transitive_labels(c, linker_flags, pkg_config_libs=[], pkg_config_cflags=[], shared=False):
```
peterebden: same here, this function doesn't need `pkg_config_cflags`

src/parse/rules/cc_rules.build_defs
```diff
@@ -822,34 +850,43 @@ def _binary_cmds(c, linker_flags, pkg_config_libs, extra_flags='', shared=False,
...
     """Applies commands from transitive labels to a cc_binary, cc_test or cc_shared_object rule."""
     def apply_transitive_labels(name):
         labels = get_labels(name, 'cc:')
         linker_prefix = '' if CONFIG.LINK_WITH_LD_TOOL else '-Wl,'
         flags = [linker_prefix + l[3:] for l in labels if l.startswith('ld:')]
-        flags.extend(['`pkg-config --libs %s`' % l[3:] for l in labels if l.startswith('pc:')])
+        if pkg_config_cflags:
+            flags.extend(['`pkg-config --cflags %s`' % l[4:] for l in labels if l.startswith('pcc:')])
+        elif pkg_config_libs:
+            flags.extend(['`pkg-config --libs %s`' % l[3:] for l in labels if l.startswith('pc:')])
```
peterebden: and this line shouldn't be conditional

test/cc_rules/clang/BUILD
```diff
@@ -52,7 +52,7 @@ cc_library(
 cc_shared_object(
     name = "so_test",
     srcs = ["so_test.cc"],
-    pkg_config_libs = ["python3"],
+    pkg_config_cflags = ["python3"],
```
peterebden: You might need to add something like `ldflags = "-undefined dynamic_lookup"` on OSX; according to https://blog.tim-smith.us/2015/09/python-extension-modules-os-x/ that seems to be needed (the OSX linker is a constant source of little differences like this)

BNMetrics: I tried to append this linker flag on the actually build_rule, however, it doesn't seem to have worked... :(

src/parse/rules/c_rules.build_defs
```diff
@@ -27,8 +27,10 @@ def c_library(name:str, srcs:list=[], hdrs:list=[], private_hdrs:list=[], deps:l
       compiler_flags (list): Flags to pass to the compiler.
       linker_flags (list): Flags to pass to the linker; these will not be used here but will be
                            picked up by a c_binary or c_test rule.
-      pkg_config_libs (list): Libraries to declare a dependency on using pkg-config. Again, the ldflags
-                              will be picked up by c_binary or c_test rules.
+      pkg_config_libs (list): Libraries to declare a dependency on using `pkg-config --libs`. Again, the ldflags
```
peterebden: This isn't quite correct; it does also imply --cflags sometimes too.

Suggest leaving this one as it was before.

src/parse/rules/c_rules.build_defs
```diff
@@ -27,8 +27,10 @@ def c_library(name:str, srcs:list=[], hdrs:list=[], private_hdrs:list=[], deps:l
...
+                              will be picked up by cc_binary or cc_test rules.
+      pkg_config_cflags (list): Libraries to declare a dependency on using `pkg-config --cflags`. Again, the ldflags
```
peterebden: The second sentence is misleading (there are no ldflags in this case). Maybe just leave it out.

test/cc_rules/clang/BUILD
peterebden: Can you change this one to be like the other one too please? (i.e. do the CONFIG.OS check)

PR #367, opened 2018-07-20
Allowing xml coverage report, and added new flag --coverage_xml_report

BNMetrics:
- xml coverage report will now be able to be output to `plz-out/log/coverage.xml`
- To specify xml coverage report path, use the flag  `--coverage_xml_report` with custom path specified, like so  `--coverage_xml_report=/tmp/coverage.xml`

### DIFF
--- a/src/test/xml_coverage.go
+++ b/src/test/xml_coverage.go
@@ -43,7 +43,7 @@ func parseXMLLines(lines []xmlCoverageLine) []core.LineCoverage {
 }
 
 // Covert the Coverage result to XML bytes, ready to be write to file
-func CoverageResultToXML(sources []core.BuildLabel, coverage core.TestCoverage) []byte {
+func coverageResultToXML(sources []core.BuildLabel, coverage core.TestCoverage) []byte {
 	linesValid := 0
 	linesCovered := 0
 	validFiles := coverage.OrderedFiles()
@@ -55,12 +55,12 @@ func CoverageResultToXML(sources []core.BuildLabel, coverage core.TestCoverage) 
 	}
 
 	// Get the list of packages for <package> tag in the coverage xml file
-	var packages []Package
+	var packages []pkg
 	for label, coverage := range coverage.Tests {
 		packageName := strings.Replace(label.String(), ":", "/", -1)
 
 		// Get the list of classes for <class> tag in the coverage xml file
-		var classes []Class
+		var classes []class
 		classLineRateTotal := 0.0
 		for className, lineCover := range coverage {
 			// Do not include files in coverage report if its not valid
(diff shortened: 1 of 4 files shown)


src/test/xml_coverage.go
```diff
@@ -38,14 +42,108 @@ func parseXMLLines(lines []xmlCoverageLine) []core.LineCoverage {
 	return ret
 }
 
+// Covert the Coverage result to XML bytes, ready to be write to file
+func CoverageResultToXML(sources []core.BuildLabel, coverage core.TestCoverage) []byte {
```
peterebden: nit: this probably shouldn't be exported (it is fine for it to be a separate function, but not sure anything outside the package would want this)

src/test/xml_coverage.go
```diff
@@ -38,14 +42,108 @@ func parseXMLLines(lines []xmlCoverageLine) []core.LineCoverage {
...
+
+	topLevelLineRate := float64(linesCovered) / float64(linesValid)
+
+	// Create the coverage object based on the data collected
+	coverageObj := Coverage{Packages: packages, LineRate: formatFloatPrecision(topLevelLineRate, 4),
+		LinesCovered: int(linesCovered), LinesValid: int(linesValid),
+		Timestamp: int(time.Now().UnixNano()) / int(time.Millisecond),
+		Sources:   sourcesAsStr}
+
+	// Parse struct to xml bytes
```
peterebden: "Serialise", not "Parse"

src/test/xml_coverage.go
```diff
@@ -61,3 +159,55 @@ type xmlCoverageLine struct {
 	Hits   int `xml:"hits,attr"`
 	Number int `xml:"number,attr"`
 }
+
+// Coverage struct for writing to xml file
```
peterebden: This (and following types) should not be exported, since they are only for serialisation

src/test/xml_coverage.go
```diff
@@ -38,14 +42,108 @@ func parseXMLLines(lines []xmlCoverageLine) []core.LineCoverage {
 	return ret
 }
 
+// Covert the Coverage result to XML bytes, ready to be write to file
+func CoverageResultToXML(sources []core.BuildLabel, coverage core.TestCoverage) []byte {
+	linesValid := 0
+	linesCovered := 0
+	validFiles := coverage.OrderedFiles()
+
+	// get the string representative of sources
```
peterebden: are you sure this is correct? I would have thought the sources element in the XML would represent the actual file paths?

src/test/xml_coverage.go
```diff
@@ -38,14 +42,108 @@ func parseXMLLines(lines []xmlCoverageLine) []core.LineCoverage {
...
+	if err != nil {
+		log.Fatalf("Failed to parse to xml: %s", err)
+	}
+	covReport := []byte(xml.Header + string(xmlBytes))
+	return covReport
+}
+
+
+// Get the line coverage info, returns: list of lines covered, num of covered lines, and total valid lines
+func GetLineCoverageInfo(lineCover []core.LineCoverage) ([]Line, int, int) {
```
peterebden: this definitely shouldn't be exported

src/test/xml_coverage.go
BNMetrics: Yes, that's usually the case when running coverage. However I thought we could have it as a please target, like so `test/python_rules/pytest:all`? Since we are running it through please, so it would be easy to tell what we run to cover. However I could make it represent the path like `test/python_rules/pytest`, and leaving out the `all` bit  if you'd like?

PR #369, opened 2018-07-25
pipenv uses virtualenv wtihout getsitepackages so I am including it in pex_main.py if it is not found.

jeffols:
This may be an odd use case.

I am using a combination of please and pipenv to straddle the line between using please for releases and pipenv for development.

pipenv uses virtualenv wtihout getsitepackages so I am including it in pex_main.py if it is not found.

I need to run plz from within the virtual environment because I wrote a tool that will generate 90% of the BUILD file for me, but I need to be in the environement for it to read the packages.

### DIFF
--- a/tools/please_pex/pex_main.py
+++ b/tools/please_pex/pex_main.py
@@ -3,7 +3,6 @@
 from importlib import import_module
 import os
 import runpy
-import site
 import sys
 
 PY_VERSION = int(sys.version[0])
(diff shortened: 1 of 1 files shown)


peterebden:
I'm a little dubious about the idea of maintaining a parallel implementation of `getsitepackages`, but I see the motivation. Guess it doesn't do any harm and does fix it for your use case (& hopefully others too!). Thanks!

PR #377, opened 2018-08-03
Make scipy work

peterebden:
Turns out that we are adding slightly too many `__init__.py` files, as discussed on the mailing list. Skip adding them in inappropriate cases.

cc @jeffols , this should fix your issue. Sorry it took a while for me to figure it out...


jeffols:
Awesome, thanks!

-Jeff

BNMetrics:
LGTM!

PR #382, opened 2018-08-13
Moves UnitTest++ main out of plz core

peterebden:
Basically we add pleasings as a subrepo that's available by default and get it all from there instead.
I prefer that since there is no clear "standard" so I don't think plz should nominate one too strongly (and I don't like having BUILD rules that echo out the contents of files).

Most of the rest of this is fixing up various minor bugs and whatnot around subrepos.


.plzconfig
```diff
@@ -39,7 +39,7 @@ defaultdbgcppflags = --std=c++11 -g3 -DDEBUG -Wall -Wextra -Werror -Wno-unused-p
 ; it detects Java is installed and it can build them locally.
 javacworker = //tools/java:javac_worker
 junitrunner = //tools/java:junit_runner
-jarcattool = //tools/jarcat
+jarcattool = @//tools/jarcat
```
BNMetrics: Why is it a subrepo?

peterebden: it's not really; this makes it "absolute" so all subrepos use the one from this repo. Not super happy about having to do it but not sure how best to handle it otherwise.

PR #385, opened 2018-08-13
Fixed the issue with #375

BNMetrics:
We appear to not have anything to check the PATH in passEnv, so I diy-ed a little


peterebden:
Looks good, can you add a test for this please?

Issue #383, opened 2018-08-13
`plz query reverseDeps` does not filter to include specified labels


agenticarus:
Expected behaviour:
`plz query --include foo reverseDeps //some/target` excludes all targets that don't have label `foo`.

Actual behaviour:
All targets that depend on //some/target are returned.

`--exclude` similarly does nothing.

peterebden:
I think it's filtering the original targets (as in if you provide `//some/package:all` that would be filtered) but not the outputs which is more intuitive here.

Could you take a look please Luna?

agenticarus:
Possibly related:

`plz query --include foo alltargets` seems to filter whereas
`plz query --include foo alltargets //some/target1 //some/target2` does not.

peterebden:
nope that always happens - --include and --exclude apply to pseudo-labels like `/...` and `:all` but not to specific targets.

PR #387, opened 2018-08-14
Fix to the issue with 383

BNMetrics:
[#383](https://github.com/thought-machine/please/issues/383)

- It looks like ReserseDeps does not check for include and exclude in `src/query/revers_deps.go`
test command: `plz run src:please -- query --exclude go reverseDeps //src/core:core`

### DIFF
--- a/src/query/reverse_deps.go
+++ b/src/query/reverse_deps.go
@@ -1,24 +1,23 @@
 package query
 
 import (
+			"core"
 	"fmt"
 	"sort"
-
-	"core"
 )
 
 // ReverseDeps For each input label, finds all targets which depend upon it.
-func ReverseDeps(graph *core.BuildGraph, labels []core.BuildLabel) {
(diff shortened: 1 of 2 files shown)


src/query/reverse_deps.go
```diff
@@ -1,24 +1,23 @@
...
-
-	"core"
 )
 
 // ReverseDeps For each input label, finds all targets which depend upon it.
-func ReverseDeps(graph *core.BuildGraph, labels []core.BuildLabel) {
-
-	uniqueTargets := make(map[core.BuildLabel]struct{})
+func ReverseDeps(state *core.BuildState, labels []core.BuildLabel) {
+	uniqueTargets := make(map[core.BuildLabel]*core.BuildTarget)
```
peterebden: I don't know the labels are very useful (the target has them anyway); might be worth turning this into a `map[*core.BuildTarget]struct{}`

BNMetrics: Good point, yeah

PR #389, opened 2018-08-15
Enabled abbreviation syntax for subrepos, ticket 386

BNMetrics:
- now supports abbreviation syntax for build target with the same subrepo and package name. for example: `@unittest_cpp//:unittest_cpp` can be shortened to `@unittest_cpp`
- Also added documentation in `subrepo` section

### DIFF
--- a/docs/subrepos.html
+++ b/docs/subrepos.html
@@ -51,3 +51,17 @@ cc_test(
 <p>Subrepos also underpin <a href="cross_compiling.html">cross-compiling</a> and share the same
   syntax; you can use that to reference architectures as well. There is currently some ambiguity
   here and so it is best not to define subrepo names that match cross-compile architectures.</p>
+
+<p>If the subrepo and the package names are the same, for example, <code>@unittest_cpp//:unittest_cpp</code>,
+    the build label reference to the target can be abbreviated, like so:</p>
+
+<pre><code>subinclude("@pleasings//go:go_bindata")
+
+cc_test(
+    name = "my_test",
+    ...
+    deps = [
+        "@unittest_cpp",
+    ],
+)
+</code></pre>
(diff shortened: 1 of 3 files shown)


docs/subrepos.html
```diff
@@ -51,3 +51,17 @@ <h2>Using subrepos</h2>
 <p>Subrepos also underpin <a href="cross_compiling.html">cross-compiling</a> and share the same
   syntax; you can use that to reference architectures as well. There is currently some ambiguity
   here and so it is best not to define subrepo names that match cross-compile architectures.</p>
+
+<p>If the subrepo and the package names are the same, for example, <code>@unittest_cpp//:unittest_cpp</code>,
+    the build label reference to the target can be abbreviated, like so:</p>
+
+<pre><code>subinclude("@pleasings//go:go_bindata")
```
peterebden: remove the `subinclude` bit, it's not really germane to the example

src/core/build_label_test.go
```diff
@@ -94,6 +94,16 @@ func TestSubrepoLabel(t *testing.T) {
 	assert.EqualValues(t, BuildLabel{PackageName: "", Name: ""}, label.SubrepoLabel())
 }
 
+func TestParseBuildLabelParts(t *testing.T) {
+	target1 := "@unittest_cpp//:unittest_cpp"
+	targetNewSyntax := "@unittest_cpp"
+	pkg, name, subrepo := parseBuildLabelParts(target1, "/", nil)
+	pkg2, name2, subrepo2 := parseBuildLabelParts(targetNewSyntax, "/", nil)
+	assert.Equal(t, pkg, pkg2)
```
peterebden: probably better to assert the values directly; technically you could pass this test with an implementation that always returned "" for the subrepo for example.

PR #391, opened 2018-08-16
Provide a way to filter an arbitrary list of targets

agenticarus:
Frequently I want to perform some plz operation that returns a list of targets, and then filter those targets for certain properties (e.g. labels).

Plz itself has `include` and `exclude` flags but these do not seem to be applied to output lists.


VERSION
```diff
@@ -1 +1 @@
-13.1.0
+13.1.1-beta1
```
peterebden: not sure it's worth having a beta for a patch release?

src/please.go
```diff
@@ -333,6 +333,13 @@ var opts struct {
 				Targets []core.BuildLabel `positional-arg-name:"targets" description:"Targets to query" required:"true"`
 			} `positional-args:"true"`
 		} `command:"roots" description:"Show build labels with no dependents in the given list, from the list."`
+		Filter struct {
+			IncludeLabels    []string `long:"in" description:"Include any targets with matching labels"`
```
peterebden: can't these just use the global `include` and `exclude` flags?

src/query/filter.go
```diff
@@ -0,0 +1,32 @@
+package query
+
+import (
+	"core"
```
peterebden: nit: local imports like core should go in a separated group after stdlib ones.

src/query/filter.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+	"cli"
+)
+
+// Filter takes the list of BuildLabels and checks which ones match the label selectors passed in.
+// If no includeLabels are provided, the default behaviour is to include all labels.
+func Filter(graph *core.BuildGraph, labels core.BuildLabels, includeLabels []string, excludeLabels []string) {
+	matcher := func(target *core.BuildTarget) bool {
```
peterebden: I think `ExpandOriginalTargets` will do this for you if you set the top-level flags?

src/please.go
```diff
@@ -614,6 +621,11 @@ var buildFunctions = map[string]func() bool{
 			query.Roots(state.Graph, opts.Query.Roots.Args.Targets)
 		})
 	},
+	"filter": func() bool {
+		return runQuery(true, opts.Query.Filter.Args.Targets, func(state *core.BuildState) {
```
peterebden: I think you can get away with `false` for the first argument (i.e. you do not need a full dependency parse, just having the definitions of these targets will suffice)

Issue #390, opened 2018-08-15
plz watch should share persistent workers with its inferior processes


peterebden:
Right now `plz watch` sets up the inotify watches etc and spawns subprocesses when a build is needed. It would be cool (and a lot faster for certain workflows) if it could provide those subprocesses with any persistent workers it had started.

PR #394, opened 2018-08-17
Addressed the issue #390, now worker persist in plz watch

BNMetrics:
- Changed the way //src/watch:watch works, it now takes the actual function to run subsequent build that's being watched instead of running subprocesses.
- Did some refactor to the main please.go. Separated the main function to its specific initBuild() function, which sets up the flags and returns a command to call the buildFunctions. 
- Separated `watch` out of `buildFunctions` map and into its specific function, as it needs to call the subsequent `buildFunctions` .

Everything should work as expected, and the worker will persist! 
Test it out with: `plz run src:please -- watch //test/python_rules:worker_test --show_all_output --nocache`


src/watch/stub.go
```diff
@@ -7,4 +7,4 @@ package watch
 import "core"
 
 // Watch is a stub implementation of the real function in watch.go, this one does nothing.
-func Watch(state *core.BuildState, labels []core.BuildLabel, run bool) {}
+func Watch(state *core.BuildState, labels core.BuildLabels, run bool, runWatchededBuild func(args []string)) {}
```
peterebden: `runWatchedBuild`

src/watch/watch.go
```diff
@@ -26,7 +24,7 @@ const debounceInterval = 50 * time.Millisecond
 // Watch starts watching the sources of the given labels for changes and triggers
 // rebuilds whenever they change.
 // It never returns successfully, it will either watch forever or die.
-func Watch(state *core.BuildState, labels []core.BuildLabel, run bool) {
+func Watch(state *core.BuildState, labels core.BuildLabels, run bool, runWatchededBuild func(args []string)) {
```
peterebden: I don't think it's ideal to pass command-line arguments back and forth as an interface; among other things this package doesn't know how to reconstruct them.

I think the callback function should probably take a list of the build labels that this package thinks it should build, and hook back into the build process with those (assuming that's even necessary... maybe we just always build the same set every time?).

src/please.go
```diff
@@ -616,6 +609,21 @@ var buildFunctions = map[string]func() bool{
 	},
 }
 
+// Run "plz watch.."
+// This guy is a bit different than the rest as it gets the function from the other commands to run as well
```
peterebden: It'd be nice if this was not different from the rest... can it not be a function like the others that basically builds then calls Watch to watch stuff?

peterebden: I don't see why that matters? It can still call other functions in this map from inside itself, there's nothing stopping that?

src/please.go
```diff
@@ -702,7 +705,7 @@ func newCache(config *core.Configuration) core.Cache {
 }
 
 // Please starts & runs the main build process through to its completion.
-func Please(targets []core.BuildLabel, config *core.Configuration, prettyOutput, shouldBuild, shouldTest bool) (bool, *core.BuildState) {
+func Please(targets []core.BuildLabel, config *core.Configuration, prettyOutput bool, shouldBuild, shouldTest bool) (bool, *core.BuildState) {
```
peterebden: nit: extra `bool` is unnecessary

src/please.go
```diff
@@ -733,7 +736,8 @@ func Please(targets []core.BuildLabel, config *core.Configuration, prettyOutput,
 	state.SetIncludeAndExclude(opts.BuildFlags.Include, opts.BuildFlags.Exclude)
 	parse.InitParser(state)
 	build.Init(state)
-	if config.Events.Port != 0 && shouldBuild {
+
+	if config.Events.Port != 0 && state.NeedBuild {
```
peterebden: guess this doesn't really matter, but these are equivalent right? you assign `state.NeedBuild = shouldBuild` back on line 726.

src/please.go
```diff
@@ -995,7 +999,23 @@ func main() {
...
+	runWatchedBuild = func(args []string) {
+		command := initBuild(args)
+		buildFunctions[command]()
+	}
+
 	if !buildFunctions[command]() {
 		os.Exit(7) // Something distinctive, is sometimes useful to identify this externally.
 	}
+
+	metrics.Stop()
```
peterebden: these should happen before the if statement above - they should be stopped on any exit, even if unsuccessful.

src/please.go
```diff
@@ -995,7 +999,23 @@ func main() {
...
 
+	return command
+}
+
+
+
+func main() {
+	command := initBuild(os.Args)
+
+	runWatchedBuild = func(args []string) {
```
peterebden: I still don't think this should need to receive `args`. If you reuse the options you already have, all the flags will probably just get honoured automatically, no?

src/please.go
```diff
@@ -676,6 +679,35 @@ func please(tid int, state *core.BuildState, parsePackageOnly bool, include, exc
...
+
+	for i, label := range labels {
+		if state.Graph.TargetOrDie(label).IsTest {
+			if i == 0 {
+				opts.Test.Args.Target = label
+			}
+			opts.Test.Args.Args = append(opts.Test.Args.Args, label.String())
+		}
+
+		if i == len(labels) - 1 {
```
peterebden: what is this for? looks like an odd piece of logic

src/please.go
```diff
@@ -676,6 +679,35 @@ func please(tid int, state *core.BuildState, parsePackageOnly bool, include, exc
...
 
+// set the watch
+func setWatchedTarget(state *core.BuildState, labels []core.BuildLabel) string {
+	if opts.Watch.Run {
+		if len(labels) > 1 {
+			opts.Run.Parallel.PositionalArgs.Targets = opts.Watch.Args.Targets
+		} else {
+			opts.Run.Args.Target = opts.Watch.Args.Targets[0]
+		}
+		return "run"
```
peterebden: I don't think this is always right - if there is more than one label it needs to return "parallel"

src/please.go
```diff
@@ -676,6 +679,35 @@ func please(tid int, state *core.BuildState, parsePackageOnly bool, include, exc
 	}
 }
 
+// set the watch
+func setWatchedTarget(state *core.BuildState, labels []core.BuildLabel) string {
+	if opts.Watch.Run {
+		if len(labels) > 1 {
+			opts.Run.Parallel.PositionalArgs.Targets = opts.Watch.Args.Targets
```
peterebden: shouldn't this be ` = labels`?

src/please.go
```diff
@@ -676,6 +679,35 @@ func please(tid int, state *core.BuildState, parsePackageOnly bool, include, exc
...
+		}
+		return "run"
+	}
+
+	for i, label := range labels {
+		if state.Graph.TargetOrDie(label).IsTest {
+			if i == 0 {
+				opts.Test.Args.Target = label
+			}
+			opts.Test.Args.Args = append(opts.Test.Args.Args, label.String())
```
peterebden: you need to empty this slice somewhere, or it will grow unboundedly the more times the watch kicks in.

peterebden: not done?

It may be cleaner to do something like
```
if allTargetsAreTests(labels) {
    opts.Test.Args.Target = labels[0]
    opts.Test.Args.Args = labels[1:]
    return "test"
}
```

src/please.go
```diff
@@ -1005,7 +1040,23 @@ func main() {
...
+func main() {
+	command := initBuild(os.Args)
+
+	runWatchedBuild = func(watchedProcessName string) {
+		buildFunctions[watchedProcessName]()
+	}
+	success := buildFunctions[command]()
+
+	metrics.Stop()
+	build.StopWorkers()
```
peterebden: can you confirm that these still get stopped even if you `plz run` a target? In that case I don't think it will ever hit this line.

src/watch/watch.go
```diff
@@ -5,10 +5,7 @@ package watch
 
 import (
 	"fmt"
-	"os"
-	"os/exec"
-	"path"
-	"strings"
+		"path"
```
peterebden: indentation looks wrong. gofmt?

src/watch/watch_test.go
```diff
@@ -3,20 +3,8 @@ package watch
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
-
-	"core"
-)
+		)
 
 func TestCommands(t *testing.T) {
```
peterebden: Delete this whole test if there's nothing left in it.

src/please.go
BNMetrics: Oh, I have moved the call to this function to the `watch` function, so it only gets called once and no longer needs clean up.

src/please.go
BNMetrics: Yes! I can definitely confirm that.

test/python_rules/BUILD
```diff
@@ -147,3 +147,11 @@ python_test(
     ],
     deps = ["//third_party/python:scipy"],
 )
+
+
+python_test(
+    name = "worker_test",
+    srcs = ["worker_test.py"],
+    worker = "//test/workers:worker",
+
```
peterebden: nit: not formatted. please run buildifier on it

test/python_rules/worker_test.py
```diff
@@ -0,0 +1,10 @@
+import unittest
+
+"""
```
peterebden: Docstring should be at the top of the file, no?

BNMetrics: ooopssy!

Issue #393, opened 2018-08-17
coverage.xml generated by `plz cover` for `python_test` target is not compatible with IntelliJ


jameshiew:
To reproduce, run `plz cover //path/to/python_test:target`

Console output is fine, like:
```
1 test target and 10 tests run in 0s; 10 passed. Total time 8.96s.
Coverage results:
path/to/python_test:
  target.py 25/34 lines, 73.5%
Total coverage:  25/34 lines, 73.5%
```

`coverage.xml`
```xml
<?xml version="1.0" encoding="UTF-8"?>
<coverage line-rate="0.7353" branch-rate="0" lines-covered="25" lines-valid="34" branches-covered="0" branches-valid="0" complexity="0" version="" timestamp="1534503093452">
	<sources>
		<source>//path/to/python_test:target</source>
	</sources>
	<packages>
		<package name="//path/to/python_test/target" line-rate="0.7353" branch-rate="0" complexity="0" line-count="0" line-hits="0">
			<classes>
				<class name="path/to/python_test/target.py" filename="path/to/python_test/target.py" line-rate="0.7353" branch-rate="0" complexity="0">
					<methods></methods>
					<lines>
						<line number="0" hits="1"></line>
						<line number="3" hits="1"></line>
						<line number="4" hits="0"></line>
						<line number="7" hits="1"></line>
						<line number="27" hits="1"></line>
						<line number="28" hits="1"></line>
						<line number="31" hits="1"></line>
						<line number="32" hits="1"></line>
						<line number="38" hits="1"></line>
						<line number="40" hits="1"></line>
						<line number="43" hits="1"></line>
						<line number="44" hits="1"></line>
						<line number="45" hits="1"></line>
						<line number="46" hits="1"></line>
						<line number="47" hits="1"></line>
						<line number="48" hits="1"></line>
						<line number="53" hits="1"></line>
						<line number="57" hits="1"></line>
						<line number="62" hits="1"></line>
						<line number="63" hits="1"></line>
						<line number="65" hits="1"></line>
						<line number="69" hits="1"></line>
						<line number="72" hits="1"></line>
						<line number="76" hits="0"></line>
						<line number="81" hits="0"></line>
						<line number="82" hits="0"></line>
						<line number="84" hits="0"></line>
						<line number="88" hits="0"></line>
						<line number="93" hits="0"></line>
						<line number="94" hits="0"></line>
						<line number="95" hits="0"></line>
						<line number="98" hits="1"></line>
[...]

PR #395, opened 2018-08-20
Fixed the reported issue regarding #393

BNMetrics:
This PR is related to the ticket [393](https://github.com/thought-machine/please/issues/393)
- Added absolute path in coverage.xml generation, as intellij only recognize absolute path. 
- Added relative path to class names, same reason as above

### DIFF
--- a/src/test/xml_coverage.go
+++ b/src/test/xml_coverage.go
@@ -52,7 +52,7 @@ func coverageResultToXML(sources []core.BuildLabel, coverage core.TestCoverage) 
 	// get the string representative of sources
 	var sourcesAsStr []string
 	for _, source := range sources {
-		sourcesAsStr = append(sourcesAsStr, path.Join(core.RepoRoot, "/", source.PackageName))
+		sourcesAsStr = append(sourcesAsStr, path.Join(core.RepoRoot, source.PackageName))
 	}
 
 	// Get the list of packages for <package> tag in the coverage xml file
--- a/src/test/xml_coverage.go
+++ b/src/test/xml_coverage.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 	"math"
+		"os"
 )
 
 func parseXMLCoverageResults(target *core.BuildTarget, coverage *core.TestCoverage, data []byte) error {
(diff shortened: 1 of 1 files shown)


src/test/xml_coverage.go
```diff
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 	"math"
+		"os"
```
peterebden: please run `gofmt` on this file (and all Go files)

src/test/xml_coverage.go
```diff
@@ -48,16 +49,22 @@ func coverageResultToXML(sources []core.BuildLabel, coverage core.TestCoverage)
 	linesCovered := 0
 	validFiles := coverage.OrderedFiles()
 
+	cwd, err := os.Getwd()
```
peterebden: I don't think you should find the working directory here. Use `core.RepoRoot` if you need to know it - but I don't think you should be encoding absolute paths into the output file.

BNMetrics: I just tried it out, using relative path would show 0 covered in intellij. :(

peterebden: Intellij is terrible :(

I'm not sure what is best then. These files can get shared through the cache, having absolute paths in them is not a good idea.

agenticarus: Perhaps you could write them with relative paths and have a separate post-process step that outputs the file for intellij that doesn't get cached (this is likely to help if we ever want to support other IDEs or formats too)?

peterebden: Hmmm on further thought the output file is *not* cached so I guess that's OK.

I still think it's rubbish that Intellij doesn't handle relative paths though.

src/test/xml_coverage.go
```diff
@@ -51,13 +51,13 @@ func coverageResultToXML(sources []core.BuildLabel, coverage core.TestCoverage)
 	// get the string representative of sources
 	var sourcesAsStr []string
 	for _, source := range sources {
-		sourcesAsStr = append(sourcesAsStr, source.String())
+		sourcesAsStr = append(sourcesAsStr, core.RepoRoot+"/"+source.PackageName)
```
peterebden: use `path.Join` to join paths rather than concatenating.

src/test/xml_coverage.go
```diff
@@ -51,13 +52,13 @@ func coverageResultToXML(sources []core.BuildLabel, coverage core.TestCoverage)
 	// get the string representative of sources
 	var sourcesAsStr []string
 	for _, source := range sources {
-		sourcesAsStr = append(sourcesAsStr, source.String())
+		sourcesAsStr = append(sourcesAsStr, path.Join(core.RepoRoot, "/", source.PackageName))
```
peterebden: you don't need the "/" if you're using JoinPath - let it add it for you if necessary.

PR #404, opened 2018-09-06
Added classpath as an argument to jarcat

BNMetrics:
As the manifest file to compile scala to jar files will need a classpath referencing to the scala std library

### DIFF
--- a/tools/jarcat/main.go
+++ b/tools/jarcat/main.go
@@ -66,6 +66,7 @@ var opts = struct {
 		PreambleFrom          string            `long:"preamble_from" description:"Read the first line of this file and use as --preamble."`
 		PreambleFile          string            `long:"preamble_file" description:"Concatenate zip file onto the end of this file"`
 		MainClass             string            `short:"m" long:"main_class" description:"Write a Java manifest file containing the given main class."`
+		ClassPath			  string			`long:"class_path" description:"Write a Java manifest file containing the given class-path."`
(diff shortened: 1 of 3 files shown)


tools/jarcat/main.go
```diff
@@ -161,6 +162,8 @@ func main() {
 	f.Align = opts.Zip.Align
 	f.Prefix = opts.Zip.Prefix
 
+	must(f.AddManifest(opts.Zip.MainClass, opts.Zip.ClassPath))
```
peterebden: this should still be conditional - before it only wrote it if main_class was passed, now it should only write if main_class or class_path are passed.

tools/jarcat/zip/writer.go
```diff
@@ -338,8 +338,20 @@ func (f *File) AddInitPyFiles() error {
 }
 
 // AddManifest adds a manifest to the given zip writer with a Main-Class entry (and a couple of others)
-func (f *File) AddManifest(mainClass string) error {
-	manifest := fmt.Sprintf("Manifest-Version: 1.0\nMain-Class: %s\n", mainClass)
+func (f *File) AddManifest(mainClass string, classPath string) error {
+
+	if mainClass == "" && classPath == "" {
```
peterebden: not sure about this. I think this function should always write the manifest, the decision of whether to do it or not should be done by the caller.

peterebden:
Is this still a thing we need? The Scala stuff is working fine without, right?

BNMetrics:
Yeah, we don't actually need this anymore

PR #406, opened 2018-09-07
Fixed the python_wheel issue

BNMetrics:
The problem was, if patches were passed, the `$SRC` environment is no longer valid, as patches will be appended to `srcs`.

### DIFF
--- a/src/parse/rules/python_rules.build_defs
+++ b/src/parse/rules/python_rules.build_defs
@@ -472,18 +472,16 @@ def python_wheel(name:str, version:str, hashes:list=None, package_name:str=None,
         licences = licences if licences else None,
     )
 
-    if patch:
-        patches = [patch] if isinstance(patch, str) else patch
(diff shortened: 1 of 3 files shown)


src/parse/rules/python_rules.build_defs
```diff
@@ -472,16 +472,18 @@ def python_wheel(name:str, version:str, hashes:list=None, package_name:str=None,
         licences = licences if licences else None,
     )
 
-    cmd = ['$TOOL x $SRC']
+    if patch:
+        patches = [patch] if isinstance(patch, str) else patch
+        cmd = ['$TOOL x $SRCS'] + [f'patch -p0 --no-backup-if-mismatch < $(location {p})' for p in patches]
```
peterebden: I don't think this is right still? if there is a patch $SRCS will contain both the patch and the wheel file and we don't want it to extract both.

Issue #405, opened 2018-09-06
name conflict when building go binary


iwinux:
## Meta

```
$ cat .plzconfig
[please]
version = 13.1.0

$ find . -type f
./smoke/BUILD
./smoke/main.go
```

## Problem

With the following `BUILD` file:

```python
go_binary(
    name='smoke',
    srcs=['main.go'],
)
```

and `main.go`:

```go
package main

func main() {
}
```

When I run `plz run`, it fails to build:

```
Build stopped after 70ms. 1 target failed:
    //smoke:smoke
Error building target //smoke:smoke: exit status 1
/usr/local/Cellar/go/1.10.3/libexec/pkg/tool/darwin_amd64/link: cannot create /src/plz-out/tmp/smoke/smoke._build/smoke: open /src/plz-out/tmp/smoke/smoke._build/smoke: is a directory
```

But if I change `name='smoke'` to any other value, it builds as expected.

BNMetrics:
Fixed in PR #407

PR #407, opened 2018-09-08
Fix for #405

BNMetrics:
Fixed the issue for the name conflicts when the output name is the same as the package name.
This is because the `go_library` created as a dependency for `go_binary` would generate a directory in plz-out/tmp/, then the `go tool link` tries to generate a binary output file under the same name, thus the conflict. 

To test it, do this command:
`plz run //test:test`

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -329,8 +329,8 @@ func moveOutputs(state *core.BuildState, target *core.BuildTarget) ([]string, bo
 	changed := false
 	tmpDir := target.TmpDir()
 	outDir := target.OutDir()
-	for tmp, output := range target.GetOutputsMapping(target.Outputs()) {
-		tmpOutput := path.Join(tmpDir, tmp)
+	for _, output := range target.Outputs() {
+		tmpOutput := path.Join(tmpDir, target.GetTmpOutput(output))
 		realOutput := path.Join(outDir, output)
 		if !core.PathExists(tmpOutput) {
 			return nil, true, fmt.Errorf("Rule %s failed to create output %s", target.Label, tmpOutput)
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -329,13 +329,8 @@ func moveOutputs(state *core.BuildState, target *core.BuildTarget) ([]string, bo
 	changed := false
 	tmpDir := target.TmpDir()
 	outDir := target.OutDir()
-	for _, output := range target.Outputs() {
-		var tmpOutput string
-		if output == target.Label.PackageName {
-			tmpOutput = path.Join(tmpDir, fmt.Sprintf(core.TmpOutputFormat, output))
-		} else {
-			tmpOutput = path.Join(tmpDir, output)
-		}
+	for tmp, output := range target.GetOutPutsMapping(target.Outputs()) {
+		tmpOutput := path.Join(tmpDir, tmp)
 		realOutput := path.Join(outDir, output)
 		if !core.PathExists(tmpOutput) {
 			return nil, true, fmt.Errorf("Rule %s failed to create output %s", target.Label, tmpOutput)
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -332,7 +332,7 @@ func moveOutputs(state *core.BuildState, target *core.BuildTarget) ([]string, bo
 	for _, output := range target.Outputs() {
 		var tmpOutput string
 		if output == target.Label.PackageName {
-			tmpOutput = path.Join(tmpDir, fmt.Sprintf("%s.out", output))
+			tmpOutput = path.Join(tmpDir, fmt.Sprintf(core.TmpOutputFormat, output))
 		} else {
 			tmpOutput = path.Join(tmpDir, output)
 		}
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -330,7 +330,12 @@ func moveOutputs(state *core.BuildState, target *core.BuildTarget) ([]string, bo
 	tmpDir := target.TmpDir()
 	outDir := target.OutDir()
 	for _, output := range target.Outputs() {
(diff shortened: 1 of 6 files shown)


src/core/build_env.go
```diff
@@ -14,6 +15,10 @@ var ExpandHomePath func(string) string = fs.ExpandHomePath
 // A BuildEnv is a representation of the build environment that also knows how to log itself.
 type BuildEnv []string
 
+// TmpOutputFormat is a Temporary formatting for outputs, this is used to avoid name conflicts
+// when package directory name and output are the same
+var TmpOutputFormat = "%s.out"
```
peterebden: this should be private

src/core/build_env.go
```diff
@@ -106,6 +114,20 @@ func BuildEnvironment(state *BuildState, target *BuildTarget) BuildEnv {
 	return env
 }
 
+// Get the outputs for the environment variable
+// Check if each output has the same name as the package, this avoids the name conflict issue with go link tool
```
peterebden: please don't blame `go link`, it's not its fault.

src/parse/rules/go_rules.build_defs
```diff
@@ -283,7 +283,7 @@ def go_binary(name:str, srcs:list=[], asm_srcs:list=[], out:str=None, deps:list=
     cmds, tools = _go_binary_cmds(static=static)
     return build_rule(
         name=name,
-        srcs=[f':_{name}#lib'],
+        srcs=[lib],
```
peterebden: nice!

src/build/build_step.go
```diff
@@ -330,7 +330,12 @@ func moveOutputs(state *core.BuildState, target *core.BuildTarget) ([]string, bo
 	tmpDir := target.TmpDir()
 	outDir := target.OutDir()
 	for _, output := range target.Outputs() {
-		tmpOutput := path.Join(tmpDir, output)
+		var tmpOutput string
+		if output == target.Label.PackageName {
```
peterebden: this logic seems duplicated from the other new function. I think it should probably be a member function on `BuildTarget`

src/core/build_target.go
```diff
@@ -459,6 +463,38 @@ func (target *BuildTarget) NamedOutputs(name string) []string {
 	return nil
 }
 
+// GetOutPutsMapping returns a mapping of temporary output(plz-out/tmp/) to real output(plz-out/bin/)
+// Check if each output has the same name as the package, this avoids the name conflict issue
+func (target *BuildTarget) GetOutPutsMapping(parseOutputs []string) map[string]string {
+	var outputMapping = make(map[string]string, len(parseOutputs))
+	var tmpOutput []string
```
peterebden: you don't need this variable?

src/core/build_target.go
```diff
@@ -459,6 +463,38 @@ func (target *BuildTarget) NamedOutputs(name string) []string {
 	return nil
 }
 
+// GetOutPutsMapping returns a mapping of temporary output(plz-out/tmp/) to real output(plz-out/bin/)
```
peterebden: `GetOutputsMapping` ("output" is one word so the p shouldn't be capitalised)

src/core/build_target.go
```diff
@@ -31,6 +31,10 @@ const DefaultBuildingDescription = "Building..."
 const buildDirSuffix = "._build"
 const testDirSuffix = "._test"
 
+// TmpOutputFormat is a Temporary formatting for outputs, this is used to avoid name conflicts
+// when package directory name and output are the same
+const tmpOutputFormat = "%s.out"
```
peterebden: probably just use this inside the function, don't think it needs a separate constant.

src/build/build_step.go
```diff
@@ -329,8 +329,8 @@ func moveOutputs(state *core.BuildState, target *core.BuildTarget) ([]string, bo
 	changed := false
 	tmpDir := target.TmpDir()
 	outDir := target.OutDir()
-	for _, output := range target.Outputs() {
-		tmpOutput := path.Join(tmpDir, output)
+	for tmp, output := range target.GetOutPutsMapping(target.Outputs()) {
```
peterebden: Be careful with maps because they iterate in indeterminate order - I don't think it matters here but it would if we hashed them in the map's order.

Might be easier to just get both slices and iterate in parallel, then you don't need the mapping function.

BNMetrics: But what if the order of the different slices don't coordinate? like the tmpOutput doesn't map to the real output? I don't know if the order of the elements in the map actually matters, because it's the output for one specific target.

peterebden: As discussed, let's try having a function that maps a real output to the temp one, and skip having the map completely.

src/core/build_target.go
```diff
@@ -459,6 +459,44 @@ func (target *BuildTarget) NamedOutputs(name string) []string {
 	return nil
 }
 
+// GetOutputsMapping returns a mapping of temporary output(plz-out/tmp/) to real output(plz-out/bin/)
```
peterebden: I think you can remove this now - don't think we call it anywhere?

src/core/build_target.go
```diff
@@ -459,6 +459,44 @@ func (target *BuildTarget) NamedOutputs(name string) []string {
...
+
+	return outputMapping
+}
+
+
+// GetTmpOutput takes the original output filename as an argument, and returns a temporary output
+// filename(plz-out/tmp/) if output has the same name as the package, this avoids the name conflict issue
+func (target *BuildTarget) GetTmpOutput(parseOutput string) string {
+	if parseOutput == target.Label.PackageName {
+		return fmt.Sprintf("%s.out", parseOutput)
```
peterebden: nit: just write this as `parseOutput + ".out"` - that's slightly more efficient since there's no format string to parse.

PR #408, opened 2018-09-11
Adds support for "build providers"

peterebden:
Basically it can hit up an external service to determine what a BUILD file for a directory would look like. A basic example is provided for Go code.

Ideally this should make it easier / possible to gradually introduce plz into a repo by stitching in BUILD files where needed but deriving them for existing bits of code. Could also imagine doing the same for implicitly converting existing build metadata - e.g. Makefiles or pom.xml, although that doesn't seem like great fun.


src/build/build_step.go
```diff
@@ -675,3 +677,39 @@ func removeIfSymlink(name string) {
...
+	if !resp.Success {
+		return nil, fmt.Errorf("Error building target %s: %s", target.Label, out)
+	}
+	// Okay, now we might need to do something locally too...
+	if localCmd != "" {
+		out2, err := runBuildCommand(state, target, localCmd, inputHash)
+		return append([]byte(out+"\n"), out2...), err
+	}
+	return []byte(out), nil
+}
```
BNMetrics: Am I missing something@? because this function is already in master from I have seen?

BNMetrics: Yeah, that makes sense!

PR #412, opened 2018-09-21
Adds a pom.xml for the javac worker stuff

peterebden:
We download this in a couple of other places by downloading plz tarballs and extracting the thing we want. Seems a lot easier if we just uploaded it to maven so we could use a maven_jar.


peterebden:
Good spot, should 100% be `build.please` in the name. Derp.

Agreed it would be nice to provide sources / javadoc; will think about those tomorrow (i.e. when I'm on a machine with Java installed...).

peterebden:
Right, javadoc and source jars are now provided. Will punt the build_def out to pleasings shortly, I suspect.

PR #413, opened 2018-09-28
Add `plz query changed`

agenticarus:
`plz query changes` is a powerful tool that requires performing multiple builds and changing the state of the workspace - `plz query changed` is a fairly direct port from PantsBuild of a similar but less accurate change detector that does not require extra builds or mutating the directory state.

It will not pick up all changes (such as adding existing files to existing targets) but I intend this to be useful as a best-effort "what have my local changes modified, and what tests should I run?" step

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -92,11 +92,10 @@ var opts struct {
 	VisibilityParse  bool   `description:"Parse all targets that the original targets are visible to. Used for some query steps." no-flag:"true"`
 
 	Build struct {
-		Prepare    bool `long:"prepare" description:"Prepare build directory for these targets but don't build them."`
-		Shell      bool `long:"shell" description:"Like --prepare, but opens a shell in the build directory with the appropriate environment variables."`
-		ShowStatus bool `long:"show_status" hidden:"true" description:"Show status of each target in output after build"`
-		Args       struct {
-			// Inner nesting is necessary to make positional-args work :(
+		Prepare    bool     `long:"prepare" description:"Prepare build directory for these targets but don't build them."`
+		Shell      bool     `long:"shell" description:"Like --prepare, but opens a shell in the build directory with the appropriate environment variables."`
+		ShowStatus bool     `long:"show_status" hidden:"true" description:"Show status of each target in output after build"`
+		Args       struct { // Inner nesting is necessary to make positional-args work :(
 			Targets []core.BuildLabel `positional-arg-name:"targets" description:"Targets to build"`
 		} `positional-args:"true" required:"true"`
 	} `command:"build" description:"Builds one or more targets"`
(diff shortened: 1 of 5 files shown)


src/please.go
```diff
@@ -339,6 +339,11 @@ var opts struct {
 				Targets []core.BuildLabel `positional-arg-name:"targets" description:"Targets to filter"`
 			} `positional-args:"true"`
 		} `command:"filter" description:"Filter the given set of targets according to some rules"`
+		Changed struct {
+			Since            string `long:"since" description:"Calculate changes since this tree-ish/scm ref (defaults to current HEAD/tip)."`
+			DiffSpec         string `long:"diffspec" description:"Calculate changes contained within given scm spec (commit range/sha/ref/etc)."`
+			IncludeDependees string `long:"include-dependees" default:"none" description:"Include direct or transitive dependees of changed targets."`
```
peterebden: FYI if this is an enum (which it looks like it is?) then you can do things like `choice:"none" choice:"direct"` to enforce it & get helpful help.

src/query/changed.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// ChangedLabels returns all BuildLabels that have changed according to the given request.
+func ChangedLabels(state *core.BuildState, request ChangedRequest) []core.BuildLabel {
+	workspaceChangedFiles := changedFiles(request.Since, request.DiffSpec)
+	if len(workspaceChangedFiles) == 0 {
+		return []core.BuildLabel{}
+	}
+
+	labels := make([]core.BuildLabel, 0)
```
peterebden: nit: I probably wouldn't bother with `make` if it's got an initial capacity of 0, just `[]core.BuildLabel{}` is a little simpler

src/query/changed.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	return scm.ChangedFiles(since, true, "")
+}
+
+func targetsForChangedFiles(graph *core.BuildGraph, files []string, includeDependees string) []*core.BuildTarget {
+	addresses := make([]*core.BuildTarget, 0)
+	for _, target := range graph.AllTargets() {
+		for _, source := range target.Sources {
+			if source.Label() == nil {
```
peterebden: you might find `target.HasSource` or `target.HasAbsoluteSource` useful - it should handle the source loops here for you.

src/query/changed.go
```diff
@@ -0,0 +1,94 @@
...
+	i := 0
+	for k := range dependents {
+		keys[i] = k
+		i++
+	}
+
+	return keys
+}
+
+func visit(dependents map[*core.BuildTarget]struct{}, target *core.BuildTarget, f func(*core.BuildTarget) []*core.BuildTarget) {
```
peterebden: you might want to track whether you have seen targets already in here? some other query things do that, it can make a big difference when dep trees become large.

agenticarus:
Agreed

PR #415, opened 2018-09-28
Basic structure of LSP and the initialization part

BNMetrics:
As titled.
- Rough draft stage
- Many things are subjected to change


tools/build_langserver/langserver/handler.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func (h *LsHandler) handleExit(ctx context.Context, request *jsonrpc2.Request) (result interface{}, err error) {
+	h.handleShutDown(ctx, request)
+	h.conn.Close()
+	return nil, nil
+}
+
+func (h *LsHandler) handleCancel(ctx context.Context, request *jsonrpc2.Request) (result interface{}, err error) {
+	// TODO(bnmetrics): rethink this, I'm probably doing this wrong lol
```
peterebden: Yeah it looks like it will need more complexity - like the server will need to keep a map of request id -> context cancel function, and this function would cancel the appropriate one.

Tbh sounds like a classic thing we can put off for later :)

tools/build_langserver/langserver_main.go
```diff
@@ -0,0 +1,75 @@
...
+
+// TODO(bnmetrics): also think about how we can implement this with .build_defs as well
+// TODO(bnmetrics): Make the Usage part better
+var log = logging.MustGetLogger("build_langserver")
+
+var opts = struct {
+	Usage     string
+	Verbosity cli.Verbosity `short:"v" long:"verbosity" default:"warning" description:"Verbosity of output (higher number = more output)"`
+
+	Mode string `short:"m" long:"mode" default:"stdio" description:"Mode of the language server communication (stdio|tcp)"`
```
peterebden: FYI the flags thing can do this for you:
```
choice:"stdio" choice:"tcp" description:"Mode of the language server communication"
```
It'll enforce the appropriate values & generate nice help too.

tools/build_langserver/langserver_main.go
```diff
@@ -0,0 +1,75 @@
...
+	handler := langserver.NewHandler()
+
+	serve(handler)
+}
+
+func serve(handler jsonrpc2.Handler) {
+	switch opts.Mode {
+	case "stdio":
+		// TODO(bnmetrics): complete this
+	case "tcp":
```
peterebden: aren't these the wrong way around? isn't this doing stdin/stdout?

Also if you do the flags thing, you can probably skip the switch and default case and just do
```
if opts.Mode == "tcp" {
} else {
    // stdio stuff
}
```
because you know it's always one of the two.

tools/build_langserver/outline.txt
```diff
@@ -0,0 +1,72 @@
+LSP
+----
+
+It should work the same way as any of our other tools, such as `jarcat`.
+
+command: plz init-lsp
```
peterebden: I think we should make it `plz tool lsp`, that's normally how we get other processes. It's pretty easy to add a new one to that set (look in src/tool).

tools/build_langserver/langserver/handler.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/sourcegraph/jsonrpc2"
+	"tools/build_langserver/lsp"
+	"gopkg.in/op/go-logging.v1"
+)
+
+var log = logging.MustGetLogger("lsp")
+
+// NewHandler creates a BUILD file language server handler
+func NewHandler() jsonrpc2.Handler {
+	// TODO: need to rethink this
```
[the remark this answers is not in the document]
BNMetrics: Yeah :) I think it should be alright. It was classic me not being confident in what I wrote XD

tools/build_langserver/langserver_main.go
BNMetrics: Yes! I moved things around and got things out of order..oops!
the flag actually take care of the invalid options? Nice!

tools/build_langserver/outline.txt
BNMetrics: :+1:

tools/build_langserver/langserver_main.go
```diff
@@ -0,0 +1,75 @@
...
+// TODO(bnmetrics): also think about how we can implement this with .build_defs as well
+// TODO(bnmetrics): Make the Usage part better
+var log = logging.MustGetLogger("build_langserver")
+
+var opts = struct {
+	Usage     string
+	Verbosity cli.Verbosity `short:"v" long:"verbosity" default:"warning" description:"Verbosity of output (higher number = more output)"`
+
+	Mode string `short:"m" long:"mode" default:"stdio" description:"Mode of the language server communication (stdio|tcp)"`
+	Host string `short:"h" long:"host" default:"127.0.0.20" description:"TCP host to communicate with"`
```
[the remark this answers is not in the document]
BNMetrics: Let's speak in person about this! :)

PR #416, opened 2018-09-30
Release tarballs as xz as well as gz

peterebden:
xzipped tarballs are 40% smaller (srsly - 9MB vs 15MB). For the foreseeable future we'll still upload gzipped versions as well since there's very little penalty to doing so.

For now compression is done by filtering through xz since that produces better output - hopefully the Go one will improve in future and we can get rid of that.


src/parse/rules/misc_rules.build_defs
```diff
@@ -349,7 +349,7 @@ def remote_file(name:str, url:str|list, hashes:list=None, out:str=None, binary:b
 
 
 def tarball(name:str, srcs:list, out:str=None, deps:list=None, subdir:str=None, gzip:bool=True,
-            test_only:bool=False, visibility:list=None, labels:list&features&tags=[]):
+            xzip:bool=False, test_only:bool=False, visibility:list=None, labels:list&features&tags=[]):
```
agenticarus: Rather than passing gzip/xzip/some new format in future, could we guess it from the out name?

peterebden: We could although by default it infers the out name from the other properties - and I kind of prefer being explicit than getting into awkward cases like it not compressing because we didn't recognise `.tgz` or something, or surprises like `.tar.bz2` being uncompressed because we don't do bzip2.

PR #417, opened 2018-10-01
Improve surefire reporting in cases where only a few tests are run

agenticarus:
We also fix the profiling at the same time as this was currently not closing the file correctly. It should now work even in the case of please exiting with failure (from the top level)

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -732,7 +732,7 @@ func doTest(targets []core.BuildLabel, surefireDir cli.Filepath, resultsFile cli
 	os.RemoveAll(string(resultsFile))
 	os.MkdirAll(string(surefireDir), core.DirPermissions)
 	success, state := runBuild(targets, true, true)
-	test.CopySurefireXmlFilesToDir(state.Graph, string(surefireDir))
+	test.CopySurefireXmlFilesToDir(state, string(surefireDir))
 	test.WriteResultsToFileOrDie(state.Graph, string(resultsFile))
 	return success, state
 }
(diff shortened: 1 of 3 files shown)


src/please.go
```diff
@@ -1038,30 +1038,33 @@ func initBuild(args []string) string {
...
 		}()
 	}
+
+	return command
+}
+
+func main() {
+	command := initBuild(os.Args)
+
+	var cpuFile *os.File
```
peterebden: I wonder if it would be neater to kick the whole lot into another function so you can defer things in there, and have it return an exit code that `main` can pass to `os.Exit`. Would be nicer to be able to use `defer` rather than dancing around the potential `os.Exit`.

src/test/surefire.go
```diff
@@ -9,10 +9,10 @@ import (
 )
 
 // CopySurefireXmlFilesToDir copies all the XML test results files into the given directory.
-func CopySurefireXmlFilesToDir(graph *core.BuildGraph, surefireDir string) {
+func CopySurefireXmlFilesToDir(state *core.BuildState, surefireDir string) {
 	outputDirs := make(map[string]struct{})
-	for _, target := range graph.AllTargets() {
-		if target.IsTest && !target.NoTestOutput {
+	for _, target := range state.Graph.AllTargets() {
```
peterebden: possibly this should be something like `state.ExpandOriginalTargets()` rather than iterating the whole graph, which might include targets that you didn't ask it to test?

src/please.go
agenticarus: oh yeah, how did this ever think it would work?

agenticarus:
I'm not sure why the `watch` logic was in a different order but it didn't look super important - I'd appreciate extra eyes on that part though!

agenticarus:
That test passes for me locally

PR #419, opened 2018-10-02
Remade how cancellation works

BNMetrics:
- Added a requestStore object in handler, this struct will handle storing the requestID mapped to the cancellation function.
- Request inside of requestStore is a struct, we can also use this later on if we were to store anything related to this specific request.

p.s. Sub contexts are awesome!

### DIFF
--- a/tools/build_langserver/langserver/handler.go
+++ b/tools/build_langserver/langserver/handler.go
@@ -123,11 +123,11 @@ func (h *LsHandler) handleInit(ctx context.Context, req *jsonrpc2.Request) (resu
 	}, nil
 }
 
-func (h *LsHandler) handleInitialized(ctx context.Context, request *jsonrpc2.Request) (result interface{}, err error) {
+func (h *LsHandler) handleInitialized(ctx context.Context, req *jsonrpc2.Request) (result interface{}, err error) {
 	return nil, nil
 }
 
-func (h *LsHandler) handleShutDown(ctx context.Context, request *jsonrpc2.Request) (result interface{}, err error) {
+func (h *LsHandler) handleShutDown(ctx context.Context, req *jsonrpc2.Request) (result interface{}, err error) {
 	h.mu.Lock()
 	if h.IsServerDown {
 		log.Warning("Server is already down!")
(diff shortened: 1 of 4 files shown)


tools/build_langserver/langserver/handler.go
```diff
@@ -137,25 +141,24 @@ func (h *LsHandler) handleExit(ctx context.Context, request *jsonrpc2.Request) (
...
-		**comments from Pebers**:
-				Yeah it looks like it will need more complexity -
-				like the server will need to keep a map of request id -> context cancel function,
-				and this function would cancel the appropriate one.
-
-				Tbh sounds like a classic thing we can put off for later :)
-	 */
-	if request.Params == nil {
+	// Is there is no param with Id, or if there is no requests stored currently, return nothing
+	if request.Params == nil || h.requestStore.requests == nil {
```
peterebden: why would `h.requestStore.requests` be nil? doesn't seem like something you should have to worry about here.

tools/build_langserver/langserver/handler.go
```diff
@@ -137,25 +141,24 @@ func (h *LsHandler) handleExit(ctx context.Context, request *jsonrpc2.Request) (
...
 		return nil, nil
 	}
 
 	var params lsp.CancelParams
 	if err := json.Unmarshal(*request.Params, &params); err != nil {
-		return nil, nil
+		return nil, &jsonrpc2.Error{
+			Code: lsp.RequestCancelled,
+			Message: fmt.Sprintf("Cancellation of request(id: %s) failed", request.ID.String()),
```
peterebden: FYI `.String()` is unnecessary here (sprintf will do that for you with a %s).

tools/build_langserver/langserver/handler.go
```diff
@@ -137,25 +141,24 @@ func (h *LsHandler) handleExit(ctx context.Context, request *jsonrpc2.Request) (
...
+		}
 	}
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
+
+	defer h.requestStore.Cancel(jsonrpc2.ID{
```
peterebden: can't you just do `requestStore.Cancel(params.ID)`? also probably you don't need to defer this.

tools/build_langserver/langserver/request_store.go
```diff
@@ -0,0 +1,57 @@
...
+type request struct {
+	id     string
+	cancel func()
+}
+
+func (rs *requestStore) Store(ctx context.Context, req *jsonrpc2.Request) context.Context {
+	ctx, cancel := context.WithCancel(ctx)
+
+	rs.mu.Lock()
+	if rs.requests == nil {
```
peterebden: it'd be easier to initialise this at construction time, then you don't have to worry if it's nil.

tools/build_langserver/langserver/request_store.go
```diff
@@ -0,0 +1,57 @@
...
+		rs.requests = make(map[jsonrpc2.ID]request)
+	}
+	rs.mu.Unlock()
+
+	rs.mu.Lock()
```
peterebden: don't unlock the mutex then lock it again...

tools/build_langserver/langserver/request_store.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (rs *requestStore) Store(ctx context.Context, req *jsonrpc2.Request) context.Context {
+	ctx, cancel := context.WithCancel(ctx)
+
+	rs.mu.Lock()
+	if rs.requests == nil {
+		rs.requests = make(map[jsonrpc2.ID]request)
+	}
+	rs.mu.Unlock()
```
peterebden: defer this unlock (that should be your default - it's like with in python)

tools/build_langserver/langserver/request_store.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// returns the sub-context for the specific request.ID
+	return ctx
+}
+
+// Cancel method removes the id from the requests map and calls cancel function of the request
+func (rs *requestStore) Cancel(id jsonrpc2.ID) {
+	if rs.requests != nil {
+		req := rs.requests[id]
```
peterebden: you need to handle this not being in the map. If it's not then the next line will panic (because req.cancel will be nil)

tools/build_langserver/langserver/handler.go
BNMetrics: Yes! oops! I forgot that was already a Jsonrpcs.ID object!

tools/build_langserver/langserver/handler.go
BNMetrics: Oh! I mostly wanted to check if the map is empty

peterebden: nil isn't the same as an empty map :)

But I think that map should be totally encapsulated by the requestStore object - it would be nicer if the handler only interacted with it through its public interface.

tools/build_langserver/langserver/request_store.go
BNMetrics: o.o, I changed my code, and your comment is still here o.o

PR #421, opened 2018-10-04
Replace html brackets with entities in python_wheel documentation

agenticarus:
image

### DIFF
--- a/src/parse/rules/python_rules.build_defs
+++ b/src/parse/rules/python_rules.build_defs
@@ -413,9 +413,9 @@ def python_wheel(name:str, version:str, hashes:list=None, package_name:str=None,
     support on different platforms, this rule is a simple wrapper around curl and unzip.
     Unless otherwise specified, the wheels are expected to adhere to common naming schemes,
     such as:
-      <package_name>-<version>[-<os>-<arch>].whl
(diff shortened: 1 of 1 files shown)


agenticarus:
As expected, this has degraded the output of `plz help`.

```
[please] ./plz-out/bin/src/please help python_wheel
python_wheel is a built-in build rule in Please. Instructions for use & its arguments:

python_wheel(name, version, hashes, package_name, outs, post_install_commands, patch, licences, test_only, repo, zip_safe, visibility, deps, name_scheme, strip)

Downloads a Python wheel and extracts it.

    This is a lightweight pip-free alternative to pip_library which supports cross-compiling.
    Rather than leaning on pip which is difficult to achieve reproducible builds with and
    support on different platforms, this rule is a simple wrapper around curl and unzip.
    Unless otherwise specified, the wheels are expected to adhere to common naming schemes,
    such as:
      &lt;package_name&gt;-&lt;version&gt;[-&lt;os&gt;-&lt;arch&gt;].whl
      &lt;package_name&gt;-&lt;version&gt;[-&lt;os&gt;_&lt;arch&gt;].whl
      &lt;package_name&gt;-&lt;version&gt;.whl
```

Perhaps passing the docstrings through an html escape step would be better.

PR #422, opened 2018-10-04
Change output files to use consistent file names.

peterebden:
Rather than encoding the hash into the name of the test results file, instead tag them with xattrs (in a similar manner to build outputs) so we still verify they match, but it's easier to find them again (doesn't need mad directory walking shenanigans).

Makes things easier for the surefire code since it doesn't have to walk everything, it can just load the test results specifically.

### DIFF
--- a/src/test/surefire.go
+++ b/src/test/surefire.go
@@ -10,22 +10,10 @@ import (
 
 // CopySurefireXmlFilesToDir copies all the XML test results files into the given directory.
 func CopySurefireXmlFilesToDir(state *core.BuildState, surefireDir string) {
-	outputDirs := make(map[string]struct{})
 	for _, label := range state.ExpandOriginalLabels() {
 		target := state.Graph.TargetOrDie(label)
 		if state.ShouldInclude(target) && target.IsTest && !target.NoTestOutput {
-			outputDir := target.OutDir()
-			if !core.PathExists(outputDir) {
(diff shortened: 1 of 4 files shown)


src/test/surefire.go
```diff
@@ -10,31 +10,22 @@ import (
...
-	for outputDir := range outputDirs {
-		fs.Walk(outputDir, func(path string, isDir bool) error {
-			if !isDir {
-				bytes, _ := ioutil.ReadFile(path)
-				if looksLikeJUnitXMLTestResults(bytes) {
-					surefireResult := filepath.Join(surefireDir, filepath.Base(path))
-					if err := fs.CopyOrLinkFile(path, surefireResult, 0644, true, true); err != nil {
-						log.Errorf("Error linking %s to %s - %s", surefireResult, path, err)
+			if path := target.TestResultsFile(); fs.PathExists(path) {
+				fs.Walk(path, func(path string, isDir bool) error {
```
agenticarus: Do we still need to walk, or is testResultsFile really a directory?

peterebden: It really is a directory for both Python and Java tests; they write one file per class or smthn.

Found that out the hard way, the first version didn't walk and I didn't get many results files :(

agenticarus: Alright, LGTM

PR #423, opened 2018-10-05
hover & Implementation of Analyzer

BNMetrics:
I have got most of the hover functionality working, there will still need clean up, especially analyzer_test.go, comments to the code need to be massively improved.

Analyzer:
- Basically a wrapper around `asp.parser`. It should be the object that handles all the lookups to the ast.
-  `Analyzer.Builtins` stores a maps of builtins(rules, builtin functions)
- In the future, `Analyzer.parser` should not be directly called from `handler` functions. 

Hover.go:
I have got most of the hovering content working, including:
- hovering to rule calls (`go_library`, `go_test`)
- And arguments in the function calls to rule definitions
- There are still several bits I need to reconsider: getting the content from nexted calls, such as `srcs = glob(["*.go"])`

Grammar.go:
- I have added a `String()` member function to `OptimisedExpression`, so I'd be able to extract the default arguments for reconstructing the definition(see `Analyzer.go/RuleDef`)
- I will need to get the values for `ValueExpression` as well to check on the type of args that are passed in to BUILD file function calls.


src/parse/asp/builtins.go
```diff
@@ -279,6 +279,8 @@ func isType(obj pyObject, name string) bool {
 		return name == "list"
 	case pyDict:
 		return name == "dict"
+	case *pyConfig:
+		return name == "CONFIG"
```
peterebden: possibly this should be "config"... the variable is capitalised since it's "constant" but the type would be more consistent as lower-case.

src/parse/asp/grammar.go
```diff
@@ -94,6 +94,21 @@ type OptimisedExpression struct {
 	Config string
 }
 
+// String is a string representation of OptimisedExpression
```
peterebden: Why do you need to string these? The AST types don't have a .String() for the most part and this type isn't meant to be used outside this package. You shouldn't need it for the LSP...

src/parse/asp/lexer.go
```diff
@@ -37,6 +37,7 @@ func (tok Token) String() string {
 }
 
 // A Position describes a position in a source file.
+// All properties in Position are zero indexed
```
peterebden: are they not 1-indexed?

src/parse/asp/parser.go
```diff
@@ -32,6 +32,8 @@ type Parser struct {
 	interpreter *interpreter
 	// Stashed set of source code for builtin rules.
 	builtins map[string][]byte
+
+	statements []*Statement
```
peterebden: What is this? I'm a bit unclear why the parser would have a list of statements in it.

src/parse/asp/parser.go
```diff
@@ -73,6 +78,18 @@ func (p *Parser) MustLoadBuiltins(filename string, contents, encoded []byte) {
 	}
 }
 
+// GetAllBuiltins returns all the builtin rules.
+// *Assuming Parser.LoadBuiltins or Parser.MustLoadBuiltins* has been called
+func (p *Parser) GetAllBuiltins() map[string][]byte {
```
peterebden: Let's do this for now... but I think in the longer term I'd like to get away from using the builtin stuff on the parser for a more generic interface. These accessors feel like they're exposing internals more than I'd like.

src/parse/init.go
```diff
@@ -23,6 +23,12 @@ type aspParser struct {
 
 // newAspParser creates and returns a new asp.Parser.
 func newAspParser(state *core.BuildState) *asp.Parser {
```
peterebden: This function isn't useful? it's just a clone of GetParserWithBuiltins?

src/parse/init.go
```diff
@@ -23,6 +23,12 @@ type aspParser struct {
...
+	p := GetParserWithBuiltins(state)
+	return p
+}
+
+// GetParserWithBuiltins returns a asp.Parser object with all the builtins loaded
```
peterebden: It's idiomatic to call this `NewParser...` if it's creating a new one. `Get` implies that one already exists.

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+
+	"parse/asp"
+	"tools/build_langserver/lsp"
+	"io/ioutil"
+	"fmt"
+)
+
 // TODO(bnmetrics): This file should contain functions to retrieve builtin and custom definitions of build defs
+
+// Analyzer is a wrapper around asp.Analyzer
```
peterebden: you mean `asp.Parser`

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+// This is being loaded into a handler on initialization
+type Analyzer struct {
+	parser *asp.Parser
+	BuiltIns map[string]*RuleDef
```
peterebden: I think this should probably be private - I can imagine us needing it to be threadsafe later on.

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+	ArgMap map[string]*Argument
+}
+
+// Argument is a wrapper around asp.Argument,
+// this is used to store the argument information for specific rules,
+// and it also tells you if the argument is required
+type Argument struct {
+	*asp.Argument
+	definition string
+	required bool
```
peterebden: This is already implicit in the `asp.Argument` - if `Value` is nil then there's no default and it's required.

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+}
+
+// Identifier is a wrapper around asp.Identifier
+// Including the starting line and the ending line number
+type Identifier struct {
+	*asp.IdentStatement
```
peterebden: Bit surprised this wraps an IdentStatement. What are you doing with them? Function definitions obviously aren't IdentStatements...

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+	StartLine int
```
peterebden: I think you might want these to be `asp.Position`s or similar? Wouldn't you need the column too?

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+	a := &Analyzer{
+		parser:parser,
+	}
+	a.BuiltInsRules()
+
+	return a
+}
+
+// BuiltInsRules gets all the builtin functions and rules as a map, and store it in Analyzer.BuiltIns
+func (a *Analyzer) BuiltInsRules() {
```
peterebden: This should probably be private too - you only use it when initialising an Analyzer

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+	statementMap := make(map[string]*RuleDef)
+
+	for _, statement := range a.parser.GetAllBuiltinStatements() {
+		// Saves FuncDefs into the statementMap if it's a None private rule
+		if statement.FuncDef != nil && !strings.HasPrefix(statement.FuncDef.Name, "_") {
```
peterebden: I wonder if we should promote the underscore-private business to a more official concept (like maybe add a bool to the FuncDef or something). Thoughts?

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	stmt, err := a.parser.ParseData(bytecontent, filepath)
+	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
```
peterebden: you've already done this :)

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+		//fmt.Println(val.Ident.Action.Call)
+		if val.Ident != nil {
+			ident := &Identifier{
+				IdentStatement: val.Ident,
+				// -1 from asp.Statement.Pos.Line, as lsp position requires zero index
+				StartLine: val.Pos.Line - 1,
+			}
+			idents = append(idents, ident)
+
+			// fillin End for the previous iterated identifier
```
peterebden: don't think this is correct, the last identifier might have been any number of lines above.

Finding the end positions is Very Difficult though I think...

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+	ruleDef := &RuleDef{
+		FuncDef:funcDef,
+		ArgMap: make(map[string]*Argument),
+	}
+
+	var headerSlice []string
+	header := "def " + funcDef.Name + "("
+	line := header
+	argLen := len(funcDef.Arguments)
+	if argLen > 0 {
```
peterebden: Don't think you need this. If there are no arguments then you'd never enter the for loop.

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+		for i, arg := range funcDef.Arguments {
+			// Check if it a builtin type method, and reconstruct header if it is
+			if i == 0 && arg.Name == "self" {
+				line = arg.Type[0] + "." +funcDef.Name + "("
+			} else {
+				// Fill in the ArgMap
+				argString := getArgument(arg)
+				required := true
+				if strings.Contains(argString, "=") {
```
peterebden: yeah look at the Value, that will give you this more elegantly.

You could also write it inline, like 
```
required: arg.Value == nil,
```

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+					required = false
+				}
+				ruleDef.ArgMap[arg.Name] = &Argument{
+					Argument: &arg,
+					definition: argString,
+					required:required,
+				}
+
+				// Add string to Header
+				if len(line) < 86 {
```
peterebden: 86?

Probably pull this out to a named constant, that would make it clearer why it has that value.

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,199 @@
...
+
+	return ruleDef
+}
+
+func getArgument(argument asp.Argument) string {
+	argType := strings.Join(argument.Type, "|")
+
+	// Get the default value for optional arguments
+	defaultVal := ""
+	if argument.Value != nil && argument.Value.Optimised != nil {
```
peterebden: Ah I see.

Probably not much we can do for now. Definitely think we should separate the builtin stuff - nothing outside `asp` should be using `Optimised` at all, we should probably embed the original files into the lsp and have the parser re-parse them (and not optimise).

That might make this reconstruction business easier too - you could just read the original lines in the file...

tools/build_langserver/langserver/analyzer_test.go
```diff
@@ -0,0 +1,99 @@
...
+	fmt.Println(err)
+	fmt.Println("Luna", stmt[0].Action.Call.Arguments[0].Value.Val)
+	fmt.Println(stmt)
+
+}
+
+func TestNewRuleDef(t *testing.T) {
+	a := newAnalyzer()
+	// Test header the definition for build_rule
+	buildRule := a.parser.GetAllBuiltinStatements()[0].FuncDef
```
peterebden: This is a bit fragile... there's no real guarantee that `build_rule` is the first builtin (or that `startswith` is the tenth). This also breaks if we change its definition (although I guess they are all subject to that).

tools/build_langserver/langserver/hover.go
```diff
@@ -29,29 +32,72 @@ func (h *LsHandler) handleHover(ctx context.Context, req *jsonrpc2.Request) (res
...
+			Code: jsonrpc2.CodeParseError,
+			Message: fmt.Sprintf("fail to parse Build file %s", uri),
+		}
+	}
+
+	lineContent := fileContent[position.Line]
+
+	var contentString string
+	// check if the hovered line is an build target indentifier definition
+	if strings.Contains(lineContent, ident.Name) {
```
peterebden: why do you need to check this? doesn't `IdentFromPos` imply that this is always true (that the identifier you found will be in the line you think it is?

I think the thing you really do need to check here is that the identifier is in your set of builtins.

tools/build_langserver/langserver/hover_test.go
```diff
@@ -0,0 +1,26 @@
...
+	filepath := path.Join(core.RepoRoot, "tools/build_langserver/langserver/BUILD")
+	uri := lsp.DocumentURI("file://" + filepath)
+	analyzer := newAnalyzer()
+	position := lsp.Position{
+		Line: 18,
+		Character: 3,
+	}
+
+	_, err := getHoverContent(ctx, analyzer, uri, position)
+	t.Log(err)
```
peterebden: this is just logging things, it's not asserting anything?

tools/build_langserver/langserver/hover_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"path"
+	"testing"
+	"tools/build_langserver/lsp"
+
+)
+
+func TestGetHoverContent(t *testing.T) {
+	core.FindRepoRoot()
+	ctx := context.Background()
+	filepath := path.Join(core.RepoRoot, "tools/build_langserver/langserver/BUILD")
```
peterebden: I don't think this works - you can't read your own build file in a test. You should add a specific test file with what you need in it.

tools/build_langserver/langserver/utils_test.go
```diff
@@ -67,6 +72,54 @@ func TestEnsureURL(t *testing.T) {
 	assert.Equal(t, uri, lsp.DocumentURI("file://"+string(currentFile)))
 }
 
+func TestReadFile(t *testing.T) {
+	ctx := context.Background()
+	filepath := path.Join(core.RepoRoot, "tools/build_langserver/langserver/BUILD")
```
peterebden: whoa whoa whoa, this isn't right.

This file isn't one of the inputs to the test, and you're reading outside your test directory. Add a test file and pull it into the test using `data`.

tools/please_pex/pytest.py
```diff
@@ -1,5 +1,6 @@
 import os
-
+import sys
+sys.stdin.read()
```
peterebden: um, I assume this isn't deliberate?

src/parse/asp/lexer.go
BNMetrics: oops. yes

src/parse/asp/parser.go
BNMetrics: This is for the bultin statements, as `LoadBuiltins()` already loads the builtins and saves into the Parser.builtin as a map of byte slices, it also loads the builtins as statements. In Analyzer, I use the statements to get the information about the builtin rules. I think since `LoadBuiltins()` already create these statements for the builtin functions and rules, we might as well save these statement for later uses as well.

src/parse/init.go
BNMetrics: oops, I moved things around then forgot to move it back

tools/build_langserver/langserver/analyzer.go
BNMetrics: oops x 3 lol

tools/build_langserver/langserver/analyzer.go
BNMetrics: For now, I won't need the column, because with hover, I'm seeing if the hovering line is within the `StartLine` and `Endline` of each statement. I might needed it later, but we will see.

tools/build_langserver/langserver/analyzer.go
BNMetrics: Yes, I think I like the idea of prefix `_` for private one, and at the same time funcDef could probably have an attribute called `IsPrivate`. If a function does have prefix `_`, we'd have `IsPrivate` as true. What do you think?

tools/build_langserver/langserver/analyzer_test.go
BNMetrics: Yeah, the test is a mess, I intent to change that later :P

tools/please_pex/pytest.py
BNMetrics: LOL. no ignore that lol

tools/build_langserver/langserver/analyzer.go
peterebden: Yup exactly. Should save checking for the same idiom in multiple places.

src/parse/asp/grammar_parse.go
```diff
@@ -144,44 +144,51 @@ func (p *parser) parseStatement() *Statement {
 	s := &Statement{}
 	tok := p.l.Peek()
 	s.Pos = tok.Pos
+
+	var endPos Position
 	switch tok.Value {
 	case "pass":
 		s.Pass = true
-		p.l.Next()
+		endPos = p.l.Next().EndPos()
```
peterebden: I wonder if you can do this generically? like either the lexer or the parser could remember the previous position and then you grab it at the end of this function?

src/parse/asp/grammar_parse.go
```diff
@@ -194,19 +201,24 @@ func (p *parser) parseStatements() []*Statement {
...
 			break
 		}
 	}
-	p.next(EOL)
-	return r
+	endPos := p.next(EOL).EndPos()
+
+	//TODO(bnm): This is a bit hacky, but I'm not sure if there is a better way :(
+	if endPos.Column == 1 {
+		endPos = r.Values[len(r.Values) - 1].Pos
```
peterebden: Not convinced this is OK. What if `r.Values` is empty?

src/parse/asp/parser_test.go
```diff
@@ -350,50 +470,107 @@ func TestMultilineStringQuotes(t *testing.T) {
 #include "UnitTest++/UnitTest++.h"
 "`
 	assert.Equal(t, expected, statements[0].Ident.Action.Assign.Val.String)
+
+	t.Log(statements[0].EndPos)
```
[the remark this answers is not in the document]
BNMetrics: oops. I left that in lol

tools/build_langserver/langserver/hover.go
BNMetrics: This is more like checking if the hovered line is the start of the `Ident`. for example, if I hover over the line `go_library(`, then this condition would be true, in this case, I'd return the content for the definition, and the docstring.

However, if I hover over a line as so `deps = [`, then I'd be returning the content for the argument.

This is because the hovering content is per line based.

tools/build_langserver/langserver/analyzer.go
```diff
@@ -1,3 +1,203 @@
...
+
+// IdentFromPos gets the Identifier given a lsp.Position
+func (a *Analyzer) IdentFromPos(uri lsp.DocumentURI, position lsp.Position) (*Identifier, error) {
+	idents, err := a.IdentFromFile(uri)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, ident := range idents {
+		if position.Line >= ident.StartLine && position.Line <= ident.EndLine {
```
peterebden: do you not need to check columns here as well?

src/parse/init.go
```diff
@@ -13,16 +13,16 @@ import (
...
-	state.Parser = &aspParser{asp: newAspParser(state)}
+	state.Parser = &aspParser{asp: NewAspParser(state)}
 }
 
 // An aspParser implements the core.Parser interface around our asp package.
 type aspParser struct {
 	asp *asp.Parser
 }
 
-// newAspParser creates and returns a new asp.Parser.
```
peterebden: nit: this should probably stay private now you don't need to call it any more.

tools/build_langserver/langserver/hover.go
```diff
@@ -29,29 +33,158 @@ func (h *LsHandler) handleHover(ctx context.Context, req *jsonrpc2.Request) (res
...
+	//
+	//	}
+	//}
+
+	return contentString
+}
+
+func ContentFromNestedCall(analyzer *Analyzer, identArg asp.CallArgument, lineContent string, position lsp.Position) string {
+	nestedIdent := identArg.Value.Val.Ident
+	withInRange := bool(position.Line >= identArg.Value.Pos.Line-1 &&
```
peterebden: nit: you don't need the `bool` here

tools/build_langserver/langserver/hover_test.go
```diff
@@ -0,0 +1,65 @@
...
+	core.FindRepoRoot()
+	ctx := context.Background()
+	filepath := path.Join(core.RepoRoot, "tools/build_langserver/langserver/test_data/example.build")
+	uri := lsp.DocumentURI("file://" + filepath)
+
+	analyzer := newAnalyzer()
+	// Test hovering on the function call
+	content, err := getHoverContent(ctx, analyzer, uri, lsp.Position{Line: 0, Character: 3})
+	assert.Equal(t, nil, err)
+	expected := "def go_library(name:str, srcs:list, asm_srcs:list=None, hdrs:list=None, out:str=None, deps:list=[],\n" +
```
peterebden: be a bit careful with stuff like this in tests; this couples it quite closely to the specifics of the rule.

tools/build_langserver/langserver/hover_test.go
```diff
@@ -0,0 +1,65 @@
...
+		"               visibility:list=None, test_only:bool&testonly=False, complete:bool=True,\n" +
+		"               _needs_transitive_deps=False, _all_srcs=False, cover:bool=True,\n" +
+		"               filter_srcs:bool=True, _link_private:bool=False, _link_extra:bool=True)\n\n" +
```
peterebden: This popup shouldn't include the private arguments (the ones prefixed with underscores), they aren't for users to call.

tools/build_langserver/langserver/analyzer.go
BNMetrics: No, not at this stage, because I would need to get the whole identifier, like a function call(build target), then the detailed analysis to specific column is done at the hover/completion level

tools/build_langserver/langserver/hover_test.go
BNMetrics: This was directly from the rule definition itself. We would need to reconstruct the definition again if we were to remove the private arguments. Thoughts?

PR #431, opened 2018-10-16
Reorder CI steps a little

peterebden:
Basically this should do the slow & expensive OSX step only after the Linux ones pass - very often we run that anyway and discover the same failure that the Linux ones do, burning up our monthly allowance along the way :)

Also simplified the Docker images a bit & got rid of the deb stuff (not sure they are really useful for anything right now).


tools/images/ubuntu_alt/Dockerfile
```diff
@@ -4,20 +4,16 @@ MAINTAINER peter.ebden@gmail.com
 # Most dependencies; Python, Java, Clang and Ruby (only used for fpm)
 RUN apt-get update && \
     apt-get install -y python3 python3-dev python3-pip openjdk-11-jdk-headless \
-    curl unzip git locales pkg-config zlib1g-dev \
-    clang libunittest++-dev ruby ruby-dev rubygems golint && \
+    curl unzip git locales pkg-config zlib1g-dev clang && \
```
agenticarus: This is now missing golint - maybe keep these lists in alphabetical order?

tools/images/ubuntu/Dockerfile
```diff
@@ -1,19 +1,15 @@
 FROM ubuntu:bionic
 MAINTAINER peter.ebden@gmail.com
 
-# Most dependencies; Go, Python, Java, Clang and Ruby (only used for fpm)
+# Go, Python, Java and other dependencies.
 RUN apt-get update && \
     apt-get install -y python3 python3-dev python3-pip openjdk-8-jdk-headless \
-    curl git locales pkg-config zlib1g-dev \
-    clang libunittest++-dev ruby ruby-dev rubygems golang-go golint && \
+    curl git locales pkg-config zlib1g-dev golang-go golint && \
```
agenticarus: Looks like we need unzip here too (it maybe been pulled in transitively before)

tools/images/ubuntu_alt/Dockerfile
peterebden: That was deliberate - I thought I didn't need it. Have updated CircleCI so I really don't (there's no point linting in two different steps)

tools/images/ubuntu/Dockerfile
peterebden: Yeah, weird. Maybe one of the Ruby packages needed it for some reason.

agenticarus:
Guess we could even pull the lint step out entirely and only run the tests if that passes.

peterebden:
The failures aren't to do with lint. I'm not sure what is causing them though...

peterebden:
OK, gonna merge this - I think it will make the build more stable (or it will ruin everything, who knows). Happy to discuss the lint step etc a bit more later.

BNMetrics:
Yes! I definitely thought my PR broke some of those tests...but it wasn't related to my PR...

PR #435, opened 2018-10-24
Some cleanup to Hover

BNMetrics:
- All hover functionality should be working now.
- Including those naked `if`, `for` statements


tools/build_langserver/langserver/analyzer.go
```diff
@@ -139,35 +170,99 @@ func (a *Analyzer) IdentFromFile(uri lsp.DocumentURI) ([]*Identifier, error) {
...
 
-			ident := &Identifier{
-				IdentStatement: stmt.Ident,
-				Type:           identType,
-				// -1 from asp.Statement.Pos.Line, as lsp position requires zero index
-				StartLine: stmt.Pos.Line - 1,
-				EndLine:   stmt.EndPos.Line - 1,
+// statementFromAsp recursively finds asp.IdentStatement and asp.Expression in the ast
+// and returns a valid Statement pointer if within range
+func (a *Analyzer) statementFromAst(v reflect.Value, position lsp.Position) (*Statement, error) {
```
peterebden: You might consider moving some of this to `src/parse/asp/util.go`; that's a good place for generic utility functions.

At some point we could probably optimise it a bit too (like I think right now it basically walks the whole AST top-down, it could skip through the `[]*Statement` entries rapidly until it finds one at the right position)

tools/build_langserver/langserver/analyzer.go
```diff
@@ -139,35 +170,99 @@ func (a *Analyzer) IdentFromFile(uri lsp.DocumentURI) ([]*Identifier, error) {
...
 			}
-			idents = append(idents, ident)
 		}
 	}
+	return nil, nil
+}
+
+func (a *Analyzer) identFromStatement(stmt *asp.Statement) *Identifier {
+	// get the identifier type
+	var identType string
```
peterebden: I'm a little dubious about this; does it really help over checking the members of `stmt.Ident` directly?

I guess I'm a little anti this kind of thing since Go doesn't have first-class enums.

src/core/config.go
```diff
@@ -527,6 +527,23 @@ func (config *Configuration) GetBuildEnv() []string {
...
+// GetTags returns a list of string represent the properties of CONFIG object
+func (config *Configuration) GetTags() []string {
+	var tags []string
+	v := reflect.ValueOf(config).Elem()
+	for i := 0; i < v.NumField(); i++ {
+		if field := v.Field(i); field.Kind() == reflect.Struct {
+			for j := 0; j < field.NumField(); j++ {
+				if tag := field.Type().Field(j).Tag.Get("var"); tag != "" {
+					tags = append(tags, tag)
+					fmt.Println(field.Field(j), tag)
```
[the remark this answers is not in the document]
BNMetrics: oops

tools/build_langserver/langserver/analyzer.go
BNMetrics: Personally I think it's a bit more readable to do that instead of going to the members, I mostly use this in the hover content part, where I check the type of `Ident` to return the respective hover content.

tools/build_langserver/langserver/analyzer.go
BNMetrics: Let's have a chat about this over slack.

src/core/config.go
```diff
@@ -527,6 +527,23 @@ func (config *Configuration) GetBuildEnv() []string {
...
+				}
+			}
+		}
+	}
+	return tags
```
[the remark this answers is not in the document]
peterebden: OK as discussed let's not.

src/parse/asp/util.go
```diff
@@ -55,3 +58,78 @@ func FindArgument(statement *Statement, args ...string) *CallArgument {
...
+// Here we only care about statement with Idents and Expressions
+// In the future we can also add more fields in this struct
+type StatementLite struct {
+	Statement  *Statement
+	Expression *Expression
+}
+
+// StatementFromAst recursively finds asp.IdentStatement and asp.Expression in the ast
+// and returns a valid statement pointer if within range
+func StatementFromAst(v reflect.Value, position Position) (*StatementLite, error) {
```
peterebden: Hmmm on second thought - rather than having the custom type, just return `(*Statement, *Expression, error)`

In fact, do you need the error? I can't see a branch where it is non-nil?

BNMetrics:
Will merge now, happy to discuss later

PR #437, opened 2018-10-25
Generate IntelliJ project files for Java (and Scala) targets

agenticarus:
Happy to discuss this in further depth (including if core is the best place for this).

The basic idea is to map `java_library` and `maven_jar` targets to IntelliJ modules and libraries, and create the chain of dependencies between them. Libraries cannot depend on anything so there needs to be intermediate modules created for each third_party `maven_jar` target, which can then depend on other modules or libraries as desired.

This introduces a set of new magic labels (notably `package_prefix` which tells IntelliJ to pretend the module is rooted at that location, meaning it doesn't need to traverse the tree). Other tools like Pants handle this by forcing all Java code to be under a set of predefined paths like `src/java` from which it can determine the package structure.

I'm a little sad about the checking of the magic "scala_library" rule in here, as the scala rules are not defined in this repository, but I can't see a nice way of making the scala builddefs tie in to this generation process (in the absence of a plugin system).

As it stands, this works "well enough" but I would probably want to improve Protobuf support before making this more widely available. It's good enough for day-to-day use at the moment though.


src/parse/rules/java_rules.build_defs
```diff
@@ -276,6 +281,8 @@ def java_test(name:str, srcs:list, resources:list=None, data:list=None, deps:lis
       jvm_args (str): Arguments to pass to the JVM in the run script.
     """
     timeout, labels = _test_size_and_timeout(size, timeout, labels)
+    if not labels:
+        labels = []
```
peterebden: Hmmm probably `_test_size_and_timeout` should return an empty list, not None? and/or the argument to this function should default to an empty list. Seems we should be able to avoid needing this check.

agenticarus:
Yeah maybe we need an IDE Server Protocol...

PR #446, opened 2018-11-02
added workspace store and fs methods handlers

BNMetrics:
With both completion and hover, we will need a way to track changes in document-in-editing.
- I have added a workspaceStore struct to save documents in editing
- Also added file system requests handlers to update the document store.
- Some of the existing file reading will need to be changed, because we are currently only reading the saved files, which is not going to work.

I'm not sure how deleting line request are being send for other editors, I noticed a bug with deleting line in the intellij LSP plugin, (handled in workspace_store.go, line 95-98), however, after testing it out with various cases, it's still not exactly ideal. Let me know if you have any thoughts on this.


tools/build_langserver/langserver/completion.go
```diff
@@ -29,22 +30,29 @@ func (h *LsHandler) handleCompletion(ctx context.Context, req *jsonrpc2.Request)
 	log.Info("completion with param %s", req.Params)
 	documentURI, err := EnsureURL(params.TextDocument.URI, "file")
 	if err != nil {
+		message := fmt.Sprintf("invalid documentURI '%s' for method %s", documentURI, completionMethod)
+		log.Fatal(message)
```
peterebden: you sure you want a fatal log here? seems dramatic.

tools/build_langserver/langserver/fs.go
```diff
@@ -0,0 +1,74 @@
...
+		h.workspace.Store(params.TextDocument.URI, params.TextDocument.Text)
+		return nil, nil
+	case "textDocument/didChange":
+		var params lsp.DidChangeTextDocumentParams
+		if err := json.Unmarshal(*req.Params, &params); err != nil {
+			return nil, err
+		}
+
+		err := h.workspace.TrackEdit(params.TextDocument.URI, params.ContentChanges)
+		if err != nil {
```
peterebden: FYI in a case like this you could just do `return nil, h.workspace.TrackEdit(params.TextDocument.URI, params.ContentChanges)`

tools/build_langserver/langserver/fs.go
```diff
@@ -0,0 +1,74 @@
...
+			return nil, err
+		}
+		return nil, nil
+	case "textDocument/didClose":
+		var params lsp.DidCloseTextDocumentParams
+		if err := json.Unmarshal(*req.Params, &params); err != nil {
+			return nil, err
+		}
+		delete(h.workspace.documents, params.TextDocument.URI)
```
peterebden: I would probably put something like this into `h.workspace` as a function, even if it's simple, so it manages its internal data structures. Will make it easier if you need to keep an eye on it (e.g. you might want to introduce locking around map accesses for thread-safety).

tools/build_langserver/langserver/workspace_store.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+type workspaceStore struct {
+	rootURI lsp.DocumentURI
+
+	documents map[lsp.DocumentURI]*document
+}
+
+type document struct {
+	text       []string
```
peterebden: Probably useful commenting these members a bit

tools/build_langserver/langserver/fs.go
BNMetrics: That's a good idea actually, it would be more consistent to the other workspace methods as well

tools/build_langserver/langserver/analyzer.go
```diff
@@ -440,6 +440,17 @@ func (a *Analyzer) IsBuildFile(uri lsp.DocumentURI) bool {
 	return a.State.Config.IsABuildFile(base)
 }
 
+// IsBuildDefFile takes a uri path and check if it's a valid .build_defs file
```
peterebden: I'm not sure we should be doing this based on the extension? Calling them `.build_defs` is just a convention

tools/build_langserver/langserver/analyzer.go
```diff
@@ -440,6 +440,17 @@ func (a *Analyzer) IsBuildFile(uri lsp.DocumentURI) bool {
...
+func (a *Analyzer) IsBuildDefFile(uri lsp.DocumentURI) bool {
+	filepath, err := GetPathFromURL(uri, "file")
+	if err != nil {
+		return false
+	}
+
+	base := path.Base(filepath)
```
peterebden: nit: not necessary

tools/build_langserver/langserver/completion.go
```diff
@@ -26,25 +27,32 @@ func (h *LsHandler) handleCompletion(ctx context.Context, req *jsonrpc2.Request)
...
+		log.Error(message)
 		return nil, &jsonrpc2.Error{
 			Code:    jsonrpc2.CodeInvalidParams,
-			Message: fmt.Sprintf("invalid documentURI '%s' for method %s", documentURI, completionMethod),
+			Message: fmt.Sprintf(message),
 		}
 	}
-	if !h.analyzer.IsBuildFile(documentURI) {
+	// Check if the input file is a build file or a build_defs file
+	if !h.analyzer.IsBuildFile(documentURI) || h.analyzer.IsBuildDefFile(documentURI) {
```
peterebden: Yeah I'm not sure the LPS should worry about this. Isn't it up to the client to decide what files it uses the server for?

tools/build_langserver/langserver/completion.go
```diff
@@ -26,25 +27,32 @@ func (h *LsHandler) handleCompletion(ctx context.Context, req *jsonrpc2.Request)
...
 	}
 
-	supportSnippet := h.init.Capabilities.TextDocument.Completion.CompletionItem.SnippetSupport
-
 	h.mu.Lock()
-	itemList, err := getCompletionItemsList(ctx, h.analyzer, supportSnippet, documentURI, params.Position)
+	itemList, err := h.getCompletionItemsList(ctx, documentURI, params.Position)
+
+	if err != nil {
+		return nil, err
```
peterebden: WHOA

This is why you should always use `defer` for unlocking a mutex - if an error happens here you never unlock it and your process is probably borked forever.

tools/build_langserver/langserver/completion_test.go
```diff
@@ -3,39 +3,57 @@ package langserver
...
+var handler = LsHandler{
+	repoRoot:  core.RepoRoot,
+	workspace: newWorkspaceStore(lsp.DocumentURI(core.RepoRoot)),
+	analyzer:  analyzer,
+	init: &lsp.InitializeParams{
+		RootURI: lsp.DocumentURI(core.RepoRoot),
+		Capabilities: lsp.ClientCapabilities{
+			TextDocument: lsp.TextDocumentClientCapabilities{
+				Completion: lsp.Completion{
+					CompletionItem: struct {
```
peterebden: this should probably be named somewhere if you want to reuse it.

tools/build_langserver/langserver/fs.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"context"
+	"encoding/json"
+
+	"tools/build_langserver/lsp"
+
+	"github.com/sourcegraph/jsonrpc2"
+)
+
+func (h *LsHandler) handleFSRequests(ctx context.Context, req *jsonrpc2.Request) (result interface{}, err error) {
```
peterebden: why are these "file system methods"? they all seem to be named "textDocument/something"

tools/build_langserver/langserver/workspace_store.go
```diff
@@ -0,0 +1,160 @@
...
+// Close method corresponds to "textDocument/didClose"
+// This removes the document stored in workspaceStore
+func (ws *workspaceStore) Close(uri lsp.DocumentURI) error {
+	if _, ok := ws.documents[uri]; !ok {
+		return fmt.Errorf("document %s did not open", uri)
+	}
+
+	ws.mu.Lock()
+	delete(ws.documents, uri)
+	ws.mu.Unlock()
```
peterebden: DEFER

tools/build_langserver/langserver/workspace_store.go
```diff
@@ -0,0 +1,160 @@
...
+
+	ws.mu.Lock()
+	for _, change := range contentChanges {
+		newText, err := ws.applyChange(doc.textInEdit, change)
+		if err != nil {
+			return err
+		}
+		doc.textInEdit = newText
+	}
+	ws.mu.Unlock()
```
peterebden: DEFER DEFER DEFER

tools/build_langserver/langserver/completion.go
```diff
@@ -26,26 +27,19 @@ func (h *LsHandler) handleCompletion(ctx context.Context, req *jsonrpc2.Request)
...
 	}
 
-	supportSnippet := h.init.Capabilities.TextDocument.Completion.CompletionItem.SnippetSupport
-
 	h.mu.Lock()
-	itemList, err := getCompletionItemsList(ctx, h.analyzer, supportSnippet, documentURI, params.Position)
-	h.mu.Unlock()
+	itemList, err := h.getCompletionItemsList(ctx, documentURI, params.Position)
+	defer h.mu.Unlock()
```
peterebden: defer immediately after the call to `Lock`. That way if `getCompletionItemsList` panics it will still unlock.

tools/build_langserver/langserver/utils.go
```diff
@@ -190,6 +190,26 @@ func LooksLikeDictAttr(str string) bool {
...
+// Beginning of the buildlabel must have a quote
+// end of the string must not be anything other than quotes or characters
+func ExtractBuildLabel(str string) string {
+	re := regexp.MustCompile(`("(\/\/|:)\w+(\/\w+)*([:]\w*)?"?$)`)
+	matched := re.FindString(strings.TrimSpace(str))
+	if matched == "" {
+		return ""
+	}
+
+	if matched[0] == '"' {
```
peterebden: just use 
```
strings.Trim(matched, `"`)
```

tools/build_langserver/langserver/utils_test.go
```diff
@@ -192,6 +192,26 @@ func TestLooksLikeDictAttr(t *testing.T) {
 	assert.False(t, LooksLikeDictAttr("{\"foo\":2}.keys()"))
 }
 
+func TestExtractBuildLabel(t *testing.T) {
+	label := ExtractBuildLabel("target = \"//src/cache/blah:hello")
```
peterebden: FYI Go has raw strings that would be nicer here: 
```
`target = "//src/cache/blah:hello`
```

BNMetrics: nice!

PR #452, opened 2018-11-08
Completion with Builtin rules and variables

BNMetrics:
- Enhanced completion so it supports all the builtin build rules and BUILD file variables
- removed private arguments from ruleDef headers
- Fixed a bug within workspaceStore...as edit did not work on the first line of the file...

p.s. There are some changes relating to signature/Help call.. Was going to put it in the next PR, but i really needed some functions in this PR..


tools/build_langserver/langserver/analyzer.go
```diff
@@ -205,6 +214,101 @@ func (a *Analyzer) AspStatementFromFile(uri lsp.DocumentURI) ([]*asp.Statement,
...
+			}
+			vars[i.Name] = variable
+		}
+	}
+
+	return vars
+}
+
+func getVarType(valExpr *asp.ValueExpression) string {
+	v := reflect.ValueOf(valExpr).Elem()
```
peterebden: I'm a bit dubious that reflection is the best thing to use here? Do you really want to access any possible string field out of the ValueExpression (as opposed to for example reading the string / fstring fields?)

BNMetrics: So basically, I would need only the type of the variable here, so I can later use it to figure out the correct completion to return. for example, if I have a variable: `blah = "my string"`. later in the file, when I use `blah.`, then the LPS will need to know what type of variable `blah` is. I mean I could also do `if, else if` statement to see which field of `asp.ValueExpression` is not `nil`. I feel like it's a bit tedious. Unless we have any other way of figuring out the value of the variable(`*asp.Identifier`)? Any thoughts?

tools/build_langserver/langserver/analyzer.go
```diff
@@ -171,17 +181,17 @@ func newRuleDef(content string, stmt *asp.Statement) *RuleDef {
...
+					Definition: getArgString(arg),
+					Required:   arg.Value == nil,
 				}
 			}
 		}
 	}
 
-	ruleDef.Header = strings.TrimSuffix(strings.Join(headerSlice, "\n"), ":")
+	header := strings.TrimSuffix(strings.Join(headerSlice, "\n"), ":")
+	ruleDef.Header = removePrivateArgFromHeader(header)
```
BNMetrics: here

src/parse/asp/util.go
```diff
@@ -66,40 +66,64 @@ func StatementOrExpressionFromAst(stmts []*Statement, position Position) (statem
...
+			return &stmt, nil
+		}
+	}
+
+	return nil, nil
+}
+
+// WalkAST is a generic function that walks through the ast recursively,
+// it accepts a callback for any operations
+func WalkAST(v reflect.Value, callback func(v reflect.Value) interface{}) interface{} {
```
peterebden: I'm not sure I like all the `reflect.Value`s in the parameters here, it's a bit awkward to make the caller use `reflect`.

Feel like it should be something like
```
func WalkAST(stmts []*Statement, callback func(stmt *Statement))
```
maybe with a paired form for expressions (or possibly take two different callbacks for them?)

tools/build_langserver/langserver/analyzer.go
peterebden: as discussed I like the idea of constructing it in the analyzer...

BNMetrics: In that case, we can just leave the header as it, as i constructed the Repr here now

src/parse/asp/interpreter.go
```diff
@@ -165,30 +165,24 @@ func (i *interpreter) pkgConfig(pkg *core.Package) *pyConfig {
...
-		expr := v.Interface().(*Expression)
-		if constant := i.scope.Constant(expr); constant != nil {
-			expr.Optimised = &OptimisedExpression{Constant: constant} // Extract constant expression
-			expr.Val = nil
-		} else if expr.Val != nil && expr.Val.Ident != nil && expr.Val.Call == nil && expr.Op == nil && expr.If == nil && expr.Val.Slice == nil {
-			if expr.Val.Property == nil && len(expr.Val.Ident.Action) == 0 {
-				expr.Optimised = &OptimisedExpression{Local: expr.Val.Ident.Name}
-			} else if expr.Val.Ident.Name == "CONFIG" && len(expr.Val.Ident.Action) == 1 && expr.Val.Ident.Action[0].Property != nil && len(expr.Val.Ident.Action[0].Property.Action) == 0 {
-				expr.Optimised = &OptimisedExpression{Config: expr.Val.Ident.Action[0].Property.Name}
+	callback := func(ASTstruct interface{}) interface{} {
```
peterebden: nit: the variable name should start lowercase (i.e. `astStruct` or similar)

tools/build_langserver/langserver/analyzer.go
peterebden: Yeah but you aren't reading the type of variable, you're reading what the name is in the struct, which includes things that aren't types (like `Ident` or `Lambda` and things like that).

Issue #451, opened 2018-11-08
python binaries broken inside third_party


matthen:
If I have in my config 

```
[python]
moduledir = "third_party.python"
```

and add a python_binary anywhere inside `third_party/python`, then running it fails with the message:

```
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "plz-out/bin/third_party/python/binary.pex/__main__.py", line 238, in <module>
  File "plz-out/bin/third_party/python/binary.pex/__main__.py", line 230, in run
  File "plz-out/bin/third_party/python/binary.pex/__main__.py", line 203, in interact
  File "plz-out/bin/third_party/python/binary.pex/__main__.py", line 212, in main
  File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 201, in run_module
    mod_name, mod_spec, code = _get_module_details(mod_name)
  File "/usr/local/Cellar/python/3.6.4_4/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 153, in _get_module_details
    code = loader.get_code(mod_name)
AttributeError: 'ModuleDirImport' object has no attribute 'get_code'
```

moving the binary outside of third party, or removing moduledir from the config fixes it.
this is in 13.2.1

peterebden:
Oh interesting. I suppose the simple answer is "put the binary somewhere else", but I don't imagine it will be anything fundamental. Luna, do you mind having a look please?

matthen:
Cool, yeah I moved it elsewhere but thought it was a bug worth reporting

BNMetrics:
This issue has been fixed in #458

PR #458, opened 2018-11-12
fixed the issue in #451

BNMetrics:
The reason why this bug occurs when having python_binary inside of third_party/python is because we use our own spec loader `ModuleDirImport` within the third_party/python, instead of the builtin python one `_frozen_importlib_external.SourceFileLoader`. 
By adding a `get_code` method in `ModuleDirImport` should solve the problem.

### DIFF
--- a/tools/please_pex/pex_main.py
+++ b/tools/please_pex/pex_main.py
@@ -1,6 +1,7 @@
 """Zipfile entry point which supports auto-extracting itself based on zip-safety."""
 
 from importlib import import_module
+import zipfile
 import os
 import runpy
 import sys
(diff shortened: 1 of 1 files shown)


tools/please_pex/pex_main.py
```diff
@@ -136,6 +136,19 @@ def load_module(self, fullname):
...
 
+    def get_code(self, fullname):
+        from pkgutil import read_code
+
+        module_path = fullname.replace('.', '/')
+        if zipfile.is_zipfile(sys.argv[0]):
+            zf = zipfile.ZipFile(sys.argv[0])
+            for name in zf.namelist():
+                if name.startswith(module_path):
+                    with open(name, 'rb') as f:
```
peterebden: not sure I follow the logic here... it walks through the whole zipfile and imports the first thing it finds that has a prefix of the thing they asked for?

I'm not sure we should do any zipfile stuff in here at all, can we just delegate to the builtin implementation?

matthen:
Nice

PR #465, opened 2018-11-15
made subinclude work!!!

BNMetrics:
- Added subinclude in analyzer.
- Currently hover works with subinclude!!
- Still need to update the methods accordingly


tools/build_langserver/langserver/analyzer.go
```diff
@@ -369,14 +376,103 @@ func (a *Analyzer) BuildLabelFromContent(ctx context.Context,
...
+
+	currentPkg, err := PackageLabelFromURI(uri)
+	if err != nil {
+		log.Warning("fail to load package from uri %s: %s", uri, err)
+	}
+	for _, stmt := range stmts {
+		if stmt.Ident != nil {
+			ident := a.identFromStatement(stmt)
+			if ident.Type == "call" && ident.Name == "subinclude" && len(ident.Action.Call.Arguments) > 0 {
+				includeLabel := ident.Action.Call.Arguments[0].Value.Val.String
```
peterebden: this can panic if the argument is not a string... e.g. 
```
something = "//build_defs:whatever"
subinclude(something)
```
which is technically legal, if unusual. Best check that `.Value.Val` is not nil.

tools/build_langserver/langserver/analyzer.go
```diff
@@ -386,23 +482,33 @@ func (a *Analyzer) VariablesFromContent(content string, pos lsp.Position) map[st
 			continue
 		}
 
-		var varType string
-		if i.Type == "assign" {
-			varType = getVarType(i.Action.Assign.Val)
-		} else if i.Type == "augAssign" {
-			varType = getVarType(i.Action.AugAssign.Val)
+		variable := a.VariableFromIdent(i)
```
peterebden: nit: it's idiomatic to do this inline in the if
```suggestion
		if variable := a.VariableFromIdent(i); variable != nil {
```

tools/build_langserver/langserver/analyzer.go
BNMetrics: Not sure if I can panic here, because that would actually shut down the LPS. however I could log it and highlight it if `.Value.Val` is nil

peterebden: yes that's what I meant, you shouldn't panic, but the previous implementation might have :)

PR #471, opened 2018-11-19
added return type for ast

BNMetrics:


src/parse/asp/grammar.go
```diff
@@ -42,7 +42,9 @@ type FuncDef struct {
 	// Not part of the grammar. Used to indicate internal targets that can only
 	// be called using keyword arguments.
 	KeywordsOnly bool
-	IsPrivate    bool
+	// allowed return types of the FuncDef
+	Returns   []string
```
peterebden: I wonder if this should just be a single `string`? We don't have any cases where we return different types and it would make the parser simpler to just read a single identifier.

src/parse/asp/grammar_parse.go
```diff
@@ -227,6 +227,15 @@ func (p *parser) parseFuncDef() *FuncDef {
 		}
 	}
 	p.next(')')
+
+	if tok := p.l.Peek(); tok.Value != ":" {
+		p.next('-')
+		p.next('>')
+
+		tok := p.oneofval("[", "bool", "str", "int", "list", "dict", "function", "config", "buildrule")
```
peterebden: remove `"["` and `"buildrule"` (we don't have a separate type for build rules).

src/parse/asp/grammar_parse.go
```diff
@@ -227,6 +227,15 @@ func (p *parser) parseFuncDef() *FuncDef {
 		}
 	}
 	p.next(')')
+
+	if tok := p.l.Peek(); tok.Value != ":" {
```
peterebden: might be slightly nicer to check `tok.Value == "-"` since the -> is optional.

src/parse/asp/objects.go
```diff
@@ -556,6 +559,10 @@ func (f *pyFunc) Call(s *scope, c *Call) pyObject {
 	if ret == nil {
 		return None // Implicit 'return None' in any function that didn't do that itself.
 	}
+	if f.returnType != "" && ret.Type() != f.returnType {
+		return s.Error("Invalid return type %s to function %s, expecting %s", ret.Type(), f.name, f.returnType)
```
peterebden: nit: replace `to` with `from`

src/parse/asp/grammar_parse.go
BNMetrics: oops!

src/parse/asp/grammar.go
```diff
@@ -42,7 +42,9 @@ type FuncDef struct {
 	// Not part of the grammar. Used to indicate internal targets that can only
 	// be called using keyword arguments.
 	KeywordsOnly bool
-	IsPrivate    bool
+	// allowed return type of the FuncDef
+	Return    string
```
peterebden: super nit, please move `Return` before `KeywordsOnly` in the struct; the current layout is inefficient (there is a Go linter for this but we don't use it during CI or anything)

PR #472, opened 2018-11-20
Added handlers for "textDocument/publishDiagnostic"

BNMetrics:
there are still some clean up to do. but it's mostly working now!


tools/build_langserver/langserver/diagnostics.go
```diff
@@ -0,0 +1,341 @@
...
+	if err := conn.Notify(ctx, "textDocument/publishDiagnostics", params); err != nil {
+		return err
+	}
+	return nil
+}
+
+type diagnosticsPublisher struct {
+	queue *queue.PriorityQueue
+
+	stored map[lsp.DocumentURI]*diagnosticStore
```
peterebden: why do you need this?

bit concerned that you might need locks around it...

tools/build_langserver/langserver/handler.go
```diff
@@ -70,6 +71,18 @@ func (h *LsHandler) Handle(ctx context.Context, conn *jsonrpc2.Conn, req *jsonrp
 		req.Method != "initialzed" && req.Method != "shutdown" {
 		ctx = h.requestStore.Store(ctx, req)
 		defer h.requestStore.Cancel(req.ID)
+
+		go func() {
+			t, err := h.diagPublisher.queue.Get(1)
```
peterebden: I would probably stick most of this into a method on `diagPublisher` rather than directly accessing its queue here. Keeps the code here shorter but also makes it easier for it to manage things.

tools/build_langserver/langserver/handler.go
```diff
@@ -70,6 +71,18 @@ func (h *LsHandler) Handle(ctx context.Context, conn *jsonrpc2.Conn, req *jsonrp
...
+			if err != nil {
+				log.Warning("fail to get diagnostic publishing task")
+			}
+			if len(t) > 0 {
+				task := t[0].(taskDef)
+
+				h.publishDiagnostics(conn, task.content, task.uri)
```
peterebden: do you not need a loop here? basically loop forever, consuming off the queue and processing them?

tools/build_langserver/langserver/diagnostics.go
BNMetrics: Yeah, that's what I was thinking, I might not even need this anymore

Issue #478, opened 2018-11-23
Binary filegroup permissions


tonickkozlov:
I'm using please to build a go binary that accesses some static files.

I'm facing the following issue:
there's a rule that encapsulates some static files for a go server:
```
filegroup(
    name="ui",
    srcs=[
        "dist",
        "index.html",
    ],
    binary=True, # marked binary to be seen by resulting go_binary
)
```
This rule is relied on by a `go_binary` rule in parent directory.

The issue arises when re-building the binary filegroup rule:

```
./pleasew build //canary/ui:bin
```
```
    //canary/ui:bin
remove plz-out/bin/canary/ui/dist/main.js: permission denied
```

Looks like the permissions set on the output folder are not correct:
```
ls -l plz-out/bin/canary/ui
total 8
dr-xr-xr-x  3 anton  staff   96 23 Nov 15:48 dist
-r-xr-xr-x  2 anton  staff  505 22 Nov 13:27 index.html
```
Notice lack of write permission on the `dist` directory for the owner.

If the rule would not be marked as binary, the permission would be there:
```
ls -l plz-out/gen/canary/ui
total 8
drwxr-xr-x  3 anton  staff   96 23 Nov 15:53 dist
-r-xr-xr-x  3 anton  staff  505 22 Nov 13:27 index.html
```

Please help me solve the permission issue of figure out what's the right way to produce a filegroup that can be accessed by a go_binary.

Thanks!

peterebden:
`binary` outputs are usually just one file, but that isn't really formally enforced.

In this case I think we are applying permissions incorrectly, we shouldn't remove write permission on directories. Can you take a look please Luna?

BNMetrics:
HI @tonickkozlov ,

This bug has been fixed in #480 ! :)

tonickkozlov:
This is great, thank you!

PR #480, opened 2018-11-26
Issue fixes for #478

BNMetrics:


src/build/build_step.go
```diff
@@ -464,7 +465,20 @@ func calculateAndCheckRuleHash(state *core.BuildState, target *core.BuildTarget)
 	// Set appropriate permissions on outputs
 	if target.IsBinary {
 		for _, output := range target.FullOutputs() {
-			if err := os.Chmod(output, target.OutMode()); err != nil {
+			// Walk through the output,
+			// if the output is a directory,apply output mode to the file instead of the directory
+			err := filepath.Walk(output, func(path string, info os.FileInfo, err error) error {
```
peterebden: please use `fs.Walk` instead of `filepath.Walk` - it is faster (the Go team know the current implementation is suboptimal but can't change the interface on their package)

src/fs/fs.go
```diff
@@ -113,3 +113,13 @@ func WriteFile(fromFile io.Reader, to string, mode os.FileMode) error {
 	// And move it to its final destination.
 	return os.Rename(tempFile.Name(), to)
 }
+
+// IsDirectory checks if a given path is a directory
```
peterebden: Do you need this? Don't think you use it?

src/build/build_step.go
```diff
@@ -464,7 +464,19 @@ func calculateAndCheckRuleHash(state *core.BuildState, target *core.BuildTarget)
 	// Set appropriate permissions on outputs
 	if target.IsBinary {
 		for _, output := range target.FullOutputs() {
-			if err := os.Chmod(output, target.OutMode()); err != nil {
+			// Walk through the output,
+			// if the output is a directory,apply output mode to the file instead of the directory
+			err := fs.Walk(output, func(path string, isDir bool) error {
+				if isDir && path == output {
```
peterebden: don't think this is right? surely you should never chmod it if `isDir`?

would suggest something like
```
if isDir {
    return nil
}
return os.Chmod(path, target.OutMode())
```

BNMetrics: That's true, it's likely we'd have more nested dir as well

src/fs/fs.go
```diff
@@ -113,3 +113,13 @@ func WriteFile(fromFile io.Reader, to string, mode os.FileMode) error {
...
+func IsDirectory(path string) bool {
+	info, err := os.Stat(path)
+	if err != nil {
```
peterebden: FYI you could one line this:
```
return err == nil && info.IsDir()
```
Not 100% sure about the case where `err != nil` but I guess there's not much to be done.

BNMetrics: I think it should be fine with that one liner, if it's err != nil, we'd return false anyways

PR #481, opened 2018-11-27
added completion for argname and local src file names

BNMetrics:
- added completion support on argument names for function calls, especially on build rules
- Added completion for local src files

- Possible snippet support later on for arguments

### DIFF
--- a/tools/build_langserver/langserver/utils.go
+++ b/tools/build_langserver/langserver/utils.go
@@ -6,6 +6,7 @@ import (
 	"core"
 	"fmt"
 	"fs"
+	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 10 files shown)


tools/build_langserver/langserver/utils.go
```diff
@@ -65,6 +65,30 @@ func GetPathFromURL(uri lsp.DocumentURI, pathType string) (documentPath string,
...
+func LocalFilesFromURI(uri lsp.DocumentURI) ([]string, error) {
+	fp, err := GetPathFromURL(uri, "file")
+	if err != nil {
+		return nil, err
+	}
+
+	var files []string
+
+	fpDir := filepath.Dir(fp)
+	err = fs.Walk(fpDir, func(name string, isDir bool) error {
```
peterebden: You probably don't need a recursive `Walk` here - probably `ioutil.ListDir` is enough.

BNMetrics: Oh! let me give that a go!

PR #482, opened 2018-11-28
Fixed typo in lexicon entry for glob rule

f-z:


peterebden:
Looks like Maven rejected a bunch of requests, how odd. Will rerun it.

peterebden:
that looks better. Thanks for the fix!

PR #484, opened 2018-11-29
Added handler for "textDocument/references" in LPS

BNMetrics:
- As textDocument/references needs to use `query.GetRevDepsLabels`, I needed some functionalities in `please.go`. I have extracted them into `src/plz`.
- As discussed, it might be nicer to extract most of the logic from please.go into its separate thing, so we can use it later for other things.
- I have started to extract some of the buildfunctions into the separate package as well, probably need a bit more thoughts on how we are going to pass the opts fields.
- In the new module I made `plz/init.go`, I have added a new struct to store these options.


src/plz/init.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	// Wait until they've all exited, which they'll do once they have no tasks left.
+	go func() {
+		wg.Wait()
+		close(state.Results) // This will signal MonitorState (below) to stop.
+	}()
+
+	// Draw stuff to the screen while there are still results coming through.
+	success := output.MonitorState(state, config.Please.NumThreads,
```
peterebden: we should really refactor this too at some point :) defo not this PR though

src/plz/build_functions.go
```diff
@@ -0,0 +1,356 @@
+package plz
+
+//
```
peterebden: leave this out for now, can always add it later once it's ready

tools/build_langserver/langserver/reference.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return core.BuildLabel{}, err
+	}
+
+	rel, err := filepath.Rel(core.RepoRoot, filepath.Dir(fp))
+	if err != nil {
+		return core.BuildLabel{}, err
+	}
+
+	labelStr := fmt.Sprintf("//%s:%s", rel, def.BuildDefName)
```
peterebden: this is a bit ugh. Use `core.NewBuildLabel` instead of printing it & parsing again.

tools/build_langserver/langserver/reference_test.go
```diff
@@ -0,0 +1,83 @@
...
+	// TODO(bnm): We need to think of a better way to run handler.getReferences in root directory instead of Chdir here
+	// save the current testing directory
+	testDir, err := os.Getwd()
+	assert.NoError(t, err)
+
+	// Change to repoRoot before we run the test
+	err = os.Chdir(core.RepoRoot)
+	assert.NoError(t, err)
+
+	locs, err := h.getReferences(ctx, uri, lsp.Position{Line: 0, Character: 4})
```
peterebden: I suspect this might be related to the test failures; it isn't really correct to have a test that goes off and walks through the rest of the repo and it might fight with other builds on some targets.

tools/build_langserver/langserver/reference.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil, err
+	}
+
+	label, err := getCoreBuildLabel(def, uri)
+	if err != nil {
+		return nil, err
+	}
+
+	//Ensure we do not get locked out
+	h.analyzer.State.NeedBuild = false
```
peterebden: I'm not sure you should persist the state object on the analyzer, that may cause some weird results if it gets reused concurrently.

Why not create a new one each time?

Issue #483, opened 2018-11-28
go_get() + extra_outs doesn't work well atm...


sean-:
This took a while to dig through.  One of the things we've run into is importing `timestamp.proto` reliably.  Ideally we don't want to copy this file into our repo and pass an additional protoc flag around.  What we would like to do is do something like:

```python
go_get(
    name = "protobuf-gogo",
    get = "github.com/gogo/protobuf",
    revision = "07eab6a8298cf32fac45cceaac59424f98421bbc",
    install = [
        "gogoproto",
        "jsonpb",
        "proto",
        "protoc-gen-gogo/descriptor",
        "protoc-gen-gogo/generator",
        "protoc-gen-gogo/generator/internal/remap",
        "protoc-gen-gogo/plugin",
        "sortkeys",
        "types",
        "vanity",
    ],
    extra_outs = [
        "protobuf/protobuf/google/protobuf/timestamp.proto",
    ],
)
```

We've since backed away from this (we were trying to use `extra_outs` like an "`extra_files`" where we copy along source files from the `go_get()` and can then refer to them elsewhere), but one of the things that we ran into was `extra_outs` in `go_get()` does not work as expected.  We traced it back to the following:

https://github.com/thought-machine/please/blob/385108c0f025679e3b9fc894258c3c2df4aa8e4b/src/parse/rules/go_rules.build_defs#L516-L555

Specifically, https://github.com/thought-machine/please/blob/385108c0f025679e3b9fc894258c3c2df4aa8e4b/src/parse/rules/go_rules.build_defs#L516

is reused later:

https://github.com/thought-machine/please/blob/385108c0f025679e3b9fc894258c3c2df4aa8e4b/src/parse/rules/go_rules.build_defs#L554

and that obviously expands incorrectly.  `outs` needs to be looped over and append to `cmd` just like is happening to `outs` on line 555.

https://github.com/thought-machine/please/blob/385108c0f025679e3b9fc894258c3c2df4aa8e4b/src/parse/rules/go_rules.build_defs#L554-L555

---

Unrelated, but if there is an idiomatic way to include something like `timestamp.proto` without copying it around, I'm all ears.

peterebden:
Thanks for the detail - makes sense to me. Would you mind taking a look please, Richard?

Re the proto thing, I have not come up with a great solution yet, but agree it would be nice to have one - it is fairly common to find situations like that these days. It's a bit hard to see a good way to dovetail `proto_library` and `go_get` nicely but would love to have one.

[...]

PR #487, opened 2018-11-30
expanding $OUTS in go_get

goddenrich:
addresses issue #483

### DIFF
--- a/src/parse/rules/go_rules.build_defs
+++ b/src/parse/rules/go_rules.build_defs
@@ -549,10 +549,10 @@ def go_get(name:str, get:str|list, repo:str='', deps:list=[], exported_deps:list
     if binary:
         outs = ['bin/' + name]
     else:
+        outs = [f'pkg/{CONFIG.OS}_{CONFIG.ARCH}/{out}' for out in outs]
         # Outputs are created one directory down from where we want them.
         # For most it doesn't matter but the top-level one will get lost.
-        cmd += [' if [ -f ${OUTS}.a ]; then mkdir -p $OUTS && mv ${OUTS}.a $OUTS; fi']
-        outs = [f'pkg/{CONFIG.OS}_{CONFIG.ARCH}/{out}' for out in outs]
+        cmd += [f' if [ -f {out}.a ]; then mkdir -p {out} && mv {out}.a {out}; fi' for out in outs]
     return build_rule(
         name = name,
         outs = outs,


sean-:
Thanks for doing this.  I thought it'd be that simple after walking through this in the issue, but I didn't want to own writing a test to validate that this doesn't get broken in the future.  Cheers.

src/parse/rules/go_rules.build_defs
```diff
@@ -549,10 +549,10 @@ def go_get(name:str, get:str|list, repo:str='', deps:list=[], exported_deps:list
     if binary:
         outs = ['bin/' + name]
     else:
+        outs = [f'pkg/{CONFIG.OS}_{CONFIG.ARCH}/{out}' for out in outs]
         # Outputs are created one directory down from where we want them.
         # For most it doesn't matter but the top-level one will get lost.
-        cmd += [' if [ -f ${OUTS}.a ]; then mkdir -p $OUTS && mv ${OUTS}.a $OUTS; fi']
-        outs = [f'pkg/{CONFIG.OS}_{CONFIG.ARCH}/{out}' for out in outs]
+        cmd += [f' if [ -f {out}.a ]; then mkdir -p {out} && mv {out}.a {out}; fi' for out in outs]
```
sean-: Can we through single quotes around `{out}.a` so that filenames with spaces aren't broken?

```python
         cmd += [f' if [ -f "{out}.a" ]; then mkdir -p "{out}" && mv "{out}.a" "{out}"; fi' for out in outs] 
```

goddenrich: > Thanks for doing this. I thought it'd be that simple after walking through this in the issue, but I didn't want to own writing a test to validate that this doesn't get broken in the future. Cheers.

yea the tests are a little more complex than the fix. Still thinking about how best to do them. Also good point about the quotes will modify when I create the pr for the tests

BNMetrics: @sean- we actually have been having this problem with the space in filenames. For example, if a repository is inside of a directory(or any parent directory) with space in names, the build doesn't seem to work. The obvious solution would be having quotes on all file names, however, it can introduce other issues as well.

sean-: The use of a shell script as the intermediate target for build rules is clever, but also problematic at the same time.  I don't have a better solution to using quotes everywhere (and escaping the values that are later quoted, which also seems to be a missing function call).

One possibility is to change to an accumulated list of argv arrays that are passed to `cmd.Exec()` or some such.  That would be a big refactor, but ultimately probably the right thing to do.  And a PITA for everyone.

I don't have answers here.  We looked into using quotes and found it introduced other subtle problems and punted.  ¯\_(ツ)_/¯

PR #488, opened 2018-11-30
Added handler for "textDocument/rename" in LPS

BNMetrics:


tools/build_langserver/langserver/analyzer.go
```diff
@@ -887,3 +924,25 @@ func aspListToStrSlice(listVal *asp.List) []string {
...
+	}
+
+	defer func() {
+		if r := recover(); r != nil {
+			log.Warning("error occurred parsing build label")
+			err = r.(error)
+		}
+	}()
+
+	return core.NewBuildLabel(rel, def.BuildDefName), err
```
peterebden: FYI I think there's a `TryNewBuildLabel` or smthn that returns an error instead of panicking, that'd save the whole recover thing.

tools/build_langserver/langserver/rename.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	edits, err := h.getRenameEdits(ctx, params.NewName, documentURI, params.Position)
+
+	return edits, nil
```
peterebden: why do you drop `err`? also is this not a compile error because `err` is unused?

tools/build_langserver/langserver/rename.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, err
+	}
+	renamingLabel, err := h.analyzer.BuildLabelFromCoreBuildLabel(ctx, coreLabel)
+	if err != nil {
+		return nil, err
+	}
+
+	revDeps, err := h.analyzer.RevDepsFromCoreBuildLabel(coreLabel, uri)
+	if err != nil {
+		log.Info("error occurred computing the reverse dependency of %")
```
peterebden: this doesn't look right (Github recognises that something is up, didn't know it did that!)

tools/build_langserver/langserver/rename.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, err
+	}
+
+	documentURI, err := getURIAndHandleErrors(params.TextDocument.URI, renameMethod)
+	if err != nil {
+		return nil, err
+	}
+
+	h.mu.Lock()
+	defer h.mu.Unlock()
```
peterebden: careful holding the mutex around a big function like `getRenameEdits`. Not saying it's wrong but locking for shorter periods of time might be more performant & help avoid deadlocks.

tools/build_langserver/langserver/rename.go
BNMetrics: oops. lol this is what happens when I don't re-read my cold

Issue #489, opened 2018-12-01
[please] Location setting causes builds to hang indefinitely.


kormat:
Very minimal example:
`/BUILD`:
```
go_binary(
    name = "main",
    srcs = ["main.go"],
)
```
`/main.go`:
```
package main

func main() {}
```
`/.plzconfig`:
```
[please]
Location = /home/diamond/.please ; Note that this _is_ the dir `please` is installed into

[build]
PATH = /usr/local/go/bin:/usr/local/bin:/usr/bin:/bin

[go]
ImportPath = github.com/kormat/plzbldfoo2
```

When i run `build`, it hangs:
```
$ plz build -v debug
16:46:21.900   DEBUG: Found repo root at /home/diamond/tmp/plz/plz-issue-location
16:46:21.900   DEBUG: Reading config from /etc/plzconfig...
16:46:21.900   DEBUG: Reading config from /home/diamond/.please/plzconfig...
16:46:21.900   DEBUG: Reading config from /home/diamond/tmp/plz/plz-issue-location/.plzconfig...
16:46:21.901   DEBUG: Reading config from /home/diamond/tmp/plz/plz-issue-location/.plzconfig_linux_amd64...
16:46:21.901   DEBUG: Reading config from /home/diamond/tmp/plz/plz-issue-location/.plzconfig.local...
16:46:21.901   DEBUG: Loading built-in build rules...
16:46:21.901    INFO: Total cache size: 0 B
16:46:21.947   DEBUG: Parser initialised
16:46:21.947   DEBUG: Attempting to acquire lock plz-out/.lock...
16:46:21.947   DEBUG: Acquired lock plz-out/.lock
16:46:31.947  NOTICE: Build running for 10s, 1 / 2 tasks done, 0 workers busy
16:46:41.947  NOTICE: Build running for 20s, 1 / 2 tasks done, 0 workers busy
16:46:51.947  NOTICE: Build running for 30s, 1 / 2 tasks done, 0 workers busy
16:47:01.947  NOTICE: Build running for 40s, 1 / 2 tasks done, 0 workers busy
16:47:11.947  NOTICE: Build running for 50s, 1 / 2 tasks done, 0 workers busy
16:47:21.947  NOTICE: Build running for 1m0s, 1 / 2 tasks done, 0 workers busy
16:47:31.947  NOTICE: Build running for 1m10s, 1 / 2 tasks done, 0 workers busy
[...]

PR #490, opened 2018-12-02
Fix termination of subdir walking.

kormat:
If an absolute path gets passed into
`src/parse/asp/targets.go:checkSubDir` (e.g. you have `Location` set to
an absolute path in your `.plzconfig`), it fails to terminate as it will
walk up to `/` and then get into an infinite loop. This change just adds
detection for `/`, to terminate the loop.

Fixes https://github.com/thought-machine/please/issues/489


BNMetrics:
Will merge this now, if anyone has any additional suggestions, feel free to comment. @peterebden

PR #491, opened 2018-12-02
Add definitions support to go_binary() and go_test()

sean-:
Pass in definitions to `go tool link` (the `-X` flag).  This is useful for passing in git SHAs or other build-time information to a go binary.

The name `definitions` was used instead of `constants` because this terminology is aligned with the verbiage used in the `go tool link` command:

```
  -X definition
    	add string value definition of the form importpath.name=value
```

Because you can assign a value to a `const` or to a `var.  From a UX perspective, I think accuracy and consistency with the Go project is more important than being colloquial (e.g. "build-time constants").

```python
go_binary(
    name = "test_dict",
    srcs = ["main.go"],
    definitions = {
        "main.Hello": "dict",
        "main.World": "nice to see you",
    },
)

go_binary(
    name = "test_list",
    srcs = ["main.go"],
    definitions = [
        "main.Hello=test_list",
        "main.World=\"works\"",
    ],
)

go_binary(
    name = "test_str",
    srcs = ["main.go"],
    definitions = "main.Hello=world",
)

# go_test() also has support for definitions. 
```

```go
package main

var (
  Hello string
  World string
)

func main() {
	fmt.Printf("Hello %q %q\n", Hello, World)
}
```

```
% plz build //...
Build finished; total time 1.65s, incrementality 100.0%. Outputs:
//:pleasings:
  plz-out/gen/pleasings
//:test_dict:
  plz-out/bin/test_dict
//:test_list:
  plz-out/bin/test_list
//:test_str:
  plz-out/bin/test_str
Hello "dict" "nice to see you"
Hello "test_list" "works"
Hello "world" ""
```

This is a precursor to being able to run and set values using something like:

```python
go_binary(
    name = "test",
    srcs = ["main.go"],
    definitions = {
        "main.GitCommit": exec(cmd = ["git", "rev-parse", "--short", "HEAD"]),
        "main.GitBranch": exec(cmd = "git symbolic-ref -q --short HEAD"),
        "main.GitSummary": exec(cmd = "git describe --tags --dirty --always"),
        "main.GitState": "dirty" if len(exec("git status --porcelain")) > 0 else "",
    },
)
```

A helper function that that will populate these values will come shortly (once #492 is merged).  This type of information is useful for `myprog --version` or `myprog version` commands, and monitoring if you emit the git-sha of your running program along with your metric data.

Specifying a generic `definitions` that is passed down to `go tool link`'s `-X` flag seemed preferable to allowing users to pass in random linker flags.

PR #492, opened 2018-12-02
exec(): add a new builtin named exec()

sean-:
Captures the output of stdout (or stderr) and returns the value as a string to the caller (if the call was successful).

With #491 it is now possible to embed git SHAs and build information into binaries.

```go
% cat main.go
package main

import "fmt"

var (
	GitBranch string = "unknown branch"
	GitStatus        = "unknown status"
	GitCommit        = "unknown commit"
	TestVal          = "missing test value"
)

func main() {
	fmt.Println("Branch", GitBranch)
	fmt.Println("Status", GitStatus)
	fmt.Println("Commit", GitCommit)
	fmt.Printf("TestVal: %q\n", TestVal)
}
```

```python
gitBranch = exec(cmd = ["git", "symbolic-ref", "-q", "--short", "HEAD"])
gitStatus  = gitStatus  = "dirty" if len(exec(cmd = "git status --porcelain")) > 0 else ""
gitCommit  = exec(cmd = "git rev-parse --short HEAD")

go_binary(
    name = "test_stderr",
    srcs = ["main.go"],
    definitions = {
        "main.GitBranch": gitBranch,
        "main.GitCommit": gitCommit,
        "main.GitStatus": gitStatus,
        "main.TestVal": exec(
            cmd = ["sh", "-c", "echo stdout1; echo 1>&2 stderr; echo stdout2"],
            stdout = False,
            stderr = True,
        ),
    },
)

go_binary(
    name = "test_stdout",
    srcs = ["main.go"],
    definitions = {
        "main.GitBranch": gitBranch,
        "main.GitCommit": gitCommit,
        "main.GitStatus": gitStatus,
        "main.TestVal": exec(
            cmd = ["sh", "-c", "echo stdout1; echo 1>&2 stderr; echo stdout2"],
            stdout = True,
            stderr = False,
        ),
    },
)

go_binary(
    name = "test_all",
    srcs = ["main.go"],
    definitions = {
        "main.GitBranch": gitBranch,
        "main.GitCommit": gitCommit,
        "main.GitStatus": gitStatus,
        "main.TestVal": exec(
            cmd = ["sh", "-c", "echo stdout1; echo 1>&2 stderr; echo stdout2"],
            stdout = True,
            stderr = True,
        ),
    },
)

# go_binary(
#     name = "test_fail",
#     srcs = ["main.go"],
#     definitions = {
#         "main.GitBranch": gitBranch,
#         "main.GitCommit": gitCommit,
#         "main.GitStatus": gitStatus,
#         "main.TestVal": exec(cmd = "false"),
#     },
# )
```

```
% ~/go/src/github.com/thought-machine/please/plz-out/bin/src/please rebuild //... && ./plz-out/bin/test_stdout && ./plz-out/bin/test_stderr && ./plz-out/bin/test_all
Build finished; total time 1.42s, incrementality 0.0%. Outputs:
//:pleasings:
  plz-out/gen/pleasings
//:test_all:
  plz-out/bin/test_all
//:test_stderr:
  plz-out/bin/test_stderr
//:test_stdout:
  plz-out/bin/test_stdout
Branch master
Status dirty
Commit 26ea52b
TestVal: "stdout1\nstdout2"
Branch master
Status dirty
Commit 26ea52b
TestVal: "stderr"
Branch master
Status dirty
Commit 26ea52b
TestVal: "stdout1\nstderr\nstdout2"
```

### DIFF
--- a/docs/lexicon.html
+++ b/docs/lexicon.html
@@ -240,6 +240,61 @@
       </ul>
     </p>
 
+    <h3><a name="exec">exec</a></h3>
+
+    <p><pre class="rule"><code>exec(cmd, stdout=True, stderr=False)</code></pre></p>
+
+    <p>Executes a command and returns the output as a string if the command's
+      exit code returns <code>0</code>. If <code>stdout</code>
+      and <code>stderr</code> are both set to true, the output from both stdout
+      and stderr is interleaved.</p>
+
+    <p>For example:
+      <ul>
+        <li><code>exec(cmd = ["git", "symbolic-ref", "-q", "--short", "HEAD"])</code> -> <code>master</code></li>
+        <li><code>exec(cmd = "git rev-parse --short HEAD")</code> -> <code>26ea52b</code></li>
+        <li><code>exec(cmd = ["sh", "-c", "echo stdout1; echo 1>&2 stderr; echo stdout2"], stdout = False, stderr = True)</code> -> <code>stderr</code></li>
+        <li><code>exec(cmd = ["sh", "-c", "echo stdout1; echo 1>&2 stderr; echo stdout2"], stderr = True)</code> -> <code>stdout1\nstderr\nstdout2</code></li>
+      </ul>
+    </p>
+
+    <table>
+      <thead>
+      <tr>
+	<th>Argument</th>
+	<th>Default</th>
+	<th>Type</th>
+	<th></th>
+      </tr>
+      </thead>
+      <tbody>
+
+      <tr>
+	<td>cmd</td>
+	<td></td>
+	<td>str or list</td>
+	<td>Command to be executed.  Command does not invoke the system shell
+	  and does not expand any glob patterns or handle other expansions,
+	  pipelines, or redirections typically done by shells.</td>
+      </tr>
+
+      <tr>
+	<td>stdout</td>
+	<td>True</td>
+	<td>bool</td>
+	<td>Capture and return stdout.</td>
+      </tr>
+
+      <tr>
+	<td>stderr</td>
+	<td>False</td>
+	<td>bool</td>
+	<td>Capture and return stderr.</td>
+      </tr>
+
+      </tbody>
+    </table>
+
 
     <h2><a name="cc">C and C++ rules</a></h2>
 
--- a/docs/lexicon.html
+++ b/docs/lexicon.html
@@ -245,9 +245,11 @@
     <p><pre class="rule"><code>exec(cmd, stdout=True, stderr=False)</code></pre></p>
 
     <p>Executes a command and returns the output as a string if the command's
-      exit code returns <code>0</code>. If <code>stdout</code>
+      exit code returns <code>0</code>.  If <code>stdout</code>
       and <code>stderr</code> are both set to true, the output from both stdout
-      and stderr is interleaved.</p>
+      and stderr is interleaved.  The string form of <code>cmd</code> strips
+      whitespace from the individual arguments before calling the function.  The
+      list form of <code>cmd</code> leaves the input arguments untouched.</p>
 
     <p>For example:
       <ul>
@@ -275,7 +277,10 @@
 	<td>str or list</td>
 	<td>Command to be executed.  Command does not invoke the system shell
 	  and does not expand any glob patterns or handle other expansions,
-	  pipelines, or redirections typically done by shells.</td>
+	  pipelines, or redirections typically done by shells. NOTE: The command
+	  must be idempotent (i.e. deterministic and side-effect free) otherwise
+	  all promises with regards to reproducibility of the build state are
+	  invalid and the repository must be rebuilt.</td>
       </tr>
 
       <tr>
@@ -290,6 +295,15 @@
 	<td>False</td>
 	<td>bool</td>
 	<td>Capture and return stderr.</td>
+      </tr>
+
+      <tr>
+	<td>cache</td>
+	<td>True</td>
+	<td>bool</td>
+	<td>Cache the output.  By default the output from commands is cached
+	  based on its input.  Commands are assumed to be idempotent and
+	  side-effect free.</td>
       </tr>
 
       </tbody>
--- a/docs/lexicon.html
+++ b/docs/lexicon.html
@@ -306,6 +306,16 @@
 	  side-effect free.</td>
       </tr>
 
+      <tr>
+	<td>key_extra</td>
+	<td></td>
+	<td>str</td>
+
+	<td>Extra key material to add to the cache key.  Example uses of this
+	  include when the working directory, package, or cross-compilation
+	  target invalidates the output of a given command.</td>
+      </tr>
+
       </tbody>
     </table>
 
--- a/src/parse/asp/builtins.go
+++ b/src/parse/asp/builtins.go
@@ -172,6 +172,7 @@ func doExec(s *scope, args []pyObject) pyObject {
 	wantStdout := args[1].IsTruthy()
 	wantStderr := args[2].IsTruthy()
 	cacheOutput := args[3].IsTruthy()
+	keyExtra := args[4].String()
 
 	if !wantStdout && !wantStderr {
 		return s.Error("exec() must have at least stdout or stderr set to true, both can not be false")
@@ -189,7 +190,7 @@ func doExec(s *scope, args []pyObject) pyObject {
 	}
 
 	// The cache key is tightly coupled to the operating parameters
-	key := execMakeKey(s, cmdArgs, wantStdout, wantStderr)
+	key := execMakeKey(keyExtra, cmdArgs, wantStdout, wantStderr)
 
 	// Only get cached output if this call is intended to be cached.
 	var completedPromise bool
@@ -235,7 +236,7 @@ func doExec(s *scope, args []pyObject) pyObject {
 	}
 
 	if cacheOutput {
-		execSetCachedOutput(key, cmdArgs, outStr)
+		execSetCachedOutput(keyExtra, key, cmdArgs, outStr)
 		completedPromise = true
 	}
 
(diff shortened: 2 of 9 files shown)


docs/lexicon.html
```diff
@@ -240,6 +240,61 @@ <h3><a name="canonicalise">canonicalise</a></h3>
...
+	<th></th>
+      </tr>
+      </thead>
+      <tbody>
+
+      <tr>
+	<td>cmd</td>
+	<td></td>
+	<td>str or list</td>
+	<td>Command to be executed.  Command does not invoke the system shell
```
csdigi: Not sure if it is worth calling out specifically, but might be a good idea to add a caution that the command should be deterministic (and reproducible with the build state), otherwise bad things w.r.t. caching and incrementality will happen. Obviously this is similar to other gotchas (such as putting timestamps in files) but every little helps.

src/parse/asp/builtins.go
```diff
@@ -155,6 +159,57 @@ func buildRule(s *scope, args []pyObject) pyObject {
...
+		cmdArgs = strings.Fields(string(cmdIn.(pyString)))
+	} else if isType(cmdIn, "list") {
+		pl := cmdIn.(pyList)
+		cmdArgs = make([]string, 0, pl.Len())
+		for i := 0; i < pl.Len(); i++ {
+			cmdArgs = append(cmdArgs, pl[i].String())
+		}
+	}
+
+	ctx, cancel := context.WithCancel(context.TODO())
```
csdigi: I guess a potential use for this context could be allowing a timeout arg to this exec so that the user can get a build interrupted if they specify a timeout (to give them better output / errors in case the command is executing forever).

src/parse/asp/builtins.go
```diff
@@ -155,6 +159,57 @@ func buildRule(s *scope, args []pyObject) pyObject {
...
+	defer cancel()
+
+	// TODO(seanc@): Memoize the output based on cmdArgs
```
csdigi: This is a good idea (both for performance and to prevent the reliance on changing commands), probably reinforces the point that we should call out the deterministic contract on the command output explicitly in the documentation above (as it in the future may be cached and reused)

src/parse/asp/builtins.go
```diff
@@ -155,6 +159,57 @@ func buildRule(s *scope, args []pyObject) pyObject {
 	return pyString(":" + target.Label.Name)
 }
 
+// doExec fork/exec's a command and returns the output as a string.  exec
+// accepts either a string or a list of commands and arguments.
+func doExec(s *scope, args []pyObject) pyObject {
+	cmdIn := args[0]
+	wantStdout := args[1].IsTruthy()
+	wantStderr := args[2].IsTruthy()
```
csdigi: We should probably explicitly error if these are both set to False, as it will cause a fall-through in the switch below and the attempted execution of a command with no required result (which can only mean people have made a mistake or are trying to do unprincipled "bad things").

docs/lexicon.html
```diff
@@ -240,6 +240,61 @@ <h3><a name="canonicalise">canonicalise</a></h3>
       </ul>
     </p>
 
+    <h3><a name="exec">exec</a></h3>
+
+    <p><pre class="rule"><code>exec(cmd, stdout=True, stderr=False)</code></pre></p>
+
+    <p>Executes a command and returns the output as a string if the command's
+      exit code returns <code>0</code>. If <code>stdout</code>
```
csdigi: We might want to mention that the spaces are trimmed here (to save someone an annoying debugging journey if they ever wanted it :) )

src/parse/asp/builtins.go
sean-: Correct.  I wasn't sure how to pull down a timeout and use it here.  I've switched to using a timeout from the state for now.  It's not entirely correct, but I'm not sure where to get the current target from the `scope`.

src/parse/asp/builtins.go
sean-: Done, though I'm not wild about the complexity this introduced.

docs/lexicon.html
sean-: Updated the docs to reflect this concern.

sean-:
Minus tests, this looks to be working atm.  I have cherry-picked the commits from #491 for the time being but am happy to revert and force-push those out once this is ready to land.

src/parse/asp/builtins.go
csdigi: Thanks, I was actually just agreeing with your aspirations, did not mean for you do it it straight away :), good effort!

src/parse/asp/exec.go
```diff
@@ -0,0 +1,165 @@
...
+				memoizedCmds[key] = promise
+				return "", false, promise
+			}
+
+			if data, ok := rawVal.(*execData); ok {
+				promise.cv.L.Unlock()
+				memoizedCmdsLock.Unlock()
+				return data.Value(), true, nil
+			} else {
+				panic(fmt.Sprintf("unknown type in execValue: %T", val))
```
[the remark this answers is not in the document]
sean-: This will only panic if there is a programmer error and we save something in the `memoizedCmds` map (e.g. we store an `int` or `struct{}` vs something that satisfies the type).  I'm happy to to move it to a `log.Fatalf()`, but have it as a `panic()` because it is a developer error, not a user error.

src/parse/asp/builtins.go
sean-: I can add that in a second pass if you'd like.  I found *a* way of pulling in some timeout value, but I'm not sure it's right.

src/parse/asp/builtins.go
sean-: I didn't like what I pushed re: promises and reworked it into something that I actually think is significantly more readable and maintainable.

src/parse/asp/exec.go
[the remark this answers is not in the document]
sean-: I completely refactored this to be something significantly easier to understand: there is a map of cached values protected by a `sync.RWMutex` and a list of promises.  Cache misses get a promise that will return the resulting value from the concurrent command.  Future cache hits will just return the cached value itself.

sean-:
@csdigi , @peterebden , and @BNMetrics can you give this a once over?  I'm happy with this atm and want to see it land now.  The only todo that I think is meaningful for future consideration is the comment I made on https://github.com/thought-machine/please/pull/492/files#diff-78d373427efb3b10c9fb727b273ebac5R101 where I made a passing reference to using scope and potentially opening up the door to having `cache` move from being `cache:bool=True` to `cache:(bool:str)=True` where a valid string argument could be "package" or "global" or something else that we can potentially use as an input key with `scope`.  I don't want to block on that now, but I figured food for thought in case it was meaningful to anyone.

That said, I've tested this with concurrent accesses to `sleep 10; echo foo` across different `BUILD` files and have traced the output where I watch promises launch, wait, and sleep.  I think we're in an okay place with this now.

sean-:
Also, my next PR needs to start with a question:

```python
gitBranch = exec(cmd = ["git", "symbolic-ref", "-q", "--short", "HEAD"])
gitStatus  = gitStatus  = "dirty" if len(exec(cmd = "git status --porcelain")) > 0 else ""
gitCommit  = exec(cmd = "git rev-parse --short HEAD")
```

Would there be an objection to making these standard functions, like `git_branch()`, `git_status()`, `git_sha(short:bool=True)`, etc?  It seems like those helper functions would cross the threshold of widespread re-usability and use to a number of people.  ?  If not we can just hold onto those as a local import.  But.  They seem like they'd be useful enough to a large enough pool of developers.  ?

sean-:
I've rebased this PR on `master` in order to remove #491 from this PRs history.

sean-:
Same.  I'm anxious to see it land and a new release shipped.  :~]

FWIW, I used the following `BUILD` file for local testing and copied it into multiple directories and timed the result as a way of validating that caching was working as expected.  The following runs in ~15s (expected).  If caching wasn't working, this would be ~5s * 4 * num directories.

```python
gitBranch = exec(cmd = ["git", "symbolic-ref", "-q", "--short", "HEAD"])
gitStatus  = "dirty" if len(exec(cmd = "git status --porcelain")) > 0 else ""
gitCommit  = exec(cmd = "git rev-parse --short HEAD")

go_binary(
    name = "test_stderr",
    srcs = ["main.go"],
    definitions = {
        "main.GitBranch": gitBranch,
        "main.GitCommit": gitCommit,
        "main.GitStatus": gitStatus,
        "main.TestVal": exec(
            cmd = ["sh", "-c", "echo stdout1; sleep 5; echo 1>&2 stderr; echo stdout2"],
            stdout = False,
            stderr = True,
        ),
    },
)

go_binary(
    name = "test_stdout",
    srcs = ["main.go"],
    definitions = {
        "main.GitBranch": gitBranch,
        "main.GitCommit": gitCommit,
        "main.GitStatus": gitStatus,
        "main.TestVal": exec(
            cmd = ["sh", "-c", "echo stdout1; sleep 5; echo 1>&2 stderr; echo stdout2"],
            stdout = True,
            stderr = False,
        ),
    },
)

go_binary(
    name = "test_all",
    srcs = ["main.go"],
    definitions = {
        "main.GitBranch": gitBranch,
        "main.GitCommit": gitCommit,
        "main.GitStatus": gitStatus,
        "main.TestVal": exec(
            cmd = ["sh", "-c", "echo stdout1; sleep 5; echo 1>&2 stderr; echo stdout2"],
            stdout = True,
            stderr = True,
        ),
    },
)

go_binary(
    name = "test_all_duplicate_cmd",
    srcs = ["main.go"],
    definitions = {
        "main.GitBranch": gitBranch,
        "main.GitCommit": gitCommit,
        "main.GitStatus": gitStatus,
        "main.TestVal": exec(
            cmd = ["sh", "-c", "echo stdout1; sleep 5; echo 1>&2 stderr; echo stdout2"],
            stdout = True,
            stderr = True,
        ),
    },
)
```

peterebden:
LGTM, nice idea!

Agreed with the sentiment about the caching being surprisingly complex. I wondered about doing something like creating synthetic build targets for them but I suspect that might just move all the complexity somewhere else.

FWIW I also like the linker definitions thing; I remember thinking that'd be useful once but obviously never got to the point of actually implementing it.

sean-:
... so... can haz merge?

The caching became more straightforward when I separated our values from promises and gave users control over a cache mixin. But. Yes. Who knew adding additional dimensions to a dependency graph would be complicated. ;)

Issue #493, opened 2018-12-03
`plz run` a gentest build rule fails with cryptic output


RichardoC:
Where c is a gentest
$ plz run //a/b:c
panic: Rule //a/b:c can't use //a/b:c; //a/b:c has multiple outputs.

Could the output be
$ plz run //a/b:c
panic: gentest //a/b:c can't be run, only tested

PR #494, opened 2018-12-03
fixed the issue with ambigious panic

BNMetrics:
#493

### DIFF
--- a/src/build/command_replacements.go
+++ b/src/build/command_replacements.go
@@ -192,7 +192,7 @@ func replaceSequenceLabel(state *core.BuildState, target *core.BuildTarget, labe
 }
 
 func checkAndReplaceSequence(state *core.BuildState, target, dep *core.BuildTarget, in string, runnable, multiple, dir, outPrefix, hash, test, allOutputs, tool bool) string {
-	if allOutputs && !multiple && len(dep.Outputs()) != 1 {
+	if allOutputs && !multiple && len(dep.Outputs()) > 1 {
 		// Label must have only one output.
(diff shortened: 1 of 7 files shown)


BNMetrics:
@peterebden. I actually thought of a few options as well, including the one you suggestions. But wouldn't it be a bit ambigious. I think the `cannot_run` can also apply to any test rules, because some tests actually fail if you use `plz run` instead of `plz test`..
Let me know your thoughts!

peterebden:
Yeah it is sometimes useful to run tests with `plz run` although as you say it only works if they don't have/use any `data`.

Still don't think having an extra attribute is necessary though, I think `run_step` should just check the outputs itself.

BNMetrics:
Okay, If that's the case, let's make it simpler by checking if it's got any `data`! :)

src/build/command_replacements.go
```diff
@@ -192,7 +192,7 @@ func replaceSequenceLabel(state *core.BuildState, target *core.BuildTarget, labe
 }
 
 func checkAndReplaceSequence(state *core.BuildState, target, dep *core.BuildTarget, in string, runnable, multiple, dir, outPrefix, hash, test, allOutputs, tool bool) string {
-	if allOutputs && !multiple && len(dep.Outputs()) != 1 {
+	if allOutputs && !multiple && len(dep.Outputs()) > 1 {
```
peterebden: I think it should still panic for when there are no outputs; might be nice to special-case that though so the message is more explicit.

BNMetrics: There is another case checking `runnable` and `len(dep.Outputs()) == 0` as a special case later actually! :)

PR #497, opened 2018-12-07
please.go refactor

BNMetrics:
As discussed last week. We agreed that it might be useful to refactor please.go into it's own package `plz/`.

I have gotten the basic idea/structure down now, It's still in rough draft stage. @peterebden let me know if you like this structure, then I will refactor the rest of the please.go


src/please.go
```diff
@@ -993,14 +974,100 @@ func execute(command string) bool {
...
+	var success bool
+	if command == "build" || command == "test" ||
+		command == "run" || command == "cover" ||
+		command == "rebuild" || command == "hash" ||
+		command == "parallel" {
+		initOpts, params := getInitOpsAndParams(command)
+		success = plz.Handle(command, *initOpts, params)
+	} else {
+		success = buildFunctions[command]()
+	}
```
sean-:
```
switch command {
case "build", "test", "run", "cover", "rebuild", "hash", "parallel":
  initOpts, params := getInitOpsAndParams(command) 
  success = plz.Handle(command, *initOpts, params)
default:
  success = buildFunctions[command]()
}
```

BNMetrics: This is temporary!:)

BNMetrics: Yeah, this PR is kind of like a proof of concept. I wanted to refactor please.go into its own package, so I can use those functionalities somewhere else later.
eventually `success = buildFunctions[command]()` will be replaced by
```
initOpts, params := getInitOpsAndParams(command)
success = plz.Handle(command, *initOpts, params)
```
So none of those if else statement will live long. :)

peterebden:
I've (finally) had time to look over this; I want to have a bit more of a go at the display stuff before coming back to it though. I think it needs to be better detached which will clean up the control flow.

peterebden:
Have (finally) come back to this; I don't think we need more refactoring in plz.go at the moment am fairly happy with it right now. I'd like it to stay relatively generic and not have specific knowledge of the CLI commands.

Issue #501, opened 2018-12-10
exec() from starlark considered harmful


jpo-joyent:
With all due respect to @sean-, I think the addition of `exec()` in https://github.com/thought-machine/please/pull/492 is a mistake.

Part of the whole point of Starlark is that it takes precautions to ensure evaluation of the language is a closed system without side-effects, which (barring implementation bugs) makes it safe to evaluate untrusted build rules. With this change, a simple `plz query alltargets` (or even auto-completion of build rules from a shell) would be sufficient to compromise the machine if operating on untrusted BUILD rules.

The introduction of `exec()` breaks that (security-relevant!) promise, not to mention the whole host of reproducibility surprises it necessarily introduces.

Instead, I would strongly recommend that such powerful functionality with infinite unimagined (and possibly malicious) side-effects be limited to build execution time (bash) which can be subsequently sandboxed, rather than build graph evaluation (starlark) which currently goes to great lengths to maintain safety guarantees.

We can provide native implementations of `git_*()` rather than using `exec()` for them, and I'd be happy to send a PR for this next week when I'll have more time.

sean-:
Hrm.  This issue spun off into a spirited discussion offline (thank you @jpo-joyent for raising a number of good points with me - this is what I get for answering the question, "when do we get tagged binaries again?" 😀).

I think there are a few things that need to happen:

- [ ] `exec()` needs to be removed from the public list of registered builtins in `src/parse/asp/builtins.go` that way there's no change in the usable API at the time of the next release.
- [ ] `build_rule` needs to accept an "extra_data" argument that `go_get()` and `go_test()` can use.  This value will contain an opaque string of data that is generated by `go_get()` and passed to `build_rule()` and used as input to `fs/PathHasher` (I'm almost done with this patch).  This is required for reproducibility.
- [ ] The existing `exec()` code can be used as the backend for a native collection of `git_*()` functions that are exported and generally available for use by the world.

There are two arguments that were made that really hit home for me:

1. The `safety of build rule evaluation seems like an obvious intended property` and `exec()` at graph-traversal time invalidates that property of `plz`.
2. I do have aspirations for seeing `plz` used more widely and in places not considering its use yet, and this function would reduce `plz`'s value to something closer to `make`, which is not an analogy or comparison I wish upon anyone.

Until point 2 was made, my sense of loss-aversion didn't kick in and I was content with relying on the *NIX argument of, "everything in UNIX is a foot-gun, don't shoot your foot off."

[...]

PR #503, opened 2018-12-10
Extend build_rule to support extra_hash_data

sean-:
Add a new argument to `build_rue()` named `extra_hash_data=str|dict=""`.  This is used to pass in arbitrary data from the caller into the filesystem `PathHasher`.  This is required by `go_get()` and `go_test()` to update the hash used for storing targets when the linker flags change.

This PR fixes a bug that was introduced in #491 

Fix 2/3 from the changes outlined in the bottom of #501


peterebden:
I'm not sure this needs to go into the `PathHasher`, that is really about hashing files. Since it's an extra attribute on the target itself, why can it not be dealt with in `ruleHash` (in `incrementality.go`) rather than as part of the source hash?

I'm a little curious about what the "other side" of this looks like, i.e. how it gets consumed downstream by the things that want to use it. In many ways it looks similar to some of the C++ stuff - those rules have properties like `includes` and `defines` that are consumed downstream via the `labels` property. If we can use the same mechanism that might save a bunch of extra work.

peterebden:
Thanks, looks good now!

Just a quick question before I merge - can you explain a little bit more what the incrementality issue is? Theoretically all the observable properties are already captured by `ruleHash`, so I would have thought that changing `definitions` would change the command and that would be enough?

sean-:
! Yes it would.  Discussed offline, but I didn't realize that `RuleHash` was in play for this, which does all of this automatically.  I'm going to close this PR as it's not required (I thought it was required when I was using the `PathHasher` but isn't as the `RuleHasher` always did the right thing for me).

The one question that we discussed offline that I don't have an answer for is how do I **elide** a value from the `RuleHash()`?  I actually want to do the inverse of this logic and **not** rebuild a binary of the git sha changes.

In a monorepo, let's say I have two commands, `cmdA` and `cmdB`.  If I stamp both binaries with SHAs, I only want to force a rebuild of the binary and re-link the binary when the source material going into the binary changes.  Just because I made a git commit doesn't mean I want to re-link all of my binaries.  I only want to re-link and re-stamp my binary with the git sha that last had an affect on my binary.

If I have a stable command, I don't want to make a change to its git sha - I only want the git sha at which any of its dependencies changed.  It's like a I need a post-commit hook that writes out a manifest file somewhere that says `cmdA` was last updated in commit sha `1234abcd` and `cmdB` was just updated and uses `fdcba098`.  Knowing that the behavior or transitive dependency of `cmdA` hasn't changed is useful.  I'm not sure how to record that information other than in an external file that is checked in but contains this information.  `git` can't infer when `cmdA` or `cmdB`'s dependencies changed, but `plz` does know that... but isn't writing that out anywhere that can be later referenced.

I think what I want is a `plz commit` that writes out a manifest in each directory adjacent to `go_binary()` that I can query and as the input to a `definition`?  I'm still noodling on this.

jpo-joyent:
Actually, if you query the git graph at every build step and persist that to the incremental-tracking DB, then you already know when the *result* of BUILD files have an impact on how you build things, and don't need to care about any subsequent changes to BUILD files themselves if they don't cause rebuilds of their targets.

I think a git rev-list of transitive srcs may actually be all that's necessary.

peterebden:
Right, I see the goal, and agree it would be useful. Also agree with @jpo-joyent that having a separate command is undesirable.

Back in the day, Blaze had a genrule parameter called `stamp` that could be used to embed that kind of information (it was a bit frowned upon but obviously had its uses). We also have it although it is used for some hash stuff instead (which in hindsight I think it is probably unnecessary for). I wonder about retooling that to provide some information like git sha - although that implementation wouldn't be available at parse time to allow adjusting `definitions` or similar.

PR #504, opened 2018-12-10
Added doc for LPS :)

BNMetrics:


docs/quickstart.html
```diff
@@ -25,6 +25,17 @@ <h2>Shell completion</h2>
     <p>The script will complete subcommands and flags but most relevantly can also complete
       build labels by reading BUILD files for you.</p>
 
+    <h2>Buildfile Language Protocol Server(LPS)</h2>
+    <p>Please ships with a language server for build files. LPS follows standard
+        <a href="https://langserver.org/">language server protocol</a>
+    from VS Code.</p>
+
+    <p>To use LPS, simply download the language server protocol plugin for your favorite editor,
+    and run the binary at <code>~/.please/build_langserver</code></p>
```
peterebden: this should probably suggest `plz tool langserver` rather than the direct location?

docs/quickstart.html
```diff
@@ -25,6 +25,17 @@ <h2>Shell completion</h2>
...
+
+    <p>Currently LPS supports auto-completion(this works with build labels too!), hover, goto defition,
```
peterebden: defition -> definition.

docs/quickstart.html
BNMetrics: In that case, I will make it `plz tool lps` :)

src/tool/tool.go
```diff
@@ -58,7 +58,7 @@ func matchingTools(config *core.Configuration, prefix string) map[string]string
 		"jarcat":      config.Java.JarCatTool,
 		"javacworker": config.Java.JavacWorker,
 		"junitrunner": config.Java.JUnitRunner,
-		"langserver":  path.Join(config.Please.Location, "build_langserver"),
```
peterebden: can you leave the old one in too please? at least for now, otherwise anyone using it now will break.

BNMetrics: sure! Done!

Issue #470, opened 2018-11-18
Import cycle while building go code


kormat:
I was trying to convert an old project to mine to use please, as a trail run. Everything was going more-or-less fine until i ran into this seemingly-impossible error: `panic: found import cycle while visiting net/http`. After some hours, i managed to create a minimal reproduction case.

`/BUILD`:
```
go_binary(
    name = "contexttest",
    srcs = ["main.go"],
    deps = [ "//context"],
)
```
`/main.go`:
```
package main

import (
        _ "net/http"
)

func main() {}
```
`/context/BUILD`:
```
go_library(
    name = "context",
    srcs = ["context.go"],
    visibility = ["//..."],
)
```
`/context/context.go`:
```
package context

import (
        _ "net/http"
)
```

And here is the resulting output:
```
$ plz build
Build stopped after 300ms. 1 target failed:
    //:contexttest
Error building target //:contexttest: exit status 1
panic: found import cycle while visiting net/http

goroutine 1 [running]:
cmd/link/internal/ld.dfs(0xc00050f930, 0xc0005289f0, 0xc0005289d8)
        /usr/local/go/src/cmd/link/internal/ld/lib.go:2323 +0x280
cmd/link/internal/ld.dfs(0xc001b7edd0, 0xc0005289f0, 0xc0005289d8)
        /usr/local/go/src/cmd/link/internal/ld/lib.go:2327 +0x10a
cmd/link/internal/ld.dfs(0xc001b700d0, 0xc0005289f0, 0xc0005289d8)
        /usr/local/go/src/cmd/link/internal/ld/lib.go:2327 +0x10a
cmd/link/internal/ld.dfs(0xc001eac820, 0xc0005289f0, 0xc0005289d8)
        /usr/local/go/src/cmd/link/internal/ld/lib.go:2327 +0x10a
cmd/link/internal/ld.dfs(0xc001b660d0, 0xc0005289f0, 0xc0005289d8)
        /usr/local/go/src/cmd/link/internal/ld/lib.go:2327 +0x10a
cmd/link/internal/ld.dfs(0xc00050f930, 0xc0005289f0, 0xc0005289d8)
        /usr/local/go/src/cmd/link/internal/ld/lib.go:2327 +0x10a
cmd/link/internal/ld.dfs(0xc00050e340, 0xc0005289f0, 0xc0005289d8)
        /usr/local/go/src/cmd/link/internal/ld/lib.go:2327 +0x10a
[...]

Issue #455, opened 2018-11-11
Bootstrapping Please...


sean-:
Given that `plz` has support for `importpath`, would there be an objection to converting the `import` paths in `plz` to support canonical Go import paths so that `plz` could be `go-gettable`?

```
# in .plzconfig
[Go]
	importpath = "github.com/thought-machine/please"
```

This change may not be the only prerequisite for making `plz` "go get"-able, but it's certainly one of the first steps required.  I'm very interested in reducing the friction associated with getting `plz` used in other projects. It makes sense that `plz` remain able to be built and released by `plz`, but random developers should be able to `go get github.com/thought-machine/please/cmd/plz` and have that work.  TL;DR: I want to reduce the friction associated with consuming `plz` so it can be a more approachable build tool for the masses.

In support of the above goal of making `plz` `go get`able, I'm happy to do the above conversion but want to make sure that work would be welcome.  Making sure `plz` is generally usable in a more broad open source capacity is important for us before we fully hunker down on using `plz` and then profess it to be "the one true way" to build the software we're working on.  For instance, we're getting there, but are still deciding:

```
% find $GOPATH/${OURORG}/${OURMONOREPO} -name BUILD | grep -vE '(vendor|plz)' | xargs cat | wc
    5952    8807  128036
```

Thanks for your consideration.

sean-:
If someone cares, this is what's required to install `plz` from Go.

```go
// DependencyToolingPlease installs the plz tool into $GOPATH/bin
func DependencyToolingPlease(ctx context.Context) error {
        const plzBaseURL = "https://my.server.here/subpath/plz"
        var plzVersion semver.Version

        { // Find the version
                latestVersionURL := plzBaseURL + "/latest_version"
                req, err := http.NewRequest("GET", latestVersionURL, nil)
                if err != nil {
                        return errors.Wrap(err, "unable to create new please version request")
                }

                resp, err := ctxhttp.Do(ctx, http.DefaultClient, req)
                if err != nil {
                        return errors.Wrapf(err, "unable to download plz/latest_version %q", latestVersionURL)
                }
                defer resp.Body.Close()

                body, err := ioutil.ReadAll(resp.Body)
                if err != nil {
                        return errors.Wrapf(err, "unable to read plz/latest_version body from %q", latestVersionURL)
                }

                plzVersion, err = semver.Parse(string(bytes.TrimSpace(body)))
                if err != nil {
[...]

PR #505, opened 2018-12-11
Rewrite all import paths to their full form

peterebden:
i.e. use github.com/thought-machine/please/<whatever> everywhere.

Relates to #455 and #470 .


sean-:
`go-guru` and the jump-to-definition in editors everywhere thank you.

PR #506, opened 2018-12-11
Add native git_*() commands to plz...

sean-:
Build upon the unexposed work in #492 and add native `git_*()` commands.

- `git_branch()` - Returns a string of the current branch
- `git_commit()` - Returns a string of the git SHA of `HEAD`
- `git_show(format:str)` - Returns the output of a `git show --format={format}`, including the last commit time when using the `%ct` format verb.
- `git_state()` - Returns the output of a given repository (i.e. clean or dirty).

This PR rounds out the last of the concerns addressed in #501.


src/parse/asp/exec.go
```diff
@@ -96,6 +185,167 @@ func execGetCachedOutput(key execKey, args []string) (output string, found bool)
...
+
+// execGitShow returns the output of a git_show() command with a strict format.
+//
+// git_show() returns the output of `git show -s --format=%{fmt}`
+//
+// %ci == commit-date:
+//   `git show -s --format=%ci` = 2018-12-10 00:53:35 -0800
+func execGitShow(s *scope, args []pyObject) pyObject {
+	formatVerb := args[0].(pyString)
+	switch formatVerb {
```
peterebden: why do we need to check the format so closely? can we not just leave that up to git - or are you trying to filter out some of them?

src/parse/asp/builtins.go
```diff
@@ -80,6 +77,10 @@ func registerBuiltins(s *scope) {
 		"get":        setNativeCode(s, "config_get", configGet),
 		"setdefault": s.Lookup("setdefault").(*pyFunc),
 	}
+	setNativeCode(s, "git_branch", execGitBranch)
```
peterebden: I wonder if, rather than having four builtins here, if there was just one for running a git command and then four wrappers around it in the builtin rules that invoke it with different arguments. Thoughts?

src/parse/asp/exec.go
sean-: We're definitely trying to filter some of them out.  I didn't want to permit any of the ASCII control characters or audit for possible side effects from potential expansions, etc.  I'm okay relaxing this in the future, but I haven't dug into `git show`'s code to see what or how this could be abused.  For instance, I don't know what in `git show`, and by extension parts of `git config` let someone pass in a `PAGER` or some other value that would be interpreted as a shell command.

src/parse/asp/builtins.go
sean-: I thought about that, too.  Something like a `git()` command with a list of sub-commands, like:

```python
def git(command:list, args:list=None) -> str:
    pass
```

Where `command` is a whitelist of commands.  Maybe?  That seemed like a huge can of worms to me.  In most cases, you want to interrogate a git repository and report out various bits of information.  Judging by prior art, in the common case it looks like most people seem to be interested in a handful of commands and pieces of information, not all of the functionality available to them via `git()`.  From a usability perspective, it seemed like there were maybe only 3-4 things that people would want to regularly use (see [`govvv`](https://github.com/ahmetb/govvv) and [`goreleaser`](https://github.com/goreleaser/goreleaser/blob/e08cc5003097b71b17e0ec86c1354de9792832c7/www/content/build.md#L46)).

peterebden: Yeah, I sort of imagined exposing only the specific commands (`git_commit` and so forth), although I don't *think* there is a suitable mechanism for hiding the underlying `git` function.

I guess that's an implementation detail. Can always change that later since we're agreed on the exposed interface.

src/parse/asp/exec.go
peterebden: ugh right, makes sense then :)

src/parse/asp/targets.go
```diff
@@ -159,6 +163,43 @@ func populateTarget(s *scope, t *core.BuildTarget, args []pyObject) {
 	}
 }
 
+// addExtraHashData creates a suitable []byte of data that can be used by
+// ExtraHashData.  If a pyString is passed in, return the string itself.  If a
+// dict is passed in, sort the keys and concatenate the key and value and return
+// it as a response.
+func addExtraHashData(s *scope, name string, obj pyObject) []byte {
```
peterebden: do you still need this given discussion in #503?

sean-: no, this is just a byproduct of basing this branch on `gh-503b`.  I'll post an update soon.

src/parse/asp/exec.go
```diff
@@ -96,6 +185,167 @@ func execGetCachedOutput(key execKey, args []string) (output string, found bool)
...
+//
+// git_commit() returns the output of `git rev-parse --short HEAD`
+func execGitCommit(s *scope, args []pyObject) pyObject {
+	short := args[0].IsTruthy()
+
+	cmdIn := make([]pyObject, 2, 4)
+	cmdIn[0] = pyString("git")
+	cmdIn[1] = pyString("rev-parse")
+	if short {
+		cmdIn = append(cmdIn, pyString("--short"))
```
jpo-joyent: The length of `--short` depends on your git config value of `core.abbrev`. This is a reproducibility issue.

I would recommend we not support hash truncation at this stage at all. I cannot think of any case where it would not be desirable from a collision-resistance perspective to always pass the full SHA. If someone only wants to actually display part of it in a `--version` or whatever, then they can truncate when printing.

src/parse/asp/exec.go
jpo-joyent: I agree with restricting this.

src/parse/asp/exec.go
```diff
@@ -35,6 +40,90 @@ func init() {
...
 
+// doExec fork/exec's a command and returns the output as a string.  exec
+// accepts either a string or a list of commands and arguments.  The output from
+// exec() is memoized by default to prevent side effects and aid in performance
+// of duplicate calls to the same command with the same arguments (e.g. `git
+// rev-parse --short HEAD`).  The output from exec()'ed commands must be
+// reproducible.
+//
+// NOTE: Commands that rely on the current working directory must not be cached.
+func doExec(s *scope, args []pyObject) pyObject {
```
jpo-joyent: Why does doExec() still take args in python types? Is this just an artifact from when exec() was going to be exposed to starlark?

Why not just []string?

src/parse/asp/exec.go
```diff
@@ -96,6 +185,167 @@ func execGetCachedOutput(key execKey, args []string) (output string, found bool)
...
+// execGitBranch returns the output of a git_branch() command.
+//
+// git_branch() returns the output of `git symbolic-ref -q --short HEAD`
+func execGitBranch(s *scope, args []pyObject) pyObject {
+	short := args[0].IsTruthy()
+
+	cmdIn := make([]pyObject, 3, 5)
+	cmdIn[0] = pyString("git")
+	cmdIn[1] = pyString("symbolic-ref")
+	cmdIn[2] = pyString("-q")
```
jpo-joyent: The `pyString(...)` conversion seems awkward and unnecessary to me. See comment above at `doExec()`

src/parse/asp/exec.go
```diff
@@ -96,6 +185,167 @@ func execGetCachedOutput(key execKey, args []string) (output string, found bool)
...
+		cmdIn = append(cmdIn, pyString("--short"))
+	}
+	cmdIn = append(cmdIn, pyString("HEAD"))
+
+	wantStdout := True
+	wantStderr := False
+	cacheOutput := True
+	keyExtra := pyString("")
+
+	execArgs := []pyObject{
```
jpo-joyent: Same here

jpo-joyent:
The support for commit truncation seems problematic to me. If we're going to support it at all (and I believe we shouldn't) then we should at least require truncation to an explicitly specified length, and should default to full SHAs, not short.

jpo-joyent:
`rev-parse --short` can determine a length which is at-least-N and still unique (at least among the commits known to that particular instance of the repo), but I'd not like to rely on that either.

peterebden:
+1 we have seen some odd little nondeterminisms from things that rely on short revs since they can suddenly gain an extra character after additional commits. Definitely prefer for plz to use full ones.

src/parse/asp/exec.go
sean-: > The length of `--short` depends on your git config value of `core.abbrev`. This is a reproducibility issue.

Drat, I forgot about that configuration parameter.  Allowing anything that allows locale to slip through is also problematic, too (e.g. timezones).  I'm going to trim the amount of data that's included in format as a result and will dial back to a more austere set of format options.

> I would recommend we not support hash truncation at this stage at all. I cannot think of any case where it would not be desirable from a collision-resistance perspective to always pass the full SHA. If someone only wants to actually display part of it in a `--version` or whatever, then they can truncate when printing.

I'm unconvinced by any of the arguments made, but am going to take the position that there should be only one way to do things and that reproducibility is paramount (I don't see how this sha in this context could reliably used, but I digress).  This does shift the burden for formatting to the user, which I'm not wild about, but it leaves little ambiguity or chance for error due to user-specified configuration parameters in user-local git configs.  If someone does want the short SHA, they can specify `git_commit()[0:8]` and this value value will be stored in their `BUILD` files.

src/parse/asp/exec.go
sean-: Cleaned up and am using something closer to idiomatic Go now.

src/parse/asp/exec.go
sean-: I'm still doing this for now that way `doExec()` can take a `pyString()` or a `pyList()` if there are future callers into `doExec()`.  It can be moved to `[]string` without trouble.

PR #517, opened 2018-12-30
First step of upgrading third-party libraries

### LINKED ISSUE #456: Finding out of date packages...
Has there been any thought with regards to having `plz` query repositories referred to via `go_get()` and having it report when there's a new version available?  One of the things that we get from `dep` is the ability to run `dep update -ensure` and have it report when there's a new version available.  Since we've copied git SHAs into various `BUILD` files, it would be really, *really* great if there were a way for `plz` to query the remote site and see if something has been updated or slipped.  For instance, right now we're pulling in `133` different externally referenced repositories via `go_get()`.  Are we expected to manage that by hand?

Has any thought gone into this particular problem?

Right now, I'm concerned that this is going to be a manual process to update every dependency by hand.  In the short term, since we just pinned everything, this isn't a huge problem, but I'd like to establish a periodic job to track updates to these libraries so we can stay current with everyone's updates.  I'm of half-a-mind to suggest that `plz-out/go/src/` be a symlink to `vendor/` so we can diff changes to our dependencies and still use `dep`, but that still leaves us with the problem of updating our `revision` SHAs.

Any commentary or light on this subject would be most appreciated.  Thanks in advance.

peterebden:
Implements `plz upgrade`, which allows rules to declare how we should detect updates for them. 

Sample implementations provided for Java and Python (most of the relevant code is in pleasings), will add one for Go once I liberate some code from my work machine.

Right now all it does is prints out the new rules, in future we should extend it to actually rewrite the BUILD files appropriately (including merging the new and old rules).

Partly resolves #456


agenticarus:
Any desire in revisiting this?

peterebden:
Sort of - I'd like to have the feature, it's just a loooong way from the top of my priority list right now :(

PR #518, opened 2019-01-01
Look up tools via PATH

peterebden:
This fixes a (fairly old now) request in #455: look up the various tools via PATH rather than insisting on them being in the install location.

### DIFF
--- a/src/core/config.go
+++ b/src/core/config.go
@@ -564,7 +564,7 @@ func (config *Configuration) getBuildEnv(expanded bool) []string {
 		// but really external environment variables shouldn't affect this.
 		// The only concession is that ~ is expanded as the user's home directory
 		// in PATH entries.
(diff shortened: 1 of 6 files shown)


src/core/config.go
```diff
@@ -506,35 +505,70 @@ func (config *Configuration) ContainerisationHash() []byte {
...
+		env = append(env, pair)
+	}
+
+	// from the user's environment based on the PassEnv config keyword
+	path := false
+	for _, k := range config.Build.PassEnv {
+		if v, isSet := os.LookupEnv(k); isSet {
+			if k == "PATH" {
+				// plz's install location always needs to be on the path.
+				v += ":" + maybeExpandHomePath(config.Please.Location)
```
agenticarus: Do we want to ensure plz's path goes first here, to prevent randomly installed binaries with the same name clashing?

peterebden: Good idea!

PR #520, opened 2019-01-07
User cache dir default

### LINKED ISSUE #519: Move plz-cache to TMPDIR if they're on the same device...
One of the things that I've noticed since moving to `plz` is the rather excessive amount of backup thrash on macOS due to TimeMachine™.  This got me thinking: would it be possible or welcome to move `.plz-cache` to `$TMPDIR` *IF AND ONLY IF* `.plz-cache` is on the same filesystem?  Using `syscall.Stat()` it looks like it should be possible to get the `Dev` which we could then compare to see if `$PWD` is on the same filesystem as `$TMPDIR`, and if it is, change `plz-cache` to be somewhere inside of `$TMPDIR`.  Thoughts?

peterebden:
Fixes #519 

Ignores the hard bit about how we clean old ones by just not worrying about it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
    build-linux:
      working_directory: ~/please
      docker:
-       - image: thoughtmachine/please_ubuntu:latest
+       - image: thoughtmachine/please_ubuntu:20190107
      environment:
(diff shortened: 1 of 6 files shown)


peterebden:
Great, looks like it's time for a Go upgrade...

PR #538, opened 2019-01-17
Fix Go coverage for hyphens in path

### LINKED ISSUE #537: plz cover chokes on packages with hyphens in their name...
One of the protobuf files that we use has a hyphen in its name.  When we attempt
to run `plz cover` on this file, we run into a problem with coverage (but not
`plz test`).

```
% plz cover //foo/bar-baz/...
Build stopped after 4.11s. 1 target failed:
    //foo/bar-baz/service:_service_test#main_lib
Error building target //foo/bar-baz/service:_service_test#main_lib: exit status 1
foo/bar-baz/service/service_test_main.go, line 205: undefined: "github.com/myorg/myrepo/pb/int/bar-baz".GoCover_host
foo/bar-baz/service/service_test_main.go, line 205: undefined: net_pb_go in net_pb_go.Count
foo/bar-baz/service/service_test_main.go, line 205: undefined: "github.com/myorg/myrepo/pb/int/bar-baz".GoCover_host
foo/bar-baz/service/service_test_main.go, line 205: undefined: net_pb_go in net_pb_go.Pos
foo/bar-baz/service/service_test_main.go, line 205: undefined: "github.com/myorg/myrepo/pb/int/bar-baz".GoCover_host
foo/bar-baz/service/service_test_main.go, line 205: undefined: net_pb_go in net_pb_go.NumStmt
```

When I look in the generated file (`service_test_main.go`) at line 205, I see
the following:

```
  coverRegisterFile("pb/int/bar-baz/bar-baz_pb.go", _cover30.GoCover_bar-baz_pb_go.Count[:], _cover30.GoCover_bar-baz_pb_go.Pos[:], _cover30.GoCover_bar-baz_pb_go.NumStmt[:])
```

Which obviously doesn't compile.  In looking into
`tools/please_go_test/gotest/find_cover_vars.go` and at `coverVar()`, it looks
like there is no escaping and the raw filename is being used instead of the
package name itself.  Is there a helper method that we can use that will pull in
the right package here?

peterebden:
Bit nervous that the space of allowable characters in filenames is a lot higher than that of valid Go identifiers - maybe it should be replacing everything outside a set...

Fixes #537

### DIFF
--- a/tools/please_go_test/gotest/find_cover_vars.go
+++ b/tools/please_go_test/gotest/find_cover_vars.go
@@ -3,6 +3,8 @@ package gotest
 
 import (
 	"go/build"
+	"go/parser"
+	"go/token"
 	"io/ioutil"
 	"path"
 	"path/filepath"
@@ -70,7 +72,12 @@ func findCoverVars(filepath, importPath string, srcs []string) ([]CoverVar, erro
 			if ok, err := build.Default.MatchFile(dir, name); ok && err == nil {
 				// N.B. The scheme here must match what we do in go_rules.build_defs
 				v := "GoCover_" + strings.Replace(name, ".", "_", -1)
-				ret = append(ret, coverVar(dir, importPath, v))
+				fs := token.NewFileSet()
+				ast, err := parser.ParseFile(fs, path.Join(dir, name), nil, parser.PackageClauseOnly)
+				if err != nil {
+					return nil, err
+				}
+				ret = append(ret, coverVar(dir, importPath, v, ast.Name.Name))
 			}
 		}
 	}
@@ -86,7 +93,7 @@ func contains(needle string, haystack []string) bool {
 	return false
 }
 
-func coverVar(dir, importPath, v string) CoverVar {
+func coverVar(dir, importPath, v, pkg string) CoverVar {
 	log.Info("Found cover variable: %s %s %s", dir, importPath, v)
 	f := path.Join(dir, strings.TrimPrefix(v, "GoCover_"))
 	if strings.HasSuffix(f, "_go") {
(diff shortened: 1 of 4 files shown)


tools/please_go_test/gotest/find_cover_vars.go
```diff
@@ -69,7 +69,7 @@ func findCoverVars(filepath, importPath string, srcs []string) ([]CoverVar, erro
 		} else if strings.HasSuffix(name, ".go") && !info.IsDir() && !contains(path.Join(dir, name), srcs) {
 			if ok, err := build.Default.MatchFile(dir, name); ok && err == nil {
 				// N.B. The scheme here must match what we do in go_rules.build_defs
-				v := "GoCover_" + strings.Replace(name, ".", "_", -1)
+				v := "GoCover_" + strings.Replace(strings.Replace(name, ".", "_", -1), "-", "_", -1)
```
sean-: That works, too.  I haven't ventured into this area of the code but was thinking there was a https://golang.org/pkg/go/ast/#File -like struct that could be passed around and used to get the correct `package` name and then there would be a helper function that would filter out invalid characters.  For now, how about:

```go
replacements := []string{
    ".", "_",
    "-", "_",
}
r := strings.NewReplacer(replacements...)
v := "GoCover_" + r.Replace(name)
```

That way future additions are easier to add?

peterebden: I did look at getting the AST earlier on but turns out we don't actually need the package name - they get renamed on import into main. I guess we could use AST parsing to find the names of the cover vars and avoid this - but I'd rather not reparse it really.

Good idea on the Replacer. Not sure I've actually used one of those before...

PR #543, opened 2019-01-28
Exclude only files with _src.jar suffix

agenticarus:
This caught me out because I created a scala_library with name 'src' and it didn't get included in my app. I think '_src.jar' should be a suitable replacement here. Or we could just not include the src jars in the dependency somehow?


peterebden:
I'm not 100% sure we need that whole feature any more. It used to be needed to filter the source jars from third-party maven_jar rules but that now happens by a different mechanism whereby they don't even turn up as srcs.

Might be better to remove it completely since there will always be some set of filenames that this could be surprising for.

agenticarus:
Well the `*_src.jar` files do still show up in the tree so there needs to be some sort of filtering

agenticarus:
image

agenticarus:
Agreed - by fixing up the scala rules, the _src.jar  files no longer show up, so removing this exclude from jarcat entirely seems like a good plan

peterebden:
I think the failures are just due to the branch being a ways behind master. Probably CI should merge it in...

PR #549, opened 2019-02-03
Experimenting with /// syntax for subrepos.

peterebden:
As discussed the other evening, am playing with what a `///` prefix would look like for subrepos.

I think it's a bit less "alien" than Bazel's @ but maybe a bit less obvious what's going on since there can be quite a lot of slashes.

Also debating how one should indicate something in the root repo - right now `@//tools/jarcat` does that, I don't think I like `/////tools/jarcat` and would prefer `///tools/jarcat`, but that is a little less flexible for some of the cross-compiling stuff that refers to targets in the local package and would have no way of doing that any more.

### DIFF
--- a/docs/dependencies.html
+++ b/docs/dependencies.html
@@ -55,21 +55,22 @@
 )
 </code></pre></p>
 
-<p>Rules within subrepos can be referenced using an <code>@</code> prefix on rules, anywhere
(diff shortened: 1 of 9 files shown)


RichardoC:
LGTM

PR #558, opened 2019-02-15
Making licences field visible in Python & Java build defs

geonaut:
Adding licence fields to please build_defs


agenticarus:
LGTM

PR #565, opened 2019-02-26
Use Go 1.12 in Docker image

peterebden:
Looks like we need to make the asm stuff more complex for it to work now.


peterebden:
Oh great so `go tool asm` basically requires a step with `-gensymabis` which doesn't exist in 1.11 🙄

PR #580, opened 2019-04-01
Add cobol_library, cobol_binary rules.

peterebden:
Thought Machine is targeting the banking industry so we find it useful to have first-class support for a language our clients are comfortable with. Furthermore, introducing this into Please will make it more readily usable for a wider selection of users who we think could benefit from a stable language with a long pedigree.


Tatskaari:
Nice to see us adopting more sensible technologies. I wonder if we have considered adding Erlang support. It's a mature language that has seen widespread adoption in the telecoms industry where distributed, fault tolerant and highly available system are key. While kubernetes offers some benefits in these regards, Erlangs standard libraries includes ORP which provide first class support for this out of the box. The advantage of this over kubernetes is that it doesn't require a linux kernel opening up the opportunity to run on the banks existing mainframe infrastructure. I think the banking industry could also benefit from Erlangs standard number type being arbitrary precision.

csdigi:
Finally we can support mature and stable technologies. Do we want to add a debug flavour to enable the runtime error checking (cobc -d)? Huge productivity gain incoming...

peterebden:
Good idea, thanks Will, have added the debug flavour.

> I wonder if we have considered adding Erlang support. 

Nope, but I have a hole in the roadmap about this time next year... ;-)

peterebden:
Unfortunately efforts are being refocused on `fortran_library` instead...

agenticarus:
Disappointing.

PR #586, opened 2019-04-08
cached suites have junit property cached set to true

goddenrich:
This pr sets the suite property in the junit output cached to true if the test suite was cached. If the property doesn't exist or is set to false the suite is assumed to not have been cached.

### DIFF
--- a/src/test/xml_results.go
+++ b/src/test/xml_results.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"strconv"
 	"time"
 
 	"github.com/thought-machine/please/src/core"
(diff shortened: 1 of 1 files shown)


src/test/xml_results.go
```diff
@@ -96,9 +98,26 @@ func toCoreTestSuite(xmlTestSuite jUnitXMLTestSuite) core.TestSuite {
 	return testSuite
 }
 
+func toCoreCached(properties jUnitXMLProperties) bool {
+	cached := false
+	for _, prop := range properties.Property {
+		if prop.Name == "cached" {
+			p, err := strconv.ParseBool(prop.Value)
+			if err == nil {
+				cached = p
```
peterebden: nit: one-line it all here:
```
if p, err := strconv.ParseBool(prop.Value); err == nil {
    return p
}
```

PR #588, opened 2019-04-09
Set _GOPATH correctly when cross-compiling.

### LINKED ISSUE #585: _GOPATH incorrectly set for cross-compiled go_library targets
By defining `_GOPATH` [here](https://github.com/thought-machine/please/blob/v13.5.1/src/parse/rules/go_rules.build_defs#L7), the variable seems to pick up the host `CONFIG.OS` and `CONFIG.ARCH`. The result is that setting `--arch` does not override the paths in `_GOPATH`; builds subsequently fail.

This appears to be fixable by inlining the definition of `_GOPATH` into `_go_library_cmds` and `_go_binary_cmds`. I'll put a PR together in a bit, if nobody beats me to it first :-P

katzdm:
Resolves #585.

### DIFF
--- a/src/parse/rules/go_rules.build_defs
+++ b/src/parse/rules/go_rules.build_defs
@@ -4,7 +4,6 @@ Go has a strong built-in concept of packages so it's probably a good idea to mat
 rules to Go packages.
 """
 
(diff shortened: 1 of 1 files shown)


peterebden:
Nice, thanks!

PR #589, opened 2019-04-09
Use Host-OS to determine whether to workaround pip+darwin bug.

### LINKED ISSUE #581: pydistutils.cfg workaround not executed on MacOS when target-architecture is not "darwin"
[This workaround](https://github.com/thought-machine/please/blob/deef4d381b50ce4ac99afb1c975dcb525f8b4088/src/parse/rules/python_rules.build_defs#L342-L348) works excellently, to prevent the [notorious](https://stackoverflow.com/questions/24257803/distutilsoptionerror-must-supply-either-home-or-prefix-exec-prefix-not-both) `must supply either home or prefix/exec-prefix -- not both` error caused by brew's system-wide `distutils.cfg` file.

However, attempting to build with `--arch` skips this, instead using the [linux codepath](https://github.com/thought-machine/please/blob/deef4d381b50ce4ac99afb1c975dcb525f8b4088/src/parse/rules/python_rules.build_defs#L337-L341).

Ideally, the `echo "[install]\nprefix=" > .pydistutils.cfg` hack is executed whenever the _host OS_ is Darwin, rather than the target OS.

katzdm:
To do this, exposes two new build config entries, `CONFIG.HOSTOS` and `CONFIG.HOSTARCH`. In particular, `HOSTOS` is required for cross-compiling pip libraries from Darwin.

Resolves #581.

### DIFF
--- a/src/cli/flags.go
+++ b/src/cli/flags.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


peterebden:
Nice - I don't think `pip_library` will actually cross-compile in any useful way (I don't think you can get pip to do that?) but I guess it works for pure-Python libraries anyway after this change.

Issue #584, opened 2019-04-05
go_get with --arch=linux_x86 fails


katzdm:
Given a `go_get` rule in package `//third_party/go`:

```
go_get(
    name = "ratelimit",
    get = "gopkg.in/bsm/ratelimit.v1",
)
```

Running `plz build //third_party/go:ratelimit` on MacOS will succeed. However, running:

`plz build --arch=linux_x86 //third_party/go:ratelimit`

will fail. This is because the expected `ratelimit.v1.a` artifact goes into `{TMP_DIR}/pkg/linux_386/gopkg.in/bsm/ratelimit.v1`, but plz expects to find the artifact in a `linux_x86` directory. Consequently, running the same command with `_386` architecture:

`plz build --arch=linux_386 //third_party/go:ratelimit`

will succeed.

katzdm:
Hey @peterebden - Thanks for addressing this one! Unfortunately, I believe there is a second, related issue. `go_library` targets don't seem to find libraries fetched by `go_get`, when cross-compiling with `--arch=linux_x86`. More precisely, if I create a target `//my_lib` to depend on the target from my original issue description, like:

```
go_library(
    name = "my_lib",
    srcs = ["my_lib.go"],
    deps = ["//third_party/go:ratelimit"],
)
```

Then this fails with `plz build --arch=linux_x86 //my_lib`:
> Error building target ///linux_x86//my_lib:my_lib: exit status 1
> my_lib/my_lib.go, line 9, column 2: can't find import: "gopkg.in/bsm/ratelimit.1/..."

Note that this continues to work, when compiling with `--arch=linux_386`. I assume this `OS` -> `GOOS` change probably needs to be made somewhere else, as well.

peterebden:
Ah yeah fair - hadn't gotten as far as testing that yet. I think you're probably right (and hopefully it is not hard).

peterebden:
FWIW f9129b4 should fix that error - but most libraries will fail by missing imports because we don't provide a cross-compiled stdlib for them. Would be nice to do that at some point but it's a bit more work than the fixes here :)

katzdm:
True, but you provided excellent guidance on http://plz.build for how to work around that, which I've already followed :-)

> Go can produce that via something like GOOS="linux" GOARCH="386" go install -i std.

Worked like a charm.

katzdm:
[...]

PR #591, opened 2019-04-10
Fix go_get --arch=linux_x86.

katzdm:
Fix last remaining issues with `go_library` and `go_binary` from #584.


peterebden:
Hmmm not sure what CircleCI is doing with those builds... will try to restart it.

peterebden:
Looks like they had a bit of an outage on workflows - all good again now.

Issue #572, opened 2019-03-20
Rules containing multiple secrets only substitute "~" in first secret


katzdm:
This might be due to the regex [here](https://github.com/thought-machine/please/blob/f0b21e617ceeebad85555282d0650355de5b9e46/src/fs/home.go#L10), coupled with how secrets are joined [here](https://github.com/thought-machine/please/blob/f0b21e617ceeebad85555282d0650355de5b9e46/src/core/build_env.go#L98).

Somewhat separately - it might be good if `secrets` could be specified as a `dict` rather than a `list`, with the items then published to the `cmd` as `KEY=replaced_value` environment variables - Thoughts?

PR #593, opened 2019-04-10
Implement named secrets; replace '~' when present in multiple secrets.

katzdm:
Addresses #572.


katzdm:
Oh right...The test I wrote ain't gonna quite fly (assuming the existence of a `.vimrc`). Let me rework this.

katzdm:
@peterebden , is the `maven_test` failure related to this?

test/secrets/BUILD
```diff
@@ -1,10 +1,29 @@
...
+echo "[ \"$SECRETS_AD_SECRETS\" = \"/bin/sh $SECRETS_HOME\" ] || \
+(echo "Failed: $SECRETS_HOME $SECRETS_AD_SECRETS"; exit 1)" > $OUT'''
+
+gentest(
+    name = "test_tildes_replaced",
+    outs = ["test_tildes_replaced.sh"],
+    cmd = tildes_replaced_test,
+    no_test_output = True,
+    secrets = {
+        "home": ["~"],
```
peterebden: Not sure we should do this in a test; it will end up needing to hash the entire home directory, which could get very large.
I see the motivation though; not sure what we can add instead that we'd be sure was going to be there.

peterebden:
Thanks, looks good, just concerned about what the behaviour of the test will be!

test/secrets/BUILD
katzdm: > Thanks, looks good, just concerned about what the behaviour of the test will be!

Yeah, I originally had this as `~/.bashrc`, but even that wasn't guaranteed to exist on CirrusCI haha...If you'd like, I can remove the test for now, and have a bit less coverage?

peterebden: We could just add a sneaky `touch ~/.bashrc` in the `.cirrus.yml`? Would be nice to keep this test case.

katzdm: Done, with tests passing 🎉

PR #595, opened 2019-04-15
Use subrepo with addDep().

### LINKED ISSUE #594: Post-build add_dep rule not compatible with cross-compilation
Given a genrule `custom_target` with a `post_build` function that executes `add_dep(target, dep)`, building `plz build --arch=ARCH` will add a dependency on `dep` instead of `'@ARCH//' + canonicalize(dep)`, resulting in an error that the dependency `dep` cannot be found in the build-graph.

See further details in [this comment](https://gitter.im/please-build/Lobby?at=5cb4b55a0aad6350198b53d6) on gitter.

katzdm:
Fixes #594 .


peterebden:
Makes sense!

PR #596, opened 2019-04-16
Allow config_setting rules to switch based on OS.

katzdm:
Small change to allow rules to `select` on conditions like:

```
config_setting(
    name = "linux",
    values = {"os": "linux"},
)
```

### DIFF
--- a/src/parse/rules/config_rules.build_defs
+++ b/src/parse/rules/config_rules.build_defs
@@ -61,6 +61,8 @@ def _config_on(name, value):
         return value.upper() in CONFIG
     elif name == 'crosstool_top':
         return False  # Android build - not clear right now how we would best represent this.
+    elif name == 'os':
+        return CONFIG.OS == value.lower()
     else:
         raise ParseError('Unknown config_setting key %s' % name)
 


katzdm:
Motivation here is to select different `outs` for `python_wheel`, based on which wheel is downloaded - Compiled shared objects might be named differently (or not present at all) in different releases of the same library (e.g. `_cffi_backend.cpython-37m-x86_64-linux-gnu.so` vs `_cffi_backend.cpython-37m-darwin.so`).

matthen:
(in this case we would usually do `post_install_commands = ["mv _cffi_backend.*.so _cffi_backend.so"]`

katzdm:
@matthen , thanks for the tip! That solved another issue I was having as well.

I'll leave it to maintainers to decide whether we want to merge this or not - I still suspect that switching on `os` is useful (I recall it being required many times with Bazel), but we can put it off for now if the change feels unneeded.

peterebden:
Yup I think this is useful - I only ever wrote a minimal set of those config settings for what I needed for compatibility at the time but this seems sensible to have.

PR #598, opened 2019-04-17
Speed up go_get: Shallow clone, don't fetch tags.

katzdm:
When `go_get` falls back to fetching a package via `git clone`, we can speed things up by fetching the revision needed, and ignoring all other tags/refs/history.


peterebden:
Nice idea!

PR #602, opened 2019-04-20
plz/watch: Increase debounce interval to prevent double-builds.

katzdm:
With one of my services I'm testing with, I noticed that the service would sometimes be restarted twice, for the same file-change. Doubling the debounce interval fixes the issue.

### DIFF
--- a/src/watch/watch.go
+++ b/src/watch/watch.go
@@ -23,7 +23,7 @@ import (
 
 var log = logging.MustGetLogger("watch")
 
-const debounceInterval = 50 * time.Millisecond
+const debounceInterval = 100 * time.Millisecond
 
 // A CallbackFunc is supplied to Watch in order to trigger a build.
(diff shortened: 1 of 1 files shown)


katzdm:
I was able to reproduce this issue locally on my Mac, but resolved it by:
- `brew install protobuf`
- Bumped version of `protoc` to `3.7.1` in `//third_party/proto/BUILD`

Seems like this test failure is maybe a problem with the CircleCI Docker image, though? Some research suggests that adding `protobuf` or `protobuf-dev` to [the image](https://hub.docker.com/r/thoughtmachine/please_ubuntu/dockerfile) might help.

Not sure why this is just happening now, though :-)

peterebden:
Computers, huh :)
Not sure how that test ever passed in that image. That's a classic developer excuse...

This change makes sense, anyway. There's obviously a tradeoff but 100ms is still pretty reasonable, given there isn't an obviously better way of resolving the actual problem.

PR #603, opened 2019-04-22
Small fixes for Go cross-compilation and plz-watch.

katzdm:
- Use `GOARCH` of `amd64` when `ARCH` is `x86-64`.
- Don't cease watching when a target's execution context is terminated before the target can be started.

Total mystery as to why `--arch=linux_x86-64` was ever working with Go in the first place, for me..

### DIFF
--- a/src/cli/flags.go
+++ b/src/cli/flags.go
@@ -280,6 +280,8 @@ func (arch *Arch) XArch() string {
 func (arch *Arch) GoArch() string {
 	if arch.Arch == "x86" {
 		return "386"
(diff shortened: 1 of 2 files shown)


katzdm:
Hey @peterebden - Just wanted to ping this one, in case it slipped your notice. Do these changes look reasonable to you? Would be great to see these fixes merged and released alongside the other recently merged PRs.

Thanks in advance for taking a look!

peterebden:
Sorry, yes, it had :) Thanks for the fixes!

PR #606, opened 2019-05-02
Add 'licences' field to go_get build def

henryaj:
Would be handy to be able to specify these manually.

At some point, could look at auto-detecting them (much like the `pip_library` rule, though I guess we'd have to do some clever parsing of a `LICENSE{.md}` file in the pulled Go repo).


peterebden:
Yup, I have idly thought of that before; soft-matching licences is not easy, and historically there wasn't anything sufficient out there for Go; https://github.com/google/licenseclassifier looks promising though.

Thanks!

PR #608, opened 2019-05-06
Small go_get improvement.

katzdm:
Include the last part of the repo in the naming formula for `go_get` downloads. As previously written, `golang.org/x/crypto` and `golang.org/x/sync` could not exist within the same `go_get` rule with revisions declared. Omitting revisions worked, but caused a fallback to the slow-path of `git clone`.

### DIFF
--- a/src/parse/rules/go_rules.build_defs
+++ b/src/parse/rules/go_rules.build_defs
@@ -660,10 +660,10 @@ def _go_github_repo_cmd(name, get, repo, revision):
     else:
         dest = parts[-1]
     remote_rule = remote_file(
-        name = name,
+        name = name + '-' + parts[-1],
         _tag = 'download',
         url = f'https://{repo}/archive/{revision}.zip',
-        out = name + '.zip',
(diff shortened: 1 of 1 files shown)


src/watch/watch.go
```diff
@@ -73,6 +73,7 @@ func Watch(state *core.BuildState, labels core.BuildLabels, callback CallbackFun
 			// Kill any previous process.
 			cancel()
 			ctx, cancel = context.WithCancel(parentCtx)
+			build(ctx, state, labels, callback)
```
peterebden: Not sure about this bit; delaying the build may be useful to make sure the changes get a chance to be written to disk before starting. It sort of feels wrong to me to be building sooner to try to prevent rebuilds...

katzdm: Fair enough; reverted this change such that this only includes the `go_get` fix!

peterebden:
Thanks, let's revisit the watch stuff separately :+1:

PR #610, opened 2019-05-07
Add "srcs" attribute to python_binary.

katzdm:
This allows binaries with multiple python source files to be defined, without explicitly defining a separate `python_library` target.

### DIFF
--- a/src/parse/rules/python_rules.build_defs
+++ b/src/parse/rules/python_rules.build_defs
@@ -92,7 +92,7 @@ def python_library(name:str, srcs:list=[], resources:list=[], deps:list=[], visi
     )
 
 
-def python_binary(name:str, main:str, resources:list=[], out:str=None, deps:list=[],
+def python_binary(name:str, main:str, srcs:list=[], resources:list=[], out:str=None, deps:list=[],
                   visibility:list=None, zip_safe:bool=None, strip:bool=False, interpreter:str=None,
                   shebang:str='', labels:list&features&tags=[]):
     """Generates a Python binary target.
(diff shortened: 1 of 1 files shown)


src/parse/rules/python_rules.build_defs
```diff
@@ -125,6 +126,14 @@ def python_binary(name:str, main:str, resources:list=[], out:str=None, deps:list
     cmd = '$TOOLS_PEX -s "%s" -m "%s" --zip_safe' % (shebang, CONFIG.PYTHON_MODULE_DIR)
     pre_build, cmd = _handle_zip_safe(cmd, zip_safe)
 
+    assert main not in srcs, "main file should not be included in srcs"
+    if srcs:
```
peterebden: I'm not sure you need a separate rule here - you should be able to just add them in on line 139 I think?

katzdm: That's what I'd initially tried, but I got a strange error message that led me to suspect something was having difficulty determining where to find `main`. Seemed easier to do things this way than to figure out how pex's actually work ;)

katzdm: Let me know whether you'd like me to try to get this working without an additional target, or if you think this is fine as-is.

src/scm/scm.go
```diff
@@ -26,7 +26,7 @@ type SCM interface {
 
 // New returns a new SCM instance for this repo root.
 // It returns nil if there is no known implementation there.
-
+//
 // If there is no known implementation it returns a stub.
```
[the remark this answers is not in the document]
peterebden: Don't worry, I've fixed this on master.

src/parse/rules/python_rules.build_defs
katzdm: Ah, I figured out the issue - I had added the `srcs` to `pex_rule` and `lib_rule`, but it was only needed for `lib_rule`  :-) Fixed now!

peterebden:
Thanks!

PR #616, opened 2019-05-12
Update hash validation

### LINKED ISSUE #615: Support checksum validation during auto-update
`plz` currently supports a field `Version` in the `[Please]` section of the configuration, which causes `plz` to auto-update to the provided release, and facilitates tracking the same release across an organization.

As per [discussion on gitter](https://gitter.im/please-build/Lobby?at=5cd4423cda34620ff91160e2), it would be great to support a `VersionChecksum` field. If present, during auto-update, `plz` will verify that a hash of the contents downloaded matches the digest provided in the `VersionChecksum` field; the update will fail if there is a mismatch. The expectation is that an engineer updating the `Version` field of an org-wide `.plzconfig` file will update `VersionChecksum` in the same commit.

This requires two pieces:
- Support for the `VersionChecksum` field, and integration with `plz update`,
- Update the release process, such that a checksum is published for each binary generated by a release (Linux, MacOS, etc)

peterebden:
Support checksums on downloaded updates, and upload during releases. Resolves #615.

I'm about as confident as I can be that this will be OK, but can't be 100% until we test it end-to-end.


katzdm:
Thanks for the quick turnaround on this one - Excited for it to land!

peterebden:
No worries - just trying to work out what Cirrus is upset about...

PR #617, opened 2019-05-13
Allow go_get definitions under other path locations then third_party/go

namtzigla:
In our repo we have a pretty large number of go dependencies and in order to avoid creating a single large file that will contain all deps we decided to change the traditional approach and split the in files using the following pattern: `third_party/go/[host]/[user]:[dep]`, but this started to throw errors like this one:
```
Error building target //commons/service:_go_test#lib: exit status 1
commons/service/driver_test.go, line 12, column 2: can't find import: "gopkg.in/DATA-DOG/go-sqlmock.v1"
```

This PR contain the changes that we adopted in order to overcome this issue.

### DIFF
--- a/src/parse/rules/go_rules.build_defs
+++ b/src/parse/rules/go_rules.build_defs
@@ -550,7 +550,7 @@ def go_get(name:str, get:str|list, repo:str='', deps:list=[], exported_deps:list
 
     # Now compile it in a separate step.
     cmd = [
-        f'export GOPATH="{CONFIG.GOPATH}" GO111MODULE="off"',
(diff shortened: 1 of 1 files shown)


peterebden:
Right, we use GOPATH from the config entry for the same thing, but it's a little less than ideal having to explicitly configure that.

FWIW I've been thinking for a while that we probably need to move away from calling `go get` directly, but haven't had time to finish up a replacement yet. This looks like a nice improvement on the status quo though!

namtzigla:
About go_get improvements, if you have time to put it in a doc/ticket I like to help you with that especially that we need it too.

Thanks!

Issue #605, opened 2019-05-02
Canceling plz build with CTRL+C does not terminate child processes


katzdm:
This is similar to the issue resolved with 594a0ec9a384cbbabc7f78171e93782e46493cd1, but applied to `plz build`.

For instance, suppose we have a build rule, in package `//third_party/go`:
```
go_get(
    name = "x",
    get = [
        "golang.org/x/crypto/...",
        ...
    ],
)
```

And we run `plz build //third_party/go:x` - Then, while the rule is fetching, the user cancels with `CTRL+C`. Subsequently running `ps` will reveal a potentially large number of processes still running (`git clone`, `/usr/local/bin/go get`, etc).

This becomes more harmful when the user tries to quickly re-execute the same `plz build` invocation - If the processes are still executing, then an error will be thrown regarding `plz` unable to get a lock. `plz clean` does not solve this; waiting for (or manually killing) the abandoned subprocesses is required.

katzdm:
Sorry, just got around to verifying this - As far as I can tell, this was fixed by #620. Thanks!

PR #620, opened 2019-05-19
Centralise subprocess management

peterebden:
Cleans a lot of code out of the overlarge `core` package. Hopefully will improve #605 by tracking all extant subprocesses and attempting to kill them on receiving a signal.

Still leaving in defense-in-depth stuff like PDEATHSIG where possible, but this should help on non-Linux platforms where that is unavailable.


katzdm:
Yaaaaaaas 🎉

PR #621, opened 2019-05-20
Explicit feature passing in the behave test runner integration

geonaut:
Adding an optional flag & method for extracting multiple feature files from a python_test() build target, so not all feature files have to reside in the same folder (e.g. where API versioning requires parallel test suites)


tools/please_pex/behave.py
```diff
@@ -1,17 +1,33 @@
...
 def get_features_dir():
     for i in TEST_NAMES:
         file_name = i.split('/')[-1]
         if '.feature' in file_name:
-            return os.path.dirname(i)
+            return [os.path.dirname(i)]
+
+
+def get_all_feature_files():
+    files = []
```
peterebden: nit: you can one-line this function using a list comprehension

PR #624, opened 2019-05-21
Allow for non-flat tarballs.

### LINKED ISSUE #622: Feature-request: Add option to tarball rule, to not flatten source files.
As discussed [here](https://gitter.im/please-build/Lobby?at=5ce2c72f63dea422b4b633d0) in the Gitter channel.

Reproducing the original proposal:
> As currently implemented, the `tarball(...)` rule flattens all input files into a single subdirectory. I would like to use `plz` to create tarballs with internal directory structures - Would we be open to adding a `flatten` flag to the `tarball` rule (with a default value of `True`, for compatibility)? Seems like the bulk of this work would be within the `jarcat` tool.
>
> Motivation: I would like to build tarballs that encompass the build context for a Docker image. A deployment tool, separate and distinct from `plz`, would consume this build-context as a post-build step, using it to build an image and upload it to an image repository. Given that these processes are side-effect-intensive, `plz` is not a good choice of tool to execute them. Docker's native support for tarballs makes this a good choice of format to pass the intermediate artifact.
>
> While this is already possible, the flattening of directory structures leads to an asymmetry between file locations in the repository vs runtime, as well as introducing possible conflicts between files with common base names, but distinct directories.

katzdm:
Adds two new options to the `tarball()` rule, via two new arguments to `jarcat tar`:

- `flatten`: If `True`, flattens the internal tarball directory structure. `True` by default, for compatibility. 

- `strip_prefix`: If `flatten` is `False`, strips this prefix from the internal directory structure of the tarball.

For non-flat tarballs, filepaths are therefore built as:
    `PATH = PREFIX + STRIP(STRIP_PREFIX, SRC_PATH)`

Closes #622

### DIFF
--- a/src/parse/rules/misc_rules.build_defs
+++ b/src/parse/rules/misc_rules.build_defs
@@ -390,7 +390,8 @@ def remote_file(name:str, url:str|list, hashes:list=None, out:str=None, binary:b
 
 
 def tarball(name:str, srcs:list, out:str=None, deps:list=None, subdir:str=None, gzip:bool=True,
-            xzip:bool=False, test_only:bool=False, visibility:list=None, labels:list&features&tags=[]):
+            xzip:bool=False, flatten:bool=True, strip_prefix:str='',
+            test_only:bool=False, visibility:list=None, labels:list&features&tags=[]):
     """Defines a rule to create a tarball containing outputs of other rules.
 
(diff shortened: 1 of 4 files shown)


peterebden:
Thanks!

PR #625, opened 2019-05-21
Allow specifying python interpreter options through global config.

katzdm:
Motivation: I would like for all `pex` files to be run with [`-S`](https://docs.python.org/3/using/cmdline.html#id3), to
disable installed site packages (e.g. those installed with `pip`). This
helps preserve hermeticity and reproducibility between the local dev
environment and CI, prod, etc.

I sort of think that `-S` should be the default behavior for any built
`pex` executable, but that would be a nontrivial breaking change. This at
least facilitates such a setting, without requiring authors to add an
explicit `shebang`/`interpreter` field to every `python_binary`/`python_test`
rule.

### DIFF
--- a/src/core/config.go
+++ b/src/core/config.go
@@ -278,6 +278,7 @@ func DefaultConfiguration() *Configuration {
 	config.Python.DefaultInterpreter = "python3"
 	config.Python.TestRunner = "unittest"
 	config.Python.UsePyPI = true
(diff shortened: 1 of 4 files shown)


peterebden:
Yep, makes sense - I have thought about having `-S` before, can't remember if there was a reason why I never did that but it definitely seems like the correct direction.

PR #637, opened 2019-05-30
Propagate zip_safe through transitive dependency edges.

katzdm:


peterebden:
I'm slightly dubious on this - it's a regression if your binary consumes some genrule etc that is generated using a non-zip-safe library, but that library itself isn't built into the final binary.

I guess that is rare enough that this is probably a better option - and I guess you have a use case that is basically the reverse where it works better?

katzdm:
> I guess you have a use case that is basically the reverse where it works better?

My use case is `pip_library` and `python_wheel` rules that (AFAICT) are inherently zip-unsafe, by virtue of installing a `.so` file (`cffi`, `cryptography`, etc). I would like for all binaries consuming such libraries to transitively also be zip-unsafe.

This also seems to be in line with the [documentation](https://please.build/lexicon.html#python_binary):
> by default it will be safe unless any of the transitive dependencies are themselves marked as not zip-safe.

Thoughts?

peterebden:
Yep, doesn't that happen already though? Setting `True` here lets it look through intermediate binaries but even without it the search should be transitive, it's just blocked by certain kinds of edges.

katzdm:
I've been observing cases where this behavior isn't working (hence my attempt to correct). I reduced it to a test case in which my binary, built with latest plz, failed while trying to call a cffi function (could not find the `.so`), but succeeded when built with a development build with `all = True`. The same binary runs well when `zip_safe` is explicitly set on the `python_binary`.

katzdm:
I'll try to put a reproducing case together, to demonstrate :-)

katzdm:
@peterebden , Thanks for pushing back on this! Upon further investigation, this was an issue on my side - I had a `genrule` wrapping a `python_wheel` rule that lacked an `output_is_complete` annotation :-) Closing this, thanks!

PR #641, opened 2019-06-05
Non-xattr fallbacks

### LINKED ISSUE #548: xattrs not available in nix builds
This is more of a heads-up than an issue or feature request.

I wanted to try packaging "please" for [nixpkgs](https://nixos.org/nixpkgs), but it's going to be difficult since [nix disables xattrs within builds](https://github.com/NixOS/nix/blob/master/src/libstore/build.cc#L2531-L2537) It appears the reasons for disabling them would not apply to "please" (other than desire to support filesystems that lack xattrs), but they are disabled at any rate.

peterebden:
Allows disabling these, and falls back to a file-based scheme (similar but not identical to the one we had before xattrs) in such a case. I think it avoids the issues we had with that one by writing one file per output rather than one per target, although there's a teensy bit more overhead as a result.

Resolves #548

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
    build-linux:
      working_directory: ~/please
      docker:
-       - image: thoughtmachine/please_ubuntu:20190226
+       - image: thoughtmachine/please_ubuntu:20190605
      environment:
        PLZ_ARGS: "-p --profile ci"
      steps:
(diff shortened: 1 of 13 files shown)


peterebden:
Thanks!

I'm pretty sure this needs the tyres kicking a bit harder given the build failure... hopefully it's nothing big.

peterebden:
Fortunately it wasn't :)

Issue #522, opened 2019-01-07
Remove --keep_going


peterebden:
I can't remember what I wanted the flag for but it is not very useful; it mostly seems to go on to other targets that then fail because their dependencies aren't built. That isn't the desired behaviour but I don't know if it's worth fixing; maybe we should just remove the flag & associated code if there is no real use case for it.

Definitely counts as a breaking change.

PR #655, opened 2019-06-11
Remove flag --keep_going and associated code

r0yce:
Attempting to complete #522

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -53,7 +53,6 @@ var opts struct {
 		Config     string            `short:"c" long:"config" env:"PLZ_BUILD_CONFIG" description:"Build config to use. Defaults to opt."`
 		Arch       cli.Arch          `short:"a" long:"arch" description:"Architecture to compile for."`
 		RepoRoot   cli.Filepath      `short:"r" long:"repo_root" description:"Root of repository to build."`
-		KeepGoing  bool              `short:"k" long:"keep_going" description:"Don't stop on first failed target."`
 		NumThreads int               `short:"n" long:"num_threads" description:"Number of concurrent build operations. Default is number of CPUs + 2."`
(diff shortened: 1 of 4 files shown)


r0yce:
apologies for continuing to work on this PR after the initial request.

i was going back and forth not knowing if i should be removing funcs not being used or not..

after removing the exported method, in the core package, state.DelayedKillAll() i thought that I should add it back given that is it indeed exported.
however, it isn't being used at the moment by any other package so i've decided to keep the change as is. I then realized that the state.DelayedKillAll()
method required the use of state.anyRunningTasks() so i went ahead and removed that method as well after checking to see if anything else called it.

should exported functions/methods be removed if not used, or should they remain? if the former, i belive this work to be complete.

src/please.go
```diff
@@ -643,14 +642,10 @@ var buildFunctions = map[string]func() bool{
 		return true
 	},
 	"changes": func() bool {
-		// Temporarily set this flag on to avoid fatal errors from the first parse.
```
peterebden: Now I see this, I'm slightly worried about losing this. I think the condition on line 98 of shell_output.go will now pass, which would cause the build to fail early here which isn't what we want.

I guess ideally we should refactor out that `os.Exit` call, it's obviously not very friendly...

peterebden: OK I'm going to merge this and then try to factor that into something nicer than calling `os.Exit`.

r0yce:
> Thanks for the PR!

Thank you for approving the PR. Apologies for going dark after the PR; some things came up I had to take care of. Anyway this was a pleasant experience and look forward to contributing more to the project in the future!

I agree on the not so friendly `os.Exit` refactor.

Thanks again for making my first PR for open source a great experience :)

peterebden:
No worries, I'm glad you enjoyed it! :)

Issue #654, opened 2019-06-10
Deadlock when cross-compiling all targets in all subpackages


katzdm:
Reproduced on on MacOS, and Docker-on-Mac. Quickest way to reproduce is probably:

```
cd `mktemp -d`
please init
touch BUILD
please build @NotEven_ARealSubrepo//...
```

I tracked the issue a bit, but didn't quite get to the bottom of it before the end of the day. The problem is that [findOriginalTask](https://github.com/thought-machine/please/blob/master/src/plz/plz.go#L144) waits for the target `//:NotEven_ARealSubrepo` via [WaitForBuiltTarget](https://github.com/thought-machine/please/blob/master/src/core/state.go#L590). This function adds a `parse` task to the task-queue, but ultimately the `Parse()` function exits (through a [highly specific special-case](https://github.com/thought-machine/please/blob/master/src/parse/parse_step.go#L120-L124)) without adding a `build` task, or any subsequent tasks, to the task-queue. This appears to end us up in a state where the task queue is empty, but `WaitForBuiltTarget()` is still spinning.

katzdm:
The practical use case blocked by this bug is use of `plz test --arch=linix_amd64 //pkg/...` .

PR #656, opened 2019-06-11
Don't deadlock when cross-compiling all subpackages.

katzdm:
Addresses #654. By setting the subrepo target to `TargetBuilt`, we close the associated channel and unblock dependors waiting for the subrepo to be instantiated.


peterebden:
Thanks for the fix!

PR #657, opened 2019-06-13
Add support for uploading test results to a remote server.

peterebden:


goddenrich:
Looks good to me. At some point might be good to be able to send partial results though XML might not be the best format for that.

PR #668, opened 2019-06-25
Cache contents of expanded zip-unsafe PEX files for subsequent runs.

katzdm:
Implements a caching scheme for zip-unsafe PEX python executables. In particular, the unzipped+compiled contents of PEX files are persisted into subdirectories of `~/.pex`. This yields significantly better performance on subsequent invocations of a PEX.

The motivating example for this PR is an internal command-line tool that utilizes the [boto3](https://github.com/boto/boto3) python library, which is [infamously](https://github.com/boto/boto3/issues/849#issuecomment-442795689) zip-unsafe.

When exploding a zip-unsafe PEX, we do the following:
1. Acquire a lock on a file with a unique name derived from plz's injected `$STAMP`.
2. Check whether the file is empty. Nonempty lockfile indicates that the cache-directory corresponding to the PEX has already been written to and made ready for use. In this case, release the lock and proceed with execution.
3. If the file is empty, then explode the PEX archive into the unique directory `~/.pex/pex-$STAMP/`, and use the `compileall` module to precompile its contents. Write a string to the lockfile before releasing it, to let future invocations know that the cache contents are ready for use.
4. Rewrite `sys.path` and proceed with program invocation.

An environment variable, `PEX_NOCACHE`, is introduced to disable caching. When this variable is present, the PEX is instead extracted to `TEMP_DIR`, and the contents (and lockfile) are erased on program-exit.

This  requires changes in a few other locations:
* `python_binary` is marked with `stamp=True`, to provide a deterministic mapping from "built PEX" to "cache directory".
* The `$STAMP` variable must be piped through `pex_tool`, to replace the macro `__PEX_STAMP__` in `pex_main.py`
* I extend the command-environment of all `plz run` executions with `PEX_NOCACHE=true`, in order to prevent caching of one-off `plz run`-invocations, every time a change is tested with an executable.

### DIFF
--- a/src/run/run_step.go
+++ b/src/run/run_step.go
@@ -139,12 +139,20 @@ func run(ctx context.Context, state *core.BuildState, label core.BuildLabel, arg
 // environ returns an appropriate environment for a command.
 func environ(config *core.Configuration, setenv bool) []string {
 	env := os.Environ()
+	for _, e := range adRunEnviron {
+		env = addEnv(env, e)
+	}
 	if setenv {
 		for _, e := range core.GeneralBuildEnvironment(config) {
 			env = addEnv(env, e)
 		}
 	}
 	return env
+}
+
+// adRunEnviron returns values that are appended to the environment for a command.
+var adRunEnviron = []string{
+	"PEX_NOCACHE=true",
 }
 
 // addEnv adds an env var to an existing set, with replacement.
--- a/tools/please_pex/pex_main.py
+++ b/tools/please_pex/pex_main.py
@@ -1,5 +1,6 @@
 """Zipfile entry point which supports auto-extracting itself based on zip-safety."""
 
+import fcntl
 from importlib import import_module
 import zipfile
 import os
@@ -54,6 +55,7 @@ sys.path = [PEX_PATH] + sys.path
 MODULE_DIR = '__MODULE_DIR__'
 ENTRY_POINT = '__ENTRY_POINT__'
 ZIP_SAFE = __ZIP_SAFE__
+PEX_STAMP = '__PEX_STAMP__'
 
 
 class SoImport(object):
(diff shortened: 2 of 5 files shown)


src/run/run_step.go
```diff
@@ -139,13 +139,22 @@ func run(ctx context.Context, state *core.BuildState, label core.BuildLabel, arg
...
+		env = addEnv(env, e)
+	}
 	if setenv {
 		for _, e := range core.GeneralBuildEnvironment(config) {
 			env = addEnv(env, e)
 		}
 	}
 	return env
 }
+// ad_run_environ returns values that are appended to the environment for a command.
```
peterebden: probably just make this a top-level variable instead of a function.

also naming nit, should be camelCase for Go.

tools/please_pex/pex_main.py
```diff
@@ -141,27 +143,65 @@ def get_code(self, fullname):
...
 
+def pex_basepath(temp=False):
+    if temp:
+        import tempfile
+        return tempfile.mkdtemp(dir=os.environ.get('TEMP_DIR'), prefix='pex_')
+    else:
+        return os.path.expanduser('~/.pex')
+
+def pex_uniquedir():
+    return f'pex-{PEX_STAMP}'
```
peterebden: don't wanna spoil your fun but can we avoid fstrings here please? would be nice to keep compatibility with older pythons unless we have a strong need.

tools/please_pex/pex_main.py
```diff
@@ -141,27 +143,65 @@ def get_code(self, fullname):
...
     This is primarily used for binary extensions which can't be imported directly from
     inside a zipfile.
     """
-    import contextlib, shutil, tempfile, zipfile
+    import contextlib
+
+    @contextlib.contextmanager
+    def pex_lockfile(basepath, uniquedir):
+        # Acquire the lockfile.
+        lockfile_path = os.path.join(basepath, f'.lock-{uniquedir}')
```
peterebden: same here

peterebden: do you need the lockfile to be different to the directory? wondering whether you could lock on the directory instead of the file and use its presence instead of writing to it?

Doesn't really matter a huge amount, just curious. I'm not 100% sure you can `flock()` a directory...

src/run/run_step.go
```diff
@@ -147,6 +150,13 @@ func environ(config *core.Configuration, setenv bool) []string {
 	return env
 }
 
+// adRunEnviron returns values that are appended to the environment for a command.
+func adRunEnviron() []string {
```
peterebden: probably make this just a top-level variable instead of a function.

also it should be camelCase for Go...

katzdm: Discussed offline in [gitter](https://gitter.im/please-build/Lobby?at=5d128b9a527a392800c1103e).

> I wanted to keep a separate state-bit for "This directory is ready for use", which lives outside the directory itself.

tools/please_pex/pex_main.py
katzdm: Discussed offline in [gitter](https://gitter.im/please-build/Lobby?at=5d128b9a527a392800c1103e).

> I wanted to keep a separate state-bit for "This directory is ready for use", which lives outside the directory itself.

agenticarus:
Could we please not introduce yet another dot directory in users' home dirs? This cache directory could well live under ~/.cache/pex or similar (for bonus points it could even respect $XDG_CACHE_HOME).

katzdm:
> Could we please not introduce yet another dot directory in users' home dirs? This cache directory could well live under ~/.cache/pex or similar (for bonus points it could even respect $XDG_CACHE_HOME).

Thanks for the suggestion! Just mailed #670 to move the cache to `~/.cache/pex`.

Issue #660, opened 2019-06-20
Cannot build dependencies from subrepo with --arch


katzdm:
Most easily reproduced with the following:

```
cd `mktemp -d`
plz init
cat <<EOF >BUILD
genrule(
    name = "make-a-file",
    cmd = "touch \$OUT",
    out = "outfile",
)

genrule(
    name = "move-xcompile-file",
    cmd = "mv \$SRCS \$OUT",
    srcs = ["@pleasings//:make-a-file"],
    out = "macosx-outfile",
)
EOF
plz build //:move-xcompile-file --arch=linux_amd64
```

Error message is:
```
Build stopped after 260ms. 1 target failed:
    ///pleasings_linux_amd64//:make-a-file
Subrepo darwin_amd64_linux_amd64 is not defined
```

I believe this is related to how `SubrepoArchName()` is written [here](https://github.com/thought-machine/please/blob/master/src/core/package.go#L119-L125), though I didn't see an "obvious" way to fix it.

PR #671, opened 2019-06-27
Add arch() subrepo-rule, for adding user-defined arch-subrepos.

katzdm:
Partially addresses #660. In the medium-term future, we might consider a more fundamental restructuring of subrepos/archs (perhaps before a hypothetical 15.0.0 release?)

This PR introduces a new subrepo-rule, `arch()`. It can be used as:

```
arch(
    name = "linux",
    os = "linux",
    arch = "amd64",
)
```

This allows for user-defined arch-subrepos, which remain distinct from plz's implicitly instantiated arch-subrepos. The motivating use for this is a target that packages a collection of binary distributions spanning multiple architectures.

In the process, we introduce a few small semantic changes - A `subrepo()` rule **_may_** instantiate an already-existing subrepo, **_if_** the declared subrepo is identical to the one already registered. This avoids a fatal assertion-failure in the event that `plz` parses the package that declares a subrepo, from within the same subrepo. We implement this subrepo-equality-check in the `MaybeAddSubrepo` function, and use this instead of `AddSubrepo` (with the belief that this should be nonbreaking, and that any new breakages ought to be discovered).

### DIFF
--- a/rules/subrepo_rules.build_defs
+++ b/rules/subrepo_rules.build_defs
@@ -173,3 +173,24 @@ def github_repo(name:str, repo:str, revision:str, build_file:str=None, hashes:st
         config = config,
         bazel_compat = bazel_compat,
     )
+
+def arch(name:str, os:str, arch:str):
+    """ Defines an architecture subrepo.
+
+    This is a convenience alias to the more general subrepo rule. It allows for a user-defined
+    subrepo for cross-built rules targeting a specific architecture. This lets users define
+    architectures that won't conflict with the subrepos implicitly instantiated by plz.
+
+    Args:
+      name: Name of the rule.
+      os:  Target operating system (e.g. "darwin")..
+      arch: Target architecture (e.g. "amd64").
+    """
+    arch_str = f'{os}_{arch}'
+    assert arch_str.count('_') == 1, "Character '_' disallowed in OS/Arch names"
+
+    subrepo(
+        name = name,
+        path = "",
+        arch = arch_str,
+    )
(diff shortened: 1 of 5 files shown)


rules/subrepo_rules.build_defs
```diff
@@ -173,3 +173,13 @@ def github_repo(name:str, repo:str, revision:str, build_file:str=None, hashes:st
         config = config,
         bazel_compat = bazel_compat,
     )
+
+def arch(name:str, os:str, arch:str):
```
peterebden: can you add a docstring please? that gets parsed to generate output for `plz help` & the docs etc.

src/parse/asp/builtins.go
```diff
@@ -758,17 +759,28 @@ func subrepo(s *scope, args []pyObject) pyObject {
 		state.Config.Bazel.Compatibility = true
 		state.Config.Parse.BuildFileName = append(state.Config.Parse.BuildFileName, "BUILD.bazel")
 	}
+
+	isCrossCompile := s.pkg.Subrepo != nil && s.pkg.Subrepo.IsCrossCompile
+	if args[5] != None { // arg 5 is arch-string, for arch-subrepos.
+		var arch cli.Arch
+		givenArch := string(args[5].(pyString))
+		if err := arch.UnmarshalFlag(givenArch); err != nil {
+			log.Fatalf("Could not interpret architecture '%s' for subrepo '%'", givenArch, name)
```
peterebden: wow didn't know Github did this, but it has spotted that the second format parameter is missing the s. Cute!

Issue #664, opened 2019-06-24
Return stable version from `/latest_version`


steakunderscore:
## Problem

Currently requesting `https://get.please.build/latest_version` returns `14.0.0-beta1`. It should really only be returning stable releases.

## Context

This url is being used for [automating updating](https://aur.archlinux.org/cgit/aur.git/tree/update-version.sh?h=please-bin) the version on [Arch AUR](https://aur.archlinux.org/packages/please-bin/). However this should be set to using a stable version only. Currently it can't be used for that as doesn't return stable releases only.

The other places `/latest_version` is being used is calling `plz update` and ruining `curl https://get.please.build | bash`. Both of these should also be using stable releases by default.

## Solution

- [ ] `/latest_version` should return stable versions only.
- [ ] A new `/latest_prerelease_version` should return stable, alpha and beta versions, whichever was most recently built.
- [ ] Update docs

peterebden:
For bonus points, add a `--prerelease` flag or similar to `plz update` which lets it opt into updating to the `latest_prerelease_version`.

agenticarus:
This seems to be working for latest_version at least.

PR #673, opened 2019-06-30
Set latest_version correctly

steakunderscore:
Set latest_version to stable versions and latest_prerelease_version to
last built version.

This is for issue #664

### DIFF
--- a/tools/misc/gen_release.py
+++ b/tools/misc/gen_release.py
@@ -154,8 +154,11 @@ class ReleaseGen:
 
     def upload_sftp(self, artifacts, signatures, checksums):
         """Uploads artifacts to get.please.build via sftp."""
-        with open('latest_version', 'w') as f:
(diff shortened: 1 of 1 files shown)


tools/misc/gen_release.py
```diff
@@ -154,8 +154,11 @@ def trigger_build(self, token, project):
 
     def upload_sftp(self, artifacts, signatures, checksums):
         """Uploads artifacts to get.please.build via sftp."""
-        with open('latest_version', 'w') as f:
+        with open('latest_prerelease_version', 'w') as f:
```
peterebden: You will need to update the instructions in the batch file below as well - i.e. line 171 needs to be within a similar check for `self.is_prerelease`, etc.

steakunderscore: Oops, fixed.

PR #675, opened 2019-07-08
Streamline `plz query changes`

peterebden:
Basically gets rid of the piping file list in and the command flags.

This is less flexible (the old one didn't really know about the SCM so you could use whatever you wanted) but since we have a concept of SCM in plz now we might as well use it. It's also a lot more intuitive without the piping of files in which was easy not to realise that you had to do to get sensible results. If someone wants to support say hg or svn it should be easy enough to write an implementation.


peterebden:
It's more correct than `query changed` though - without doing checkouts & comparing `changed` can't introspect some things.

It may be worth combining them into one with a flag or something along those lines though.

PR #676, opened 2019-07-08
Retain file attributes upon PEX extraction

noxxious:
Due to a longstanding bug in the Zipfile implementation in core Python (https://bugs.python.org/issue15795) the extraction does not retain file attributes, i.e. executable flags are removed.
This PR adds a fix for it.


peterebden:
Sorry, just realised I forgot to approve / merge this after we talked the other day!

PR #678, opened 2019-07-11
Python coverage is broken after update to new PEX cache location

noxxious:
Update the coverage locations for Python to follow new PEX cache location. #670

### DIFF
--- a/tools/please_pex/pex_main.py
+++ b/tools/please_pex/pex_main.py
@@ -154,6 +154,13 @@ def pex_uniquedir():
     return 'pex-%s' % PEX_STAMP
 
 
+def pex_paths():
+    no_cache = os.environ.get('PEX_NOCACHE')
+    no_cache = no_cache and no_cache.lower() == 'true'
+    basepath, uniquedir = pex_basepath(no_cache), pex_uniquedir()
+    pex_path = os.path.join(basepath, uniquedir)
+    return pex_path, basepath, uniquedir, no_cache
+
 def explode_zip():
     """Extracts the current pex to a temp directory where we can import everything from.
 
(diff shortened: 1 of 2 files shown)


peterebden:
Thanks!

drsjmiddleton:
I think this is causing an issue on containers that don't have mkdir privileges.

`
 File "/admin_app_server.pex/__main__.py", line 184, in _explode_zip
  File "/opt/tm/tools/python/3.6.2-12/usr/lib/python3.6/os.py", line 210, in makedirs
    makedirs(head, mode, exist_ok)
  File "/opt/tm/tools/python/3.6.2-12/usr/lib/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/.cache'
`

peterebden:
I think that was introduced as a side-effect of #668 (since it now caches under ~ instead of always into a tmp dir). I guess we could give it a fallback, but there's a reasonable argument that those containers should do something sensible with HOME as well.

drsjmiddleton:
Accepting the fact I'm bias because this breaks something I want to use - it still seems a bit harsh. I don't think a build system should be opinionated about how you set up your containers. Maybe catch the PermissionError and waive your right to caching?

peterebden:
Agreed - we certainly try not to be, and I'm speculating on possible solutions here. There is certainly a limiting case though - if you have a zip-unsafe .pex it must have *somewhere* writable that it can explode itself to.

The ideal solution is most likely to try `~/.cache` and fall back to somewhere under `$TMPDIR` if that fails (and then just fail hard if that doesn't work either).

noxxious:
Another option is to support the explode option on PEX to do explode the zip during container build.
@drsjmiddleton you could use the PEX_NOCACHE=true environment variable to always use tmpdir, but that still requires mkdir permissions on TMP.

drsjmiddleton:
I'm probably revealing my profound lack of knowledge regarding Linux permissions but I'm still slightly confused as to how home isn't available to the (presumably) root user attempting to explode the zip. 

But I guess structurally, since this code seems to increase performance rather than function, you should step down through increasingly unperformant options, where the last is the least performant but which will still function in the most hostile (unpermissive) build environment.

noxxious:
root user's home is '/' if not defined otherwise, but having a home would not help with no mkdir permissions.
The in question does not just increase performance in production environment. In some of the case the pex zip has to be exploded (when not zip safe data files,etc. are present) as noted by @peterebden.

My suggestion would be to add an option to explode during the container build, otherwise using non-pex style build, like Bazel, but that requires heavy coding.

peterebden:
Agreed, a change like that to python_binary would be a lot of work and (very) breaking.
I think the right thing here is a fallback to $TMPDIR - I'm on holiday with limited internet access for the next week, but might be able to get to it sometime, otherwise will try to take a look if someone else fixes it.
In the meantime you might be able to work around by setting $HOME within the container.

drsjmiddleton:
See https://github.com/thought-machine/please/pull/681

PR #684, opened 2019-07-18
Fix to fall back to TEMP is home does not have write access

drsjmiddleton:
Duplicate to rerun CI tests

### DIFF
--- a/tools/please_pex/pex_main.py
+++ b/tools/please_pex/pex_main.py
@@ -158,11 +158,12 @@ class ModuleDirImport(object):
 
 
 def pex_basepath(temp=False):
-    if temp:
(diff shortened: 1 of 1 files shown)


tools/please_pex/pex_main.py
```diff
@@ -158,12 +158,13 @@ def get_code(self, fullname):
 
 
 def pex_basepath(temp=False):
-    if temp:
+    if temp or not os.access('~/.cache/pex', os.W_OK):
```
peterebden: same comment as other PR, this will return `False` if `~/.cache/pex` doesn't exist yet; it may be better to try to extract to $HOME and fall back if that fails.

(It would also need an `expanduser` as well, it doesn't do that implicitly)

PR #685, opened 2019-07-22
Initial implementation of Google remote_execution API for storage

peterebden:
This only adds an incomplete cache implementation, not actual execution; will look at that more in the future (although this is about 80% of the work anyway, calling the execution bit is pretty simple once you've got all the blobs uploaded). Really intending this to start investigating whether it's a good option for us.

As predicted, it is a fairly complex protocol, but two 10 hour flights were just about enough to wrap my head around most of it. Major open concerns with it are:
 - Output directories need to be explicitly flagged versus output files; we do not require that at present and cannot reliably infer it. AFAICT Bazel "solves" this by not allowing `genrule` to output directories.
 - We can probably negotiate SHA1 as a digest for the storage API, but it looks like execution only allows one (https://github.com/bazelbuild/remote-apis/blob/e7282cf0f0e16e7ba84209be5417279e6815bee7/build/bazel/remote/execution/v2/remote_execution.proto#L1441) which seems to be server-set. We might need to be able to globally support SHA256 as a hash (which is mostly just undesirable because it's slower).
 - Will likely need some redoing of our internal cache API to match how it wants to work. I think that may not be a bad thing anyway though. I'm leaving that out of this PR for now since it's quite intrusive.

Obviously the benefit of doing it this way rather than coming up with our own protocol is interaction with existing servers (I'm particularly keen on buildbarn but there are other implementations too); that is likely a big enough win to be worth some of the (mostly temporary) pain.


src/remote/action.go
```diff
@@ -0,0 +1,178 @@
...
+			Properties: []*pb.Platform_Property{
+				{
+					Name:  "OSFamily",
+					Value: translateOS(target.Subrepo),
+				},
+				// We don't really keep information around about ISA. Can look at adding
+				// that later if it becomes relevant & interesting.
+			},
+		},
+		// We have to run everything through bash since our commands are arbitrary.
```
agenticarus: There's something in the docs about "toolchain" rules for things like bash, might be worth seeing how that stuff works?

src/cache/cache.go
```diff
@@ -13,32 +13,35 @@ import (
...
 	mplex := &cacheMultiplexer{}
-	if config.Cache.Dir != "" && !remoteOnly {
-		mplex.caches = append(mplex.caches, newDirCache(config))
+	if state.Config.Cache.Dir != "" && !remoteOnly {
+		mplex.caches = append(mplex.caches, newDirCache(state.Config))
 	}
-	if config.Cache.RPCURL != "" {
-		cache, err := newRPCCache(config)
+	if state.Config.Remote.URL != "" {
+		mplex.caches = append(mplex.caches, newRemoteCache(state))
```
agenticarus: Feels a little odd this takes the whole state when the other caches take the config - what does the remote cache setup need from the state at this point?

src/cache/rex_cache.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (rc *rexCache) Store(target *core.BuildTarget, key []byte, files ...string) {
+	log.Debug("Storing %s in remote cache...", target.Label)
+	if err := rc.client.Store(target, key, cacheArtifacts(target, files...)); err != nil {
+		log.Warning("Error storing artifacts in remote cache: %s", err)
+	}
+}
+
+func (c *rexCache) StoreExtra(target *core.BuildTarget, key []byte, file string) {
+	log.Debug("Storing %s: %s in remote cache...", target.Label, file)
```
agenticarus: ~Missing an implementation here?~ Ah, I see there's no way to retrieve these - maybe remove the log line?

src/cache/stub.go
```diff
@@ -7,9 +7,14 @@ package cache
...
 	"fmt"
+
 	"github.com/thought-machine/please/src/core"
 )
 
 func newRPCCache(config *core.Configuration) (*httpCache, error) {
 	return nil, fmt.Errorf("Config specifies RPC cache but it is not compiled")
 }
+
+func newRemoteCache(state *core.BuildState) core.Cache {
```
agenticarus: Why does this file exist - can we not just compile in the remote caches for every use?

src/please.go
```diff
@@ -453,7 +453,8 @@ var buildFunctions = map[string]func() bool{
 					config.Cache.RPCURL = ""
 					config.Cache.HTTPURL = ""
 				}
-				clean.Clean(config, newCache(config), !opts.Clean.NoBackground)
+				state := core.NewBuildState(1, nil, 4, config)
```
agenticarus: This feels a bit odd too - what are these magic numbers?

src/remote/blobs.go
```diff
@@ -0,0 +1,335 @@
...
+		Requests:     reqs,
+	})
+	if err != nil {
+		return err
+	}
+	// TODO(peterebden): this is not really great handling - we should really use Details
+	//                   instead of Message (since this ends up being user-facing) and
+	//                   shouldn't just take the first one. This will do for now though.
+	for _, r := range resp.Responses {
+		if r.Status.Code != 0 {
```
agenticarus: Is there an enumerated set of response codes? I'm guessing `0` means Success?

src/remote/remote.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+// init is passed to the sync.Once to do the actual initialisation.
+func (c *Client) init() {
+	c.err = func() error {
+		// TODO(peterebden): We may need to add the ability to have multiple URLs which we
+		//                   would then query for capabilities to discover which is which.
+		conn, err := grpc.Dial(c.state.Config.Remote.URL.String(),
+			grpc.WithTimeout(dialTimeout),
+			grpc.WithInsecure(),
```
agenticarus: Shouldn't this depend on the URL passed to it?

src/remote/remote.go
```diff
@@ -0,0 +1,326 @@
...
+		// Query the server for its capabilities. This tells us whether it is capable of
+		// execution, caching or both.
+		ctx, cancel := context.WithTimeout(context.Background(), dialTimeout)
+		defer cancel()
+		resp, err := pb.NewCapabilitiesClient(conn).GetCapabilities(ctx, &pb.GetCapabilitiesRequest{
+			InstanceName: c.instance,
+		})
+		if err != nil {
+			return err
+		} else if lessThan(&apiVersion, resp.LowApiVersion) || lessThan(resp.HighApiVersion, &apiVersion) {
```
agenticarus: Don't need the `else` - can split into two blocks

src/remote/remote.go
```diff
@@ -0,0 +1,326 @@
...
+// A Client is the interface to the remote API.
+//
+// It provides a higher-level interface over the specific RPCs available.
+type Client struct {
+	actionCacheClient pb.ActionCacheClient
+	storageClient     pb.ContentAddressableStorageClient
+	bsClient          bs.ByteStreamClient
+	initOnce          sync.Once
+	state             *core.BuildState
+	err               error // for initialisation
```
agenticarus: Do we want to check this in method calls and break early?

src/remote/remote_test.go
```diff
@@ -0,0 +1,271 @@
...
+
+	"github.com/thought-machine/please/src/core"
+)
+
+func TestInit(t *testing.T) {
+	c := newClient()
+	assert.NoError(t, c.CheckInitialised())
+}
+
+func TestBadAPIVersion(t *testing.T) {
```
agenticarus: It's not clear how setting the major versions to `1` here creates the expected output

src/remote/remote_test.go
```diff
@@ -0,0 +1,271 @@
...
+	cachedContents, err := ioutil.ReadFile("plz-out/gen/package/out1.txt")
+	assert.NoError(t, err)
+	assert.Equal(t, contents, cachedContents)
+}
+
+func newClient() *Client {
+	config := core.DefaultConfiguration()
+	config.Build.Path = []string{"/usr/local/bin", "/usr/bin", "/bin"}
+	// Can't use NewDefaultBuildState since we need to modify the config first.
+	state := core.NewBuildState(1, nil, 4, config)
```
agenticarus: Here are those magic numbers again

src/remote/remote_test.go
```diff
@@ -0,0 +1,271 @@
...
+	}, nil
+}
+
+func (s *testServer) Reset() {
+	s.DigestFunction = []pb.DigestFunction_Value{
+		pb.DigestFunction_SHA1,
+		pb.DigestFunction_SHA256,
+	}
+	s.LowApiVersion = semver.SemVer{Major: 2}
+	s.HighApiVersion = semver.SemVer{Major: 2, Minor: 1}
```
agenticarus: Ah, I see how this affected the first test now - feels like this might be neater in the construction of server, and done in a `setUp` method?

src/remote/utils.go
```diff
@@ -0,0 +1,88 @@
...
+	return digest
+}
+
+// digestMessageContents is like DigestMessage but returns the serialised contents as well.
+func digestMessageContents(msg proto.Message) (*pb.Digest, []byte) {
+	b, err := proto.Marshal(msg)
+	if err != nil {
+		// Not really sure if there is a valid possibility to bring us here (given that
+		// the messages in question have no required fields) so assume it won't happen :)
+		log.Fatalf("Failed to marshal message: %s", err)
```
agenticarus: I'm not a fan of `logger.Fatal` - os.Exit is a bit of a blunt instrument. Could we panic() instead?

src/remote/utils.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	sum := sha1.Sum(b)
+	return &pb.Digest{
+		Hash:      hex.EncodeToString(sum[:]),
+		SizeBytes: int64(len(b)),
+	}, b
+}
+
+// lessThan returns true if the given semver instance is less than another one.
+func lessThan(a, b *semver.SemVer) bool {
```
agenticarus: Somewhat surprised this doesn't exist in the SemVer package

src/remote/action.go
peterebden: I'm not sure there's much we can do to communicate that - it mentions it around the `Platform` proto but all we can do there are communicate key-value pairs.

Will probably ask this one (and one or two others) on the mailing list at some point.

src/cache/cache.go
peterebden: It needs the PathHasher since it needs to hash files individually (just having the overall target hash isn't granular enough for the API). That lives on the state so it doesn't get persisted between subsequent runs within the process (e.g. for `plz watch` or `plz query changed` etc).

src/cache/rex_cache.go
peterebden: Yeah I realised after this line that I'd need to do a refactor of the whole API for this (which I think will be a bit neater overall though). I'll take the message out for now.

src/cache/stub.go
peterebden: Just wanted to avoid the dependencies during bootstrap - it adds a bunch of extra Go modules that have to be downloaded & compiled but never actually used.

src/please.go
peterebden: Number of threads & verbosity (can't actually remember why that's in there tbh, it feels like it shouldn't be).

Agreed it is a mess - I want to clean up those constructors at some point, they aren't really super useful as they are.

src/remote/blobs.go
peterebden: I think it's the same as the gRPC ones. Inexplicably it is not actually an enum though - a comment in the file says it's a member of `google.rpc.Code` but it's actually just an `int32`. Not sure what's going on but I'll use one of those.

src/remote/remote.go
peterebden: no, grpc urls typically don't have a protocol so we can't infer it. I'll add a TODO, it really needs to be specified in the config (I expect we will want the ability to pass a CA at least for TLS).

src/remote/remote.go
peterebden: Ah damn I fixed that but must be on another branch. Yes, good spot :)

src/remote/remote_test.go
peterebden: I'll add a comment.

src/remote/remote_test.go
peterebden: yup, same as above; as the comment mentions, NewDefaultBuildState would be simpler but we have to pass in the config first. Again, needs a wide-ranging refactor in its own PR.

src/remote/remote_test.go
peterebden: `setUp` method? this isn't Python... :)

src/remote/utils.go
peterebden: I'm not really a fan of spitting out a big stacktrace at the user if they can't do anything with it. As the comment mentions though, I'm not sure what can even cause it to get here...

src/remote/utils.go
peterebden: It's a proto package so doesn't have any helper methods like that (and they didn't add any ala the `ptypes` stuff). I guess maybe they don't want to maintain them for every language, although agreed this was rather less pleasant than using `coreos/semver` :(

PR #687, opened 2019-07-23
Rewriting cache interface

peterebden:
This makes it a lot easier to work with the remote API; IMO it is a bit cleaner overall without the "Extra" functions. Will also be useful for some other things (e.g. have some plans around the HTTP cache that would depend on this too).

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -122,7 +122,7 @@ func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget) (err
 			postBuildOutput, err := loadPostBuildOutput(target)
 			if err != nil {
 				log.Warning("Missing post-build output for %s; will rebuild.", target.Label)
-			} else if err := runPostBuildFunctionIfNeeded(tid, state, target, postBuildOutput); err != nil {
+			} else if err := runPostBuildFunction(tid, state, target, postBuildOutput, ""); err != nil {
 				log.Warning("Error from post-build function for %s: %s; will rebuild", target.Label, err)
 			}
 		}
@@ -202,7 +202,7 @@ func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget) (err
 			if metadata := state.Cache.Retrieve(target, cacheKey); metadata != nil {
 				storePostBuildOutput(target, metadata.Stdout)
 				postBuildOutput = string(metadata.Stdout)
-				if err := runPostBuildFunctionIfNeeded(tid, state, target, postBuildOutput); err != nil {
+				if err := runPostBuildFunction(tid, state, target, postBuildOutput, ""); err != nil {
 					return err
 				} else if retrieveArtifacts() {
 					return writeRuleHash(state, target)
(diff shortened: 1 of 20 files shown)


src/build/build_step.go
```diff
@@ -191,9 +198,11 @@ func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget) (err
 		// Note that ordering here is quite sensitive since the post-build function can modify
 		// what we would retrieve from the cache.
 		if target.PostBuildFunction != nil && !haveRunPostBuildFunction {
-			log.Debug("Checking for post-build output file for %s in cache...", target.Label)
-			if state.Cache.RetrieveExtra(target, cacheKey, target.PostBuildOutputFileName()) {
-				if postBuildOutput, err = runPostBuildFunctionIfNeeded(tid, state, target, postBuildOutput); err != nil {
+			log.Debug("Checking for post-build output for %s in cache...", target.Label)
+			if metadata := state.Cache.Retrieve(target, cacheKey); metadata != nil {
```
agenticarus: Just wanted to check here - I think this creates a new scoped metadata and leaves the external one unchanged - did you mean to replace it?

src/build/build_step.go
```diff
@@ -191,9 +198,11 @@ func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget) (err
...
+				storePostBuildOutput(target, metadata.Stdout)
+				postBuildOutput = string(metadata.Stdout)
+				if err := runPostBuildFunctionIfNeeded(tid, state, target, postBuildOutput); err != nil {
```
agenticarus: Given we know that PostBuildFunction is not nil here (line 193) can we just inline the call here? (See also line 125 for identical concern)

src/cache/dir_cache.go
```diff
@@ -182,12 +180,18 @@ func (cache *dirCache) storeFile(target *core.BuildTarget, out, cacheDir string)
...
+	if needsPostBuildFile(target) {
+		outs = append(outs, target.PostBuildOutputFileName())
+	}
+	if !cache.retrieveFiles(target, key, "", outs) {
+		return nil
+	}
+	if needsPostBuildFile(target) {
+		return loadPostBuildFile(target)
+	}
+	return &core.BuildMetadata{}
```
agenticarus: This looks weird - why are there no contents?

src/cache/http_cache.go
```diff
@@ -69,21 +73,27 @@ func (cache *httpCache) StoreExtra(target *core.BuildTarget, key []byte, file st
...
-	retrieved := false
-	for _, out := range cacheArtifacts(target) {
-		if !cache.RetrieveExtra(target, key, out) {
-			return false
+	var metadata *core.BuildMetadata
+	for _, out := range target.Outputs() {
+		if !cache.retrieveOne(target, key, out) {
+			return nil
+		}
+		metadata = &core.BuildMetadata{}
```
agenticarus: Don't need to do this in the loop?

src/build/build_step.go
peterebden: Nope it's deliberate - I don't want to overwrite the external one with nil and if it gets out of this scope it won't matter anyway (because we won't be using the cached version).

src/build/build_step.go
peterebden: yeah I think you're right

src/cache/dir_cache.go
peterebden: there's no stdout because no post-build, and this cache implementation doesn't store any timestamps

src/cache/http_cache.go
peterebden: it's covered by the comment above (it assumes that a target with no outputs can't be retrieved).

FWIW a rewrite of http_cache is coming which will fix this silliness, but want to do it after this interface change (which will make it a lot easier)

PR #688, opened 2019-07-23
Cleanup NewBuildState()

peterebden:
Basically stop having all these opaque `1, nil, 4` things everywhere.
Also ripped out the `Verbosity` setting which was only used in a couple of places and frankly seems pretty unimportant.

### DIFF
--- a/src/core/config.go
+++ b/src/core/config.go
@@ -209,6 +209,7 @@ func DefaultConfiguration() *Configuration {
 	config.Please.Autoclean = true
 	config.Please.DownloadLocation = "https://get.please.build"
 	config.Please.NumOldVersions = 10
+	config.Please.NumThreads = runtime.NumCPU() + 2
 	config.Parse.BuiltinPleasings = true
 	config.Parse.GitFunctions = true
 	config.Build.Arch = cli.NewArch(runtime.GOOS, runtime.GOARCH)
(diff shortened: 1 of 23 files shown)


src/core/config.go
```diff
@@ -209,6 +209,7 @@ func DefaultConfiguration() *Configuration {
 	config.Please.Autoclean = true
 	config.Please.DownloadLocation = "https://get.please.build"
 	config.Please.NumOldVersions = 10
+	config.Please.NumThreads = runtime.NumCPU() + 2
```
agenticarus: Does this work appropriately in Docker containers? Would GOMAXPROCS be more appropriate?

src/follow/grpc_test.go
```diff
@@ -42,7 +42,9 @@ func TestClientToServerCommunication(t *testing.T) {
 	// Note that the ordering of things here is pretty important; we need to get the
 	// client connected & ready to receive events before we push them all into the
 	// server and shut it down again.
-	serverState := core.NewBuildState(5, nil, 4, core.DefaultConfiguration())
+	config := core.DefaultConfiguration()
+	config.Please.NumThreads = 5
```
agenticarus: Necessary? Could you just use DefaultBuildState here?

src/core/config.go
peterebden: Docs say runtime.NumCPU is "the number of logical CPUs usable by the current process" which implies it might be aware of cgroups etc; I think it is probably the best option (GOMAXPROCS is probably going to end up using it anyway)

src/follow/grpc_test.go
peterebden: nope, this test is specifically testing that this value gets propagated so it needs to be something that is not (usually) the default.

PR #689, opened 2019-07-24
HTTP cache rewrite

### LINKED ISSUE #556: Make HTTP cache compatible with Cirrus
Cirrus have a remote cache (https://cirrus-ci.org/examples/#remote-cache) which is compatible with a bunch of systems; make our HTTP cache match that.

It's possible that no change will be needed; I suspect we might need to get rid of the multipart stuff in favour of using tarballs though.

peterebden:
Resolves #556 

Basically rewrites the protocol to use tarballs instead of multipart (what was I thinking with that...) so hence it's a very simple PUT/GET which can be used with nginx webdav or similar.

Doesn't work with bazel-remote which is a lot stricter; we can finesse the hash length issue if needed but it also requires a Content-Length header which is awkward - we obviously don't know how long it's gonna be until we've built the gzipped tarball, and I don't want to buffer the whole thing in memory since it could be quite large.

Am turning it on for Cirrus so we'll see whether that works when this runs...


tools/images/http_cache/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:bionic
+MAINTAINER peter.ebden@gmail.com
+
```
agenticarus: worth adding a VOLUME directive for `/var/www`

peterebden: Dunno if necessary, am mostly using this as a throwaway rather than anything serious

peterebden:
Fair point on reclaiming space, there is a bit less info but the server could still take a stab based on age / access patterns.

Mostly expecting this to be used with hosted services (cirrus is a motivating example) which presumably manage to clean artifacts in some way.

tools/images/http_cache/Dockerfile
peterebden: OK on further thought have done - doesn't matter much but I suppose it is better (also fixed some permissions issues along the way)

src/cache/http_cache.go
```diff
@@ -18,178 +19,186 @@ import (
...
+		return nil, err
+	}
+	defer gzr.Close()
+	tr := tar.NewReader(gzr)
+	for {
+		hdr, err := tr.Next()
+		if err != nil {
+			if err == io.EOF {
+				return loadPostBuildFile(target), nil
+			}
```
tmthrgd: There’s a `return` missing here. If `tr.Next` returns a non-`io.EOF` error, the line below will panic.

peterebden: Good spot, thank you!

PR #698, opened 2019-08-01
Redo some internals of how tasks are handled

peterebden:
Basically expose them as channels rather than a function interface to get the next task.

This hides a few more details (like the taskType enum) and also allows the caller to be more flexible about how it routes events, which will be needed to have an asymmetric worker pool.


src/core/state.go
```diff
@@ -42,13 +42,19 @@ const (
...
+	Type     taskType
 }
 
 func (t pendingTask) Compare(that queue.Item) int {
 	return int((t.Type & priorityMask) - (that.(pendingTask).Type & priorityMask))
 }
 
+// A LabelPair is the type returned for parse tasks
+type LabelPair struct {
+	Label, Dependor BuildLabel
```
agenticarus: Dependor is not good english, use dependee or dependent depending on which you mean here (or even go as far as DependedOnBy or DependsOn)

peterebden: According to Wiktionary it's the first-person singular present passive indicative of dependo, so maybe it's good Latin instead...

PR #707, opened 2019-08-15
fix: when using testbootstrap the zip_safe pre-build used wrong cmd

noxxious:
This was an odd find. When the custom test bootstrap is used the under certain zip-unsafe conditions the cmd was reverted back to the old one (containing --include-test-deps) as the pre-build lambda retained the cmd without the bootstrap changes.


peterebden:
Nice thanks, I think that's a bit cleaner overall too.

PR #708, opened 2019-08-16
Fix url parsing for ips

henryaj:
I don't like this terribly, but the GRPC client expects IP addresses in the form 1.2.3.4:5555, whereas url.Parse() will fail if they're not in the form //1.2.3.4:5555.

@peterebden any better ideas?

### DIFF
--- a/src/core/config.go
+++ b/src/core/config.go
@@ -390,10 +390,10 @@ type Configuration struct {
 		Upload          cli.URL      `help:"URL to upload test results to (in XML format)"`
 	}
 	Remote struct {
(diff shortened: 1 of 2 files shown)


henryaj:
Arr, so looks like https://github.com/thought-machine/please/blob/master/src/core/config.go#L647-L648 is the reason URL parsing doesn't trip up on IPs when they're pass in as a command-line arg (the call to `url.Parse` is skipped).

henryaj:
We store URLs as strings under the hood, so maybe we just ditch the URL parsing altogether...? (per @agenticarus' suggestion)

peterebden:
I'm happy with that. Not sure we are getting much from it really, especially with gRPC URLs having no protocol so e.g. `plz-cache` is a valid URL.

PR #713, opened 2019-08-21
Break out Go test main templating to a separate rule

peterebden:
This lets users take advantage of the templating in more arbitrary ways (for example building a non-test binary from it) without exposing any internal details (e.g. having them deal with the please_go_test binary directly).


fsiddiqia:
Thanks for looking into this Pebers!

PR #717, opened 2019-08-28
Set exit codes to contain a bit more detail

peterebden:
As mentioned in the file:
0 -> success
1 -> general failure
2 -> a target failed to build
7 -> a test failed

The last one is staying (slightly awkwardly) as 7 for compatibility. We can't change it until plz 15...


sean-:
This doesn't cover testing, but if there are other exit codes within `plz` that you want to standardize around, FYI: https://godoc.org/github.com/sean-/sysexits

peterebden:
Thanks, will take a look at them!

PR #721, opened 2019-08-29
Improve error handling and logging when using a remote cache

henryaj:
- Print stacktrace in debug logs on panic
- Fail fast if remote execution is enabled but remote end doesn't support it
- Gracefully handle invalid responses from build server (e.g. when expected proto fields are left nil)
- Print helpful error if remote build exits with non-zero code

In time I'd like to improve all this such that remote build failures are non-fatal - please should just fall back to building locally I guess.

### DIFF
--- a/src/cache/dir_cache.go
+++ b/src/cache/dir_cache.go
@@ -1,4 +1,4 @@
-// Diretory-based cache.
+// Directory-based cache.
 
 package cache
 
--- a/src/remote/blobs.go
+++ b/src/remote/blobs.go
@@ -311,6 +311,9 @@ func (c *Client) downloadBlobs(f func(ch chan<- *blob) error) error {
 
 // retrieveByteStream receives a file back from the server as a byte stream.
 func (c *Client) retrieveByteStream(b *blob) error {
+	if b.Digest == nil {
+		return fmt.Errorf("can't retrieve byte stream from nil digest")
+	}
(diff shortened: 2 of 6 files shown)


src/cache/dir_cache.go
```diff
@@ -1,4 +1,4 @@
-// Diretory-based cache.
+// Directory-based cache.
```
peterebden: much shame:(

src/remote/blobs.go
```diff
@@ -311,6 +311,9 @@ func (c *Client) downloadBlobs(f func(ch chan<- *blob) error) error {
 
 // retrieveByteStream receives a file back from the server as a byte stream.
 func (c *Client) retrieveByteStream(b *blob) error {
+	if b.Digest == nil {
+		return fmt.Errorf("can't retrieve byte stream from nil digest")
```
peterebden: ooh dear does this happen? it probably shouldn't...

henryaj: Maybe - it's to do with reading [these digest fields](https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/remote_execution.proto#L870-L885) which I guess are optional, strictly speaking.

src/cache/dir_cache.go
henryaj: :smug:

PR #724, opened 2019-09-02
Optimise go_get a bit for downstream consumers

peterebden:
This is a little obtuse but basically makes things faster for `go_binary` et al which are currently receiving all the sources for all third-party dependencies when they actually only need the binary artifacts.

By using `srcs` rather than `deps` it skips the recursive search; we need an extra `require` for `go_get` itself which does need both.


peterebden:
N.B. This does not necessarily fix everything; it still pulls in sources for dependencies of dependencies which are a bit harder to sort out.

PR #725, opened 2019-09-03
pip somewhat imposes user site by default

noxxious:
This should fix the disttools exception "must supply either home or prefix/exec-prefix".


noxxious:
Ok, so this is a pure mess on the pip's side if the pip config has user=True, when this behaviour is apparent. I'm using Manjaro Linux, which recently started setting it as global option.
Links to the discussion (still no end in sight:):
https://github.com/pypa/pip/issues/1668
https://github.com/pypa/pip/issues/5702

The workaround:
https://wiki.archlinux.org/index.php/Python_package_guidelines#pip

PR #736, opened 2019-09-18
Allow home dir to be overridden when building remotely

henryaj:
Also conditionally exclude local directories from PATH when building remotely.


src/core/config.go
```diff
@@ -570,18 +571,43 @@ func (config *Configuration) getBuildEnv(includePath bool) []string {
 		if v, isSet := os.LookupEnv(k); isSet {
 			if k == "PATH" {
 				// plz's install location always needs to be on the path.
-				v = ExpandHomePath(config.Please.Location) + ":" + v
+				if config.Remote.URL != "" {
```
peterebden: I wonder if we might be able to get away without this stuff by passing the desired home directory into ExpandHomePath... will try something along those lines

PR #737, opened 2019-09-19
Don't use local home dir for remote

peterebden:
Thoughts?


peterebden:
Yep this won't fix the unfound blobs, unfortunately :(

Issue #734, opened 2019-09-17
Can't link cc_shared_object on macos


SanZhiYuan:
Hi, thanks for the building tool, here I got some issues with cc_shared_object

plz.build version: 14.1.3 & 14.1.2
OS: macos mojave 10.14
Clang: version 8.0.1 (tags/RELEASE_801/final)

everything is fine with cc_library, 
but I get this error when using cc_shared_object

**Build stopped after 5.07s. 1 target failed:
    //:main1
Error building target //:main1: exit status 1
Undefined symbols for architecture x86_64:
  "test1()", referenced from:
      _main in _main1#lib.a(main.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)**

>file libtest1.so
**libtest1.so: Mach-O 64-bit dynamically linked shared library x86_64**

after adding some linker_flags, the link stage passed, but run `main` failed

**dyld: Library not loaded: /Users/dev/plz/plz-out/tmp/libtest1._build/libtest1.so
  Referenced from: /Users/dev/plz/./plz-out/bin/main2
  Reason: image not found**

thanks for your help

a simple demo here to reproduce the case.

[plz.tar.gz](https://github.com/thought-machine/please/files/3619721/plz.tar.gz)

handicraftsman:
Same problem on GNU/Linux

peterebden:
I think this is a fairly normal consequence of how shared objects work; plz doesn't try to automatically set the linker search path or rpath. So to fix the compile error you need a linker_flag like `-L.` (which I guess you figured out), and to make it run you either need to set the rpath to `$ORIGIN` or set `$LD_LIBRARY_PATH` when running the binary, since the current directory is not usually searched by the dynamic linker.

I guess you have a more profound use case in mind than this example, but as this shows cc_shared_object requires a bit of marshalling around at runtime, so in general cc_library or cc_static_library are easier to manage.

SanZhiYuan:
> I think this is a fairly normal consequence of how shared objects work; plz doesn't try to automatically set the linker search path or rpath. So to fix the compile error you need a linker_flag like `-L.` (which I guess you figured out), and to make it run you either need to set the rpath to `$ORIGIN` or set `$LD_LIBRARY_PATH` when running the binary, since the current directory is not usually searched by the dynamic linker.
> 
[...]

PR #748, opened 2019-10-01
Dynamically add shared object deps

peterebden:
This addresses the request in #734 , that dependencies on `cc_shared_object` should get picked up automatically at build time.

Still some issues though:
 - need to properly exclude own labels from search (the `if not shared` clause isn't really right)
 - default output name for `cc_shared_object` should be `lib{name}.so` not `{name}.so`, right now you
   have to make sure it gets named correctly. Would like to fix but this is a breaking change.
 - need to set RPATH on the resulting binary, or set LD_LIBRARY_PATH when running. Don't think I want to change that implicitly but it's not hard to add to `ldflags`.


peterebden:
OK not sure why those tests are failing; they don't do that locally for me. Might come back to this one later.

PR #749, opened 2019-10-01
Introduce PEX_CACHE_DIR envvar to override default

jbarcsfx:
We have identified a scenario launching pex files on AWS Lambda is
failing when launched second and more times.
AWS Lambda reuses interpreter and current setup with zip-unsafe pexes
isn't handled properly, because subsequent invocations override sys.path
again, not to mention that modules have already been loaded and their
physial location removed.

Nonetheless, providing a way to specify cache dir solves this problem --
as we cannot possibly use read-only ~/ dir on lambdas, specifying the
dir ourselves (/tmp/pex in my case) solves the issue. And provides a
little bit more configurability


peterebden:
Looks good, thanks!

PR #762, opened 2019-10-21
Changing .pydistutils.cfg to setup.cfg per specs

jbarcsfx:
Oh boy, this caused me some time to find and debug.
We populate [install] prefix= in .pydistutils.cfg file local to
wheel build directory in order to circumvent brew's messed up python.

This, however, didn't work properly for me. Sometime between 14.0.0 and
14.1.7 --isolated has been introduced and it _might_ be the culprit:
see, according to specs we should populate setup.cfg, not
.pydistutils.cfg:

https://github.com/python/cpython/blob/master/Lib/distutils/dist.py#L343

setup.cfg is loaded from cwd, while .pydistutils.cfg is loaded from
$HOME iff --isolated is not specified (and in our case it is)

### DIFF
--- a/rules/python_rules.build_defs
+++ b/rules/python_rules.build_defs
@@ -368,7 +368,7 @@ def pip_library(name:str, version:str, hashes:list=None, package_name:str=None, 
         # https://github.com/Homebrew/brew/blob/master/docs/Homebrew-and-Python.md suggests fixing with --install-option
         # but that prevents downloading binary wheels. This is more fiddly but seems to work.
         # Unfortunately it does *not* work similarly on the Debian problem :(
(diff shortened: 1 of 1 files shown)


jbarcsfx:
@peterebden You seem to have changed setup.cfg to .pydistutil.cfg some time ago, I wonder what's the intended course of action given that --isolated prevents ~/.pydistutil.cfg from being loaded.
Also, the failed test doesn't seem like it's my doing

peterebden:
IIRC the pydistutil thing was added a fairly long time ago, before --isolated was added for OSX. It was all a series of hacks trying to persuade pip to behave in a sensible way - pretty sure it did fix something at the time but may well be incorrect now --isolated is introduced (also pip itself is a moving target...).

Anyway, this makes sense to my (limited) knowledge of it. Thanks for the investigation!

PR #765, opened 2019-10-24
python2 compatibility in please_pex

sfirmery:
Fix issue with python2 when `zip_safe` is False.


sfirmery:
Yes, i still have some python2 code. And upgrading to the last version of please break them.
I hope to migrated them to python3 soon.

PR #768, opened 2019-11-02
Add PassUnsafeEnv config option

### LINKED ISSUE #677: Add ability to exclude specific variables from cache-hashes
We use [CircleCI](https://circleci.com) as our CI tool, and I spent the last day or so debugging an issue with the cache (that is, it isn't using it). Finally I realized please includes environment variables in computing cache keys.

This is problematic, as to use Docker in a build in CircleCI (i.e. to use packer) you run the `setup_remote_docker` build step which allocates a docker executor for you and sets environment variables (`DOCKER_HOST`, `DOCKER_CERT_PATH`, etc). These are different every run (even when you re-run a build on the same commit), and so the always results in cache-misses.

I understand (and generally agree with) including environment variables in cache keys (as they very much can affect the build). However there *are* exceptions (like this) where these values do not meaningfully affect the output (another example would be proxy settings).

Would it be a reasonable request to allow whitelisting certain `PassEnv`-ed environment variables from the cache-key computation? I imagine this would be implimented like excluding them from the hash computation, or normalizing them to some magic value.

jacobwgillespie:
Fixes #677.

---

This adds a new config option, `PassUnsafeEnv`, which operates like `PassEnv`, passing the values of environment variables to build rules.  However unlike `PassEnv`, those values are not used in `ruleHash` when generating hashes, so their values can change without invalidating the hash.

Examples could include environment variables that configure the build system but do not affect the generated output, for instance HTTP proxy configuration, or the `DOCKER_HOST` example provided by @ThisGuyCodes in #677.

My personal use-case is we have a system for fetching artifacts that operates behind an authenticating proxy, so our users need to have a session token in their environment (fetched from a SAML IdP).  That session token environment variable does not affect the file downloaded from the artifact host and thus doesn't affect the output of the build, but does differ between our users and does change as users renew their session.


rulatir:
What is preventing this from being merged?

Tatskaari:
Hey, thank you for your patience. I admit we were dragging our feet a bit on this one because we weren't entirely comfortable with premise this sets. Saying that I can see your use case and I think it's legitimate. 

We've recently introduced a build metadata file that we can use to identify when these "unsafe" variables change and log a warning. I think this mitigates a lot of our concerns because if a build does go wrong, at least you'll get a warning to point you in the right direction.

jacobwgillespie:
Nice, yeah thank you for the merge, and also thank you for the careful consideration, I know this particular change is complicated. ❤️

PR #770, opened 2019-11-04
Make remote client timeout configurable

henryaj:
Was digging into why a please job is hanging downloading artifacts from the rex cache; this won't fix it but seemed like a useful addition regardless.

### DIFF
--- a/src/remote/blobs.go
+++ b/src/remote/blobs.go
@@ -59,7 +59,7 @@ func (c *Client) uploadBlobs(f func(ch chan<- *blob) error) error {
 			req.BlobDigests[i] = b.Digest
 			m[b.Digest.Hash] = b
 		}
-		ctx, cancel := context.WithTimeout(context.Background(), reqTimeout)
+		ctx, cancel := context.WithTimeout(context.Background(), c.reqTimeout)
 		defer cancel()
 		resp, err := c.storageClient.FindMissingBlobs(ctx, req)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


henryaj:
This will also enforce the timeout not just for an individual download operation, but for a whole build target - previously downloads could stall indefinitely because although the individual blob download op respected the 2 min timeout, the overall 'download artifacts for this build target' op did not.

src/remote/blobs.go
```diff
@@ -274,40 +274,52 @@ func (c *Client) byteStreamDownloadName(digest *pb.Digest) string {
...
-		modes[b.Digest.Hash] = b.Mode
-		if b.Digest.SizeBytes > c.maxBlobBatchSize || !c.canBatchBlobReads {
-			// This blob individually exceeds the size, have to use this
-			// ByteStream malarkey instead.
-			if err := c.retrieveByteStream(b); err != nil {
-				return err
+		digests := []*pb.Digest{}
+		filenames := map[string]string{}  // map of hash -> output filename
+		modes := map[string]os.FileMode{} // map of hash -> file mode
+		var totalSize int64
```
agenticarus: I don't see where this amount ever gets increased during normal blob processing

src/remote/blobs.go
```diff
@@ -274,40 +274,52 @@ func (c *Client) byteStreamDownloadName(digest *pb.Digest) string {
...
 // should close it when finished.
-func (c *Client) downloadBlobs(f func(ch chan<- *blob) error) error {
+func (c *Client) downloadBlobs(ctx context.Context, f func(ch chan<- *blob) error) error {
 	ch := make(chan *blob, 10)
+	done := make(chan struct{})
 	var g errgroup.Group
-	g.Go(func() error { return f(ch) })
+	go func() error {
+		defer close(done)
+		g.Go(func() error { return f(ch) })
```
agenticarus: Do we need to start this in this goroutine? might be a little clearer to do it outside?

henryaj: It's a bit fiddly, but I want to make sure this is killed if we hit the context timeout - hence the `select` on `<-done` and `<-ctx.Done()`. Open to suggestions of a better way to do this though.

src/remote/blobs.go
henryaj: Good point. Perhaps https://github.com/thought-machine/please/pull/770/files#diff-cf387a5ec5198b6baaf902dd7ef5860dR305 ought to be `totalSize += b.Digest.SizeBytes`?

PR #780, opened 2019-11-26
Patch out glog

peterebden:
As far as I can tell there is no sensible interface for controlling this library, and it is printing over the top of our own output. I'll raise something upstream asking to switch to a less awful library.


henryaj:
LGTM, though you need to buildify by the looks of things

PR #784, opened 2019-12-04
Ensure binary rules are made executable by rex backend

henryaj:
Fixes problems we were seeing with .pex files not being executable.

Is there a good reason we weren't doing this already?

### DIFF
--- a/src/remote/action.go
+++ b/src/remote/action.go
@@ -123,9 +123,6 @@ func (c *Client) buildTestCommand(target *core.BuildTarget) (*pb.Command, error)
 
 // getCommand returns the appropriate command to use for a target.
 func (c *Client) getCommand(target *core.BuildTarget) string {
-	if target.TestOnly {
-		return "true"
-	}
 	if target.IsRemoteFile {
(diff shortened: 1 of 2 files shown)


src/remote/action.go
```diff
@@ -135,7 +135,11 @@ func (c *Client) getCommand(target *core.BuildTarget) string {
 		}
 		return cmd
 	}
-	return target.GetCommand(c.state)
+	cmd := target.GetCommand(c.state)
+	if target.IsBinary {
+		return "(" + cmd + ") && chmod +x $OUT"
```
peterebden: might need to be $OUTS (don't think we necessarily guarantee binary targets have a single output, although it is a good idea for them to)

peterebden:
> Is there a good reason we weren't doing this already?

I made plz enforce it itself at some point early on (because it was annoying to keep forgetting it and easy for it to do). Unfortunate we can't do that here :(

henryaj:
This is causing problems with `sh_cmd` targets (which use a heredoc, so can't be appended to); will need to rethink a bit

henryaj:
Ready for review.

PR #787, opened 2019-12-06
Update java build def to support remote execution

henryaj:
When using remote execution, `$OUT` is a relative path (which the backend expects to be relative to `$TMP_DIR`). This changes the java build rule to get the file's abspath and move it to the expected location.

`|| true` is required for the rule to build locally as the file is already in the right place.

### DIFF
--- a/src/remote/action.go
+++ b/src/remote/action.go
@@ -54,9 +54,12 @@ func (c *Client) buildCommand(target *core.BuildTarget, inputRoot *pb.Directory,
 	}
 	// We can't predict what variables like this should be so we sneakily bung something on
 	// the front of the command. It'd be nicer if there were a better way though...
-	const commandPrefix = "export TMP_DIR=\"`pwd`\" && export OUT=\"$TMP_DIR/$OUT\" && "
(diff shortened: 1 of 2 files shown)


src/remote/action.go
```diff
@@ -54,9 +54,12 @@ func (c *Client) buildCommand(target *core.BuildTarget, inputRoot *pb.Directory,
 	}
 	// We can't predict what variables like this should be so we sneakily bung something on
 	// the front of the command. It'd be nicer if there were a better way though...
-	const commandPrefix = "export TMP_DIR=\"`pwd`\" && "
+	var commandPrefix = "export TMP_DIR=\"`pwd`\" && "
 	// TODO(peterebden): Remove this nonsense once API v2.1 is released.
 	files, dirs := outputs(target)
+	if len(target.Outputs()) == 1 { // $OUT is relative when running remotely; make it absolute
+		commandPrefix += "export OUT=\"$TMP_DIR/$OUT\" && "
```
peterebden: FYI you could use a raw string for this which removes the need for escaping the inner quotes.

henryaj: :excellent:

PR #788, opened 2019-12-06
Fix "not a directory" errors

peterebden:
We're double-downloading some things because of the weird nested Tree proto stuff. This fixes some of the "not a directory" errors.

Also limit the number of blobs in a single request arbitrarily since some servers seem unable to cope well, even if still under the size limit.

### DIFF
--- a/src/remote/remote.go
+++ b/src/remote/remote.go
@@ -359,14 +359,8 @@ func (c *Client) Retrieve(target *core.BuildTarget) (*core.BuildMetadata, error)
 			tree := &pb.Tree{}
 			if err := c.readByteStreamToProto(dir.TreeDigest, tree); err != nil {
 				return err
-			}
-			if err := c.downloadDirectory(dirPath, tree.Root); err != nil {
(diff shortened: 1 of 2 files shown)


src/remote/remote.go
```diff
@@ -359,15 +359,9 @@ func (c *Client) Retrieve(target *core.BuildTarget) (*core.BuildMetadata, error)
 			tree := &pb.Tree{}
 			if err := c.readByteStreamToProto(dir.TreeDigest, tree); err != nil {
 				return err
-			}
-			if err := c.downloadDirectory(dirPath, tree.Root); err != nil {
+			} else if err := c.downloadDirectory(dirPath, tree.Root); err != nil {
 				return err
 			}
```
henryaj: Not sure I understand how the `if...else` here is different from before.

peterebden: oh it's not, this is after some other changes were made and then unmade.

The actual fix is deleting lines 366-370.

PR #789, opened 2019-12-09
add human-readable scm identifier

VJftw:
Hello!

This PR adds an `SCM_DESCRIBE` environment variable that contains a "human-readable" identifier for the current git commit (https://git-scm.com/docs/git-describe) that is accessible from build targets that have `stamp = True` set.

We desire to use this alongside `SCM_REVISION` e.g.:
```
go_binary(
    ...
    definitions = {
        "github.com/my-user/my-project/pkg/version.BuildSCMVersion": "${SCM_REVISION}",
        "github.com/my-user/my-project/pkg/version.BuildHumanVersion": "${SCM_DESCRIBE}",
    },
    stamp = True,
    ...
)
```

This allows us to build binaries that can provide output in the format `<git describe --always> (SHA: <git rev-parse HEAD>)`:
```bash
$ my-binary version
Version: 1.2.3 (SHA: 6a6ce0a3930bc7e110d255308215991e4dd2543e)
# or a in-development version
$ my-binary version
Version: 1.2.3-1038-ge09a588f (SHA: ge09a588f930bc7110d255308215991e4dd2543e)
```

### DIFF
--- a/src/scm/scm.go
+++ b/src/scm/scm.go
@@ -14,6 +14,8 @@ var log = logging.MustGetLogger("scm")
 
 // An SCM represents an SCM implementation that we can ask for various things.
 type SCM interface {
+	// DescribeIdentifier returns the string that is a "human-readable" identifier of the given revision.
+	DescribeIdentifier(sha string) string
 	// CurrentRevIdentifier returns the string that specifies what the current revision is.
 	CurrentRevIdentifier() string
 	// ChangesIn returns a list of modified files in the given diffSpec.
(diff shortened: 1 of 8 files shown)


src/scm/scm.go
```diff
@@ -14,6 +14,8 @@ var log = logging.MustGetLogger("scm")
 
 // An SCM represents an SCM implementation that we can ask for various things.
 type SCM interface {
+	// DescribeIdentifier returns the string that is a "human-readable" identifier of the given revision.
+	DescribeIdentifier(sha string) string
```
peterebden: super nit: `revision string` (the passed value might not be a SHA hash).

test/stamp/stamp_test.sh
```diff
@@ -1,6 +1,13 @@
 #!/usr/bin/env bash
 
-if [ "`$DATA`" = "12345" ]; then
+git_revision=$(git rev-parse HEAD)
```
peterebden: this test should not execute git commands; this would fail if it were sandboxed, for example.

VJftw: aww, have updated now :)

PR #790, opened 2019-12-09
Fix query revdeps to work when reading from stdin

peterebden:


peterebden:
There is a general fix already for most commands (in `plz.go` it checks against BuildLabelStdin - just realised I could probably share some code there actually) but that doesn't work for this one because it parses the whole repo then processes a different set of labels.

PR #795, opened 2019-12-16
Display RPC bytes in/out

peterebden:
This is a generic mechanism for measuring how much data is flowing, it complements rather than replaces #794 

Have bumped the sdks dep with a small change that allows us to specify the gRPC DialOptions. Haven't sent it upstream yet since I still have a couple open and not sure if they want more from me :)

### DIFF
--- a/src/remote/stats.go
+++ b/src/remote/stats.go
@@ -1,0 +1,99 @@
+package remote
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"google.golang.org/grpc/stats"
+)
+
+// The length of time we keep stats around for.
+const windowDurationSeconds = 5
+const windowDuration = -windowDurationSeconds * time.Second
+
+// updateFrequency is the rate at which we update stats internally (which is independent of display updates)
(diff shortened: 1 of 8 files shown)


src/remote/stats.go
```diff
@@ -0,0 +1,98 @@
...
+		h.client.byteRateIn = h.updateStat(&h.in, &h.inmtx)
+		h.client.byteRateOut = h.updateStat(&h.out, &h.outmtx)
+	}
+}
+
+func (h *statsHandler) updateStat(stats *[]stat, mtx *sync.Mutex) int {
+	mtx.Lock()
+	defer mtx.Unlock()
+	s := *stats
+	idx := h.survivingStats(s, time.Now().Add(windowDuration))
```
henryaj: Took me a while to realise that `windowDuration` is negative... :stuck_out_tongue:

peterebden: Yeah it's a weird bit of the API; `Sub` does something quite different so you have to add a negative duration.

PR #804, opened 2019-12-31
fix(pex): use sys.version_info for PY_VERSION

sfirmery:
Fix an incorrect version condition that have been added in #765.


peterebden:
Thanks for the fix!

Issue #806, opened 2019-12-31
//test/secrets:test_tildes_replaced doesn't work on a system without ~/.bashrc


macripps:
This may happen if (for example) you only use zsh.

PR #807, opened 2019-12-31
Plumb context through from the main method

macripps:
Open to comments / revision here.

There are a few odd things here (such as passing contexts into constructors as they make remote calls) but this seems pretty simple. I've tested it with "plz plz build" but see #806 - the repo doesn't fully build on my machine any more.

This is a precursor to adding tracing (e.g. OpenTracing) to all operations for visibility/fun (see similar functionality in Pants - https://github.com/pantsbuild/pants/pull/7342 )

### DIFF
--- a/src/plz/plz.go
+++ b/src/plz/plz.go
@@ -1,6 +1,7 @@
 package plz
 
 import (
+	"context"
 	"strings"
 	"sync"
 
@@ -23,7 +24,7 @@ var log = logging.MustGetLogger("plz")
 // afterwards to find success / failure.
 // To get detailed results as it runs, use state.Results. You should call that *before*
 // starting this (otherwise a sufficiently fast build may bypass you completely).
-func Run(targets, preTargets []core.BuildLabel, state *core.BuildState, config *core.Configuration, arch cli.Arch) {
+func Run(ctx context.Context, targets, preTargets []core.BuildLabel, state *core.BuildState, config *core.Configuration, arch cli.Arch) {
 	parse.InitParser(state)
 	build.Init(state)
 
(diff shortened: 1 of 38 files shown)


agenticarus:
Not sure what's wrong with the build here - it compiles locally and the PR definitely contains the change

henryaj:
@agenticarus you need to update the method signature in `src/cache/stub.go` - that's what's causing the build to fail

peterebden:
This is a lot of plumbing (thanks for that!). I'm curious about how useful opentracing will be - maybe I'm slightly cynical after attempting Prometheus back in the day, but unsure how good a match it will be for what we do (as opposed to a request-based server which may be more what it's designed for?).

agenticarus:
That's a reasonable concern - it's more for fun than any concrete proposal at present. But I think it has the potential to show interesting behaviour around remote caching/execution etc.

src/plz/plz.go
```diff
@@ -45,13 +46,13 @@ func Run(targets, preTargets []core.BuildLabel, state *core.BuildState, config *
 	wg.Add(config.Please.NumThreads + config.Remote.NumExecutors)
 	for i := 0; i < config.Please.NumThreads; i++ {
 		go func(tid int) {
-			doTasks(tid, state, parses, builds, tests, arch, false)
+			doTasks(ctx, tid, state, parses, builds, tests, arch, false)
```
peterebden: I'm not 100% sure about this... should these be a derived context at this point (since this is where it shards out to multiple parallel processes?)

agenticarus: That sounds reasonable. What's a good way of doing that? None of the big four methods seem like what I want (which is just "give me a child context")

peterebden: hmmm good point - I'm not sure how opentracing would use the contexts to identify which thread things are on. Maybe you'd use a `ctx.WithValue` to attach something that identifies them? guess we can leave for now tho

agenticarus:
```//tools/build_langserver/lsp:lsp_test 21 tests run in 301ms; 21 passed
    TestDefinition                 PASS  20ms
    TestDefinitionStatement        PASS  20ms
    TestDefinitionBuiltin          PASS  20ms
    TestInitialize                 PASS  10ms
    TestInitializeNoURI            PASS  0s
    TestDidOpen                    PASS  10ms
    TestDidChange                  PASS  10ms
    TestDidSave                    PASS  10ms
    TestDidClose                   PASS  10ms
    TestFormatting                 PASS  10ms
    TestShutdown                   PASS  10ms
    TestCompletion                 PASS  20ms
    TestCompletionPackages         PASS  20ms
    TestCompletionInMemory         PASS  30ms
    TestCompletionPartial          PASS  10ms
    TestCompletionFunction         PASS  10ms
    TestCompletionPartialFunction  PASS  20ms
    TestDiagnostics                PASS  20ms
    TestSymbols                    PASS  10ms
    TestMoreSymbols                PASS  20ms
    TestTopLevelFunctionSymbols    PASS  10ms
1 test target and 21 tests run in 0s; 21 passed. Total time 4.77s.
```
These tests pass on my machine, not sure what's wrong with the CI server

macripps:
Going to abandon this, if it ever proves useful a more sensible approach might be to start from the bottom up and not the top down and do it one method at a time.

PR #817, opened 2020-01-16
Add returns for rules to allow reuse

noxxious:
Some of the rules were missing returns of the results, limiting their reuse and referencing.


agenticarus:
This looks like git-gone-wrong, please can you check the state of your branches?

noxxious:
Yes, will resubmit

PR #820, opened 2020-01-16
Add some extra logging around outputs

agenticarus:
We're seeing some weird behaviour where some files are coming out with 0 bytes, or just not being generated at all. This adds some more debug logging to help us track down where/why.

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -399,6 +399,7 @@ func moveOutputs(state *core.BuildState, target *core.BuildTarget) ([]string, bo
 		if err != nil {
 			return nil, true, err
 		}
+		// TODO(agenticarus): Remove this once we're happy we're never outputting empty files by accident.
 		tmpStat, tmpErr := os.Stat(tmpOutput)
 		realStat, realErr := os.Stat(realOutput)
 		if tmpErr == nil && realErr == nil {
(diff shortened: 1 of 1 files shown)


src/build/build_step.go
```diff
@@ -102,7 +102,7 @@ func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget, runR
 	var postBuildOutput string
 	if state.PrepareOnly && state.IsOriginalTarget(target.Label) {
 		if target.IsFilegroup {
-			return fmt.Errorf("Can't prepare temporary directory for %s; filegroups don't have temporary directories", target.Label)
+			return fmt.Errorf("can't prepare temporary directory for %s; filegroups don't have temporary directories", target.Label)
```
peterebden: ytho?

src/build/build_step.go
```diff
@@ -393,12 +393,17 @@ func moveOutputs(state *core.BuildState, target *core.BuildTarget) ([]string, bo
...
 		realOutput := path.Join(outDir, output)
 		if !core.PathExists(tmpOutput) {
-			return nil, true, fmt.Errorf("Rule %s failed to create output %s", target.Label, tmpOutput)
+			return nil, true, fmt.Errorf("rule %s failed to create output %s", target.Label, tmpOutput)
 		}
 		outputChanged, err := moveOutput(state, target, tmpOutput, realOutput)
 		if err != nil {
 			return nil, true, err
 		}
+		tmpStat, tmpErr := os.Stat(tmpOutput)
```
peterebden: can you add a TODO to remove this once the investigation is complete please?

src/build/build_step.go
agenticarus: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

src/build/build_step.go
agenticarus: Yeah, sure

PR #821, opened 2020-01-16
Enable running java_binaries with `--debug`

agenticarus:
I'm really very unhappy with the munging of the command, but the way we default to `java -jar` is pretty gross as well. Any suggestions appreciated.

### DIFF
--- a/src/run/run_step.go
+++ b/src/run/run_step.go
@@ -23,8 +23,8 @@ import (
 var log = logging.MustGetLogger("run")
 
 // Run implements the running part of 'plz run'.
-func Run(state *core.BuildState, label core.BuildLabel, args []string, env bool) {
-	run(context.Background(), state, label, args, false, false, env)
+func Run(state *core.BuildState, label core.BuildLabel, args []string, env bool, debug bool) {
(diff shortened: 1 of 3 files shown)


src/run/run_step.go
```diff
@@ -97,6 +97,10 @@ func run(ctx context.Context, state *core.BuildState, label core.BuildLabel, arg
 			log.Fatalf("Can't find binary %s", splitCmd[0])
 		}
 		splitCmd[0] = cmd
+		if debug {
+			// Obviously this only works if we really are a java -jar
```
peterebden: plz should not have hardcoded knowledge of Java flag incantations in it... can this not be included on the target in its debug config?

agenticarus: That would be ideal but I couldn't see how to specify that. We're creating this command using `core.ReplaceSequences(state, target, fmt.Sprintf("$(out_exe %s)", target.Label))` which sticks a `java -jar` on the front of [output], so this layer violation is already happening. The commands in `java_binary` seem to be only used for generating the jar (they seem to be jarcat commands)?

agenticarus:
This has some issues with workers that need to be thought through more closely

PR #823, opened 2020-01-16
use lower case path instead to differentiate from env variable

rabunkosar-dd:
"PATH" string causes to think that it's the shell path variable, so proposing the use of lowercase to differentiate


peterebden:
Thanks, makes sense!

PR #824, opened 2020-01-17
Do not do shell expansion for sh_cmd rule

noxxious:
sh_cmd rule looses any shell variable arguments specified as the cat is doing shell expansion.


peterebden:
I'm concerned that this is a breaking change for anyone who *is* expecting the current behaviour. With this change it wouldn't be possible to read any of the build-time env vars any more, right?

noxxious:
@peterebden, yes, you're right - so thanks for the finding. I've made a better solution to this, by documenting on how to avoid the expansion and updated the tests to verify.

peterebden:
Thanks, looks good!

PR #829, opened 2020-01-22
Simplify Jarcat input by writing to TMPDIR and then moving afterwards

macripps:
We simplify the script we generate in the following ways:

 * No need to replace characters that don't occur.
 * `source` the requested main script instead of copying it inline.
 * Write the preamble file to TMPDIR so it doesn't need to be excluded by jarcat
 * Prevent Jarcat from writing its output file until it's completed, meaning it also does not need to be excluded.
 * We use `os.Rename` where possible to do the move, but in please sandbox this will fail so we fall back to reading the zip file and writing it to the requested location.
 * Tests updated to use the temp filename
 * isPy37 no longer requires a zipfile object


macripps:
Not a fan of the duplication here now between main.go and pex.go, will look into this

PR #830, opened 2020-01-23
Use a tempfile for writing Zip file to avoid having to exclude it

macripps:
This will simplify all other rules that use Jarcat as the preamble file and the output file are written to TMPDIR first.

### DIFF
--- a/tools/jarcat/main.go
+++ b/tools/jarcat/main.go
@@ -16,6 +16,7 @@ import (
 	"gopkg.in/op/go-logging.v1"
 
 	"github.com/thought-machine/please/src/cli"
+	"github.com/thought-machine/please/src/fs"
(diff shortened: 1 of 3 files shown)


tools/jarcat/main.go
```diff
@@ -219,4 +223,23 @@ func main() {
 	for _, filename := range opts.Zip.In.Get() {
 		must(f.AddFiles(filename))
 	}
+
+	f.Close()
+
+	err = os.Rename(filename, opts.Zip.Out)
+	if err != nil {
+		// Fall back to copy
```
peterebden: can you use `fs.CopyFile` for this? save writing the same code again :)

macripps: Yup!

PR #836, opened 2020-01-27
Use the new Remote Asset API for remote_file

peterebden:
This puts more onus on the implementation (since now they have to have another service, or one that implements it in the same endpoint) but seems to be the future direction of these APIs.
It's also nicer than us trying to make up some bash / curl command for these.

### DIFF
--- a/src/remote/action.go
+++ b/src/remote/action.go
@@ -153,7 +153,7 @@ func (c *Client) buildTestCommand(target *core.BuildTarget) (*pb.Command, error)
 func (c *Client) getCommand(target *core.BuildTarget) string {
 	if target.IsRemoteFile {
 		// This isn't a real command, but it suits us to construct a pseudo-version of one.
-		cmd := "fetch " + strings.Join(target.AllURLs(), " ") + " & verify " + strings.Join(target.Hashes, " ")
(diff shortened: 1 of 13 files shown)


src/remote/action.go
```diff
@@ -152,14 +152,10 @@ func (c *Client) buildTestCommand(target *core.BuildTarget) (*pb.Command, error)
...
-		for i, s := range target.Sources {
-			urls[i] = "curl -fsSLo $OUT " + s.String()
-		}
-		cmd := strings.Join(urls, " || ")
+		// This isn't a real command, but it suits us to construct a pseudo-version of one.
+		cmd := "fetch " + strings.Join(target.AllURLs(c.state.Config), " ") + " & verify " + strings.Join(target.Hashes, " ")
 		if target.IsBinary {
-			return "(" + cmd + ") && chmod +x $OUT"
+			return cmd + " binary"
 		}
```
henryaj: How does this work?

peterebden: It doesn't really. It's not a real command but it's a lot easier to generate a BuildAction and bung it in the CAS so the rest of the code can be agnostic to how the target was executed. Basically I'm just making something up that contains the important bits of the rule (the URLs and hashes etc) so it hashes uniquely.

PR #842, opened 2020-01-29
Allow short date retrieval

macripps:
This minor change adds two more options to `git_show` in order to retrieve the author and committer date formatted according to the `--date` parameter. Currently this is hard-coded to be `short` but as nothing else uses it, I felt we didn't need to add the parameter to `git_show` itself.


peterebden:
Can we do this through `stamp` instead somehow? Am a bit dubious that these functions will cause rebuilds more often than necessary (because whenever the git commit changes they will alter the target, even if it is otherwise unaffected)

PR #844, opened 2020-01-31
Retrieve Commit Date for STAMP builds

macripps:


src/scm/git.go
```diff
@@ -161,3 +163,16 @@ func (g *git) Checkout(revision string) error {
...
 }
+
+func (g *git) CurrentRevDate(format string) string {
+	out, err := exec.Command("git", "show", "-s", "--format=%ct").CombinedOutput()
+	if err != nil {
+		return "UNKNOWN"
+	}
+	timestamp, err := strconv.ParseInt(string(out), 10, 64)
+	if err != nil {
+		return string(out)
```
peterebden: I think I see the logic, but might be worth adding a comment about why it's OK to "ignore" the error here.

src/scm/stub.go
```diff
@@ -35,3 +35,7 @@ func (s *stub) ChangedLines() (map[string][]int, error) {
 func (s *stub) Checkout(revision string) error {
 	return fmt.Errorf("Unknown SCM, can't checkout")
 }
+
+func (s *stub) CurrentRevDate(format string) string {
+	return "<unknown>"
```
peterebden: Don't strongly mind which but let's pick one of either `<unknown>` or `UNKNOWN` (as in the git impl when it can't be determined)

src/scm/git.go
macripps: I've actually made this return the error

peterebden:
You might want to rebase this on master - I've fixed the build failure (we apologise for the inconvenience)

PR #852, opened 2020-02-12
add coverage by directory to plz cover json, fixes #848

### LINKED ISSUE #848: Provide line coverage by directory
Currently the coverage result file provides percentages by file:

```
 "coverage_by_file": {
            "my/lib/one/file1.go": 100,
            "my/lib/two/file1.go": 80,
```

Might be impractical to provide coverage for all dirs, but could pass in a directory? e.g. `plz cover //my/lib/... --coverage_by_directory //my/lib/one`

thecapdan:
fixes #848

### DIFF
--- a/src/test/coverage.go
+++ b/src/test/coverage.go
@@ -142,6 +142,7 @@ func WriteCoverageToFileOrDie(coverage core.TestCoverage, filename string, incre
 
 	out.Files = convertCoverage(coverage.Files, allowedFiles)
 	out.Stats = getStats(coverage)
+	out.Directories = getDirectoryCoverage(coverage)
 	out.Stats.Incremental = incrementalStats
 	if b, err := json.MarshalIndent(out, "", "    "); err != nil {
 		log.Fatalf("Failed to encode json: %s", err)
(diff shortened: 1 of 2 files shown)


thecapdan:
Unit tested, but I still need to manually test on a large codebase. Also, should discuss whether it makes sense to amalgamate coverage for all nested directories

src/test/coverage.go
```diff
@@ -204,9 +235,10 @@ func convertCoverage(in map[string][]core.LineCoverage, allowedFiles []string) m
 
 // Used to prepare core.TestCoverage objects for JSON marshalling.
 type jsonCoverage struct {
-	Tests map[string]map[string]string `json:"tests"`
-	Files map[string]string            `json:"files"`
-	Stats stats                        `json:"stats"`
+	Tests       map[string]map[string]string `json:"tests"`
+	Files       map[string]string            `json:"files"`
+	Directories map[string]float32           `json:"directories"`
```
thecapdan: probably makes more sense to move this into the stats object

src/test/coverage.go
```diff
@@ -192,6 +193,36 @@ func getStats(coverage core.TestCoverage) stats {
...
+
+func (lines dirLines) getPercentage() float32 {
+	return 100.0 * float32(lines.covered) / float32(lines.totalCoverable)
+}
+
+func getDirectoryCoverage(coverage core.TestCoverage) map[string]float32 {
+	dirCoverage := make(map[string]float32)
+	linesByDir := make(map[string]*dirLines)
+
+	for _, file := range coverage.OrderedFiles() {
```
peterebden: why iterate over OrderedFiles? could you not iterate over just `Files` instead? Can't see how ordering would make a difference...

thecapdan: :+1:

peterebden:
I'm not too worried about codebase size; doing in-memory operations over all files / build targets etc tends to be quick (there are lots of similar passes in the various `query` operations for example).

PR #858, opened 2020-02-20
Use UploadIfMissing from SDK

peterebden:
Continuing the theme of getting rid of our code in favour of the SDK...

### DIFF
--- a/src/remote/blobs.go
+++ b/src/remote/blobs.go
@@ -2,61 +2,23 @@ package remote
 
 import (
 	"context"
-	"encoding/hex"
-	"os"
 
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/chunker"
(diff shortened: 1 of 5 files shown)


src/remote/blobs.go
```diff
@@ -1,241 +1,30 @@
...
-		blobs = append(blobs, b)
-		if len(blobs) == buffer {
-			filter(blobs)
-			blobs = blobs[:0]
-		}
-	}
-	if len(blobs) != 0 {
-		filter(blobs)
+	g.Go(func() error { return f(ch) })
+	chomks := []*chunker.Chunker{}
```
macripps: Perhaps better english would be "chunks"?

peterebden: The TM memes have infected me to the point where this makes perfect sense ;-)

PR #869, opened 2020-03-10
Allow the user to bring their own test runner

peterebden:
Basically the `test_runner_bootstrap` thing still defines whatever dependencies are needed, and you can set `test_runner` to a module path to have it import and use that.

### DIFF
--- a/tools/please_pex/custom.py
+++ b/tools/please_pex/custom.py
@@ -1,0 +1,8 @@
+import importlib
+
+
+def run_tests(args):
+    """Runs tests using a custom test runner that is selected by the user."""
+    runner, _, name = "__TEST_RUNNER__".rpartition('.')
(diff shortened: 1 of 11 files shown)


tools/please_pex/custom.py
```diff
@@ -0,0 +1,8 @@
+import importlib
+
+
+def run_tests(args):
+    """Runs tests using a custom test runner that is selected by the user."""
+    runner, _, name = "__TEST_RUNNER__".rpartition('.')
+    mod = importlib.import_module(runner)
+    return getattr(mod, name)(TEST_NAMES, args)
```
macripps: Can we check the result of `getattr` is a function? It's likely to allow us to return a nicer error message than triggering a traceback with "TypeError: 'foo' object is not callable"

PR #870, opened 2020-03-11
fix coverage for dependencies and python compatibility for py-coverage >= 4.x

noxxious:
Due to wrongly passed in deps flag the nested dependency files were filtered out of the coverage report.
In addition made the Python PEX runner compatible with coverage version >4.x as the xml_file method signature has changed and they use data file by default, which overrides the COVERAGE_FILE iwth coverage's own sqlite database.


peterebden:
Thanks!

PR #879, opened 2020-03-24
Redo internal graph model

peterebden:
Fixes a lurking issue since #864 where plz could rarely exit unsuccessfully due to not building the target it expected to.

This solves it with some fairly extensive internal re-plumbing of how we track targets & reverse dependencies; it is now a blocking model rather than an event-driven one, i.e. there is a goroutine for each target that blocks until its dependencies are built, and no explicit tracking of reverse dependencies (apart from some lazily calculated stuff for `plz query`). This seems to have no particular performance issues and may even be slightly faster (the go runtime should be happy enough with arbitrary numbers of goroutines as long as most of them are blocked).

A little unsure if it is simpler overall once all the fringe cases are taken into account. The main parts that feel awkward are having to do cycle detection during the build (otherwise it has a nasty habit of hanging) and some of the error handling stuff (which I might be able to boot out to a generic utility package - if we had a version of errgroup that eagerly reported errors it'd be a lot simpler I think). OTOH the previous setup was clearly problematic in its own way.

Currently have disabled `parse_step_test`, am unsure to what degree that can be resurrected in its current form. Will revisit later (it seems about time to put this through CI / review etc).


peterebden:
Needs more work, will reopen when ready.

PR #884, opened 2020-03-27
Extend remote execution test to run more tests

peterebden:
It is currently running more quickly than I expected - ~1 min per run, which leaves it a lot of time (it's basically free as long as it's faster than the darwin build which is 4-5 mins). So add more tests!


peterebden:
Not sure what's going wrong here. Might come back later.

PR #885, opened 2020-03-28
fix: use unified class and file names for coverage xml report

noxxious:
The XML coverage reports, had filenames trimmed to remove the target's package name, resulting in non-uniform names, braking code coverage reporting tools, such as https://codecov.io. This fix addresses that.


peterebden:
Thanks, makes sense!

Issue #877, opened 2020-03-18
java_library resources argument does not actually include resources


hnridder:
The documentation at http://please.build documents the `resources` parameter to `java_library` as _Resources to include in the .jar file_. However the resources do not actually get included in the jar, but they are put in a separate jar, which is added as a dependency. This is problematic when the purpose is to produce a jar that can be exported (and not an intermediate step to a java_binary, which will eventually pick up the resources).
I suggest to make the rule match the documentation and include the contents of the created resources jar in the output jar.

PR #890, opened 2020-04-02
Put resources in the same jar as java_library sources

macripps:
This addresses the requested issue #877. We make a slight change to the way sources for workers are specified (to allow us to not send these resources to it), and now copy the resources directly into the output tree (in _tmp).

Added an end-to-end test that passes, would value some greater check on (in particular) the change to resources_root calculation - I'm not convinced the previous documentation was correct but now seems to do the right thing.

```
Archive:  plz-out/gen/test/java_rules/issue_877/issue_877.jar
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  01-01-2001 00:00   META-INF/
       39  01-01-2001 00:00   META-INF/please_sourcemap
      615  01-01-2001 00:00   MyClass.class
       81  01-01-2001 00:00   some.txt
```

Previously this would have created a whole new Jar file just to contain the `some.txt` file.

### DIFF
--- a/rules/java_rules.build_defs
+++ b/rules/java_rules.build_defs
@@ -1,7 +1,7 @@
 """Built-in rules to compile Java code."""
 
 
-def java_library(name:str, srcs:list=None, src_dir:str=None, resources:list=None, resources_root:str=None,
+def java_library(name:str, srcs:list=None, src_dir:str=None, resources:list=[], resources_root:str=None,
                  deps:list=[], modular:bool=False, exported_deps:list=None, visibility:list=None,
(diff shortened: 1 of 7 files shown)


macripps:
> 15:26:19.630   ERROR: Error from remote worker: exec: java: not found

Is this something I've broken? Doesn't seem likely...

peterebden:
Yeah, can you bung a `java` label on your new e2e test? The FreeBSD workers don't have Java installed, which it works out, but it doesn't realise that test is a Java test.

peterebden:
I'm a little nervous about the magic, but thanks for the fix!

PR #895, opened 2020-04-04
Remote stamp lookups

peterebden:
Implements special lookup logic for targets with `stamp = True`.

Background: stamped targets get access to the current SCM revision and a couple of other things. That doesn't force a rebuild when it changes though, only if we otherwise would because the actual inputs have changed (which keeps these targets from building every commit). However this wasn't happening for remote execution; this adds some special-casing for that logic.

_Technically_ this means there is an incrementality issue where if you had a target that successfully built remotely without being stamped, then you turned it on, it would not rebuild. I think that's a weird enough situation that we don't need to care though (and arguably it sort of mirrors the "doesn't force rebuild" logic of it anyway).

### DIFF
--- a/src/remote/remote.go
+++ b/src/remote/remote.go
@@ -267,7 +267,7 @@ func (c *Client) build(tid int, target *core.BuildTarget) (*core.BuildMetadata, 
 	// If we're gonna stamp the target, first check the unstamped equivalent that we store results under.
 	// This implements the rules of stamp whereby we don't force rebuilds every time e.g. the SCM revision changes.
 	if target.Stamp {
-		if command, digest, err := c.buildAction(target, false, target.Stamp); err == nil {
+		if command, digest, err := c.buildAction(target, false, false); err == nil {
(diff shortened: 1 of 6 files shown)


src/remote/remote.go
```diff
@@ -265,10 +261,45 @@ func (c *Client) Build(tid int, target *core.BuildTarget) (*core.BuildMetadata,
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	metadata, ar, err := c.execute(tid, target, command, digest, target.BuildTimeout, false, needStdout)
+	if err != nil {
+		return metadata, ar, digest, err
+	}
+	if target.Stamp {
+		// Store results under unstamped digest too.
+		_, unstampedDigest, _ := c.buildAction(target, false, false)
```
Tatskaari, line 286: This might fail in some really funky ways

peterebden: Have redone a bit so as not to ignore errors. Also not recalculating the digests this time for mad perf gainz

peterebden:
I think the FreeBSD CI failure must be unrelated. Going to land this and address that separately since it seems nontrivial.

Issue #860, opened 2020-02-20
java_test cannot access resources in jar


iamneal:
I have a java_test target, and a java_binary target that both include a `dep` that has resources  `//path/to/resource_rule:resources` as resources.  Removing company information, it looks like this:

```
java_library(
    name = "lib",
    resources = ["//path/to/resource_rule:resources"],
    srcs = glob(["src/main/**/*.java"]),
)
java_binary(
    name = "example",
    main_class = "com.company.package.Example",
    self_executable = True,
    deps = [":lib"],
)
java_test(
    name = "test_example",
    srcs = glob(["src/test/**/*.java"]),
    resources = ["//path/to/resource_rule:resources"],
    data = ["//path/to/resource_rule:resources"],

    deps = [":lib"],
    test_package = "com.company.package",
)
```

The java_libarary `lib` rule has a class included that will read from this resource file like this:

```java
class ExampleClass {
        // matches the resource path included in the build target
        String filename = "path/to/resource_rule/Resource.tokens";
        File file = null;
        try {
            file = new File(getClass().getClassLoader()
                .getResource(filename).getFile());
        } catch (NullPointerException e) {
            System.out.println("could not find resource: " + e.getMessage());
        }
}

```
This file exists in both the executable jar and the test jar. Running `jar -tf example.jar | grep tokens`
and  `jar -tf test_example.jar | grep tokens` will output `path/to/resource_rule/Resource.tokens` for both jars.

The java_binary runs just fine, and loads the resource correctly.

The java_test gets a file not found exception.

This seems like a bug, if it is not a bug, how do I get the java_test to load my resource from the jar?

```
$ plz --version
Please version 14.3.4
```

Thanks for any help

macripps:
I've looked into this a bit closer and I think something funky is going on with the test classpath creation (although I think ideally you don't want to be including the resource twice).

A `java_test` wraps up the sources and resources in a library (`_test_example#lib.jar`). However, these files are then _also_ copied into the `test_example.jar` file. The only thing I can think of is the ClassLoader shenanigans in https://github.com/thought-machine/please-java/blob/master/junit_runner/src/build/please/main/TestMain.java#L119 are causing it to lose access to the jar file contents.

Notably, calling into ExampleClass from the test _does_ work (the resource is found), so it's only if you try to read the resource directly in the Test that it fails. It feels a bit strange to be including the resource at two separate libraries anyway, so perhaps this could be a workaround?

[...]

PR #896, opened 2020-04-04
Fix non worker javac invocation

peterebden:
If it has resources as well then the changes in #890 break the javac invocation.

Added a simple test since I realised we don't have anything covering this scenario in the repo any more (the java tools do exercise it but when they got moved out we lost it).


macripps:
Thanks! #860 might be worth adding a test case for as well? (Add the resource to the `java_test` instead of the `java_library`?)

peterebden:
Thanks, I've added a similar looking test case but it seems just to pass, which makes me think I've missed some of the subtlety of the original...

PR #897, opened 2020-04-06
Try updating FreeBSD image to 11-3

peterebden:
12 seems to have other errors, although we should try it out sometime...


peterebden:
Okay gonna bank what we have on this one

PR #898, opened 2020-04-07
Allow symlinks in jarcat tar archives

sfirmery:
Avoid error `CRITICAL: Error writing tarball: archive/tar: write too long` when sources of `tarball()` contains symlinks.


peterebden:
Thanks, makes sense! The test failures seem related though - link.txt changing into "symlink file", but not on Linux? Gotta love OS-specific issues like this :)

peterebden:
Looks like not a problem with this PR but something about how the test directory gets built by plz. Strange that it's different across OSs although I do remember having some issues with how symlinks were handled before...

peterebden:
OK looks like that's got it. Will merge this - thanks for the change!

sfirmery:
Thanks for the fix!

PR #906, opened 2020-04-12
Simplify how we do help for config entries

peterebden:
Basically get rid of all the pre-calculating and embedding and just do runtime reflection on it.

Means we follow some more Go rules due to less embedding.


peterebden:
Actually there should be nothing stopping this merging to master now...

PR #907, opened 2020-04-12
Simplify how we do help for config entries

peterebden:
Basically get rid of all the pre-calculating and embedding and just do runtime reflection on it.

Means we follow some more Go rules due to less embedding.


peterebden:
Have to admit I'm at a loss about why this would fail only on FreeBSD, or why that test is even significantly affected. I'm going to reopen this against the 1.5

PR #917, opened 2020-04-15
Specify python test runner at the build rule level.

bLiGM:
Expands https://github.com/thought-machine/please/pull/869 by letting users specify test runners on a per build rule level.


peterebden:
What's the motivation for this? Concerned that there are a bunch of things that aren't configurable at the build rule level (interpreter, module dir etc), is there really a need for mixing multiple tests in the same directory with different test runners this way?

bLiGM:
The configuration of the Python interpreter at the build rule level inspired this change. But specifically the issue I'm facing is:

I have a package which I specify the test runner for a Python test `Foo`. There is another rule in the same package called `Bar` that is spawned from someone else's build definition. That build definition creates python test for `Bar`. However it has no knowledge of my test runner (specifically of the dependencies).

```
- Foo
- Bar
    - BarTest
```

A workaround would be to move the `Bar` into a different package.  But since the two items are related I'd prefer keeping them in the same package.

peterebden:
OK makes sense; just be aware that not everything can or should be specified this way. The interpreter is actually a good example of something I would have done differently in hindsight (similarly to other 'tool' type things that typically aren't configurable at the rule level).

PR #918, opened 2020-04-18
Re-introduce fallback to copy in RecursiveLink et al

peterebden:
Got lost in #898.


peterebden:
Ignoring failure, it looks unrelated (and that test is to die soon)

PR #923, opened 2020-04-21
Istanbul fixes

tiagovtristao:
**What does this PR do?**

* It adds new TypeScript related extensions to the default cover configuration
* Fixes istanbul related code
  * Uncovered lines can't be marked as such after being first marked as covered. The current logic prevents this as the `Covered` const is higher. This solution allows the coverage of lines to be updated until a statement finally marks them as uncovered. For instance, istanbul will mark an `if` statement as covered if reached, but statements within it as uncovered if not
  * Corrects an istanbul test as a result of the fix above. See snapshots below
  * Sanitises file names properly if coverage is run within a sandbox

The first two statements and respective counts in the `istanbul_coverage_2.json` file below show how lines 16-22 are actually uncovered.

Screenshot 2020-04-21 at 19 58 43

Screenshot 2020-04-21 at 19 59 40

### DIFF
--- a/src/core/config.go
+++ b/src/core/config.go
@@ -96,8 +96,8 @@ func ReadConfigFiles(filenames []string, profiles []string) (*Configuration, err
 	}
 	setBuildPath(&config.Build.Path, config.Build.PassEnv)
 	setDefault(&config.Build.PassEnv)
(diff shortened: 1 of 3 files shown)


peterebden:
Thanks, looks good!

PR #927, opened 2020-04-27
Moved please to use github.com/bmatcuk/doublestar for globbing

### LINKED ISSUE #874: glob exclude patterns inconsistency
Initialize a new repo and create (touch) the following files:
```
src/a.not
src/a.txt
src/b.txt
src/c.txt
src/d/a.txt
src/d/b.txt
src/d/c.txt
src/d/e/a.txt
```
Create a BUILD file with the following contents:
```
filegroup(
   name = 'one',
   srcs = glob(['src/**'], exclude=['src/d/*']),
)
```
Create a directory java, copy the entire src tree there and the BUILD file as well (actually I created a symlink).
## Difference in :one and //java:one
```
$ plz query print :one //java:one
# //:one:
filegroup(
    name = 'one',
    srcs = [
        'src/a.not',
        'src/a.txt',
        'src/b.txt',
        'src/c.txt',
    ],
    timeout = 600,
)

# //java:one:
filegroup(
    name = 'one',
    srcs = [
        'src/a.not',
        'src/a.txt',
        'src/b.txt',
        'src/c.txt',
        'src/d/e/a.txt',
    ],
    timeout = 600,
)
```
I expect the same set of filenames in both cases.

## Multiple exclude patterns
Change the exclude list to `['*.not', 'src/d/*']`. Now the output is:
```
# //:one:
filegroup(
    name = 'one',
    srcs = [
        'src/a.txt',
        'src/b.txt',
        'src/c.txt',
        'src/d/a.txt',
        'src/d/b.txt',
        'src/d/c.txt',
        'src/d/e/a.txt',
    ],
    timeout = 600,
)

# //java:one:
filegroup(
    name = 'one',
    srcs = [
        'src/a.txt',
        'src/b.txt',
        'src/c.txt',
        'src/d/a.txt',
        'src/d/b.txt',
        'src/d/c.txt',
        'src/d/e/a.txt',
    ],
    timeout = 600,
)
```
It looks as if the second exclude pattern is ignored. However, if we change the order of the exclude list to `['src/d/*', '*.not']` then the output becomes
```
# //:one:
filegroup(
    name = 'one',
    srcs = [
        'src/a.txt',
        'src/b.txt',
        'src/c.txt',
    ],
    timeout = 600,
)

# //java:one:
filegroup(
    name = 'one',
    srcs = [
        'src/a.txt',
        'src/b.txt',
        'src/c.txt',
        'src/d/e/a.txt',
    ],
    timeout = 600,
)
```

## How to exclude an entire directory tree
As background information, I'm trying to exclude multiple directory trees and it would be nice if please would support this, which I don't think it does at the moment. However, due to the inconsistencies mentioned above, I'm not entirely certain what please does do and what it doesn't do.

I would like to take all files rooted at T, _except_ those that are contained in a subdirectory (or deeper) of T/I/A. Files in T/I/A themselves should be included. So something like `glob(['T/**'], exclude=['T/I/A/*/**'])`.

It would be nice if please could support this, e.g. by allowing \*\* in exclude patterns, or by treating an exclude pattern ending in / as a subtree match.

Tatskaari:
Fixes #874

### DIFF
--- a/src/fs/fs.go
+++ b/src/fs/fs.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
-	"sync"
 	"syscall"
 
 	"gopkg.in/op/go-logging.v1"
@@ -121,26 +120,8 @@ func IsDirectory(path string) bool {
 	return err == nil && info.IsDir()
 }
 
-// Memoize this to cut down on filesystem operations
-var isPackageMemo = map[string]bool{}
-var isPackageMutex sync.RWMutex
-
 // IsPackage returns true if the given directory name is a package (i.e. contains a build file)
 func IsPackage(buildFileNames []string, name string) bool {
-	isPackageMutex.RLock()
-	ret, present := isPackageMemo[name]
-	isPackageMutex.RUnlock()
-	if present {
-		return ret
-	}
-	ret = isPackageInternal(buildFileNames, name)
-	isPackageMutex.Lock()
-	isPackageMemo[name] = ret
-	isPackageMutex.Unlock()
-	return ret
-}
-
-func isPackageInternal(buildFileNames []string, name string) bool {
 	for _, buildFileName := range buildFileNames {
 		if FileExists(path.Join(name, buildFileName)) {
 			return true
--- a/src/fs/fs.go
+++ b/src/fs/fs.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"sync"
 	"syscall"
 
 	"gopkg.in/op/go-logging.v1"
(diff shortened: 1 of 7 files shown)


src/fs/fs.go
```diff
@@ -119,3 +120,31 @@ func IsDirectory(path string) bool {
...
 	return err == nil && info.IsDir()
 }
+
+// Memoize this to cut down on filesystem operations
+var isPackageMemo = map[string]bool{}
+var isPackageMutex sync.RWMutex
+
+// IsPackage returns true if the given directory name is a package (i.e. contains a build file)
+func IsPackage(buildFileNames []string, name string) bool {
+	isPackageMutex.RLock()
```
peterebden, line 130: Noted that we had this before, thanks for preserving.... but on further thought do we think it's worth it? Do we really re-glob over enough package directories to make it worth it?

PR #947, opened 2020-05-12
Improve Windows compatibility of pexes

peterebden:
Basically uses portalocker instead of fcntl so pexes don't blow up at import time.

Obviously this does not particularly help with pexes having appropriately-architectured binary bits, but for a pure-Python one it should work rather better.

### DIFF
--- a/tools/please_pex/BUILD
+++ b/tools/please_pex/BUILD
@@ -49,7 +49,6 @@ genrule(
         "//third_party/python:packaging",
         "//third_party/python:importlib_metadata",
         "//third_party/python:zipp",
-        "//third_party/python:portalocker",
     ],
     outs = ["please_pex"],
     binary = True,
--- a/tools/please_pex/BUILD
+++ b/tools/please_pex/BUILD
@@ -41,6 +41,7 @@ genrule(
         "//third_party/python:behave",
         "//third_party/python:parse",
         "//third_party/python:parse_type",
+        "//third_party/python:portalocker",
         "//third_party/python:traceback2",
         "//third_party/python:enum34",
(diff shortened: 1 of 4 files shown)


tools/please_pex/BUILD
```diff
@@ -41,13 +41,15 @@ genrule(
...
         "//third_party/python:parse_type",
+        "//third_party/python:portalocker",
         "//third_party/python:traceback2",
         "//third_party/python:enum34",
         "//third_party/python:win_unicode_console",
         "//third_party/python:colorama",
         "//third_party/python:packaging",
         "//third_party/python:importlib_metadata",
         "//third_party/python:zipp",
+        "//third_party/python:portalocker",
```
RichardoC, line 52: This was included twice

tools/please_pex/pex_main.py
```diff
@@ -1,12 +1,12 @@
 """Zipfile entry point which supports auto-extracting itself based on zip-safety."""
 
-import fcntl
 from importlib import import_module
 from zipfile import ZipFile, ZipInfo, is_zipfile
 import os
 import runpy
 import sys
 
+
```
RichardoC, line 9: Is this formatting change desired?

tools/please_pex/BUILD
peterebden: Ah thanks, good spot

tools/please_pex/pex_main.py
peterebden: Yeah PEP8 says two blank lines after imports. Not that it deeply matters much :)

PR #950, opened 2020-05-13
Simplify the pip_library build rule

Tatskaari:


Tatskaari:
We want the srcs added to plz-out/gen so we're going to do this a different way

PR #953, opened 2020-05-18
Moved post build output file to be the whole metadata file 

Tatskaari:
Also cleaned up some of the code around buildTarget

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -124,8 +124,6 @@ func prepareOnly(tid int, state *core.BuildState, target *core.BuildTarget) erro
 // 3) Actually build the rule
 // 4) Store result in the cache
 func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget, runRemotely bool) (err error) {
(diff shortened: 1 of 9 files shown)


src/build/build_step.go
```diff
@@ -301,12 +295,19 @@ func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget, runR
...
+
 	checkLicences(state, target)
 
 	if runRemotely {
 		target.SetState(core.BuiltRemotely)
 		state.LogBuildResult(tid, target.Label, core.TargetBuilt, "Built remotely")
 		return nil
+	} else if target.BuildCouldModifyTarget() {
+		if err := storeTargetMetadata(target, metadata); err != nil {
+			// TODO(jpoole): the plz-out dir doesn't seem to exist at this point when it's run remotely so storing the md
```
peterebden, line 307: As discussed, this is correct... remote runs don't create any of the intermediate outputs. Need to make sure rex keeps working.

PR #954, opened 2020-05-18
Fix TestFileGroupBinDir when umask is not defualt

zfrank:
TestFileGroupBinDir failed if umask did not have the usual value 0022.
The test is verifying that when please creates a new directory, it won't
modify its permissions.
So instead of comparing against a specific set
of permissions (which depend on the environment) it's simpler to create
a new dir on the spot and check that its permissions are the same for
the directories that please created.


peterebden:
Thanks!

PR #962, opened 2020-05-25
Cleanup concurrency control for parsing

peterebden:
This keeps the logic simple and centralised in the parser rather than having it pop up in core etc as well.

Some quick testing suggests it is about the same performance as before.


src/parse/asp/builtins.go
```diff
@@ -264,7 +264,11 @@ func subincludeTarget(s *scope, l core.BuildLabel) *core.BuildTarget {
 		s.NAssert(s.contextPkg.Target(l.Name) == nil, "Target :%s is not defined in this package; it has to be defined before the subinclude() call", l.Name)
 	}
 	s.NAssert(l.IsAllTargets() || l.IsAllSubpackages(), "Can't pass :all or /... to subinclude()")
+	// Temporarily release the parallelism limiter; this is important to keep us from deadlocking
+	// all available parser threads (easy to happen if they're all waiting on a single target which now can't start)
+	<-s.interpreter.limiter
 	t := s.state.WaitForBuiltTarget(l, pkgLabel)
+	s.interpreter.limiter <- struct{}{}
```
Tatskaari, line 271: What does this do? Seems like a poor mans mutex?

Tatskaari: Seems that's exactly what it is... 

http://www.golangpatterns.info/concurrency/semaphores

Could be nice to use some type aliasing and extension methods to make this read a bit better.

peterebden: Fair point  - done.

I could also one-line some of these by doing `defer limiter.Acquire().Release()` but personally I find that confusing enough that I'd rather not...

PR #966, opened 2020-05-28
Make tests run in parallel when the --num_runs is provided

Tatskaari:


src/core/build_target.go
```diff
@@ -140,6 +141,8 @@ type BuildTarget struct {
 	Progress float32 `print:"false"`
 	// The results of this test target, if it is one.
 	Results TestSuite `print:"false"`
+	// A mutex to control access to Results
+	ResultsMux sync.Mutex
```
peterebden, line 145: You never seem to actually call this? 

Would also suggest maybe making it private and having some method to control access to the results.

Tatskaari: It's called in test_step.go before doing anything to the Results field. I could move all the methods for updating the results to this file and make the mux private.

goddenrich:
thanks for this @Tatskaari!

PR #969, opened 2020-05-29
Add conditional logic for parsing go 1.14 output

bLiGM:
Go 1.14 slightly changes the format of the output of `go test`, such that test output occurs before the test result line, instead of afterwards. This PR adds some additional logic to the test package to take these differences into account.

### DIFF
--- a/src/test/go_results.go
+++ b/src/test/go_results.go
@@ -49,22 +49,44 @@ func parseGoTestResults(data []byte) (core.TestSuite, error) {
 					Stderr:   strings.Join(testOutput, ""),
 				})
 			} else if bytes.Equal(testResultMatches[1], []byte("SKIP")) {
-				i++ // Following line has the reason for being skipped
+				// The skip message is found at the bottom of the test output segment.
+				// Prior to Go 1.14 the test output segment follows the results line.
+				// In Go 1.14 the test output segment sits between the start line and the results line.
+
+				var outputLines []string
+				if resultLooksPriorGo114(i, lines) {
+					outputLines = getPostResultOutput(i, lines)
+				} else {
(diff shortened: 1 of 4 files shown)


src/test/go_results.go
```diff
@@ -49,19 +49,41 @@ func parseGoTestResults(data []byte) (core.TestSuite, error) {
 					Stderr:   strings.Join(testOutput, ""),
 				})
 			} else if bytes.Equal(testResultMatches[1], []byte("SKIP")) {
-				i++ // Following line has the reason for being skipped
+				// The skip message is found at the bottom of the test output segment.
+				// Prior to Go 1.14 the test output segment follows the results line.
+				// In Go 1.14 the test output segment sits between the start line and the results line.
+
+				var outputLines []string
```
peterebden, line 56: Can you pull these five lines out to a function? It's repeated below (and in general I prefer a single-assignment rather than a declaration before an if this way, although that may be my old C++ days talking)

src/test/go_results.go
```diff
@@ -85,3 +107,47 @@ func parseGoTestResults(data []byte) (core.TestSuite, error) {
...
+// Only fully applicable for failing and skipped tests
+// (as a message is guaranteed to occur before the results block).
+func resultLooksPriorGo114(currentIndex int, lines [][]byte) bool {
+	if currentIndex == 0 {
+		return false
+	}
+
+	prevLine := lines[currentIndex-1]
+	prevLineMatchesStart := testStart.FindSubmatch(prevLine)
+	if prevLineMatchesStart != nil {
```
peterebden, line 121: just `return prevLineMatchesStart != nil` here is sufficient

PR #973, opened 2020-05-30
Fix negative repeat

### LINKED ISSUE #967: Error: strings: negative Repeat count  for small terminal sizes
In our CI pipeline we have step `./pleasew update` which fails when I change the version in `.plzconfig` from 14.0.0 to something higher (I've tried 14.4.4). The error I get is:
```
14:46:29.364 CRITICAL: Failed to download Please: strings: negative Repeat count
 ```

After looking into your source code, I found that this could be caused by https://github.com/thought-machine/please/blob/master/src/cli/progress.go#L84 Indeed, when I try to reproduce the problem locally, as soon as my terminal has width less than 40 I get the same error.

I tried to go around this problem by using `./pleasew --plain_output update` but the flag does not seem to change the behaviour at all.

peterebden:
Don't try to do an interactive progress bar if it's not wide enough.

The normal interactive display also behaves badly at small sizes but that seems less easy to fix (and it at least doesn't crash).

Fixes #967

### DIFF
--- a/src/cli/progress.go
+++ b/src/cli/progress.go
@@ -19,6 +19,10 @@ type progressReader struct {
 	interactive               bool
 }
 
+// This is the minimum column width we support. Below this it becomes hard to print sensible
+// content (and later we can run into issues with repeats, see #967)
+const minCols = 40
+
 // NewProgressReader returns a new progress bar reader.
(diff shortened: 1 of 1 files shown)


j0z3f:
Thanks!

PR #974, opened 2020-06-01
Support --num_runs meaningfully for remote execution

peterebden:
Basically don't allow it to be cached (otherwise you immediately get the cached results back n times). Also honour test args.


src/remote/action.go
```diff
@@ -161,6 +161,9 @@ func (c *Client) buildTestCommand(target *core.BuildTarget) (*pb.Command, error)
 	}
 	const commandPrefix = "export TMP_DIR=\"`pwd`\" TEST_DIR=\"`pwd`\" && "
 	cmd, err := core.ReplaceTestSequences(c.state, target, target.GetTestCommand(c.state))
+	if len(c.state.TestArgs) != 0 {
```
Tatskaari, line 164: not sure the significance of this

peterebden: Arguments passed on the command line (i.e. `plz test //my:test blah blah`). The `blah blah` wasn't getting honoured before.

goddenrich:
🥇

PR #982, opened 2020-06-04
Swap to go-junit-report for parsing go test output

bLiGM:
Swap to using a 3rd party library for parsing Go test output.

This should cover more edge cases than our previous simple parser.

An additional note - by reading various PRs in `go-junit-report` around the different edge cases in Go test output. I've found that go test has a json output which should be used for parsing tasks like we're doing.  However there's discussions on the `go-junit-report` repo which suggest that there aren't much gains to be had just yet by using the go test json output.

Also the json option only exists for Go 1.10+. I think the best thing to do going forwards is to revisit this if we officially stop support Go versions < 1.10, to see if we hit any more edge cases that would be solved by using a json format parser.


src/test/results_test.go
```diff
@@ -169,11 +169,6 @@ func TestGoIgnoreUnknownOutput(t *testing.T) {
 	assert.Equal(t, 0, results.Skips())
 }
 
-func TestGoFailIfUnknownTestPasses(t *testing.T) {
```
bLiGM, line 172: I'm not sure how often we expect this particular case to come up. Go Junit report doesn't error out in this scenario - but it does 'misattribute' a failure to a Test case, so I'm not sure it's 100% correct in that regard.

peterebden: That dates back to literally before the dawn of time (initial commit). I don't think I've ever seen it actually happen though; not terribly bothered about losing this one case if everything else is still passing.

peterebden:
> Also the json option only exists for Go 1.10+. I think the best thing to do going forwards is to revisit this if we officially stop support Go versions < 1.10, to see if we hit any more edge cases that would be solved by using a json format parser.

I doubt anyone deeply cares about the older versions; we support I think 1.8+ (https://golang.org/pkg/testing/#MainStart is probably the most critical thing there) but I don't imagine lots of people deeply care about building 1.8 or 1.9 with the latest version of plz.

PR #984, opened 2020-06-05
When running repeated tests, apply a low priority for caching results.

peterebden:
Servers _might_ interpret this in a useful way.


peterebden:
I think I might leave this for now; I'm not sure I'm doing it the right way around (the proto is somewhat unintuitive about how these fields are defined) and I think I'm going to implicitly solve it another way.

PR #986, opened 2020-06-06
Remove buildid replacer

peterebden:
This doesn't actually work any more in at least go 1.14 - it appears that the check on `arg == "-buildid"` no longer works since it's now passed as `-buildid=abcd`. I suspect this is the case in 1.13 too. On the bright side it may not be necessary any more in those versions since their logic seems to have improved a bit.

The `-trimpath` flag seems better overall but forces a complete rebuild of the whole Go world which is pretty unfortunate (and seems to break some tests too). I'll log a ticket to investigate that later.


macripps:
Pleased to see the golang build tool improvements have obsoleted this!

peterebden:
Looks like -trimpath has some pretty negative consequences too (forcing rebuilds of lots of things). I'll investigate that later but the rest of this can stand (which means it can go on master now instead of waiting for the next major, since there is no Go version requirement).

PR #990, opened 2020-06-09
Fix go test parsing

peterebden:
Basically the lib is overwriting most of the useful test output which makes it quite hard to see what's gone on in your test :)
The fix appears to be fairly straightforward, fortunately; I've done it on a fork for now.


Tatskaari:
looks like your gomod package name is wrong in the fork.

peterebden:
Not the biggest problem - the "fix" wasn't :(

PR #993, opened 2020-06-09
Update protobuf to something more modern

peterebden:
Hopefully this makes freebsd happier


Tatskaari:
Might be worth updating protoc as well?

peterebden:
Good point, will do (although I think that goes away in v15...)

PR #996, opened 2020-06-10
Fix plz test when more than one test argument is provided

nkostoulas:
I've fixed the `LoadPreviousFailures` method to use the correct build label and thus resolve the issue with `plz test --failed` not working.

There was also a problem with `plz test` not working when a target was provided along with more than one test argument. (It works when a single argument is provided)

### DIFF
--- a/tools/please_go_test/gotest/write_test_main.go
+++ b/tools/please_go_test/gotest/write_test_main.go
@@ -135,6 +135,7 @@ package main
 
 import (
 	"os"
+	"strings"
 	"testing"
     "testing/internal/testdeps"
 
@@ -208,7 +209,8 @@ func main() {
 {{end}}
     testVar := os.Getenv("TESTS")
     if testVar != "" {
-        args = append(args, "-test.run", testVar)
+		testVar = strings.Replace(testVar, " ", "|", -1)
+		args = append(args, "-test.run", testVar)
     }
     os.Args = append(args, os.Args[1:]...)
 	m := testing.MainStart(testDeps, tests, nil, examples)
--- a/src/test/results.go
+++ b/src/test/results.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strings"
 
 	"github.com/thought-machine/please/src/core"
 	"github.com/thought-machine/please/src/fs"
@@ -83,7 +84,8 @@ func LoadPreviousFailures(filename string) ([]core.BuildLabel, []string) {
 	args := []string{}
 	for _, suite := range junit.TestSuites {
 		if suite.Failures > 0 {
-			labels = append(labels, core.ParseBuildLabel(suite.Name, "")) // These always have complete labels
+			labels = append(labels, core.NewBuildLabel(
+				strings.Replace(suite.Package, ".", "/", -1), suite.Name))
 			for _, c := range suite.TestCases {
 				if c.Failure != nil || c.Error != nil {
 					args = append(args, c.Name)


src/please.go
```diff
@@ -826,9 +826,9 @@ func runPlease(state *core.BuildState, targets []core.BuildLabel) {
 // One can also pass a 'failed' flag which runs the failed tests from last time.
 func testTargets(target core.BuildLabel, args []string, failed bool, resultsFile cli.Filepath) []core.BuildLabel {
 	if failed {
-		targets, args := test.LoadPreviousFailures(string(resultsFile))
+		targets, _ := test.LoadPreviousFailures(string(resultsFile))
```
peterebden, line 829: Is this not losing the "just run specific failing test cases" behaviour - it will re-run the whole target, no?

src/test/test_step.go
```diff
@@ -153,6 +153,9 @@ func test(tid int, state *core.BuildState, label core.BuildLabel, target *core.B
 		}
 	}
 
+	// Fresh set of results for this target.
+	target.StartTestSuite()
```
peterebden, line 157: I think this was picked up by Jon in #997 (thanks anyway though!)

src/please.go
nkostoulas: So if you have multiple targets it will only re-run the targets that had failures. The problem was that the args returned contain both Testcases and sub-Testcases (as the results file records both) and this was causing it to fail with some ReturnValue error. I'll investigate this further.

nkostoulas: I've reverted this.

nkostoulas:
`TESTS` env was set to the space-separated string of Test Args and given that this mirrors `go test -run` (I understand) it was causing problems whenever there was more than 1 Test Arg. I've replaced this to be set to pipe-separated to mirror an "a or b" regex.

src/test/test_step.go
```diff
@@ -310,6 +313,7 @@ func testCommandAndEnv(state *core.BuildState, target *core.BuildTarget, run int
 	if len(state.TestArgs) > 0 {
 		args := strings.Join(state.TestArgs, " ")
 		replacedCmd += " " + args
+		args = strings.Replace(args, " ", "|", -1)
```
peterebden, line 316: I don't think it's valid to just do this as a global replacement on whatever the user passed in.

Can we move this somewhere that's more specific to `--failed`?

nkostoulas: It's a problem regardless of the `--failed` flag when you provide more than one test argument to test, e.g. `plz test TARGET ARG1 ARG2`, which is what `--failed` creates internally.

peterebden: but what if I *want* to pass "ARG1 ARG2" to my test? Not all tests are Go tests and they are allowed to accept other flags, I don't think plz can have a hard policy here that it always makes this replacement.

nkostoulas: Ah right I see what you mean!
Is the `TESTS` env used in all tests or just Go? I've only seen it used in `please_go_test`. In any case I guess I can just do this replacement in the `testMainImpl` of Go tests only, since that is the problem I've observed.

tools/please_go_test/gotest/write_test_main.go
```diff
@@ -208,6 +209,7 @@ func main() {
 {{end}}
     testVar := os.Getenv("TESTS")
     if testVar != "" {
+		testVar = strings.Replace(testVar, " ", "|", -1)
```
nkostoulas, line 212: Not sure what's wrong with indentation here I seem to be using spaces with 4-space tabs.

src/test/test_step.go
peterebden: the args are definitely consumed by Python tests too, but I can't remember if they do it through that env var or through argv

tools/please_go_test/gotest/write_test_main.go
peterebden: I think gofmt uses tabs?

nkostoulas: Think this whole block is a template string thus gofmt doesn't format it. That and github uses 8-space which makes it look odd.

src/test/test_step.go
nkostoulas: Can't find a reference in the code. It seems `unittest` requires the format of args to be space separated but it might be through the command directly. There is also a `TEST_ARGS` env that is set when creating the environment but I can't see it being used either.

PR #1000, opened 2020-06-15
Update alternate image to go 1.15 beta

peterebden:
Would be nice to know it works before we get to release day...


peterebden:
No idea what's going on in the linked build (it doesn't fail locally). Maybe I'll try again when beta2 is out and it'll magically fix itself

PR #1006, opened 2020-06-18
Add ability to specify metadata alongside a build target

Tatskaari:
Added the ability to specify arbitrary python objects on build_rules via a new "metadata" parameter. This metadata can then be retrieved by the BIF get_rule_metadata which takes in a build label. 

This will become useful for toolchains where we need to specify information about the sdk e.g. JAVA_HOME


src/build/incrementality_test.go
```diff
@@ -51,6 +51,7 @@ var KnownFields = map[string]bool{
 	"TestOutputs":                 true,
 	"Stamp":                       true,
 	"OutputDirectories":           true,
+	"RuleMetadata":                true,
```
peterebden, line 54: should this be in one of the other sections of this file? looks like it doesn't actually contribute to the rule hash (which I think is OK since it all happens at parse ime)

PR #1008, opened 2020-06-20
Some fixes to caching of remote actions

peterebden:
This started out looking into getting sensible numbers for incrementality with remote execution, but ended up with some general fixes to caching during remote execution - it looks like we were trying to read metadata files but never wrote them.

Also need to thread the Cached state through the test results. Wonder if that would be better just as a property (we seem to spend a lot of time special-casing it) but that would be better in a separate PR.


src/remote/utils.go
```diff
@@ -199,7 +199,13 @@ func (c *Client) locallyCacheResults(target *core.BuildTarget, digest *pb.Digest
 	}
 	data, _ := proto.Marshal(ar)
 	metadata.RemoteAction = data
-	c.state.Cache.Store(target, c.localCacheKey(digest), nil)
+	// TODO(jpoole): Similar to retrieveTargetMetadataFromCache, it would be cleaner if we could store
+	//               into the cache from an arbitrary reader.
+	if err := build.StoreTargetMetadata(target, metadata); err != nil {
+		log.Warning("%s", err)
```
Tatskaari, line 205: Does this log line work? We need Warningf or just warning without the format string.

peterebden: Yep it does... remember this is not the same log library we use internally at TM, it always interpolates.

PR #1011, opened 2020-06-22
A start on Java toolchains

Tatskaari:
Implemented building jars and running tests using the java tool-chain.

Still to do:
1) Invoke the javac worker with the toolchains JRE
2) Host the JDKs ourselves so we can control the download link format
3) Write a convenience "please_hosted_jdk" rule that just takes a version and figures out the URL for each platform based on that. 
4) Migrate please to use java toolchan
5) Make sure all the other java rules work with the toolchain (java_runtime_image)
6) documentation

### DIFF
--- a/rules/java_rules.build_defs
+++ b/rules/java_rules.build_defs
@@ -1,7 +1,12 @@
 """Built-in rules to compile Java code."""
 
-def java_toolchain(name : str, sdk_url : str, visibility:list = ["PUBLIC"]) -> str:
+def java_toolchain(name : str, sdk_url : str|dict, visibility:list = ["PUBLIC"]) -> str:
     current_package = package_name()
+
+    sdk_url = sdk_url if isinstance(sdk_url, str) else sdk_url[CONFIG.OS]
+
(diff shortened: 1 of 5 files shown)


rules/java_rules.build_defs
```diff
@@ -25,9 +24,11 @@ def java_library(name:str, srcs:list=None, src_dir:str=None, resources:list=[],
       javac_flags (list): List of flags passed to javac.
       labels (list): Additional labels to apply to this rule.
```
peterebden, line 25: add toolchain to the docstring

rules/java_rules.build_defs
```diff
@@ -25,9 +24,11 @@ def java_library(name:str, srcs:list=None, src_dir:str=None, resources:list=[],
...
     """
+    toolchain = toolchain if toolchain else CONFIG.JAVA_TOOLCHAIN
```
peterebden, line 27: you should just be able to make `CONFIG.JAVA_TOOLCHAIN` the default for the argument, I think.

PR #1016, opened 2020-06-24
Only upload results if all runs have completed

Tatskaari:
We were getting the occasional concurrent map access in serialising the xml because each parallel run was attempting to upload the same results. This PR makes it so only the last run uploads the XML.

### DIFF
--- a/src/core/build_target.go
+++ b/src/core/build_target.go
@@ -141,6 +141,8 @@ type BuildTarget struct {
 	Progress float32 `print:"false"`
 	// The results of this test target, if it is one.
 	Results TestSuite `print:"false"`
+	// The number of completed runs
+	completedRuns int `print:"false"`
 	// A mutex to control access to Results
 	resultsMux sync.Mutex `print:"false"`
 	// Description displayed while the command is building.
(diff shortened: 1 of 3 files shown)


src/core/build_target.go
```diff
@@ -354,6 +356,15 @@ func (target *BuildTarget) TestDirs() string {
 	return path.Join(TmpDir, target.Label.Subrepo, target.Label.PackageName, target.Label.Name+testDirSuffix)
 }
 
+// CompleteRun completes a run and returns true if this was the last run
+func (target *BuildTarget) CompleteRun(state *BuildState) bool {
+	target.resultsMux.Lock()
+	defer target.resultsMux.Unlock()
+
+	target.completedRuns++
+	return target.completedRuns == state.NumTestRuns
```
goddenrich, line 365: Is this guaranteed to always be reached eg if the worker gets killed or there is an error?

peterebden: Not necessarily, something could kill the process before it gets here. It generally should though...

Tatskaari: I can put this in a defer so it is less likely to fail. I think either way it would be an exceptional case where this doesn't work.

PR #1021, opened 2020-06-24
Release notes for v15

peterebden:


ChangeLog
```diff
@@ -1,3 +1,25 @@
+Version 15.0.0
```
Tatskaari, line 1: Do we not want to change the version file?

peterebden: Done

PR #1023, opened 2020-06-25
Set required rex version to 2.1

peterebden:
This is gonna be tagged any day now...
We basically require a server with the non-specific output paths field. Our heuristic for guessing inevitably gets it wrong sooner or later and I don't want to keep the crappy code around for it.


peterebden:
This hasn't been tagged. I'll revisit if and when it is.

PR #1027, opened 2020-06-26
Implemented --rerun flag for test and cover

Tatskaari:
This can be useful if the code hasn't changed but something about the build env has


Tatskaari:
Oops, duplicate work!

PR #1028, opened 2020-06-26
Add rerun flag to test command

sagikazarmark:
This PR adds a `--rerun` flag to the test command to rerun tests for the selected targets, bypassing the cache for tests. It does not rebuild the dependencies though.

A common use case for this feature is rerunning integration tests against an external component.

## Questions / To do

- [x] How can we test this flag?
- [x] How about the `cover` command?

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -732,7 +732,7 @@ func Please(targets []core.BuildLabel, config *core.Configuration, shouldBuild, 
 	state.Watch = len(opts.Watch.Args.Targets) > 0
 	state.CleanWorkdirs = !opts.FeatureFlags.KeepWorkdirs
 	state.ForceRebuild = opts.Build.Rebuild
-	state.ForceTestRerun = opts.Test.Rerun || opts.Cover.Rerun
+	state.ForceRerun = opts.Test.Rerun || opts.Cover.Rerun
 	state.ShowTestOutput = opts.Test.ShowOutput || opts.Cover.ShowOutput
 	state.DebugTests = debugTests
 	state.ShowAllOutput = opts.OutputFlags.ShowAllOutput
--- a/src/please.go
+++ b/src/please.go
@@ -111,7 +111,7 @@ var opts struct {
 	Test struct {
 		FailingTestsOk  bool         `long:"failing_tests_ok" hidden:"true" description:"Exit with status 0 even if tests fail (nonzero only if catastrophe happens)"`
 		NumRuns         int          `long:"num_runs" short:"n" default:"1" description:"Number of times to run each test target."`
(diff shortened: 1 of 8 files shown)


sagikazarmark:
Duplicate of #1027

src/please.go
```diff
@@ -111,6 +111,7 @@ var opts struct {
 	Test struct {
 		FailingTestsOk  bool         `long:"failing_tests_ok" hidden:"true" description:"Exit with status 0 even if tests fail (nonzero only if catastrophe happens)"`
 		NumRuns         int          `long:"num_runs" short:"n" default:"1" description:"Number of times to run each test target."`
+		Rerun           bool         `long:"rerun" description:"To force rerunning tests for a target"`
```
Tatskaari, line 114: Can you also do this for Cover?

Tatskaari:
Thinking about how we could test this... if we have a test that prints something to std out, we can assert that this appears in the std-out when invoking `plz test :test --show_all_output`. If the test was cached, this log line won't appear. Please logs this stuff to std-err so be careful there. 

In //test/build_defs there's a `please_repo_e2e_test` rule. We can invoke please twice piping std-err to a file and asserting the contents.

sagikazarmark:
The breaking test does not seem to be related. I can't rerun the tests.

test/build_defs/test.build_defs
```diff
@@ -8,7 +8,7 @@ def please_repo_e2e_test(
         expected_output: dict,
 ):
     test_cmd = " && ".join(
-        ["cd $DATA_REPO", f"$TMP_DIR/$DATA_PLEASE {plz_command}"] +
+        ["cd $DATA_REPO", "export PLZ=$TMP_DIR/$DATA_PLEASE", f"$PLZ {plz_command}"] +
```
Tatskaari, line 11: I'd probably just change this to `"export PLZ=$TMP_DIR/$DATA_PLEASE", f"{plz_command}"`

I think this is more flexible.

test/cli/test_repo/package/BUILD_FILE
```diff
@@ -0,0 +1,5 @@
+gentest(
+    name = "wibble", 
+    #outs = ["wibble.txt"],
+    test_cmd = "echo wibble wibble wibble && echo '' > $RESULTS_FILE",
+)
```
Tatskaari, line 5: Let's just have `test_cmd = "echo wibble"` and set `no_test_output = True`

Tatskaari:
> The breaking test does not seem to be related. I can't rerun the tests.

Agreed, I reran it for you :)

PR #1029, opened 2020-06-26
Implement hash_filegroup in the build language

peterebden:
Gets rid of a lot of internal specialisation nonsense and makes it work for remote execution.

### DIFF
--- a/rules/misc_rules.build_defs
+++ b/rules/misc_rules.build_defs
@@ -286,16 +286,16 @@ def hash_filegroup(name:str, srcs:list=None, deps:list=None, exported_deps:list=
     return build_rule(
         name=name,
         srcs=srcs,
-        cmd='',
+        out_dirs = ['_out'],
+        cmd = 'for S in SRCS; do mv $S "${S%.*}-$(sha256sum $S)${S##*.}"; done',
         deps=deps,
         exported_deps=exported_deps,
         visibility=visibility,
-        building_description='Copying...',
+        building_description='Hashing...',
         output_is_complete=True,
         test_only=test_only,
         labels=labels,
         requires=requires,
-        _hash_filegroup=True,
     )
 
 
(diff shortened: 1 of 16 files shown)


rules/misc_rules.build_defs
```diff
@@ -286,16 +286,16 @@ def hash_filegroup(name:str, srcs:list=None, deps:list=None, exported_deps:list=
     return build_rule(
         name=name,
         srcs=srcs,
-        cmd='',
+        output_dirs = ['_out'],
+        cmd = 'mkdir _out; for S in $SRCS; do B="$(basename $S)"; mv "$S" "_out/${B%.*}-$(shasum -a 256 $S | cut -d " " -f 1 ).${B##*.}"; done',
```
Tatskaari, line 290: I have no idea what this does but I trust you that it works.

test/filegroup/BUILD
```diff
@@ -12,3 +12,14 @@ plz_e2e_test(
     name = "filegroup_concurrent_build_test",
     cmd = "plz clean //test/filegroup:gen && plz build //test/filegroup/many:all",
 )
+
+hash_filegroup(
+    name = "hashed",
+    srcs = ["hash_filegroup_test.txt"],
```
Tatskaari, line 18: Might be nice to have 2 files here so we trigger that loop in the bash

rules/misc_rules.build_defs
peterebden: It turns a filename like `script.js` into `script-a13ffe6ee73459f04e180132a3e38c3e5fc145e27b4a5a4613960b01058680aa.js`

test/filegroup/BUILD
peterebden: Yeah fair, don't quite have time now though, will revisit later.

PR #1035, opened 2020-06-30
Add rebuild flag to run command

sagikazarmark:
In some cases, I might want to run a target, but rebuild it first.

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -157,6 +157,7 @@ var opts struct {
 
 	Run struct {
 		Env      bool `long:"env" description:"Overrides environment variables (e.g. PATH) in the new process."`
+		Rebuild  bool `long:"rebuild" description:"To force the optimisation and rebuild one or more targets."`
 		Parallel struct {
 			NumTasks       int  `short:"n" long:"num_tasks" default:"10" description:"Maximum number of subtasks to run in parallel"`
 			Quiet          bool `short:"q" long:"quiet" description:"Suppress output from successful subprocesses."`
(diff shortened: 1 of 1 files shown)


sagikazarmark:
This looks like a flaky test.

sagikazarmark:
Can someone trigger the builds for me please?

Tatskaari:
Hey Mark, thanks for the PR. I'm just wondering what the justification is here. We've had a lot of requests for --rerun so thanks a load for that one however I've never seen anybody ask for --rebuild on plz run as of yet. I'm just cautious about muddying the command line interface with lots of less than useful flags.

sagikazarmark:
Well, the rationale is same as for having `--rebuild` for the build command.

Obviously, in a 100% Please compatible scenario you won't need that.

Adding the `--rebuild` flag basically is a shorthand for `plz build --rebuild && plz run` which I found myself doing a lot while playing with Please in a project which does not fully embrace the Please dependency&src system (yet).

sagikazarmark:
Here is a perfect explanation:

https://github.com/thought-machine/please/commit/118518bd176d824364ddbecfb18bc4c8b8e55fe2#diff-02f0b547c2779d25cff89672135f20e3R28

😄

peterebden:
I think it's a little different though - `plz run --rerun //my:target` will only rebuild `//my:target`, not everything along the way (whereas for `plz test` presumably just re-running the test is what's wanted). Is that what you want?

sagikazarmark:
Not sure I understand the difference between rebuilding the one target vs rerunning the tests (for one target).

Again, this is just a shorthand for `plz build --rebuild //something && plz run //something` (at least that's my understanding, correct me if I'm wrong). Since `plz run` is basically `plz build` and then executing the output, I thought it would make sense to add this flag to the run command as well.

Admittedly, this is more useful in a scenario where the build rule is not "pure" in the sense of please build caching, but so is `--rerun` for testing (eg. toolchain updated under the hood, whereas the toolchain could be part of the please build graph).

My use case for this is a custom `go_build` rule that just runs `go build` in the project root with a few parameters. I understand this is not the "please way", but it works nicely in projects where proper please adoption is yet to happen.

peterebden:
OK yep - just wanted to make sure it was going to do what you expect. SGTM then!

sagikazarmark:
Thanks!

PR #1043, opened 2020-07-02
Added doco around java toolchains

Tatskaari:


docs/template.html
```diff
@@ -111,7 +111,7 @@ <h1><a href="index.html"><img src="images/please_build_p.png"></a></h1>
       <img src="images/circle_y.png">
       <p><span class="white">Want Please?</span> Get started:</p>
       <p>&nbsp;</p>
-      <p><span class="button"><a href="quickstart.html">Yes Please</a></span></p>
+      <p><a href="quickstart.html"><span class="button">Yes Please</span></a></p>
```
Tatskaari, line 114: This makes the whole button clickable.

test/java_toolchain/src/main/java/net/thoughtmachine/bazel_compat/BUILD
```diff
@@ -0,0 +1,34 @@
+package(bazel_compatibility = True)
+
+if CONFIG.OS == "linux" or CONFIG.OS == "darwin":
+    please_java_toolchain(
```
peterebden, line 4: should these just be java_toolchain? that seems like the more idiomatic name...

Tatskaari: There's already a java_toolchain rule in Bazel so I have to alias it

PR #1045, opened 2020-07-03
Remove hacky _CREATE_OUTPUT_DIRS thing

peterebden:
Since #1024 we shouldn't need this any more (thank goodness)

### DIFF
--- a/src/remote/action.go
+++ b/src/remote/action.go
@@ -515,16 +515,6 @@ func (c *Client) buildEnv(target *core.BuildTarget, env []string, sandbox bool) 
 		env = append(env, "SANDBOX=true")
 	}
 	if target != nil {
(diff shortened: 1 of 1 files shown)


Tatskaari:
Are we sure? I think there are other rules in core3 that do this (go_mocks). If we want to remove this hack we probably need to remove the add_out built in so its not possible to add outputs from a post build function.

peterebden:
it's a little more subtle than that - it's not needed in general for adding outputs, only for build rules that fail when the output directory exists on re-execution.

peterebden:
FWIW have tested some go_mock targets and they still build fine after this change.

Tatskaari:
Okay I guess it would be better to have the rules accommodate this (i.e. by using || rather than && or something) rather than please itself.

PR #1047, opened 2020-07-05
support wildcards when using --include_file option with plz cover

thecapdan:
Example use:

```
plz cover //my/test:target --include_file=my/path/*
```

Unit tests added + tested wildcard using `plz cover` manually with several targets containing nested coverage results.

### DIFF
--- a/src/output/shell_output.go
+++ b/src/output/shell_output.go
@@ -686,9 +686,18 @@ func shouldInclude(file string, files []string) bool {
 		return true
 	}
 	for _, f := range files {
-		if file == f {
+		if file == f || wildcardMatch(file, f) {
 			return true
 		}
+	}
+	return false
+}
+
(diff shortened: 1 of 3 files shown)


src/output/shell_output.go
```diff
@@ -686,13 +686,22 @@ func shouldInclude(file string, files []string) bool {
 		return true
 	}
 	for _, f := range files {
-		if file == f {
+		if file == f || wildcardMatch(file, f) {
```
peterebden, line 689: You should use `path.Match` for this (https://golang.org/pkg/path/#Match); that takes care of a few more patterns, and basically does the work for you :)

We also have internal code for double-star globs (in `src/fs/glob.go`) although I don't think it's factored in a way that's super convenient to use here so maybe leave for now.

PR #1051, opened 2020-07-06
support pattern match when using --include_file option with plz cover

thecapdan:
Example use: `plz cover //my/test:target --include_file=my/path/*`

Might need more complex pattern to show results for nested directories, not sure if this should be default for the example above?

Unit tests added. Needs manual testing

### DIFF
--- a/src/output/shell_output.go
+++ b/src/output/shell_output.go
@@ -686,7 +686,7 @@ func shouldInclude(file string, files []string) bool {
 		return true
 	}
 	for _, f := range files {
-		if file == f {
+		if isMatch, _ := path.Match(f, file); isMatch {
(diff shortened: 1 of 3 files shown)


thecapdan:
Manually tested using various shell patterns (/* + /**/* etc). Also tested existing --include-file finctionality had not regressed

Tatskaari:
Nice one, you could do some "manual" testing in the form of a pleases repo test.

thecapdan:
> Nice one, you could also convert your "manual" testing to the form of a pleases repo test. See `//test/build_defs` for the build rule.

@Tatskaari you just blew my mind. thanks for the tip

PR #1052, opened 2020-07-07
Fix `plz hash --update` and possibly one or two other things.

peterebden:
errors.Is requires Go 1.13 but that shouldn't be a big deal I don't think.


Tatskaari:
> errors.Is requires Go 1.13 but that shouldn't be a big deal I don't think.

We're releasing this as part of a major release anyway

PR #1057, opened 2020-07-09
Add caching to circle

Tatskaari:


.circleci/config.yml
```diff
@@ -73,6 +73,9 @@ jobs:
        - save_cache:
            key: go-linux-alt-v3-{{ checksum "third_party/go/BUILD" }}
            paths: [ ".plz-cache/third_party/go" ]
+       - save_cache:
```
peterebden, line 76: you need a restore_cache directive to go along with this.

You also need to do it for each of the builds (at least linux, linux-alt and darwin)

Tatskaari: Yeah, I don't think rex needs this. Updated the other projects.

PR #1060, opened 2020-07-09
Memoises client.buildAction so it always returns the correct digest. 

Tatskaari:
The problem is that output directories change the build target in such a way that the command digests (and action digest) changes. This means that we can't retrieve the action result again once the build graph has been updated. 

This change memoises the buildAction  function so that it will always return the correct command and digest. 

This is probably not the right solution. Can discuss tomorrow.

### DIFF
--- a/src/remote/action.go
+++ b/src/remote/action.go
@@ -52,8 +52,15 @@ func (c *Client) uploadAction(target *core.BuildTarget, isTest, isRun bool) (*pb
 	return command, digest, err
 }
 
-// buildAction creates a build action for a target and returns the command and the action digest digest. No uploading is done.
+// buildAction creates a build action for a target and returns the command and the action digest. No uploading is done.
 func (c *Client) buildAction(target *core.BuildTarget, isTest, stamp bool) (*pb.Command, *pb.Digest, error) {
+	c.buildActionMutex.Lock()
+	defer c.buildActionMutex.Unlock()
+
+	if d, ok := c.buildActions[target.Label]; ok {
+		return d.command, d.actionDigest, nil
+	}
+
 	inputRoot, err := c.uploadInputs(nil, target, isTest)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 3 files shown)


Tatskaari:
This doesn't work because of stamped and unstamped digests...

src/remote/action.go
```diff
@@ -52,8 +52,15 @@ func (c *Client) uploadAction(target *core.BuildTarget, isTest, isRun bool) (*pb
 	return command, digest, err
 }
 
-// buildAction creates a build action for a target and returns the command and the action digest digest. No uploading is done.
+// buildAction creates a build action for a target and returns the command and the action digest. No uploading is done.
 func (c *Client) buildAction(target *core.BuildTarget, isTest, stamp bool) (*pb.Command, *pb.Digest, error) {
+	c.buildActionMutex.Lock()
+	defer c.buildActionMutex.Unlock()
+
+	if d, ok := c.buildActions[target.Label]; ok {
```
peterebden, line 60: I think you need to be careful to honour isTest and stamp here.

src/remote/action.go
```diff
@@ -52,8 +52,15 @@ func (c *Client) uploadAction(target *core.BuildTarget, isTest, isRun bool) (*pb
 	return command, digest, err
 }
 
-// buildAction creates a build action for a target and returns the command and the action digest digest. No uploading is done.
+// buildAction creates a build action for a target and returns the command and the action digest. No uploading is done.
 func (c *Client) buildAction(target *core.BuildTarget, isTest, stamp bool) (*pb.Command, *pb.Digest, error) {
+	c.buildActionMutex.Lock()
```
peterebden, line 57: a little unsure about holding this the whole time since we could end up waiting for non-trivial work (like hashing files or smthn).

Maybe use a sync.Map instead?

Tatskaari: Yeah will do

PR #1062, opened 2020-07-10
Fix output hashing for remote_file rules

peterebden:
They use subresource integrity on the downloaded file, so it has to be the actual hash of the file. Currently it's the hash of the action result which obviously isn't the same.

Will leave it that way in general since there is no single obvious way of calculating the hash of a bunch of directories, but for single files this should make things considerably more working.

### DIFF
--- a/src/remote/remote.go
+++ b/src/remote/remote.go
@@ -240,8 +240,8 @@ func (c *Client) Build(tid int, target *core.BuildTarget) (*core.BuildMetadata, 
 	if err != nil {
 		return metadata, err
 	}
-	hash, _ := hex.DecodeString(c.digestMessage(ar).Hash)
 	if c.state.TargetHasher != nil {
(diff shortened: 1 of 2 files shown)


src/remote/remote.go
```diff
@@ -240,8 +240,9 @@ func (c *Client) Build(tid int, target *core.BuildTarget) (*core.BuildMetadata,
 	if err != nil {
 		return metadata, err
 	}
-	hash, _ := hex.DecodeString(c.digestMessage(ar).Hash)
 	if c.state.TargetHasher != nil {
+		log.Warning("here %s %s", target, ar)
```
Tatskaari, line 244: o.O

peterebden: never happened

Tatskaari: Ahaha and you actually did a -f so it really never happened

PR #1066, opened 2020-07-10
go_test(): avoid redefined imports in generated test main files

### LINKED ISSUE #1065: go_test() generates invalid code when test code's package name is one of {os,strings,testing,testdeps}
When a `go_test()` target is defined for a test file belonging to any of the following packages, `plz build` fails when compiling the test file:

* `os`
* `strings`
* `testing`
* `testdeps`

MWE:

`strings/BUILD`:
```
go_test(
    name = "strings_test",
    srcs = ["strings_test.go"],
)
```

strings/strings_test.go`:
```go
package strings

import "fmt"

func TestOK() {
    fmt.Println("ok")
    // Output: ok
}
```

Output from `plz build //strings:strings_test`:
```
Build stopped after 410ms. 1 target failed:
    //strings:_strings_test#main_lib
Error building target //strings:_strings_test#main_lib: exit status 1
strings/strings_test_main.go, line 35, column 13: undefined: strings.Replace
```

This is because the [test main file template](https://github.com/thought-machine/please/blob/master/tools/please_go_test/gotest/write_test_main.go#L140) already imports packages with the above names from the Go standard library; e.g. in the MWE above, the following invalid test main file is generated:

```
package main

import (
        "os"
        "strings"
        "testing"
    "testing/internal/testdeps"


        strings "common/go/strings"

)

[...]
```

This bug can also manifest itself as Go complaining about the package name being redeclared when compiling the test main file.

The quickest way to fix this would be to assign aliases to the packages imported from the Go standard library in the template, to eliminate the cause of the collision.

chrisnovakovic:
When a `go_test()` target is defined for a test file belonging to any of the following packages, `plz build` fails when compiling the test file:

* `os`
* `strings`
* `testing`
* `testdeps`

This is because the test main file template already imports packages with these names from the Go standard library before the test package with an identical name is imported. Alias these imports with the prefix `_gostdlib_`, making it much less likely that `WriteTestMain()` generates invalid code because of redefined imports.

Fixes #1065.

Also corrects the indentation of the `testdeps` import in the template.

### DIFF
--- a/tools/please_go_test/gotest/write_test_main.go
+++ b/tools/please_go_test/gotest/write_test_main.go
@@ -141,28 +141,28 @@ var testMainTmpl = template.Must(template.New("main").Parse(`
 package main
 
 import (
-	"os"
-	{{if not .Benchmark}}"strings"{{end}}
-	"testing"
-    "testing/internal/testdeps"
+	_gostdlib_os "os"
+	{{if not .Benchmark}}_gostdlib_strings "strings"{{end}}
+	_gostdlib_testing "testing"
+	_gostdlib_testdeps "testing/internal/testdeps"
 
 {{range .Imports}}
 	{{.}}
 {{end}}
 )
 
-var tests = []testing.InternalTest{
+var tests = []_gostdlib_testing.InternalTest{
 {{range .TestFunctions}}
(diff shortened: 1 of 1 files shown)


peterebden:
Thanks!

PR #1071, opened 2020-07-13
Encapsulate state.OriginalTargets & add a lock

peterebden:
Saw a related panic that I've not seen before the other day. There is a data race around this guy; I had assumed it's benign but better to be safe.

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -146,9 +146,6 @@ type BuildState struct {
 	RemoteClient RemoteClient
 	// Hasher for targets
 	TargetHasher TargetHasher
-	// Targets that we were originally requested to build
-	originalTargets     []BuildLabel
-	originalTargetMutex sync.Mutex
 	// Arguments to tests.
 	TestArgs []string
 	// Labels of targets that we will include / exclude
@@ -231,6 +228,9 @@ type stateProgress struct {
 	pendingPackageMutex sync.Mutex
 	// The set of known states
 	allStates []*BuildState
+	// Targets that we were originally requested to build
+	originalTargets     []BuildLabel
+	originalTargetMutex sync.Mutex
 }
 
 // SystemStats stores information about the system.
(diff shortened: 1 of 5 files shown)


src/core/state.go
```diff
@@ -444,11 +447,9 @@ func (state *BuildState) AddOriginalTarget(label BuildLabel, addToList bool) {
...
 	}
 	if addToList {
-		// The sets of original targets are duplicated between states for all architectures,
-		// we must add it to all of them to ensure everything sees the same set.
-		for _, s := range state.progress.allStates {
-			s.OriginalTargets = append(s.OriginalTargets, label)
-		}
+		state.progress.originalTargetMutex.Lock()
+		state.progress.originalTargets = append(state.progress.originalTargets, label)
+		state.progress.originalTargetMutex.Unlock()
```
Tatskaari, line 452: Do we not care about other architectures anymore?

src/core/state.go
```diff
@@ -578,7 +579,10 @@ func (state *BuildState) ExpandOriginalTargets() BuildLabels {
 // ExpandOriginalLabels expands any pseudo-labels (ie. :all, ... has already been resolved to a bunch :all targets)
 // from the set of original labels.
 func (state *BuildState) ExpandOriginalLabels() BuildLabels {
-	return state.ExpandLabels(state.OriginalTargets)
+	state.progress.originalTargetMutex.Lock()
+	targets := state.progress.originalTargets[:]
+	state.progress.originalTargetMutex.Unlock()
```
Tatskaari, line 584: nit: could use defer here

src/please.go
```diff
@@ -468,8 +468,7 @@ var buildFunctions = map[string]func() int{
 	"gc": func() int {
 		success, state := runBuild(core.WholeGraph, false, false, true)
 		if success {
-			state.OriginalTargets = state.Config.Gc.Keep
-			gc.GarbageCollect(state, opts.Gc.Args.Targets, state.ExpandOriginalTargets(), state.Config.Gc.Keep, state.Config.Gc.KeepLabel,
+			gc.GarbageCollect(state, opts.Gc.Args.Targets, state.ExpandLabels(state.Config.Gc.Keep), state.Config.Gc.Keep, state.Config.Gc.KeepLabel,
```
Tatskaari, line 471: Just checking you meant to make this change. It's not obvious it's related.

src/core/state.go
peterebden: the `progress` sub-structure is shared between all the states so it doesn't need the loop any more.

src/core/state.go
peterebden: I don't want to hold the lock around ExpandLabels though

PR #1073, opened 2020-07-15
Fix issue with logging completed results in remote exection

Tatskaari:
We were logging a completed build in the `monitorProgress` method in remote.go as well as in build_step.go and test_step.go. This is a problem when these targets are in the pendingTargets map as the channel gets closed twice. This is the case for sub-includes and target metadata rules.

### DIFF
--- a/src/remote/remote.go
+++ b/src/remote/remote.go
@@ -708,7 +708,7 @@ func (c *Client) updateProgress(tid int, target *core.BuildTarget, metadata *pb.
 		case pb.ExecutionStage_EXECUTING:
 			c.state.LogBuildResult(tid, target.Label, core.TargetBuilding, "Building...")
 		case pb.ExecutionStage_COMPLETED:
-			c.state.LogBuildResult(tid, target.Label, core.TargetBuilding, "Completed...")
+			c.state.LogBuildResult(tid, target.Label, core.TargetBuilding, "Completed")
 		}
 	} else {
 		switch metadata.Stage {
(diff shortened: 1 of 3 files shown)


src/remote/remote.go
```diff
@@ -708,7 +708,8 @@ func (c *Client) updateProgress(tid int, target *core.BuildTarget, metadata *pb.
 		case pb.ExecutionStage_EXECUTING:
 			c.state.LogBuildResult(tid, target.Label, core.TargetBuilding, "Building...")
 		case pb.ExecutionStage_COMPLETED:
-			c.state.LogBuildResult(tid, target.Label, core.TargetBuilt, "Built")
+			// The build result will be logged in build/build_step.go
```
peterebden, line 711: Might still want to log this as `TargetBuilding` but with a slightly different message?

tools/java/BUILD
```diff
@@ -18,8 +18,8 @@ if CONFIG.OS == "linux" or CONFIG.OS == "darwin":
     java_toolchain(
         name = "toolchain",
         hashes = [
-            "81bf8bee66555b8177fb4c4dd770ee9c1d243bced4bade4661f10fb1b1cf24be",
-            "a357054504bd55cceec52a9468834da5d12448e2b7779b6a1f24367d20454a33",
+            "bcd6c66f3c8a8b78c2caf7e0006be5598c365a2245120b6e8af9096d7c495c34",
```
peterebden, line 21: why are these changing?

Tatskaari: I'm not sure

src/remote/remote.go
Tatskaari: Oh maybe, I figured it was moments away from returning back to build_step.go so it didn't really matter. Though ti would be a bit weird to log "Built" with the status `TargetBuilding` no?

peterebden: Yeah Complete sounds good (I was thinking "Execution complete" but shorter is better)

tools/java/BUILD
peterebden: I suppose because it's `downloads/latest/`, i.e. presumably they have released a patch release to corretto 11.
We should either find a pinned URL or lose these hashes (because this is gonna keep happening)

Tatskaari: YEah and 3 days ago they pushed: https://github.com/corretto/corretto-11/commit/d7aaa4ab00a18313663f71cf40e97c6f78eabab3

Tatskaari: Yeah okay they suggest using versioned releases from their github. Will do that instead.

PR #1078, opened 2020-07-16
Add a module major version to go_get rules

Tatskaari:
This allows us to deal with repositories that have a major version so the github repo structure doesn't match where we actually want the sources when building.

### DIFF
--- a/rules/go_rules.build_defs
+++ b/rules/go_rules.build_defs
@@ -625,7 +625,7 @@ def cgo_test(name:str, srcs:list, data:list=None, deps:list=None, visibility:lis
 def go_get(name:str, get:str|list, repo:str='', deps:list=[], exported_deps:list=None,
            visibility:list=None, patch:str=None, binary:bool=False, test_only:bool&testonly=False,
            install:list=None, revision:str|list=None, strip:list=None, hashes:list=None,
-           extra_outs:list=[], licences:list=None):
+           extra_outs:list=[], licences:list=None, module_major_version:str=''):
     """Defines a dependency on a third-party Go library.
 
     Note that unlike a normal `go get` call, this does *not* install transitive dependencies.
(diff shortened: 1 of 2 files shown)


peterebden:
Maybe add a mention of this in the release notes?

rules/go_rules.build_defs
```diff
@@ -625,7 +625,7 @@ def cgo_test(name:str, srcs:list, data:list=None, deps:list=None, visibility:lis
 def go_get(name:str, get:str|list, repo:str='', deps:list=[], exported_deps:list=None,
            visibility:list=None, patch:str=None, binary:bool=False, test_only:bool&testonly=False,
            install:list=None, revision:str|list=None, strip:list=None, hashes:list=None,
-           extra_outs:list=[], licences:list=None):
+           extra_outs:list=[], licences:list=None, module_major_version:str=''):
```
sagikazarmark, line 628: Shouldn't it also support a list as a bunch of other parameters (used for compiling circular dependent packages in a single rule)?

Tatskaari: God my brain hurts thinking about that. I guess we can index match the get list with this new list. I think we do this in a few other places.

Tatskaari: Ah, never mind. The code is already there for revisions and tags.

peterebden: I'd like to move away from that in the future (as Jon says it is not easy to understand and I'm not sure if / where we still use it) - maybe in favour of a more full-fledged module implementation like the one you've been working on, Mark. Hence happy to treat this as a slightly more temporary measure.

sagikazarmark: Yeah, I'm not to keen on using `go_get` that way, just wanted to point it out. 👍

PR #1081, opened 2020-07-16
Fix deadlock when getting rule metadata

Tatskaari:


src/parse/asp/builtins.go
```diff
@@ -233,12 +233,18 @@ func getRuleMetadata(s *scope, args []pyObject) pyObject {
...
 
 	t := s.state.Graph.Target(label)
+
 	if t == nil {
 		if label.Subrepo == s.pkg.SubrepoName && label.PackageName == s.pkg.Name {
 			// This is a get_rule_metadata in the same package, check the target exists.
 			log.Fatalf("Target %s is not defined in this package yet; it has to be defined before the get_rule_metadata() call", name)
 		}
-		t = s.state.WaitForBuiltTarget(label, core.NewBuildLabel(s.pkg.Name, "all"))
+		func(){
```
peterebden, line 242: I think these two are common enough that you could pull them out to a separate function even

Tatskaari: Done

PR #1087, opened 2020-07-21
Calculate combined test duration for shell output summary

thecapdan:
manually tested locally

Example output: `1 test target and 14 tests run in 1.428s; 14 passed. Total time 4.34s.`
Previously displayed: `1 test target and 14 tests run in 0s; 14 passed. Total time 4.34s.`


thecapdan:
This does create slightly confusing output for multiple targets:

```
8 test targets and 69 tests run in 18.372s; 69 passed. Total time 8.75s.
```
The combined target time is greater than the total run time because of targets running in parallel

peterebden:
Yeah we could think about how to rephrase that to make it clearer I suppose. That is more or less what it's meant to display though.

PR #1089, opened 2020-07-21
Un-sandbox pip_library rule.

### LINKED ISSUE #1088: `pip_library` build rules currently fail with requests timeout to pypi.org
All the failures are some form of `requests` timeout exception:

```
pip._vendor.requests.exceptions.ConnectionError: HTTPSConnectionPool(host='pypi.python.org', port=443): Max retries exceeded with url: /simple/click/ (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x7f0602502c18>: Failed to establish a new connection: [Errno -5] No address associated with hostname',))
```

I think it's because the current build rule to fetch pip srcs are sandboxed, so we have no chance at connecting to pypi. Likely a one line change to https://github.com/thought-machine/please/blob/e463593ac614e9464c131fab5f1a3cde7637fd92/rules/python_rules.build_defs#L403

bLiGM:
Fixes #1088

This change makes a small tweak to un-sandbox the command in `pip_library` that requires network connections to remote resources.

Not sure if we want to add a new test in `tests/python_rules` to point at a pypi dependency (e.g. see tensorflow_test - but with a pypi dependency). Could potentially flake a CI build every now and then but otherwise I'm not sure how we can catch regressions here otherwise.

### DIFF
--- a/rules/python_rules.build_defs
+++ b/rules/python_rules.build_defs
@@ -411,6 +411,7 @@ def pip_library(name:str, version:str, hashes:list=None, package_name:str=None, 
         tools = {
             'pip': [CONFIG.PIP_TOOL],
         },
+        sandbox = False,
     )
 
     cmd = '$TOOLS_JARCAT z -d -i .'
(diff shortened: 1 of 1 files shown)


Tatskaari:
>Not sure if we want to add a new test in tests/python_rules to point at a pypi dependency (e.g. see tensorflow_test - but with a pypi dependency). Could potentially flake a CI build every now and then but otherwise I'm not sure how we can catch regressions here otherwise.

I don't think we sandbox tests/builds in the normal please repo. Perhaps we should.

PR #1091, opened 2020-07-23
Tag v15

Tatskaari:


peterebden:
Last minute - would like to get #1092 in first (I think we would consider it a breaking change to remote execution, but we definitely want it internally for a couple of things so don't want to defer to a future v16)

PR #1096, opened 2020-07-25
Use golangci-lint

peterebden:
Fixed a bunch of errors; generally minor stuff but nice to have.
Have left errcheck for now since there are quite a few cases (but we should deal with those at some point).

A few weird false positives from unused and staticcheck; weirdly they don't seem to be particularly "hard" cases, or at least not obviously so.

### DIFF
--- a/rules/misc_rules.build_defs
+++ b/rules/misc_rules.build_defs
@@ -363,7 +363,7 @@ def remote_file(name:str, url:str|list, hashes:list=None, out:str=None, binary:b
             labels = labels,
         )
         if strip_prefix:
-            cmd = '"$TOOL" x $SRCS -s ' + strip_prefix
+            cmd = '"$TOOL" x $SRCS -o "$OUT" -s ' + strip_prefix
         else:
             cmd = '"$TOOL" x $SRCS -o "$OUT"'
         return build_rule(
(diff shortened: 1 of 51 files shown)


rules/misc_rules.build_defs
```diff
@@ -363,7 +363,7 @@ def remote_file(name:str, url:str|list, hashes:list=None, out:str=None, binary:b
             labels = labels,
         )
         if strip_prefix:
-            cmd = '"$TOOL" x $SRCS -o "$OUT" -s ' + strip_prefix
```
macripps, line 366: What's happened here? Do you not need the OUT file name?

sagikazarmark:
Note that some linters actually compile the packages they lint (haven't checked the list of enabled linters), so this could cause the regular go module download to be triggered. Please is not particularly heavy, so it shouldn't be a problem, but for large projects this can increase the build time. (I'm actually looking for a solution for this problem)

rules/misc_rules.build_defs
peterebden: Ah that's where that went... I thought it was on another branch.

I'll split it out to a separate PR. I don't think the current setup works basically (at least I can't make it do anything I want...)

peterebden:
> Note that some linters actually compile the packages they lint (haven't checked the list of enabled linters), so this could cause the regular go module download to be triggered.

True, we do restore the mod cache on Circle though and it seems fast enough. Locally it's a bit slow the first time but pretty quick after.
The only one that isn't tagged as 'fast' is `unused` so we could look at turning that off if it's a problem; from a little testing here it doesn't seem to make much difference though.

PR #1099, opened 2020-07-29
Fix some cache warnings

peterebden:
Various "no such file" type errors when trying to cache metadata during remote builds. These are mostly benign but might lead to some things not being cached correctly.

Problem is that we're writing the build metadata file multiple times for stamped targets. The cleanest solution (although not the least code) is to implement the TODO about being able to write a file without going through the disk first.


peterebden:
Gonna rethink this... I think there are some underlying issues with the HTTP cache that the Cirrus build is showing.

PR #1102, opened 2020-07-31
Graph parallelism improvements

peterebden:
Very much a WIP but seems worth discussing if this is an interesting direction.

The original motivation is to remove the centralised mutex in `BuildGraph` which is fairly highly contended. Earlier attempts (#864) did this but ended up with some subtle errors where things could fail.

This attempts to solve with a wider restructure. Rather than maintaining a map of reverse dependencies and trying to update it in an "event-based" way, it has a blocking model where we have a goroutine for each build target that blocks until it's ready to go.


peterebden:
This is in mostly fairly good shape now (thanks for fixing all the hard bits, Jon!).
The only part remaining is handling dependencies that are added post-build; the current model really only handles waiting on each target once. I think I might need to wait for inspiration to strike here though, it's not very obvious how best to handle that.

peterebden:
OK well that only took nine months or so to get green...

peterebden:
Given the current point in v16 I think I'll save this for 16.1.0 (and take it for more of a spin locally first).

PR #1107, opened 2020-08-03
Implement metadata store for rex instead of using the cache

Tatskaari:
Mainly just getting this out there so I can get some feedback. Probs needs some tests.

### DIFF
--- a/src/remote/build_metadata_store.go
+++ b/src/remote/build_metadata_store.go
@@ -1,0 +1,78 @@
+package remote
+
+import (
+	"encoding/gob"
+	"fmt"
+	"github.com/thought-machine/please/src/core"
+	"github.com/thought-machine/please/src/fs"
+	"os"
+	"path/filepath"
+)
+
+const pleaseCacheDirName = "please"
+const storeDirectoryName = "build-metadata-store"
+
+type buildMetadataStore interface {
+	StoreMetadata(key string, md *core.BuildMetadata) error
+	RetrieveMetadata(key string) (*core.BuildMetadata, error)
+}
+
+func newDirMDStore() *directoryMetadataStore {
+	userCacheDir, err := os.UserCacheDir()
+	if err != nil {
+		panic(fmt.Errorf("failed to find user cache dir for metadata store: %v", err))
+	}
+	dir := filepath.Join(userCacheDir, pleaseCacheDirName, storeDirectoryName)
+
+	if err := os.MkdirAll(dir, fs.DirPermissions); err != nil {
+		panic(fmt.Errorf("failed to create metadata store directory: %v", err))
+	}
+	return &directoryMetadataStore{
+		directory: dir,
+	}
+}
+
+type directoryMetadataStore struct {
+	directory string
+}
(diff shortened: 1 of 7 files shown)


src/remote/build_metadata_store.go
```diff
@@ -0,0 +1,78 @@
...
+
+type buildMetadataStore interface {
+	StoreMetadata(key string, md *core.BuildMetadata) error
+	RetrieveMetadata(key string) (*core.BuildMetadata, error)
+}
+
+func newDirMDStore() *directoryMetadataStore {
+	userCacheDir, err := os.UserCacheDir()
+	if err != nil {
+		panic(fmt.Errorf("failed to find user cache dir for metadata store: %v", err))
```
peterebden, line 23: probably `log.Panicf` or `log.Fatalf` or smthn?

src/remote/build_metadata_store.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	dir := filepath.Join(userCacheDir, pleaseCacheDirName, storeDirectoryName)
```
peterebden, line 25: Might be worth putting an extra layer in here (I quite like a two-letter prefix) so you don't just have one directory with millions of entries in it... (you'd at  least have 1/256th of them each)

Tatskaari: I *could* use the targets package and name?

src/remote/remote_test.go
```diff
@@ -261,6 +261,47 @@ func TestOutDirsSetOutsOnTarget(t *testing.T) {
...
+	md, err = store.retrieveMetadata("keep")
+	require.NoError(t, err)
+	assert.Equal(t, md, mds["keep"])
+
+	store.clean()
+
+	assert.FileExists(t, filepath.Join(storeDirectoryName, "ke", "keep"))
+
+	_, err = os.Lstat(filepath.Join(storeDirectoryName, "de", "delete"))
+	assert.True(t, os.IsNotExist(err))
```
Tatskaari, line 302: There's not assert.FileNotExists....

src/remote/build_metadata_store.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// clean will delete any cached metadata that has expired
+func (d *directoryMetadataStore) clean() {
+	_ = fs.Walk(d.directory, func(name string, isDir bool) error {
+		if isDir {
+			return nil
+		}
+
+		if md, err := loadMetadata(name); err == nil && d.hasExpired(md) {
```
peterebden, line 52: you might need to make sure this isn't trying to read a file you're still in the process of writing? or do we just think the `err == nil` will solve that?
I think it's atomic on the read side...

Tatskaari: This is all best effort. I assume it will error and just carry on happily nuking as much as it can.

peterebden: I *think* this is ok because if it's still  being written it will error  on read and won't get inside this loop...

PR #1110, opened 2020-08-04
Tag v15.0.1

Tatskaari:


ChangeLog
```diff
@@ -1,3 +1,10 @@
+Version 15.0.1
+--------------
+
+    * Go rules now write a go.mod file to plz-out to help go tooling work better with a please repo.
```
peterebden, line 4: also #1100 is probably worth noting here

PR #1114, opened 2020-08-06
Do git operations in parallel

peterebden:
`CurrentRevIdentifier` is fast (< 10ms) for git, but both `DescribeIdentifier` and `CurrentRevDate` are both slow (~0.5s on our repo here; they seem to be very repo size dependent). Might as well do them in parallel.


Tatskaari:
Bah getting EOFs on the java tests...

Issue #1111, opened 2020-08-05
Limit Max Amount of Parallel HTTP Cache Requests


fkorotkov:
It seems there is no limit on how many requests can be done simultaneously which can overwhelm either network or remote cache. It might be reasonable to limit the amount of parallel cache requests based on amount of CPUs or do like Bazel and limit by a [configurable flag with a default of 100](https://github.com/bazelbuild/bazel/blob/62475e738f1206a06dbda0d2638015c2c2fb1c88/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java#L60-L69)

Tatskaari:
Cache requests are probably not CPU bound but we can probably try and limit this to some configurable number like you suggest.

PR #1115, opened 2020-08-06
Limit the number of concurrent http cache requests

Tatskaari:
Resolves https://github.com/thought-machine/please/issues/1111


src/core/config.go
```diff
@@ -244,6 +244,7 @@ func DefaultConfiguration() *Configuration {
 	config.BuildEnv = map[string]string{}
 	config.Cache.HTTPWriteable = true
 	config.Cache.HTTPTimeout = cli.Duration(25 * time.Second)
+	config.Cache.HTTPConcurrentRequestLimit = 100
```
peterebden, line 247: 100 seems like probably too many to be useful? I would have thought say 10 concurrent requests is likely enough...

PR #1117, opened 2020-08-07
Change the way we mark the java toolchain as binary

Tatskaari:


rules/java_rules.build_defs
```diff
@@ -605,8 +605,8 @@ def _java_toolchain(name : str, jdk_url : str|dict='', jdk:str='', visibility:li
         cmd = " && ".join([
             '"$TOOL" x $SRCS -o extract_dir',
             f'mv $(dirname $(dirname $(find extract_dir/ -name javac))) {name}',
+            f"chmod +x {name}/bin/*",
         ]),
-        binary = True,
```
peterebden, line 609: you might still want this?

PR #1122, opened 2020-08-11
Make plz clean --nobackground more robust

peterebden:
Currently it uses `rm -rf` for normal `plz clean` but `os.RemoveAll` for `--nobackground` which is roughly akin to `rm -r`. This makes it more robust to read-only files etc.

Also move the code out of `//src/core` since that's big enough already.

### DIFF
--- a/src/build/BUILD
+++ b/src/build/BUILD
@@ -6,7 +6,6 @@ go_library(
     ),
     visibility = ["PUBLIC"],
     deps = [
-        "//src/cache",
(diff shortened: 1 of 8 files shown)


Tatskaari:
Seems like you need to move the tests too

peterebden:
OK I'm going to tackle the Cirrus tests in #1129 (think I know what is going on and it's not relevant to this change)

Issue #1061, opened 2020-07-09
go get quoting breaks linking


sagikazarmark:
Quoting introduced in #856 breaks linking (the linker doesn't find the necessary files).

I have been able to trace it back to #856.

The issue can be reproduced using the example in this repository: https://github.com/sagikazarmark/please-issues/tree/master/go-get-quote

PR #1124, opened 2020-08-11
Fix go_binary linking of third party dependencies outside of root #1061

brian-bice:
For reference I am using MacOS.

This issue only exists for the `go_binary` rule and third party dependencies which are added as library paths through command substitution. i.e. third party dependencies defined outside of the root `third_party/go/BUILD.plz`. The `go_library` rule works only because it's not actually quoting the file paths due to a subtle bug.

```bash
$(for i in $(find "$TMP_DIR" -name pkg -type d); do echo -n " -I \"$i\"/{CONFIG.GOOS}_{CONFIG.GOARCH} "; done)
```

The snippet above in the python `go_library` rule becomes the following in the actual shell.

```bash
$(for i in $(find "$TMP_DIR" -name pkg -type d); do echo -n " -I "$i"/darwin_amd64 "; done)
```

The escaped quote is lost and instead we get string concatenation which is probably not what was originally intended. In regards to the `go_binary` rule it **is** actually quoting the file paths. However, because it is being injected via command substitution it does not work properly. There are a couple issues here.

 Issue **1)** The output of command substitution is split on whitespace regardless of quotes within the output. So, for example, a file path which includes spaces `printargs $(echo '"foo bar\baz.txt"')` would become the following arguments.

```
arg 0: "foo
arg 1: bar\baz.txt"
```

Issue **2)** The quotes within the command substitution output are preserved. So, if we have `-L "foo/bar/pkg/darwin_amd64"` it's going to attempt to look in a directory named `"foo`, which begins with a quote, and not the correct directory `foo`.

This fix removes the need for quoting by using null separated arguments instead and providing them to the link/compile command via xargs instead of command substitution. Once I began manually testing with file paths which included spaces I noticed that the `_LINK_PKGS_CMD` also didn't account for spaces so I fixed it in a similar manner.

### DIFF
--- a/rules/go_rules.build_defs
+++ b/rules/go_rules.build_defs
@@ -6,11 +6,10 @@ rules to Go packages.
 
 # This links all the .a files up one level. This is necessary for some Go tools to find them.
 _LINK_PKGS_CMD = ' '.join([
-    'for FN in `find . -name "*.a" | sort`; do',
+    'find . -name "*.a" -print0 | sort -z | xargs -0 -n1 | while read FN; do',
     'DN=${FN%/*}; BN=${FN##*/}; if [ "${DN##*/}" == "${BN%.a}" ]; then',
-    'ln -s "$TMP_DIR"/$FN ${DN%/*}; fi; done'
+    'ln -s "$TMP_DIR/$FN" "${DN%/*}"; fi; done'
 ])
-
 
 def go_library(name:str, srcs:list, asm_srcs:list=None, hdrs:list=None, out:str=None, deps:list=[],
                visibility:list=None, test_only:bool&testonly=False, complete:bool=True,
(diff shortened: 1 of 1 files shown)


Tatskaari:
Hi Brian, thanks very much for sending this over. It sounds like what you're doing is sensible. It looks like your usage of xargs isn't valid on the version that comes bundles on ubuntu. I guess this is a difference between gnu xargs vs. whatever macos and FreeBSD use. 

If you can get that fixed, I can do some sanity checking on my end to make sure this doesn't have any regressions.

brian-bice:
Hi Tatskaari. Unfortunately I didn't realize that the `-J` option was going to be an issue until after I made the pull request. I spent some time yesterday trying to figure out how I could move forward with xargs in a portable way but ultimately came to the conclusion that it would not be possible.

I'm currently investigating some alternative approaches which won't require xargs.

Tatskaari:
Nice! I think that makes sense. This resolves #1061 doesn't it? Let me play around with this and see if it breaks anything on our side. 

I've restarted your build. Amazon sometimes cut our downloads off for the JDK. I guess they get too many from that IP address.

Tatskaari:
Oh boy, it looks like mapfile is a bash 4.0 thing but macs come with 3.2. I suppose we could switch the way we do this based on the platform we're on. What do you think @peterebden ?

brian-bice:
Maybe I can just do without `mapfile` and use a loop to load the array if this is going to be an issue.

brian-bice:
Just pushed a change to remove the mapfile requirement. It should be okay on bash 3.2 now.

peterebden:
Can you rebase this off master please? I think that will help the FreeBSD build (on Cirrus), and I'd like to get tests green on all platforms to check there aren't any weirdnesses in BSD awk or similar that we fall afoul of.

Tatskaari:
@peterebden Thanks for that. I'll make a build off this branch and test a few things here at TM before merging. 

@brian-bice many thanks for this heroic effort, it seems like it was affecting a few of our users.

brian-bice:
@peterebden @Tatskaari  I rebased with master and also ran `//rules:generate` after a test failure said to do so. It appears that `TestOutDirsSetOutsOnTarget` failed on ci. Is this failure related to my changes?

Tatskaari (approved):
Thanks Brian, much appreciated.

PR #1129, opened 2020-08-12
Fixes to HTTP cache and Cirrus builds

peterebden:


src/cache/http_cache.go
```diff
@@ -142,13 +142,14 @@ func (cache *httpCache) retrieve(target *core.BuildTarget, key []byte) (bool, er
 	resp, err := cache.client.Do(req)
 	if err != nil {
 		return false, err
-	} else if resp.StatusCode == http.StatusNotFound {
+	}
```
Tatskaari, line 145: This seems like a more interesting change. Maybe we should re-name the PR

peterebden: Done. This is ready for review now.

.cirrus.yml
```diff
@@ -6,7 +6,7 @@ env:
 
 task:
   install_script: pkg install -y bash protobuf go git python3 && python3 -m ensurepip
-  build_script: PLZ_ARGS="-o cache.httpurl:http://$CIRRUS_HTTP_CACHE_HOST" ./bootstrap.sh --exclude pip --exclude py2 --exclude no_cirrus --exclude cc
+  build_script: ./bootstrap.sh --exclude pip --exclude py2 --exclude no_cirrus
```
Tatskaari, line 9: It would be nice to get some testing of the http cache. Maybe we can write some integration tests though.

peterebden: Yep agreed; I'm not sure if this will be a permanent thing. Ideally we'd be able to re-add it since it would make Cirrus builds faster if it were working again.

PR #1130, opened 2020-08-13
Fix cc protoc rules so they actually pass hdrs to the cc_library

Tatskaari:
Before we had some annotated outputs post-build stuff that would allow us to depend on subsets of the output of the singular ptotoc rule. In the new world we have many protoc rules: one for each language. This meant there was no way to depends on the hdrs and the normal srcs anymore. 

This fixes that by adding a filter_srcs rule that just re-emits a subset of its srcs based on their extension.


rules/proto_rules.build_defs
```diff
@@ -227,11 +226,10 @@ def proto_languages():
         'cc': proto_language(
             language = 'cc',
             proto_language = 'cpp',
-            extensions = {'cc': ['.pb.cc'], 'cc_hdrs': ['.pb.h']},
             func = lambda name, srcs, deps, test_only: cc_library(
                 name = name,
                 srcs = srcs,
-                hdrs = [srcs[0] + '_hdrs'],
+                hdrs = [filter_srcs(name + '_hdrs', srcs[0], "h")],
```
Tatskaari, line 232: We'll need to do this in the toolchain repo too

Issue #1131, opened 2020-08-14
Have some better integration testing around http caching


Tatskaari:
We're disabled the http cache for cirrus. It would be nice to have some automated testing around this. 

We should be able to spin up a http cache, and write a please e2e test to check it works.

PR #1133, opened 2020-08-14
Added a http cache implementation

Tatskaari:
Might be useful to some people however I mostly just wanted an easy way to test that the HTTP caching logic on the please side works. 

Addresses #1131

### DIFF
--- a/.circleci/http_cache_test.sh
+++ b/.circleci/http_cache_test.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -eu
+
+trap 'killall http_cache' SIGINT SIGTERM EXIT
+
+DIR="${1:-/tmp/please}"
+
+# Extract the plz installation from earlier step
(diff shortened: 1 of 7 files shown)


.circleci/http_cache_test.sh
```diff
@@ -0,0 +1,23 @@
...
+ln -s "${DIR}/please" "${DIR}/plz"
+export PATH="$DIR:$PATH"
+
+# Start the server in the background
+plz run parallel -p -v notice --colour --detach //tools/http_cache:run_local
+
+# Test we can rebuild plz itself.
+plz test --profile localcache -p -v notice --colour //src/...
+
+# Run that again to make sure cache restoration works
```
peterebden, line 22: Do you want a clean or smthn in between?

Tatskaari: Ah that's a point. Clean doesn't clean the http cache so this should help actually use it.

Issue #645, opened 2019-06-07
Is there a way to bootstrap without network access?


dmgk:
Is there a way to vendor all 3rd party dependencies to avoid requiring network access during bootstrapping? We are working on FreeBSD port and the build currently fails with

```
1 error occurred:
        * Get https://github.com/kevinburke/go-bindata/archive/46eb4c183bfc1ebb527d9d19bcded39476302eb8.zip: dial tcp: lookup github.com on 1.0.0.1:53: write udp 127.0.0.1:61403->1.0.0.1:53: write: permission denied

Build stopped after 40ms. 1 target failed:
    //third_party/go:_go-bindata-go-bindata#download
1 error occurred:
        * Get https://github.com/kevinburke/go-bindata/archive/46eb4c183bfc1ebb527d9d19bcded39476302eb8.zip: dial tcp: lookup github.com on 1.0.0.1:53: write udp 127.0.0.1:61403->1.0.0.1:53: write: permission denied
```

because network access is not allowed during build phase.

swills:
I believe other package build systems have this restriction too.

peterebden:
There isn't at present, I suppose we could think about a way of generically overriding `remote_file` rules to be able to look into a local directory possibly. I very much don't want to vendor dependencies into this repo.

How does one usually handle this for Go in FreeBSD? Typically `go build` etc will make network requests as needed to pull in dependencies?

swills:
Normally, the resources will be fetched, checksums checked and then placed in the appropriate locations before the build starts.

swills:
This is still an issue.

swills:
That sounds fine, the orchestration on our side isn't a problem, it's fairly easy to do really. I didn't know you could build with `go build src/please.go` but I will look into that. (I was building with `go run -tags bootstrap src/please.go --profile bootstrap build //src:please`)

swills:
Building with `go build src/please.go` solved my problem. Thanks for educating me and sorry for the noise! So I'm wondering about these other tools. I wanted to try the `curl https://get.please.build | bash` install, but it reports: `Unknown operating system FreeBSD
` despite [the quickstart](https://please.build/quickstart.html) saying that FreeBSD is supported.

peterebden:
Yeah we don't do binary builds because it's on a different CI platform which doesn't do the releases. I don't _think_ it would be too hard to add that though (it'd just have to assume Circle updated the latest_release file).

swills:
[...]

PR #1135, opened 2020-08-18
Allow local remote files

peterebden:
Allow `remote_file` to load local files prefixed by `file://`. This addresses one aspect of #645 (although it will not necessarily cover everything since not all `go_get` rules go this way). It also happens to be something I will find useful shortly...


Tatskaari:
I guess we also need to re-write remote files if we want bootstrapping to work without an internet connection

peterebden:
Yeah you could _theoretically_ use this to solve that issue but it won't cover everything (and as you mention in it, would need quite a bit of orchestration on the client side). OTOH it is light enough a change and IMO slots fairly logically into the existing rule.

PR #1138, opened 2020-08-19
Add generated rules bindata to codebase

Tatskaari:
As well as a unit test to make sure they are up to date. This should allow building a complete please binary with `go build src/please.go`.

Should help with #645


.plzconfig.bootstrap
```diff
@@ -1,13 +1,2 @@
-; Used during initial bootstrap, mostly to get rules loaded.
```
peterebden, line 1: I think maybe you can totally remove this file (and the references to `--profile  bootstrap`, can you give that a go?

peterebden (approved):

Issue #710, opened 2019-08-20
Document Intellij command


barlock:
I'm starting to set up please with my project, in poking around the cli I found `plz ide intellij` which is great as thats my IDE. I can't figure out what it does or why I'd use it other than the cli help which is somewhat helpful.

```
[intellij command arguments]
  labels:                    Targets to include.
```

If you can describe it, perhaps I can help doc it wherever it makes sense.

agenticarus:
Hi @barlock - I wrote this feature so I'm happy to try and give some more context around it.

The intention was to provide a command that would set up an IntelliJ project structure for the given targets. Unfortunately, Go and Python were too difficult to support initially so it currently only works for Java and Scala targets.

Once you run `plz ide intellij [some java targets]` you can then open `./plz-out/intellij` in your IDE and you should find that modules have been created for all the targets and third party deps (you may need to `plz build` the third_party jar targets once). Unfortunately something seems screwy with the auto-refreshing and you may need to do a manual sync if you make large changes to your structure.

If you add a label of `package_prefix:[some package]` to your target (see [[test/intellij/src/build/please/intellij/BUILD]] for an example) then IntelliJ will automatically add that to the module definition.

There are a few wrinkles
 - IntelliJ doesn't do well with multiple modules in the same directory, so I recommend having only one java_library() target per directory. This still ends up with lots of modules, so you may find it easier to just have a single java_library() at the top level of your source tree, in a style more like Maven than Plz.
- I haven't tested it with code generation targets (like Protobuf etc) so those targets are unlikely to compile at present.

If you find any bugs, please let us know!

Tatskaari:
We've talked about this internally and this feature was never really finished. It doesn't behave like people expect and requires a lot of work to get right. We would rather approach this from the other side and have third parties support the IDEs of their choice. For example, write a please plugin for intellij. 

I'm going to remove this feature from please and close this issue when I have done so.

PR #1141, opened 2020-08-20
Remove plz ide intellij

Tatskaari:
This feature was never really completed and never worked as users expected it to. If we come back to this we will probably approach it the other way: Write a please plugin for intellij. 

cc: #710


peterebden (approved):
Aw :( 

Good to see you embracing the light and admitting that IntelliJ isn't very good though, I knew I'd persuade you eventually ;-)

macripps:
Sad times, this worked exactly as _I_ expected it to!

Tatskaari:
@macripps Me so sorry. It'll come back in some form or another. The ultimate goal would be to have a IJ plugin however maybe this can be moved to pleasings in the meantime.

PR #1143, opened 2020-08-21
Implement retries to the HTTP cache with go-retryablehttp

### LINKED ISSUE #1140: Add retry to HTTP cache
I've recently started to use a HTTP cache for please as part of moving some of my personal builds to using Gitlab CI. I've noticed that the HTTP cache doesn't seem to handle errors well - for example, if there are timeouts on reading or writing to the cache, or unexpected EOFs because the cache connection breaks for whatever reason.

I'm wondering if it would be good to have a retry for the HTTP cache on any sort of error - that way, when a connection timeout or EOF (or any other weird error) occurs, please doesn't just assume the files don't exist in the cache and then goes on to rebuild the entire tree.

Context: Right now, for Gitlab CI, I have an rclone service acting as a sidecar so that it can proxy webdav calls to S3. Ideally, it might also be good to have an S3-compatible cache as well, but retries would be good in general, so I figure that's probably the better thing to tackle here.

(I'd be happy to send an initial PR for this if this sounds like a good idea, but I wanted to run the idea by here first.)

robxu9:
A number of builds that I was doing rely on a somewhat unreliable
webdav-S3 server bridge, where occasionally timeouts happen or a file
cuts of in transit due to one error or another. Instead of immediately
pretending a file doesn't exist (or never uploading a file), retry
a certain number of times first in a controlled manner.

Closes #1140

Signed-off-by: Robert Xu <me@robxu9.com>

### DIFF
--- a/src/core/config.go
+++ b/src/core/config.go
@@ -245,6 +245,7 @@ func DefaultConfiguration() *Configuration {
 	config.Cache.HTTPWriteable = true
 	config.Cache.HTTPTimeout = cli.Duration(25 * time.Second)
 	config.Cache.HTTPConcurrentRequestLimit = 20
+	config.Cache.HTTPRetry = 4
 	if dir, err := os.UserCacheDir(); err == nil {
 		config.Cache.Dir = path.Join(dir, "please")
 	}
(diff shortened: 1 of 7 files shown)


src/core/config.go
```diff
@@ -381,6 +384,10 @@ type Configuration struct {
 		HTTPWriteable              bool         `help:"If True this plz instance will write content back to the HTTP cache.\nBy default it runs in read-only mode."`
 		HTTPTimeout                cli.Duration `help:"Timeout for operations contacting the HTTP cache, in seconds."`
 		HTTPConcurrentRequestLimit int          `help:"The maximum amount of concurrent requests that can be open. Default 20."`
+		HTTPRetry                  int          `help:"The maximum number of retries before a request will give up, if a request is retryable"`
+		HTTPRetryWaitMin           cli.Duration `help:"The minimum amount of time a retryable request will wait before retrying."`
+		HTTPRetryWaitMax           cli.Duration `help:"The maximum amount of time a retryable request will wait before retrying."`
+		HTTPLinearJitterRetry      bool         `help:"Uses linear backoff with jitter instead of exponential backoff. Minimum and Maximum should be changed if you select this."`
```
peterebden, line 390: This seems like quite a few parameters and particularly the linear jitter one is quite specific to the implementation. I'd be tempted to put them to sensible defaults and just expose the number of retries as a parameter for now.

peterebden (commented):
Thanks!

src/core/config.go
robxu9: Sounds good to me, I'll send a new commit. 👍

peterebden (approved):
LGTM!

PR #1154, opened 2020-08-24
Draft: Implemented import path for go rules 

Tatskaari:
We need to think about how to roll this out.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -17,7 +17,7 @@ notice "Building Please..."
 go run src/please.go $PLZ_ARGS build //src:please --log_file plz-out/log/bootstrap_build.log
 # Use it to build the rest of the tools that come with it.
 notice "Building the tools..."
-plz-out/bin/src/please $PLZ_ARGS build //package:installed_files --log_file plz-out/log/tools_build.log
+plz-out/bin/github.com/thought-machine/please/src/please $PLZ_ARGS build //package:installed_files --log_file plz-out/log/tools_build.log
 
 if [ $# -gt 0 ] && [ "$1" == "--skip_tests" ]; then
     exit 0
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -84,9 +84,9 @@ if [ ! -d "/usr/lib/gcc/x86_64-linux-gnu/${GCCVER%.*.*}/32" ] && [ ! -d "/usr/li
     EXCLUDES="${EXCLUDES} --exclude=x86"
 fi
 
-plz-out/bin/src/please $PLZ_ARGS ${PLZ_COVER:-test} $EXCLUDES --exclude=e2e --log_file plz-out/log/test_build.log --log_file_level 4 --trace_file plz-out/log/trace.json $@
+plz-out/bin/github.com/thought-machine/please/src/please $PLZ_ARGS ${PLZ_COVER:-test} $EXCLUDES --exclude=e2e --log_file plz-out/log/test_build.log --log_file_level 4 --trace_file plz-out/log/trace.json $@
 
 # We run the end-to-end tests separately to ensure things don't fight with one another; they are
 # finicky about some things due to running plz recursively and disabling the lock.
 notice "Running end-to-end tests..."
-plz-out/bin/src/please $PLZ_ARGS ${PLZ_COVER:-test} $EXCLUDES --include=e2e --log_file plz-out/log/e2e_build.log --log_file_level 4 $@
+plz-out/bin/github.com/thought-machine/please/src/please $PLZ_ARGS ${PLZ_COVER:-test} $EXCLUDES --include=e2e --log_file plz-out/log/e2e_build.log --log_file_level 4 $@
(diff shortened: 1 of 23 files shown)


sagikazarmark:
Is this change going to work for `go_library` rules for third-party dependencies?

Like `third-party/go/github.com/gorilla/mux/gorilla/BUILD`:

```
go_library(
    name = "gorilla",
    srcs = [], # downloaded third-party dependencies
)
```

Tatskaari:
> Is this change going to work for `go_library` rules for third-party dependencies?
> 
> Like `third-party/go/github.com/gorilla/mux/BUILD`:
> 
> ```
> go_library(
>     name = "mux",
>     srcs = [], # downloaded third-party dependencies
> )
> ```

I sense that this solution isn't flexible enough. I'm seeing how allowing complete control over the output location of a rule rather than just having it as the package prefix might work. Right now if you're in a please package `//foo/bar` and you have an `output_location` of `github.com/some/thing`, you'll output into `plz-out/gen/github.com/some/thing/foo/bar`. I don't think this is necessarily what you want all the time. I feel like a complete decoupling of output location and please project structure might be needed to get golang to really play ball. 

You can see how this breaks down a bit in `//third_party/go/zip`.

bootstrap.sh
```diff
@@ -17,7 +17,7 @@ notice "Building Please..."
 go run src/please.go $PLZ_ARGS build //src:please --log_file plz-out/log/bootstrap_build.log
 # Use it to build the rest of the tools that come with it.
 notice "Building the tools..."
-plz-out/bin/src/please $PLZ_ARGS build //package:installed_files --log_file plz-out/log/tools_build.log
+plz-out/bin/github.com/thought-machine/please/src/please $PLZ_ARGS build //package:installed_files --log_file plz-out/log/tools_build.log
```
peterebden, line 20: I don't think we should change the output location for binaries. I don't think it particularly fixes much (versus libraries, where it fixes a lot) but it's also going to be quite disruptive (whereas generally users care a lot less about library output locations)

sagikazarmark:
I rather meant the hardcoded `output_location = CONFIG.GO_IMPORT_PATH,` inside the `go_library` rule. It's probably not what we want for third-party libraries with different import paths.


And I agree, in case of third-party rules, the actual go output should be the proper go import path of the package, it should not include `third_party/go`.

As an alternative to this, I could imagine a solution that the go build tool does: because of caching, it doesn't actually save the compiled archives into a unified location, but generates a map for each dependency and the archive file. Not sure if it would be easier to implement, but it would certainly make the output location less relevant.

sagikazarmark:
The go build tool does something like this:

```
# generate import config
cat >$WORK/b138/importcfg << 'EOF' # internal
# import config
packagefile bytes=$WORK/b003/_pkg_.a
packagefile errors=$WORK/b004/_pkg_.a
packagefile io=$WORK/b013/_pkg_.a
packagefile unicode/utf8=$WORK/b018/_pkg_.a
EOF

# compile
/usr/local/go/pkg/tool/darwin_amd64/compile -importcfg $WORK/b123/importcfg
```

Something like this could work:

```
def go_importcfg(deps:list=[]):
    cmd = [
        'cat >$OUT << 'EOF' # internal',
        '# import config',
    ]

    for d in deps:
        if has_label(d, "go_archive"):
            import_path = get_metadata(d).import_path
            cmd += [f'packagefile {import_path}=$(location {d})']

    cmd += ['EOF']

    return build_rule(
        cmd = '\n'.join(cmd)
    )
```

But this is probably a different conversation, output location could be quick solution done first.

rules/go_rules.build_defs
```diff
@@ -15,7 +15,8 @@ _LINK_PKGS_CMD = ' '.join([
 def go_library(name:str, srcs:list, asm_srcs:list=None, hdrs:list=None, out:str=None, deps:list=[],
                visibility:list=None, test_only:bool&testonly=False, complete:bool=True,
                _needs_transitive_deps=False, _all_srcs=False, cover:bool=True,
-               filter_srcs:bool=True, _link_private:bool=False, _link_extra:bool=True, _abi:str=False):
+               filter_srcs:bool=True, _link_private:bool=False, _link_extra:bool=True, _abi:str=False,
+               go_import_path:str=CONFIG.GO_IMPORT_PATH):
```
sagikazarmark, line 19: How about calling it just `import_path`? The `go_` part seems redundant, since this is a go rule.

Tatskaari:
> I rather meant the hardcoded output_location = CONFIG.GO_IMPORT_PATH, inside the go_library rule. It's probably not what we want for third-party libraries with different import paths.

Oh yeah, that's just me being lazy in this draft. I've updated it now though I'm sure I've missed something. Just wanted to see what you guys thought of this approach in general before I get it working.

sagikazarmark:
Makes sense. Personally, I really-really want to use this ASAP one way or another. 😄 

That being said, it would probably make sense to explore the import config solution I outlined above. Using the metadata it would probably be somewhat cleaner and more please-like, since you wouldn't need this whole output location thing.

Tatskaari:
I've made good progress with the import config option and as Mark says, that feels much more compatible with how please works. I'm going to close this PR.

PR #1156, opened 2020-08-25
Add more string utils to asp

Tatskaari:


Tatskaari:
Peter got there first

sagikazarmark:
TBH `trim` sounds better than `remove` to me.

PR #1160, opened 2020-08-28
Removing -s flag to patch to enable patching in busybox

RichardoC:
The "-s" flag is not recognised by the patch cli included in busybox so removing it


Tatskaari (approved):
I guess at worst we might end up with some guff in the stdout of this rule under some circumstances but that seems like a low cost to pay for better portability.

PR #1161, opened 2020-08-29
Allow staying in the current directory when running a target

### LINKED ISSUE #1159: plz run in subdirectories
`plz run` is currently incapable of running a target in a subdirectory of a project. The working directory will always be the root.

While there are cases where this is definitely an advantage, in other cases it's a limitation.

I'm not sure why please works this way, so it might be historical or a deliberate decision, but it would be nice to be able to run targets in subdirectories.

WDYT?

sagikazarmark:
This is an attempt to fix #1159

I'm not particularly a fan of the `StayInDir` name, so I'm open to ideas.

Sadly, because of the nature of how please changes the directory, the implementation isn't really nice either, let me know if you want me to change it.

Last, but not least: In addition to "staying in the current directory", we could add a dir parameter as well that lets the user define which directory to change to. WDYT?

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -155,9 +155,10 @@ var opts struct {
 	} `command:"cover" description:"Builds and tests one or more targets, and calculates coverage."`
 
 	Run struct {
-		Env      bool `long:"env" description:"Overrides environment variables (e.g. PATH) in the new process."`
-		Rebuild  bool `long:"rebuild" description:"To force the optimisation and rebuild one or more targets."`
-		Parallel struct {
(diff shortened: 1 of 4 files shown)


Tatskaari (approved):
Nice one! Thanks Mark!

sagikazarmark:
@Tatskaari Thanks!

TBH, I was hoping we could improve the naming before merging, but I'm happy to provide another PR if we find a better name. `stay_in_dir` doesn't sound good to me, it was only a working title. WDYT?

PR #1168, opened 2020-09-03
Rename stay_in_dir to in_wd

### LINKED ISSUE #1164: Find a better name for "StayInDir"
I was thinking about a better name for the option introduced in #1161

Here is what I came up with:

- `--no_cd`
- `--without_cd`
- `--in_dir`
- `--in_pwd`

I'm in favor of the last two. I guess it would make sense to fix this before tagging a new release to avoid breaks. Let me know what you think.

sagikazarmark:
Fixes #1164


Tatskaari:
Thanks Mark!

PR #1171, opened 2020-09-03
Fix staying in directory

sagikazarmark:
Unfortunately the feature introduced in #1161 is non-functional: changing the directory breaks the run command, as the command is resolved relatively to the repo root. Somehow it slipped by me, sorry about that.

Here is a fix that seems to solve the problem: calculate the absolute path of the executable.

Could use some help with testing. Also, I'm not sure if conceptually this is the right solution.

### DIFF
--- a/src/core/command_replacements.go
+++ b/src/core/command_replacements.go
@@ -59,6 +59,8 @@ var locationsReplacement = regexp.MustCompile(`\$\(locations ([^\)]+)\)`)
 var exeReplacement = regexp.MustCompile(`\$\(exe ([^\)]+)\)`)
 var outExeReplacement = regexp.MustCompile(`\$\(out_exe ([^\)]+)\)`)
 var outReplacement = regexp.MustCompile(`\$\(out_location ([^\)]+)\)`)
+var absOutExeReplacement = regexp.MustCompile(`\$\(abs_out_exe ([^\)]+)\)`)
+var absOutReplacement = regexp.MustCompile(`\$\(abs_out_location ([^\)]+)\)`)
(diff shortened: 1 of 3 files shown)


sagikazarmark:
Test fails because `TempDir` was added in Go 1.15.

Do you want me to make it compatible with Go 1.14? Or can you upgrade Go in `thoughtmachine/please_ubuntu_alt`?

src/core/command_replacements.go
```diff
@@ -59,6 +59,8 @@ var locationsReplacement = regexp.MustCompile(`\$\(locations ([^\)]+)\)`)
 var exeReplacement = regexp.MustCompile(`\$\(exe ([^\)]+)\)`)
 var outExeReplacement = regexp.MustCompile(`\$\(out_exe ([^\)]+)\)`)
 var outReplacement = regexp.MustCompile(`\$\(out_location ([^\)]+)\)`)
+var absOutExeReplacement = regexp.MustCompile(`\$\(abs_out_exe ([^\)]+)\)`)
```
peterebden, line 62: Hmmmm, I'm not sure I like these becoming a replacement any build rule has access to. It seems like if it were used in an actual build rule it would only be doing bad things.
Can we just use `filepath.Abs` on the returned path in run_step.go?

peterebden (commented):
I'd really like to keep coverage on Go 1.14 until upstream drops support. It's probably not _that_ likely but nice to know we don't accidentally introduce a dependency in our Go rules.
Can we work around using `ioutil.TempDir` and deferring a deletion until that time comes please?

sagikazarmark:
Submitted a different PR with the requested changes, because it seemed easier.

PR #1172, opened 2020-09-03
Fix staying in directory

sagikazarmark:
Resubmitted #1171 in a different PR with @peterebden 's comments. It was easier than reverting the changes.

Fixes #1171


peterebden:
Thanks, looks good!

PR #1176, opened 2020-09-04
Write go.mod for non-filegroup targets too

### LINKED ISSUE #1173: go.mod is not created in plz-out
In theory, the following rule should result in Please creating a `go.mod` file in `plz-out`:

```
genrule(
    name = "go.mod",
    cmd = "",
    labels = ["go"],
)
```

But it's not created.

peterebden:
Resolves #1173

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -170,6 +170,11 @@ func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget, runR
 	var cacheKey []byte
 	var metadata *core.BuildMetadata
 
+	if target.HasLabel("go") {
+		// Create a dummy go.mod file so Go tooling ignores the contents of plz-out.
+		goModOnce.Do(writeGoMod)
+	}
+
 	if runRemotely {
 		metadata, err = state.RemoteClient.Build(tid, target)
 		if err != nil {


sagikazarmark:
@peterebden actually, I was wondering if this could be a config option.

I have a build step which does nothing but runs the linter, without any of the please cache being around. In order to run the linter, I have to manually build a go labeled rule first. With a config option, I could just run `plz lint` without building something first.

WDYT?

peterebden:
Could you solve that by labeling something in your `plz lint`?

I was a bit conservative on it since it only needs to happen once, and I assumed it normally would fairly soon for people doing Go things. 
I think a config option is a bit fiddly - I expect it would basically turn it on for all rules, in which case it's either on by default (and probably nobody cares to turn it off) or it's off by default but it's a bit subtle to find. I'd probably rather just do it always unless we have a case where it's a problem?

sagikazarmark:
Hm, I could label the download rule for the golangci-lint binary itself.

I agree that manually turning it on and off is a bit weird. I also had the idea to introduce a concept of "project type" which would implicitly enable generating that file, but I realized a project might not really have a single type.

I guess labeling go tools with the go label makes sense. Another thing we could do is creating a go.mod in plz-out if there is a go.mod in the project root, but that won't cover all the cases either (eg. monorepo with go submodules). :\

I guess there is no perfect solution. Thanks for this PR anyway, it'll help anyway. Any chance you could release another beta version?

PR #1177, opened 2020-09-04
Change the import config to use exported deps

Tatskaari:


rules/go_rules.build_defs
```diff
@@ -108,8 +112,9 @@ def go_library(name:str, srcs:list, asm_srcs:list=None, hdrs:list=None, out:str=
             building_description = 'Packing...',
             requires = ['go'],
             labels = labels,
-            provides = provides,
+            provides =  {'go': ':' + name, 'go_src': lib_rule},
             test_only = test_only,
+            exported_deps = exported_deps,
```
sagikazarmark, line 117:
```suggestion
            exported_deps = exported_deps + deps,
```

Not really nice, but fixes #1181

sagikazarmark:
Not related to this PR directly, but this seems weird:

```
    cmd = f'echo packagefile {import_path}/{name}=$PKG_DIR/{out} > $OUT'
    if pkg.endswith(name):
        cmd = f'{cmd} && echo packagefile {import_path}=$PKG_DIR/{out} >> $OUT'
```

Why do we need to append the package name to the import path? If it's provided, it should just be the right import path IMHO.

Tatskaari:
> Not related to this PR directly, but this seems weird:
> 
> ```
>     cmd = f'echo packagefile {import_path}/{name}=$PKG_DIR/{out} > $OUT'
>     if pkg.endswith(name):
>         cmd = f'{cmd} && echo packagefile {import_path}=$PKG_DIR/{out} >> $OUT'
> ```
> 
> Why do we need to append the package name to the import path? If it's provided, it should just be the right import path IMHO.
> 
> This could actually be an issue if you have a package nested inside another with the same name.

To maintain the existing behavior (we used to symlink all the .a files back one directory) I will try and remove this for v16.

rules/go_rules.build_defs
Tatskaari: Nice spot! They don't need to be exported, they just need to be deps which is nice.

peterebden (approved):
I think this makes sense. One thing to check is that it doesn't explode the fan-out too much in terms of dependencies needed to build each action?

PR #1180, opened 2020-09-05
Add an Alpine image for building static Linux binaries

peterebden:
This should make the binaries more portable. Impact on size seems negligible so far.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -302,13 +302,13 @@ workflows:
             - build-linux-alt
       - build-freebsd:
           requires:
-            - build-linux
+            - build-alpine
       - test-rex:
           requires:
-            - build-linux
+            - build-alpine
       - test-http-cache:
           requires:
-            - build-linux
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -24,11 +69,10 @@ jobs:
            path: plz-out/results
        - run:
            name: Package
-           command: ./plz-out/bin/src/please build //package:all //tools/misc:gen_release //docs:tarball -p
+           command: ./plz-out/bin/src/please build //package:signer //tools/misc:gen_release //docs:tarball -p
```
Tatskaari, line 72: I assume this is intentional and we only need the signer for some reason.

peterebden: Yeah the tarballs etc used to get built by this job; they are now built by the alpine one instead.

Tatskaari (approved):
A lot of config I'm not _that_ familiar with but it seems sensible. Thanks! :)

PR #1182, opened 2020-09-06
fix: use output_dirs wildcard to avoid conflicts between packages

noxxious:
The introduction of output_dirs broke python installs where third party packages would share some folders inside (namespaced packages), i.e. Google Cloud Python packages have the internal structure of 'google/cloud/iam', 'google/cloud/tasks', etc. 
The bug was that with multiple packages like above installed they would claim the intermediate folders and hence would invalidate builds all the time, resulting in flaky builds depending on the order the packages were built.
Earlier, 'outs' parameter could be used to cherry-pick the files to collect, but no "outs" is not used anymore by "pip_library".

The change now goes recursively through the package pip output folder and collects the files only, resulting in specific file set rather than just folders. While this is rudimentary could also work for remote builds (as output_dirs are not void remote builds), but I haven't tested that.


Tatskaari:
Oh man that's frustrating. Some pip libraries are too large to use the /** approach. They have too many files and when we pass these around as $SRCS, we hit limits for the number of arguments that can be pass through via that environment variable. 

Does the old approach pre-output directory approach suffer the same problem? We could make pip_library use `out` instead of the output directory when it is specified.

noxxious:
The 'outs' approach works, but requires a lot of manual handcrafting and cherry-picking the files. Some examples here:
`pip_library(
    name = "google_cloud_pubsub",
    package_name = "google-cloud-pubsub",
    outs = [
         "google/cloud/pubsub_v1",
    ],
    version = "1.5.0",
)

pip_library(
    name = "google_cloud_firestore",
    package_name = "google-cloud-firestore",
    outs = [
         "google/cloud/firestore.py",
         "google/cloud/firestore_v1",
         "google/cloud/firestore_v1beta1",
    ],
    version = "1.7.0",
)`

I like the result of the output_dirs/** as this does give a give a more detailed and easier declaration (potentially also allowing remote builds).  But, yes $SRCS problem is a core design one.

As this is a only an intermediate target (for the resulting wheel), so building of pex is not affected.

The core issue here is during the srcs build target with multiple libraries involved they start owning the same gen output folders (and sometimes files). We could try and prefix the gen output of srcs target with something that is library specific, like name and then strip it during the wheel build step (as srcs is only intermediate target), but I could not think of a nice way of doing it. Btw, the wheel build step does not use $SRCS or $DEPS and jarcat just takes in whatever is in tmp build folder.

Tatskaari:
> The 'outs' approach works, but requires a lot of manual handcrafting and cherry-picking the files. Some examples here:
> `pip_library(
> name = "google_cloud_pubsub",
> package_name = "google-cloud-pubsub",
> outs = [
> "google/cloud/pubsub_v1",
> ],
> version = "1.5.0",
> )
> 
> pip_library(
> name = "google_cloud_firestore",
> package_name = "google-cloud-firestore",
> outs = [
> "google/cloud/firestore.py",
> "google/cloud/firestore_v1",
> "google/cloud/firestore_v1beta1",
> ],
> version = "1.7.0",
> )`
> 
> I like the result of the output_dirs/** as this does give a give a more detailed and easier declaration (potentially also allowing remote builds). But, yes $SRCS problem is a core design one.
> 
> As this is a only an intermediate target (for the resulting wheel), so building of pex is not affected.
> 
> The core issue here is during the srcs build target with multiple libraries involved they start owning the same gen output folders (and sometimes files). We could try and prefix the gen output of srcs target with something that is library specific, like name and then strip it during the wheel build step (as srcs is only intermediate target), but I could not think of a nice way of doing it. Btw, the wheel build step does not use $SRCS or $DEPS and jarcat just takes in whatever is in tmp build folder.

I wonder if having the srcs rule is even worth it. Perhaps we can just download and output the zip in one rule. We originally decided it was kinda nice to have the srcs in plz-out/gen so people can just see what's in there. 

Alternatively we could just output the srcs into `plz-out/gen/{pkg}/{name}/...` and handle this change in file structure in the wheel rule.

noxxious:
I guess any of these options is ok as the gen output is not used (apart from the rules debug). For any normal python operations, like IDE's autocompletion I still have to setup a proper virtualenv.

Tatskaari:
I think we can solve the ide problem by linking pip libraries to `plz-out/python-env` or the like. 

Anyway, I think we can just download the sources to `plz-out/gen/{pkg}/{name}/...`. Do you want to do that? If not I don't mind doing it tomorrow.

noxxious:
I'll do it since have a repo to test as well. The python-env is a good idea but requires a bit more work in getting proper layout of a virtual env and is for another PR.

noxxious:
@Tatskaari I  think I did it. Had to namespace to plz-out/gen/{pkg}/{name}/{version} as this is more error and future prone. Also as bonus  fixed some python dependencies (had those solved in my repo anyway).

Alos noticed  that pip_library wheels are including all the transient wheels, which is not necessary as the py_binary and py_test pex rules merge the dep wheels linked in their tmp, which is all the deps anyways. This reduced pex size 3x for me, probably also improving the execution performance a bit.

Tatskaari:
@noxxious Nice! Let me have a look. In the meantime it looks like the version of tensorflow you've chosen isn't available for linux (yet?)

rules/python_rules.build_defs
```diff
@@ -336,7 +336,7 @@ def pip_library(name:str, version:str, hashes:list=None, package_name:str=None,
       version (str): Specific version of the package to install.
       hashes (list): List of acceptable hashes for this target.
       package_name (str): Name of the pip package to install. Defaults to the same as 'name'.
-      outs (list): List of output files / directories. Defaults to [name].
+      outs (list): List of output files / directories. Defaults to [name]. If subdirectory is used outs[0] is used as the prefix.
```
Tatskaari, line 339: Does it make sense to just use the rule name? That way `//third_party/pythong:foo` always outputs to `plz-out/gen/third_party/python/foo/...`

.gitignore
```diff
@@ -12,3 +12,4 @@ __pycache__
 *.iml
 /venv
 /out
+tools/jarcat/zip/*.zip
```
Tatskaari, line 15: What was creating this?

Tatskaari (approved):
Looks good to me! Just a couple queries.

.gitignore
noxxious: jarcat tests :)

rules/python_rules.build_defs
```diff
@@ -370,7 +370,7 @@ def pip_library(name:str, version:str, hashes:list=None, package_name:str=None,
         else:
             repo_flag = '-f ' + repo
 
-    target = pip_target_dir
+    target = f"{pip_target_dir}/{name}/{version}"
```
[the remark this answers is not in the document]
noxxious: yes, on the other hand you're right. If multiple versions are used that would have to reflect in the name.

.gitignore
Tatskaari: They create them in the src root? That's not good.

noxxious: yes, somehow I got the zips there. Let me try and reproduce.

rules/python_rules.build_defs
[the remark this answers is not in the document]
noxxious: As for outs[0], the outs parameter is not used really anymore, just for that only change. So a proper way would be to remove the the outs and add something like subdirectory_name. Or make use of outs as a way to further filter down what the package outputs so that it does not take everything in. That could be filtered through jarcat's include/exclude options if needed.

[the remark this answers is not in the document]
Tatskaari: I've created #1202 to do this for  v16

Tatskaari:
Thanks :)

Issue #1179, opened 2020-09-04
Subrepo tools


sagikazarmark:
I'm a huge fan of reusing tools across projects and making project tooling as standard as possible.

This is why I created my version of the pleasings repo: https://github.com/sagikazarmark/mypleasings

The repo comes with two kinds of things:

- Build rules for things like Helm charts
- A bunch of rules and targets for downloading binaries

In case of Helm, there is a `helm_lint` rule that lints a Helm chart.
It needs a Helm binary, which is downloaded in the same repo, using the `helm_binary` rule.

Relevant parts:

```starlark
CONFIG.setdefault("HELM_TOOL", "//tools/k8s:helm")

def helm_lint(name:str, visibility:list=None):
    """Lints a Helm chart.

    Args:
      name (str): Name of the rule.
      visibility (list): Visibility of the rule.
    """

    return gentest(
        name = name,
        test_cmd = f'"$(location {CONFIG.HELM_TOOL})" lint "$PKG_DIR" 2>&1',
        data = glob(["**"], hidden = True) + [CONFIG.HELM_TOOL],
        visibility = visibility,
        no_test_output = True,
    )
```

And the download target in `tools/k8s`:

```starlark
CONFIG.setdefault("HELM_VERSION", "3.3.1")

helm_binary(
    name = "helm",
    version = CONFIG.HELM_VERSION,
    visibility = ["PUBLIC"],
)
```

When running tests inside my pleasings clone, it works perfectly.

When I try to use it as a subrepo (as shown in [this example](https://github.com/sagikazarmark/please-issues/tree/master/subrepo-tools)), it complains:

```
❯ plz build
Build stopped after 360ms. 1 target failed:
    //tools/k8s:helm
//deploy/charts/hello-world:lint depends on //tools/k8s:helm, but the directory tools/k8s doesn't exist
```

I need to manually configure `HELM_TOOL` in my repo in order to make it work.

My aim is to create a rule, that can go with the "default" rule in the subrepo, but also gives the user a chance to override it.
I think this is somewher in line with your goals, to move tools out of the please distribution.

Is there a way I can do that?

Tatskaari:
package_name() should give you the package name of the current directory. You could set the config to `CONFIG.setdefault("HELM_TOOL", package_name() + ":helm")` but this assumes that they're both in the same package. It might be nice to have a subrepo_name() rule.

sagikazarmark:
Hm...they are not in the same repo, but I can play with relative paths.

[...]

PR #1190, opened 2020-09-07
Added subrepo name rule to interpreter

Tatskaari:
I wonder if cononicalise could be made to handle sub-repos as well. 

fixes: #1179

### DIFF
--- a/src/parse/asp/builtins.go
+++ b/src/parse/asp/builtins.go
@@ -44,6 +44,7 @@ func registerBuiltins(s *scope) {
 	setNativeCode(s, "join_path", joinPath).varargs = true
 	setNativeCode(s, "get_base_path", packageName)
 	setNativeCode(s, "package_name", packageName)
+	setNativeCode(s, "subrepo_name", subrepoName)
 	setNativeCode(s, "canonicalise", canonicalise)
 	setNativeCode(s, "get_labels", getLabels)
 	setNativeCode(s, "add_dep", addDep)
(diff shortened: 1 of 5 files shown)


src/parse/asp/builtins.go
```diff
@@ -578,6 +579,10 @@ func packageName(s *scope, args []pyObject) pyObject {
 	return pyString(s.pkg.Name)
 }
 
+func subrepoName(s *scope, args []pyObject) pyObject {
+	return pyString("///" + s.pkg.SubrepoName)
```
sagikazarmark, line 583: Note that `package_name` does not. Not sure if that's an inconsistency you want to care about, but it's worth noting.

Also, what if there is no subrepo name?

Ideally, I would use it like this: `subrepo_name() + package_name()` or `subrepo_name() + "//tools/k8s:helm"`.

If I want it to work in both subrepos and non-subrepos, I have to check if `subrepo_name() == "///"`.

sagikazarmark (commented):
> I wonder if cononicalise could be made to handle sub-repos as well.

I think that would be great. If `canonicalise("//tools/k8s:helm")` would return `//tools/k8s:helm` in the same repo and `///subrepo//tools/k8s:helm` when used as a subrepo, that would be the best.

src/parse/asp/builtins.go
Tatskaari: Haha my excuse is that this was a last minute PR at 7pm on a Monday.

sagikazarmark: No worries! Thanks a lot!

sagikazarmark (approved):

Issue #920, opened 2020-04-19
Make java_binary self_executable by default


peterebden:
This has some ancient historic context relating to some weird failure we had internally way back when. It is not a logical default - the default should be self-executable in the same way a python_binary or go_binary is.

This also means we can get rid of the special casing around `$(exe //java:binary)` stuff which has always been a bit gross.

PR #1197, opened 2020-09-09
Make java binary self executable

Tatskaari:
Gated behind a feature flag

Fixes: #920

### DIFF
--- a/src/core/config.go
+++ b/src/core/config.go
@@ -504,7 +504,7 @@ type Configuration struct {
 
 	FeatureFlags struct {
 		JavaBinaryExecutableByDefault bool `help:"Makes java_binary rules self executable by default. Target release version 16." var:"FF_JAVA_SELF_EXEC"`
-	}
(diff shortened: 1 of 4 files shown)


src/core/config.go
```diff
@@ -501,6 +501,10 @@ type Configuration struct {
 	PleaseLocation string
 	// buildEnvStored is a cached form of BuildEnv.
 	buildEnvStored *storedBuildEnv
+
+	FeatureFlags struct {
+		JavaBinaryExecutableByDefault bool `help:"Makes java_binary rules self executable by default. Target release version 16." var:"FF_JAVA_SELF_EXEC"`
+	}
```
peterebden, line 507: you should add some help on the struct too so that `plz halp featureflags` has a bit of explanation.

Might be worth mentioning that these generally won't remain around very long relative to other config options (i.e. they're typically enabling "preview" features for the next major).

peterebden (approved):

PR #1198, opened 2020-09-09
Make subrepo_name() work in the context of a subinclude

Tatskaari:
Doing a beta release so Mark can play around with this and see if it works an intended. 

I've also added some logging around interpreter errors so we can get a stack trace out if we have nil pointer problems.


src/parse/asp/errors.go
```diff
@@ -44,6 +45,7 @@ func fail(pos Position, message string, args ...interface{}) {
 
 // AddStackFrame adds a new stack frame to the given errorStack, or wraps an existing error if not.
 func AddStackFrame(pos Position, err interface{}) error {
+	log.Debugf("error parsing build file: %v \n%v", err, string(debug.Stack()))
```
peterebden, line 48: this will log a bunch of times (once per asp scope), maybe we should log this when the assert fails or whatever

peterebden (approved):
As discussed I'm a _little_ dubious about this since subrepo_name() behaves differently if called within a function of a subinclude than at the top level. I suppose it is strictly more useful though.

PR #1200, opened 2020-09-09
jarcat/unzip: don't attempt to create directories twice when untarring

### LINKED ISSUE #1199: jarcat fails to extract Tar archives containing directory members
For the following build target:

```
remote_file(
    name = "logstash-filter-verifier",
    out = "logstash-filter-verifier-extracted",
    extract = True,
    hashes = ["523e430c264b0501d12a92e39179286d2175d2be2324e197d11455ecc5b26b40"],
    licences = ["apache-2.0"],
    url = "https://github.com/magnusbaeck/logstash-filter-verifier/releases/download/1.5.1/logstash-filter-verifier_1.5.1_linux_amd64.tar.gz",
)
```

where `logstash-filter-verifier_1.5.1_linux_amd64.tar.gz` has the following contents:

```sh
$ tar tvf logstash-filter-verifier_1.5.1_linux_amd64.tar.gz
drwxrwxr-x magnus/magnus     0 2019-07-11 23:05 ./
-rwxr-xr-x root/root   6121442 2019-07-11 23:05 ./logstash-filter-verifier
-rw-r--r-- magnus/magnus 10173 2019-07-11 23:05 ./LICENSE
-rw-rw-r-- magnus/magnus 13163 2019-07-11 23:05 ./README.md
```

Please fails to build the target with the following error:

```
Build stopped after 560ms. 1 target failed:
    //third_party/binary:logstash-filter-verifier
Error building target //third_party/binary:logstash-filter-verifier: exit status 1
13:40:49.517 CRITICAL: Error extracting zipfile: mkdir logstash-filter-verifier-extracted: file exists
```

This appears to be because jarcat attempts to create directory members twice, and the second attempt (which uses `os.Mkdir()`) fails because a file already exists at the target path:

```go
		if err := e.makeDir(out); err != nil {
			return err
		}
		switch hdr.Typeflag {
		case tar.TypeDir:
			if err := os.Mkdir(out, 0755); err != nil {
				return err
			}
```

chrisnovakovic:
When extracting a Tar archive, don't attempt to create a directory at the target path if the Tar member being processed is a directory: `e.makeDir()` will have just created a directory at that path, so the subsequent call to `os.Mkdir()` with the same path will fail.

Fixes #1199.


chrisnovakovic:
(The new test file `tools/jarcat/unzip/test_data/xmlrunner.tar` is just `tools/jarcat/unzip/test_data/xmlrunner.zip` repackaged as a Tar archive.)

Tatskaari (approved):
Nice catch! I think this makes sense.

PR #1203, opened 2020-09-09
jarcat/unzip: create entire path when creating directory Tar members

chrisnovakovic:
Following #1200, `jarcat/unzip` no longer explicitly creates directories when processing members of Tar archives. This is problematic when extracting an archive containing an empty directory, since there is no future action that will cause that directory to be created (e.g. the creation of intermediate directories for a child file, since there are none).

Mostly revert #1200, and explicitly create directory Tar members using `os.MkdirAll()`, which avoids raising an error when the target path already exists (which is the problem #1200 intended to solve).


chrisnovakovic:
The new test files `tools/jarcat/unzip/test_data/edgecases.tar` and `tools/jarcat/unzip/test_data/edgecases.zip` have the following structure, specifically to test the scenario described in the commit message:

```
drwxrwxr-x root/root         0 2020-09-09 16:54 top_level_dir/
drwxrwxr-x root/root         0 2020-09-09 16:54 top_level_dir/empty_dir/
```

Tatskaari (approved):
Yeah this seems very safe. Thanks Chris.

PR #1205, opened 2020-09-10
Remove //:pleasings as a magic build rule

Tatskaari:
Gated behind a feature flag and with some plz init niceness


src/core/graph.go
```diff
@@ -79,6 +79,14 @@ func (graph *BuildGraph) Target(label BuildLabel) *BuildTarget {
 func (graph *BuildGraph) TargetOrDie(label BuildLabel) *BuildTarget {
 	target := graph.Target(label)
 	if target == nil {
+		// TODO(jpoole): This is just a small usability message to help with the migration from v15 to v16. We should
```
peterebden, line 82: Wouldn't this still fire afterwards if you created one called `pleasings`? or am I missing something?

src/parse/parse_step.go
```diff
@@ -166,6 +170,40 @@ func activateTarget(tid int, state *core.BuildState, pkg *core.Package, label, d
 
 // parsePackage performs the initial parse of a package.
 func parsePackage(state *core.BuildState, label, dependent core.BuildLabel, subrepo *core.Subrepo) (*core.Package, error) {
+	packageName := label.PackageName
```
peterebden, line 173: are we duplicating all this code?

src/please.go
```diff
@@ -215,6 +215,11 @@ var opts struct {
 				Options ConfigOverrides `positional-arg-name:"config" required:"true" description:"Attributes to set"`
 			} `positional-args:"true" required:"true"`
 		} `command:"config" description:"Initialises specific attributes of config files"`
+		Pleasings struct {
```
peterebden, line 218: would it be better to prompt interactively during `plz init` rather than having a separate subcommand?

Tatskaari: Ah yeah that's a good way to do this. I think this is still useful for migrating an existing repository though (once they see the warning above).

src/parse/parse_step.go
Tatskaari: Yeah, I think this makes the v16 PR very straight forward. Do you not like this approach?

src/core/graph.go
Tatskaari: This is only hit when:
1) the label they're looking for is within the pleasings sub-repo
2) we don't have any pleasings sub-repo configured
3) we're about to panic because of this

If we have a pleasings repo configured `_, ok := graph.subrepos["pleasings"]; !ok` would be false so we won't log this. 

I think this warning makes sense in that case?

peterebden: ah I see, yep, makes sense

src/parse/parse_step.go
peterebden: Just a bit worried about how long we're going to keep both side by side for. I guess we can remember to update both if we need to change one...

.plzconfig.ci
```diff
@@ -5,3 +5,8 @@ dircompress = true
 
 [buildconfig]
 ci = true
+
+; CI should disable all feature flags so we're actually testing the default distribution.
```
peterebden, line 9: I'm a little lost about why we're enabling it in the main config but disabling here. Won't that mean it is effectively disabled everywhere since all our PRs will have to pass CI?

peterebden (approved):

.plzconfig.ci
Tatskaari: Well I think the CI stuff should test the code our users are likely to run so should definitely have it disabled by default. It's enabled in the main config so when I'm testing locally, I'm testing my latest changes. 

Ideally we'd have one CI platform (linux-alt?) that has them enabled so CI tests both.

src/parse/parse_step.go
Tatskaari: Definitely a concern. Ideally we have tests that will make this obvious. Looking at the git history, this file is updated once every two or so months. That seems like it should be okay.

PR #1209, opened 2020-09-10
fix: filegroup_hash used shasum which was part of perl not coreutils

noxxious:
shasum is part of perl and it was not on the standard path on my machine (Manjaro/Arch).
In addition since the invocation was wrapped in a subshell and a string expansion the error was not detected by bash.

### DIFF
--- a/rules/misc_rules.build_defs
+++ b/rules/misc_rules.build_defs
@@ -284,8 +284,8 @@ def hash_filegroup(name:str, srcs:list=None, deps:list=None, exported_deps:list=
       requires (list): Kinds of output from other rules that this one requires.
     """
     cmd = 'mkdir _out && '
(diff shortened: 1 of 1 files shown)


rules/misc_rules.build_defs
```diff
@@ -283,11 +283,19 @@ def hash_filegroup(name:str, srcs:list=None, deps:list=None, exported_deps:list=
       test_only (bool): If true the exported file can only be used by test targets.
       requires (list): Kinds of output from other rules that this one requires.
     """
+    cmd = 'mkdir _out && '
+    cmd += 'SHA256CMD="$(command -v sha256sum || echo shasum)" && '
```
Tatskaari, line 287: sha256sum is part of coreutils right? If so, I don't think we need to have this switch.

Tatskaari (approved):

PR #1210, opened 2020-09-10
feat: add errorexit option to rule scripts execution for better control

noxxious:
Related to #1209.  Allows addressing some of the possible errors in the rule cmd's (but not all cases would be caught as this is bash anyway).


build_defs/plz_e2e_test.build_defs
```diff
@@ -10,6 +10,7 @@ def plz_e2e_test(name:str, cmd:str, pre_cmd:str=None, expected_output:str=None,
 
     def _e2e_test_cmd(cmd):
         profile_flag = '--profile ci' if CONFIG.get('CI') else ''
+        cmd='set +e; %s' % cmd
```
noxxious, line 13: This is to allow the e2e scripts to handle the errors.

sagikazarmark:
I'm currently writing an e2e rule for helm chart. It would be handy if I could make sure that the chart is uninstalled from the cluster, regardless if the test failed or not. Currently I can do that, but with this change, I won't be able to.

Any thoughts?

noxxious:
I guess, this does not cater for all the requirements, but most of the build rules are simple one/two liner scripts that might not necessary have the error handling in place (see #1209 as an example).

The options for the case of yours are similar to the please e2e testing rule and you can do `set +e` in the script also in your case I would think about the exit traps (i.e. https://www.putorius.net/using-trap-to-exit-bash-scripts-cleanly.html)

Tatskaari:
@noxxious Bit concerned that Mark has already noted that he's relying on the current behavior. We probably need a config option for this. I worry that this could be very impactful to a larger codebase and some users might not see the value vs. the cost of migrating their build rules. Saying that, this flag definitely seems more principled; we should make this the default in v16.

noxxious:
This is a suggestion as I think a stricter set of control on rules could be useful. Shell scripts are finicky in error handling.

src/remote/action.go
```diff
@@ -165,7 +165,7 @@ func (c *Client) buildTestCommand(target *core.BuildTarget) (*pb.Command, error)
 			},
 		},
 		Arguments: []string{
-			c.bashPath, "--noprofile", "--norc", "-u", "-o", "pipefail", "-c", commandPrefix + cmd,
+			c.bashPath, "--noprofile", "--norc", "-u", "-o", "pipefail", "-e", "-c", commandPrefix + cmd,
```
peterebden, line 168: I guess we should centralise this set of flags somewhere...

peterebden (commented):
I agree with Jon; I think we need a way of opting into this (given it won't modify hashes it's pretty difficult to test over a large repo).

I do like the idea though, it seems like the right direction in future. Happy to take on a bit of refactoring if you want.

src/remote/action.go
noxxious: I guess so, maybe BASH should be defined as tool as well ?

noxxious:
@peterebden I would appreciate you take it over, this was my hack at it (after noticing a bug on hash_filegroup), but this goes across multiple components that I'm not familiar with yet (c/c++ rules, e2e tests, remote builds). And I'm still learning go :).

As a general design approach I think the build tools should be very strict on catching errors by default (to help general cohort of users) and allow relaxing of that if for advanced ones in specific cases, like @sagikazarmark has. And shell scripts are finicky at this especially.

I guess for the hash_filegroup issue the best approach would the that any tools (even core ones have limitations, like 'cp' command on osx does not support some of linking options available in linux) used in the rules would be detected by plz and requested to be defined as in tool section, hence providing a clear input the 'toolchain' concept. But one can dream :)

src/remote/action.go
Tatskaari: I suppose we could have a `shell` config option (though setting this to anything other than bash would mean you can't use any of our built in rules). 

I can't quite see how we'd define bash as a tool because defining something as a tool (currently) means setting an bash env variable and we can't do that without bash.

peterebden:
> @peterebden I would appreciate you take it over

Sure - happy to :)

> I guess for the hash_filegroup issue the best approach would the that any tools (even core ones have limitations, like 'cp' command on osx does not support some of linking options available in linux) used in the rules would be detected by plz and requested to be defined in tool section, hence providing a clear input the 'toolchain' concept. But one can dream :)

You _can_ do this at present (you can put `tools = ["cp"]` and use that as normal), but we rarely bother for well-known things. It's not very flexible either - we couldn't choose between `shasum` or `sha256sum` for example. I guess we might potentially do more there at some point in the future.

peterebden:
OK I've implemented most of that - fair enough, in the end it was somewhat tricky to factor some of it given some of the packages and dependencies etc.

It's configurable globally as a default and per-target, right now it's all off by default but that should provide flexibility to opt into the new behaviour (and to opt out after we flip the default in the next major version).

Tatskaari (approved):
Looks sensible. Perhaps I can have a go at enabling this for please. We should make sure our built in rules work at least.

PR #1211, opened 2020-09-12
fix: pip_library labels not applied then zip safe

noxxious:
If the zip_safe=True then the resulting pip_library would not have any labels applied.


noxxious:
@Tatskaari I'm completely baffled by the failing builds or completely unrelated paths, can you have a look, since I cannot rerun the CI build due to permissions to verify if it's  not a flake.

Tatskaari:
Ah thanks for this. Let me have a look... you change looks pretty harmless so I doubt it's related.

noxxious:
Thanks

PR #1219, opened 2020-09-15
Make please go tests compatible with the working directory of `go test`

Tatskaari:
This should give a better migration story for people trying to migrate from go build  -> please


Tatskaari:
While this was a good proof to show that this config option works, we've decided not to change this in the please repo. I've opened #1226 which is this config option without the migration.

PR #1220, opened 2020-09-16
Make FreeBSD ignore python tests because compiling numpy/grpcio takes ages

Tatskaari:


peterebden (approved):
You could possibly also do this in the .cirrus.yml file.

Tatskaari:
> You could possibly also do this in the .cirrus.yml file.

Ah that's a better place to do this. Is there any reason to have it on the bootstrap as well?

PR #1222, opened 2020-09-16
Draft: Command replacement V2

Tatskaari:


src/core/command/parser.go
```diff
@@ -0,0 +1,75 @@
...
+	//TODO(jpoole): this one isn't documented
+	if keyword == "dir" {
+		cmd.tokens = append(cmd.tokens, dir(l))
+	}
+	if keyword == "hash" {
+		cmd.tokens = append(cmd.tokens, hash(l))
+	}
+
+	cmd.labels = append(cmd.labels, l)
+}
```
Tatskaari, line 75: This can be done via a map similar to how please.go works

PR #1224, opened 2020-09-17
Detect go importpath on plz init

sagikazarmark:
This PR helps users getting started with Go projects by automatically detecting the go import path for a module.

### DIFF
--- a/src/plzinit/init.go
+++ b/src/plzinit/init.go
@@ -1,13 +1,15 @@
 package plzinit
 
 import (
+	"bufio"
 	"fmt"
-	"gopkg.in/op/go-logging.v1"
 	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
+
+	"gopkg.in/op/go-logging.v1"
 
 	"github.com/thought-machine/please/src/assets"
 	"github.com/thought-machine/please/src/cli"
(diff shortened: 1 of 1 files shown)


Tatskaari:
I don't know if we should be including go config by default. I'm slowly making `plz init` interactive. We should probably make it prompt the user to ask then if they would like to add language support for go.

I've done something similar for the pleasings repo.

src/plzinit/init.go
```diff
@@ -1,14 +1,16 @@
...
+	"bufio"
 	"fmt"
-	"gopkg.in/op/go-logging.v1"
 	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
 
+	"gopkg.in/op/go-logging.v1"
```
Tatskaari, line 12: Thanks :)

Tatskaari (approved):

src/plzinit/init.go
sagikazarmark: Oops, that was unintentional. :)

Tatskaari: No no that's how we should be doing imports so it's good

sagikazarmark: Yeah, but I don't like these changes sneaking into other changes. I prefer at least separate commits.

PR #1225, opened 2020-09-17
Make please init golang config

Tatskaari:
I also want to make sure that if GOROOT is not set, that go is on the path that please uses (maybe prompting the user to add their goroot to the path)

### DIFF
--- a/src/plzinit/go.go
+++ b/src/plzinit/go.go
@@ -1,0 +1,79 @@
+package plzinit
+
+import (
+	"bufio"
+	"fmt"
+	"github.com/thought-machine/please/src/core"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/thought-machine/please/src/cli"
+)
+
+const goConfig = `
+[go]
+;gotool = ...
+;goroot - ... 
+`
+
+// golangConfig prompts the user and returns the go config to add to the .plzconfig
+func golangConfig(dir string) string {
+	goModule, moduleFound := findGoModule(dir)
+
+	if !moduleFound && !cli.PromptYN("Would you like to setup golang in this repository", false) {
+		return ""
+	}
+
+	config := goConfig
+
+	goOnPath, _ := core.LookPath("go", core.DefaultPath)
+	if goOnPath == "" {
+		fmt.Println("Warning: go is not found on the default path. Please configure gotool, or goroot under [go] in your .plzconfig")
+		fmt.Println("Please doesn't use the host machines $PATH variable. Instead the path is configured in .plzconfig with a default of")
+		fmt.Println("\n[build]")
+		fmt.Println("path=", strings.Join(core.DefaultPath, ":"))
+		fmt.Println("\nYou may also add {GOROOT}/go/bin to the path if you prefer")
+	}
+
+	var importPath string
+	var err error
+	if moduleFound {
+		importPath = goModule
+	} else {
+		importPath, err = cli.Prompt("Import path (i.e. go module)", "")
+	}
+	if err != nil {
+		os.Exit(1)
+		return ""
+	}
+	if importPath != "" {
+		if !moduleFound {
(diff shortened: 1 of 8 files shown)


src/plzinit/go.go
```diff
@@ -0,0 +1,67 @@
...
+
+	goModule, ok := findGoModule(dir)
+
+	if !ok && !cli.PromptYN("Would you like to setup golang in this repository", false) {
+		return ""
+	}
+
+	config := "\n[go]\n"
+
+	goRoot, err := cli.Prompt("Do you want to set GOROOT? Otherwise please will use `go env GOROOT`", os.Getenv("GOROOT"))
```
sagikazarmark, line 26: I'm not sure this is right. `GOROOT` != `go env GOROOT`. As we discussed earlier, `GOROOT` does not have to be set, the go tool can just figure it out.

Also, a custom `GOROOT` is probably a local config, chances are that in CI or on other machines it will be different.

Personally, I would not prompt for setting `GOROOT`. If it comes from the environment, Please already handles that well (fixed in #1195). Otherwise it should come from the go tool, thus should not be set at all.

src/plzinit/go.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		os.Exit(1)
+		return "" // most likely ctr-c
+	}
+	if goRoot != "" {
+		config += fmt.Sprintln("goroot = ", goRoot)
+	}
+
+	importPath, err := cli.Prompt("import path/module", goModule)
```
sagikazarmark, line 35: Hm, why does this have to be interactive. I don't see a reason why the user would NOT want to use the one in `go.mod`. Maybe only ask this question, if a module cannot be detected?

src/plzinit/go.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+// golangConfig prompts the user and returns the go config to add to the .plzconfig
+func golangConfig(dir string) string {
+	//TODO(jpoole): make this check that go is in the please path
+	//TODO(jpoole): detect if we're in a go project and suggest they enable GO_BUILD_COMPAT to help migration
+
+	goModule, ok := findGoModule(dir)
+
+	if !ok && !cli.PromptYN("Would you like to setup golang in this repository", false) {
```
sagikazarmark, line 20:
```suggestion
	if !ok && !cli.PromptYN("Would you like to setup Go in this repository", false) {
```

src/plzinit/go.go
sagikazarmark: Maybe set this in `.plzconfig.local`? But that feels unnecessary too.

sagikazarmark (commented):
TBH, I'm not sure I'm totally happy with adding interactive steps. I prefer running `plz init` without having to fill out a questionnaire, Please guessing as much values as possible (eg. Go import path).

Maybe make the interactive mode opt-out? I see why interactive mode might be the default, but I will probably not use it if I don't have to.

sagikazarmark:
Also, a quite nice terminal prompt library: https://github.com/AlecAivazis/survey

I use it in my CLI projects, works quite nicely.

src/plzinit/go.go
Tatskaari: > As we discussed earlier, GOROOT does not have to be set, the go tool can just figure it out.

It's how we find the golang SDK. If GOROOT it not set, we look in `$(go env GOROOT)/go` for the `.a`s

Very valid points though. I think I should just check that `$(which go)` is somewhere in the `path` config option and ask if they would like to add it.

src/plzinit/go.go
Tatskaari: Go point

src/plzinit/go.go
sagikazarmark: Forgot about the stdlib. That's valid.

But even then, the current goroot "detection" from the environment should be fine. Hard coding it in the config makes the project less portable.

peterebden (approved):

Issue #1075, opened 2020-07-15
Please should descend into the package directory when running go test


sagikazarmark:
The go tool executes test binaries inside the package directory, which lets tests use resources relative to the package path. Please on the other hand executes the test binary in the test run root. I think it should descend into package directory and run from there.

My use cases:

- read a file from a higher directory passed as data to the test
- read files from `testdata` in the same package passed as data to the test

WDYT?

sagikazarmark:
From `go help testflag`: 

> When 'go test' runs a test binary, it does so from within the
> corresponding package's source code directory. Depending on the test,
> it may be necessary to do the same when invoking a generated test
> binary directly.

Tatskaari:
Sounds like a reasonable idea. I've had annoyances where test data is loaded differently by my IDE and go tools vs. how please actually runs the test. I'd probably class this as a breaking change though so probably will have to be done for please 16.0.0. 

What do you think @peterebden?

peterebden:
This would be a seriously breaking change, yes. I have thought this before but I'm not very enthusiastic since it makes Go tests work differently to other languages.

Maybe we could have some way of opting into this or a config option? That would make it a lot easier in terms of an upgrade path.

sagikazarmark:
> since it makes Go tests work differently to other languages.

Well, it makes it work differently to regular Go tests, which means the test code is not compatible with both please and the Go toolkit at the same time, which sounds worse to me.

A global option sounds like an acceptable solution (I doubt anyone would want to control this per rule)

Tatskaari:
Will consider this for v16

sagikazarmark:
🎉

PR #1226, opened 2020-09-17
Added go build compatibility

Tatskaari:
Fixes: #1075

#1219 migrates please to use this config setting


src/core/config.go
```diff
@@ -429,6 +429,7 @@ type Configuration struct {
 		CgoCCTool     string `help:"Sets the location of CC while building cgo_library and cgo_test rules. Defaults to gcc" var:"CGO_CC_TOOL"`
 		FilterTool    string `help:"Sets the location of the please_go_filter tool that is used to filter source files against build constraints." var:"GO_FILTER_TOOL"`
 		DefaultStatic bool   `help:"Sets Go binaries to default to static linking. Note that enabling this may have negative consequences for some code, including Go's DNS lookup code in the net module." var:"GO_DEFAULT_STATIC"`
+		GoBuildCompat bool   `help:"Changes the behavior of the build rules to be more compatible with go build i.e. please will descend into the package directory to run unit tests so go test does." var:"GO_BUILD_COMPAT"`
```
peterebden, line 432: Can we think of a more specific name than this? It only does one aspect of go build compatibility (I guess `GoTestCompat` would be a bit closer)

peterebden (approved):

PR #1227, opened 2020-09-17
Fix plz_e2e_tests not passing PLZ_ARGS

Tatskaari:
This was breaking incrementality and causing the e2e tests to re-build java stuff because they weren't using the right profile.


peterebden:
Why wasn't the Java stuff already built by the time it got to the e2e tests?

peterebden (approved):

Issue #1212, opened 2020-09-12
Remove --nocache


peterebden:
I've spent more time trying to explain this flag than any other. Users expect it to behave like `plz build --rebuild` but it actually just disables caches (so it's actually equivalent to `-o cache.dir: -o cache.httpurl:`).

Since we now have equivalents like `--rerun`, let's save some trouble by removing it. Probably best to issue a warning with a suggestion about alternatives for a while.

Tatskaari:
I have found this kinda useful when working on please in the caching space however that's not a good reason to keep it. I can just comment out the code path.

peterebden:
Yeah, for cases like that I think using the `-o` flags is sufficient; it doesn't need to be a top-level flag given that's a fairly niche use case.

janhancic:
👍 this is a good change!

peterebden:
Bad bot!

janhancic:
plz stop stale bot

peterebden:
Surely we have done this by now...

peterebden:
Yes, in #1612 , so stale-bot  was _sort of_ right...

PR #1229, opened 2020-09-18
Log a warning when --nocache is passed

peterebden:
First step of #1212

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -72,7 +72,7 @@ var opts struct {
 
 	FeatureFlags struct {
 		NoUpdate           bool    `long:"noupdate" description:"Disable Please attempting to auto-update itself."`
-		NoCache            bool    `long:"nocache" description:"Disable caches (NB. not incrementality)"`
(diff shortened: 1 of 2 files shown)


src/please.go
```diff
@@ -72,7 +72,7 @@ var opts struct {
 
 	FeatureFlags struct {
```
Tatskaari, line 73: This is confusingly named. Feature flags are what I'm calling the things that gate breaking changes between releases. Might want to revisit this.

Tatskaari (approved):

src/please.go
peterebden: This was there first :-P

But agreed. It's only internal to this package so we can call it whatever we want.

Tatskaari: Oh okay, I assumed this affected `plz halp`

peterebden: This is the struct that defines the "Options that enable / disable certain features" section in `plz --help`, but the actual name of the struct doesn't appear there.

PR #1230, opened 2020-09-18
Fix rex not adding additional test outputs

Tatskaari:


src/remote/remote.go
```diff
@@ -480,33 +479,19 @@ func moveOutDirFilesToTmpRoot(target *core.BuildTarget, dir string) error {
...
-			if execErr == nil && err != nil {
-				return metadata, results, nil, err
-			}
-		}
+	metadata, ar, err := c.execute(tid, target, command, digest, target.TestTimeout, true, false)
+	if err != nil {
+		return metadata, err
 	}
-	return metadata, results, coverage, execErr
+	return metadata, c.client.DownloadActionOutputs(context.Background(), ar, target.TestDir(run))
```
Tatskaari, line 494: I'm wondering if c.execute returning err needs to be handled differently?

peterebden (approved):
Nice, looks cleaner

PR #1232, opened 2020-09-20
build: add an HTTP User-Agent header to all GET requests

sean-:
Fixes: #1231

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -983,13 +983,14 @@ func fetchRemoteFile(state *core.BuildState, target *core.BuildTarget) error {
 				Proxy: http.ProxyURL(state.Config.Build.HTTPProxy.AsURL()),
 			}
 		}
+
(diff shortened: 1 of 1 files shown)


sean-:
It would be great to see this in `15.3.1` because this effectively showed up as a sudden prod blocker for all use of `plz` at work.  We don't understand why this suddenly changed, but all of CI and users on our VPN are frozen out from using `plz` until this shows up in a released version of `plz`.

sean-:
fwiw, one thing we don't understand is why this works with `plz` 14.X but broke with 15.X.  We haven't bisected the problem to go back in time and figure out the change, but nothing stands out as obvious in the release notes.

Tatskaari:
Hi @sean-, I'm not sure what happened either. This change looks pretty straight forward. I'll merge this now and release 15.4.0 sometime tomorrow.

PR #1236, opened 2020-09-21
Cache lookup of missing blobs

peterebden:
A little experimentation suggests that this reduces blobs looked up to ~1/4 in the plz repo.

Only concern is memory usage since it will grow fairly continually...

There is a suspicious amount of similarity in the requests so it might do even better if we serialised the requests. I don't think that's a good idea given UploadIfMissing is also doing the uploads; potentially we could push this logic into the SDK and have that do it around just the FindMissingBlobs call.

### DIFF
--- a/src/remote/action.go
+++ b/src/remote/action.go
@@ -441,7 +441,7 @@ func (c *Client) uploadLocalTarget(target *core.BuildTarget) error {
 	for _, c := range m {
 		chomks = append(chomks, c)
 	}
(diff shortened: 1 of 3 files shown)


Tatskaari (approved):
Looks good to me! I don't think we'll have much issue with memory. It's only caching 64 bytes per src file and proto message. Even with 100k of these, that's only 6.4mb of ram.

PR #1239, opened 2020-09-22
Don't download empty blobs for stdout / stderr

peterebden:


src/remote/action.go
```diff
@@ -361,14 +361,14 @@ func (c *Client) buildMetadata(ar *pb.ActionResult, needStdout, needStderr bool)
...
+	if needStdout && len(metadata.Stdout) == 0 && ar.StdoutDigest != nil && ar.StdoutDigest.Hash != digest.Empty.Hash {
 		b, err := c.client.ReadBlob(context.Background(), digest.NewFromProtoUnvalidated(ar.StdoutDigest))
 		if err != nil {
 			return metadata, err
 		}
 		metadata.Stdout = b
 	}
-	if needStderr && len(metadata.Stderr) == 0 && ar.StderrDigest != nil {
+	if needStderr && len(metadata.Stderr) == 0 && ar.StderrDigest != nil && ar.StderrDigest.Hash != digest.Empty.Hash {
 		b, err := c.client.ReadBlob(context.Background(), digest.NewFromProtoUnvalidated(ar.StderrDigest))
```
Tatskaari, line 372: I feel like this can be done in ReadBlob

Tatskaari (approved):

src/remote/action.go
peterebden: It can be; https://github.com/bazelbuild/remote-apis-sdks/pull/184 would indeed do that.

PR #1241, opened 2020-09-22
Don't download empty blobs

peterebden:
Updates the SDK a bit with some new optimisations. Most of the code changes are resulting API changes from that.


Tatskaari (approved):

Tatskaari:
Looks like you have a couple other places you need to update

PR #1243, opened 2020-09-24
Implement entry points for tools

Tatskaari:
I think this is about ready to be merged. There're a few things that will have to follow: 

- Validate that the rule generated the entry points as output
- Make `plz run` entry point aware


sagikazarmark:
I'm trying to understand what this does, but I'm completely lost TBH.

Tatskaari:
> I'm trying to understand what this does, but I'm completely lost TBH.

I'm trying to make it so you can define a file within a rule as the "entry point" to that rule when you run it. 

Say you have the compiled go SDK that looks like:

```
$GOROOT/
- bin/go
- pkg/...
```

This will allow you to add 
```python
entry_points = {
   "go": "bin/go",
   "fmt": "bin/fmt",
}
```
to your rule and then add:

```python
tools = {
  "GO": "//tools:go_sdk|go",
  "LINT": "//tools:go_sdk|fmt",
}
```

to your rule

sagikazarmark:
Interesting, thanks for the explanation!

src/build/build_step.go
```diff
@@ -287,14 +285,13 @@ func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget, runR
 		if err != nil {
 			return err
 		}
-	}
-	// The remote action will set the output directory outs here
-	if !runRemotely {
```
peterebden, line 292: just checking this is deliberate? Not sure I see how it's related...?

src/core/build_target.go
```diff
@@ -201,6 +201,8 @@ type BuildTarget struct {
 	OutputDirectories []OutputDirectory `name:"output_dirs"`
 	// RuleMetadata is the metadata attached to this build rule. It can be accessed through the "get_rule_metadata" BIF.
 	RuleMetadata interface{} `name:"config"`
+	// EntryPoints represent named binaries within the rules output that can be targeted via //package:rule|entry_point_name
+	EntryPoints map[string]string
```
peterebden, line 205: add the `name:"entry_points"` tag, that will make `plz query print` show it nicely

src/remote/action.go
```diff
@@ -407,7 +407,7 @@ func outputsForActionResult(ar *pb.ActionResult) map[string]bool {
 // verifyActionResult verifies that all the requested outputs actually exist in a returned
 // ActionResult. Servers do not necessarily verify this but we need to make sure they are
 // complete for future requests.
-func (c *Client) verifyActionResult(target *core.BuildTarget, command *pb.Command, actionDigest *pb.Digest, ar *pb.ActionResult, verifyOutputs, isTest bool) error {
+func (c *Client) verifyActionResult(target *core.BuildTarget, command *pb.Command, actionDigest *pb.Digest, ar *pb.ActionResult, verifyRemoteBlobsExists, isTest bool) error {
```
peterebden, line 410: nit: `verifyRemoteBlobsExist`

test/build_defs/BUILD
```diff
@@ -9,3 +9,8 @@ filegroup(
     name = "build_defs",
     srcs = glob(["*.build_defs"]),
 )
+
+filegroup(
+    name = "free_bsd_config",
```
peterebden, line 14: nit: `freebsd_config`

peterebden (approved):

src/build/build_step.go
Tatskaari: Just a quick code simplification. We have an if statement above that checks runRemotely. I moved this code into the else.

Tatskaari: Yeah I just moved the `addOutputDirectoriesToBuildOutput(target)` into the else branch of the 
```
if runRemotely {
  ...
}
```

rather than having it in it's own if statement.

PR #1246, opened 2020-09-24
Add golangci-lint github action

peterebden:
This is adding https://github.com/golangci/golangci-lint-action; that should run golangci-lint much as we do now on Circle but will add appropriate comments into PRs.


Tatskaari (approved):
Cool :sunglasses:

PR #1247, opened 2020-09-24
Added test_tools to build_rule

Tatskaari:


src/parse/asp/builtins.go
```diff
@@ -149,15 +149,6 @@ func setLogCode(s *scope, name string, f func(format string, args ...interface{}
 // new objects to the build graph.
 func buildRule(s *scope, args []pyObject) pyObject {
 	s.NAssert(s.pkg == nil, "Cannot create new build rules in this context")
-	// We need to set various defaults from config here; it is useful to put it on the rule but not often so
-	// because most rules pass them through anyway.
-	// TODO(peterebden): when we get rid of the old parser, put these defaults on all the build rules and
```
peterebden, line 154: you sure this is safe? did we put the defaults in everywhere they're needed?

Tatskaari: They're on `def build_rule(...):` which everything uses underneath. I'm assuming that even though it's a built in it still respects default parameters.

Tatskaari: Ah noo It can be overriden if a caller has a default parameter like `licences = None`

peterebden: Let's definitely leave it out of this PR. This is a nontrivial feature in itself, better not to roll in unrelated stuff that could cause problems.

peterebden (approved):

PR #1248, opened 2020-09-25
Add an unused function to test linter

peterebden:
Not to be merged, just testing the new github action thingy.


peterebden:
Cool, looks like this is working.

PR #1252, opened 2020-09-28
Add PLZ_ARGS variable to pleasew

sagikazarmark:
When running please in a CI environment, copying the same arguments between scripts is error prone and tedious.

Setting a global `PLZ_ARGS` environment variable that's automatically in every invocation is a nice way around that. For example, setting plain output for every CI step is a good practice.

I realized this might not be the absolute best solution, but it's a start and can be the start of a discussion.


pleasew
```diff
@@ -69,4 +69,4 @@ for x in `ls "$DIR"`; do
 done
 ln -sf "${DIR}/please" "${LOCATION}/plz"
 echo -e >&2 "${GREEN}Should be good to go now, running plz...${RESET}"
-exec "$TARGET" "$@"
+exec "$TARGET" "$PLZ_ARGS" "$@"
```
peterebden, line 72: This is an error if it's undefined? (we set +u at the top of the script)

I think something like `"${PLZ_ARGS+}"` would help, but we also need to be sure it doesn't inject a spurious empty string into the arguments. Maybe just `${PLZ_ARGS+}` would do it?

sagikazarmark: Silly me, I haven't actually tried it **without** `$PLZ_ARGS` 😄

sagikazarmark: I think `${PLZ_ARGS:-}` will be the right syntax here.

sagikazarmark:
The breaking CI does not seem to be related.

Tatskaari:
> The breaking CI does not seem to be related.

I think you might just need to merge master in. I fixed the up to date failures yesterday.

sagikazarmark:
I rebased from master this morning.

Tatskaari (approved):

PR #1254, opened 2020-09-29
Adding aws sync to s3 website bucket

Tatskaari:
First step in migrating please.build from our LAMP + cloudflare stack to an aws stack.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,13 +69,14 @@ jobs:
            path: plz-out/results
        - run:
            name: Package
-           command: ./plz-out/bin/src/please build //package:signer //tools/misc:gen_release //docs:tarball -p
+           command: ./plz-out/bin/src/please build //package:signer //tools/misc:gen_release //docs:tarball //docs:deep-tarball -p
        - persist_to_workspace:
            root: plz-out/pkg
            paths:
              - gen_release.pex
              - release_signer
              - docs.tar.gz
+             - deep-docs.tar.gz
        - store_artifacts:
            path: plz-out/log
        - save_cache:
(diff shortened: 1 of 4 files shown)


docs/BUILD
```diff
@@ -82,6 +82,14 @@ tarball(
     labels = ["hlink:plz-out/pkg"],
 )
 
+tarball(
```
Tatskaari, line 85: The flat tarball was harder to deal with and was flattening the images into the web root for some reason.

.circleci/config.yml
```diff
@@ -270,6 +271,16 @@ jobs:
...
+   aws-sync:
+     docker:
+       - image: amazon/aws-cli
+     steps:
+       - attach_workspace:
+           at: /tmp/site
+       - run:
+           name: Upload docs
+           working_directory: /tmp/site
+           command: tar -xzf deep-docs.tar.gz && aws s3 sync docs s3://please-build && aws s3 sync linux_amd64 s3://please-build/linux_amd64 && aws s3 sync darwin_amd64 s3://please-build/darwin_amd64 && aws s3 sync freebsd_amd64 s3://please-build/freebsd_amd64
```
peterebden, line 283: You don't have all the GPG signatures etc here, those are generated in the release step

.circleci/config.yml
```diff
@@ -309,3 +320,9 @@ workflows:
           filters:
             branches:
               only: master
+      - aws-sync:
+          requires:
+            - build-linux
```
peterebden, line 325: this would require the darwin and freebsd builds as well, no?

.circleci/config.yml
Tatskaari: Ah right I see... I guess the current server just routes requests so it doesn't need to worry about that. Probably needs some thought. I guess I'll come back to the releases. We can get the docs updating first.

.circleci/config.yml
Tatskaari: Ah good catch! I've made this PR only implement syncing for docs now though.

peterebden (approved):
We'll need to sort out credentials in CircleCI presumably?

Tatskaari:
> We'll need to sort out credentials in CircleCI presumably?

I _think_ I've already done so (set some env vars as the docs suggest)

PR #1255, opened 2020-09-29
Fix python_wheel rule when wheel doesn't contain a subdirectory for t…

Tatskaari:
…he package

### DIFF
--- a/test/python_rules/pyparsing_test.py
+++ b/test/python_rules/pyparsing_test.py
@@ -1,0 +1,8 @@
+import unittest
+
+
+class PyparsingTest(unittest.TestCase):
+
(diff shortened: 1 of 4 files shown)


test/python_rules/pyparsing_test.py
```diff
@@ -0,0 +1,8 @@
+import unittest
+
+
+class NumpyTest(unittest.TestCase):
```
peterebden, line 4: NumpyTest -> PyparsingTest

peterebden (approved):

Tatskaari:
This breaks hashes... probably going to hold off on this until v16.

PR #1259, opened 2020-09-30
Support Git-formatted patches in Python build rules

chrisnovakovic:
Add support for patches generated with `git format-patch` in the `pip_library` and `python_wheel` build rules, which need to be applied with `patch -p1` rather than `patch -p0`.

### DIFF
--- a/rules/python_rules.build_defs
+++ b/rules/python_rules.build_defs
@@ -572,12 +572,33 @@ def python_wheel(name:str, version:str, hashes:list=None, package_name:str=None,
 def _patch_cmd(patch):
     """Returns a command for the given patch or patches, with OS-specific tweaks where needed."""
     patches = [patch] if isinstance(patch, str) else patch
+
+    patch_opts = '' if CONFIG.HOSTOS == 'freebsd' else '--no-backup-if-mismatch'
+
+    cmds = []
+    for patch in patches:
+        # If the patch header contains the magic Git date, this is a patch
+        # generated with `git format-patch`, and needs to be applied with
+        # `patch -p1`; otherwise, default to `patch -p0`
+        cmds += [
+            "if sed -n 1p $(location {patch}) | grep -q '^From [0-9a-f]\{{40\}} Mon Sep 17 00:00:00 2001$'; then "
+                'level=1; '
+                'else '
+                'level=0; '
+                'fi'.format(patch=patch)
+        ]
+
+        cmds += [
+            'patch -p$level {patch_opts} < $(location {patch})'.format(patch=patch,
+                                                                       patch_opts=patch_opts)
+        ]
+
     if CONFIG.HOSTOS == 'freebsd':
-        # --no-backup-if-mismatch is not supported, but we need to get rid of the .orig
-        # files for hashes to match correctly.
-        return patches, [f'patch -p0 < $(location {patch})' for patch in patches] + ['find . -name "*.orig" | xargs rm']
+        # --no-backup-if-mismatch is not supported on BSD patch but we need to
+        # get rid of the .orig files for hashes to match correctly.
+        return patches, cmds + ['find . -name "*.orig" | xargs rm']
     else:
-        return patches, [f'patch -p0 --no-backup-if-mismatch < $(location {patch})' for patch in patches]
+        return patches, cmds
 
 
 def _handle_zip_safe(name):


rules/python_rules.build_defs
```diff
@@ -572,12 +572,33 @@ def python_wheel(name:str, version:str, hashes:list=None, package_name:str=None,
...
+
+    patch_opts = '' if CONFIG.HOSTOS == 'freebsd' else '--no-backup-if-mismatch'
+
+    cmds = []
+    for patch in patches:
+        # If the patch header contains the magic Git date, this is a patch
+        # generated with `git format-patch`, and needs to be applied with
+        # `patch -p1`; otherwise, default to `patch -p0`
+        cmds += [
+            "if sed -n 1p $(location {patch}) | grep -q '^From [0-9a-f]\{{40\}} Mon Sep 17 00:00:00 2001$'; then "
```
Tatskaari, line 584: Why'd this date hard coded? 

```
jpoole [gaia] ~/xattr (add-supported-const) ♬  git format-patch master
0001-Added-constant-for-whether-xattrs-are-supported-by-t.patch
jpoole [gaia] ~/xattr (add-supported-const) ♬  cat 0001-Added-constant-for-whether-xattrs-are-supported-by-t.patch
From fce5c00c88adef11e5f39505018e47aab1c6fb12 Mon Sep 17 00:00:00 2001
From: jpoole <jpoole@thoughtmachine.net>
Date: Mon, 5 Oct 2020 13:21:15 +0100
Subject: [PATCH] Added constant for whether xattrs are supported by the
 current platform

---
 xattr_bsd.go         | 3 +++
 xattr_darwin.go      | 3 +++
 xattr_linux.go       | 3 +++
 xattr_unsupported.go | 3 +++
 4 files changed, 12 insertions(+)

diff --git a/xattr_bsd.go b/xattr_bsd.go
index 09db849..183730a 100644
```

I also seem to get a few extra lines. Don't we need to set level to match the preamble length?

peterebden (commented):
I am a bit nervous about the level of complexity here. I'd rather avoid this much shell complexity going into rules if possible (the pip ones are already bad enough from trying to work around all the other issues with that tool).

Also have some reservations against requiring `git format-patch`; the current scheme can be generated from `git diff` just fine (modulo some fiddling of paths, which I guess is what you want to avoid).

Maybe we should consider a more intelligent patching tool; I feel like I want something like [epatch](https://dev.gentoo.org/~gokturk/devmanual/head/ebuild-writing/functions/src_prepare/epatch/index.html), although I don't know what is out there and whether we want to assume its existence.

rules/python_rules.build_defs
chrisnovakovic: Patches written by `git format-patch` are in mbox format, and that specific date in the first line is used by Git to indicate that the patch was written by `format-patch`.

`-p` strips leading components from the paths of files to be patched, so `-p1` is the correct level for `git format-patch`-formatted patches (the leading `a/` and `b/` need to be stripped).

chrisnovakovic:
> Also have some reservations against requiring `git format-patch`; the current scheme can be generated from `git diff` just fine (modulo some fiddling of paths, which I guess is what you want to avoid).

Indeed - the main driver behind this was to make it easier to cherry-pick patches from upstream without having to rewrite them. I was going to do something similar for the Go rules (which are the only other ones in the default Please lexicon that have a `patch` parameter), but it seems they're already applied with `-p1`.

> Maybe we should consider a more intelligent patching tool; I feel like I want something like [epatch](https://dev.gentoo.org/~gokturk/devmanual/head/ebuild-writing/functions/src_prepare/epatch/index.html), although I don't know what is out there and whether we want to assume its existence.

Agreed - it'd be cleaner, simpler and more reusable to factor out the patching logic into a separate tool, so all rules could benefit.

Tatskaari:
I agree with Peter here I think. This is a lot of complexity to add to the build rules. Once you're patching third party deps you're pretty far down the rabbit hole. We can assume these are power users who're capable of formatting the patch. We should make sure this is documented correctly though. 

I'm going to close this PR off however if you strongly disagree then happy to revisit. :)

chrisnovakovic:
It turns out this wasn't even a good solution to the problem anyway because `git am` also has a `-p` option that I wasn't previously aware of, so the presence of the magic header isn't a guarantee that the patch can be applied with `-p1` anyway. I think @peterebden is right: consistently using something like epatch in all rules with patching functionality is probably the way to go.

PR #1264, opened 2020-10-01
Added release script that checks if the release exists already

Tatskaari:


.circleci/release_s3.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -e
```
peterebden, line 3: nit: would probably `set -eu` to ensure you don't misspell `$VERSION` anywhere...

peterebden (approved):

PR #1265, opened 2020-10-01
Adds plz fmt

peterebden:
This is still based on buildifier, although with a smol modification to support f-strings. Seems nicer to (1) support them and (2) have it all in the same tool rather than pointing to a totally different repo to get the canonical formatter.


src/format/fmt.go
```diff
@@ -0,0 +1,78 @@
...
+		return formatAll(utils.FindAllBuildFiles(config, core.RepoRoot, ""), rewrite, quiet)
+	}
+	ch := make(chan string)
+	go func() {
+		for _, filename := range filenames {
+			ch <- filename
+		}
+		close(ch)
+	}()
+	return formatAll(ch, rewrite, quiet)
```
Tatskaari, line 37: Why did you fan out here? Unless I'm mistaken there's still only one thread doing the formatting. Did you want to spin up some go routines?

docs/commands.html
```diff
@@ -315,6 +315,21 @@ <h2><a name="hash">plz hash</a></h2>
...
+
+    <p>Auto-formats existing BUILD files. You can either provide a list of files to reformat
+      or, if none are given, it will discover all BUILD files in the repository.</p>
+
+    <p>The <code>-w</code> flag rewrites existing files in-place; if not passed the
+      formatted version will be printed to stdout.</p>
+
+    <p>The implementation is currently based on a lightly modified version of
+      <a href="https://github.com/bazelbuild/buildtools">buildifier</a> which supports nearly
+      a superset of the same dialect, but lacks one or two features such as type annotations.<br/>
```
Tatskaari, line 330: Would be good to get annotations working too ;) 

We could then use this in build defs as well.

src/please.go
```diff
@@ -251,6 +252,14 @@ var opts struct {
 		} `command:"outputs" description:"Exports outputs of a set of targets"`
 	} `command:"export" subcommands-optional:"true" description:"Exports a set of targets and files from the repo."`
 
+	Format struct {
+		Quiet bool `long:"quiet" short:"q" description:"Don't print corrections to stdout"`
```
Tatskaari, line 256: I guess it's obvious that this returns a non-zero code that you can use for just doing linting?

Tatskaari (approved):
Nice!

src/format/fmt.go
peterebden: it's not fanning out, it's just changing from a slice to a channel to match what FindAllBuildFiles does

PR #1269, opened 2020-10-02
Allow query somepath to read from stdin

peterebden:
This lets you do more complex queries with it like `plz query changes --inexact | plz query somepath //thing/i/care:about -` which could be hard to answer before.

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -566,10 +566,11 @@ var buildFunctions = map[string]func() int{
 		})
 	},
 	"somepath": func() int {
-		return runQuery(true,
-			[]core.BuildLabel{opts.Query.SomePath.Args.Target1, opts.Query.SomePath.Args.Target2},
+		a := utils.ReadStdinLabels([]core.BuildLabel{opts.Query.SomePath.Args.Target1})
+		b := utils.ReadStdinLabels([]core.BuildLabel{opts.Query.SomePath.Args.Target2})
+		return runQuery(true, append(a, b...),
 			func(state *core.BuildState) {
-				query.SomePath(state.Graph, opts.Query.SomePath.Args.Target1, opts.Query.SomePath.Args.Target2)
+				query.SomePath(state.Graph, a, b)
 			},
 		)
 	},
--- a/src/please.go
+++ b/src/please.go
@@ -568,11 +568,12 @@ var buildFunctions = map[string]func() int{
 	"somepath": func() int {
 		a := utils.ReadStdinLabels([]core.BuildLabel{opts.Query.SomePath.Args.Target1})
 		b := utils.ReadStdinLabels([]core.BuildLabel{opts.Query.SomePath.Args.Target2})
-		return runQuery(true, append(a, b...),
-			func(state *core.BuildState) {
-				query.SomePath(state.Graph, a, b)
-			},
-		)
+		return runQuery(true, append(a, b...), func(state *core.BuildState) {
+			if err := query.SomePath(state.Graph, a, b); err != nil {
+				fmt.Printf("%s\n", err)
+				os.Exit(1)
+			}
+		})
 	},
 	"alltargets": func() int {
 		return runQuery(true, opts.Query.AllTargets.Args.Targets, func(state *core.BuildState) {
--- a/src/query/somepath.go
+++ b/src/query/somepath.go
@@ -1,14 +1,14 @@
 package query
 
-import "fmt"
+import (
+	"fmt"
 
-import "github.com/thought-machine/please/src/core"
+	"github.com/thought-machine/please/src/core"
+)
 
 // SomePath finds and returns a path between two targets, or between one and a set of targets.
 // Useful for a "why on earth do I depend on this thing" type query.
-func SomePath(graph *core.BuildGraph, from, to []core.BuildLabel) {
-	from = expandAllTargets(graph, from)
-	to = expandAllTargets(graph, to)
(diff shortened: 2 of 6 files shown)


src/query/somepath.go
```diff
@@ -4,60 +4,88 @@ import "fmt"
...
-	if label1.IsAllTargets() {
-		for _, target := range graph.PackageOrDie(label1).AllTargets() {
-			if querySomePath1(graph, target, label2, false) {
+func SomePath(graph *core.BuildGraph, from, to []core.BuildLabel) {
+	from = expandAllTargets(graph, from)
+	to = expandAllTargets(graph, to)
+	s := somepath{
+		memo: map[core.BuildLabel]map[core.BuildLabel]struct{}{},
+	}
+	for _, l1 := range expandAllTargets(graph, from) {
```
Tatskaari, line 15: Are we re-expanding from?

src/query/somepath.go
```diff
@@ -4,60 +4,88 @@ import "fmt"
...
+		for _, l2 := range expandAllTargets(graph, to) {
```
Tatskaari, line 16: And again for to here?

src/please.go
```diff
@@ -566,10 +566,11 @@ var buildFunctions = map[string]func() int{
 		})
 	},
 	"somepath": func() int {
-		return runQuery(true,
-			[]core.BuildLabel{opts.Query.SomePath.Args.Target1, opts.Query.SomePath.Args.Target2},
+		a := utils.ReadStdinLabels([]core.BuildLabel{opts.Query.SomePath.Args.Target1})
+		b := utils.ReadStdinLabels([]core.BuildLabel{opts.Query.SomePath.Args.Target2})
```
Tatskaari, line 570: nit: `from` and `to` would make this code more "self documenting" ;P

src/query/somepath.go
```diff
@@ -4,60 +4,88 @@ import "fmt"
...
+					fmt.Printf("  %s\n", l)
+				}
 				return
 			}
 		}
-		fmt.Printf("Couldn't find any dependency path between %s and %s\n", label1, label2)
-	} else {
-		querySomePath1(graph, graph.TargetOrDie(label1), label2, true)
 	}
+	log.Fatalf("Couldn't find any dependency path between those targets")
```
Tatskaari, line 26: Is this what we want to do here? I guess exit 1 can be useful so you can use this in bash if statements but it feels wrong to log that as a fatal error.  Maybe `SomePath(...)` should return an exit code.

src/query/somepath.go
```diff
@@ -4,60 +4,88 @@ import "fmt"
...
-	for _, target := range graph.ReverseDependencies(target2) {
-		if printSomePath(graph, target1, target) {
-			if target2.Parent(graph) != target {
-				fmt.Printf("  %s\n", target2.Label)
-			}
-			return true
+	return s.somePath2(target1, target2, m)
+}
+
+func (s *somepath) somePath2(target1, target2 *core.BuildTarget, seen map[core.BuildLabel]struct{}) []core.BuildLabel {
```
Tatskaari, line 65: This doesn't actually use `s` so maybe we can just make this `func somePath(...)`

src/query/somepath.go
```diff
@@ -4,60 +4,88 @@ import "fmt"
...
 }
 
-func querySomePath2(graph *core.BuildGraph, target1, target2 *core.BuildTarget, print bool) bool {
-	if !printSomePath(graph, target1, target2) && !printSomePath(graph, target2, target1) {
-		if print {
-			fmt.Printf("Couldn't find any dependency path between %s and %s\n", target1.Label, target2.Label)
-		}
-		return false
+type somepath struct {
+	memo map[core.BuildLabel]map[core.BuildLabel]struct{}
```
Tatskaari, line 45: memo?

perpetualjourney (approved):
Looks ok but I am :noideadog: about the please codebase so probs worth another tick from @Tatskaari

Tatskaari (commented):
Looks good, should make this command more useful!

src/query/somepath.go
peterebden: it's always best to call functions again just to be sure they've really done what they're meant to

src/query/somepath.go
peterebden: see above

src/please.go
peterebden: They don't actually have directionality though - you can pass the targets in either order

src/query/somepath.go
Tatskaari: I sense some trust issues. You should really work on that.

src/query/somepath.go
peterebden: because it's memoising?

src/query/somepath.go
peterebden: Still exiting with 1 but now just printing instead of logging (so I can leave most of the tests as they were).

That is a change but I think it's more useful to be able to tell the difference.

src/query/somepath.go
Tatskaari: Ah okay, didn't really clock that's the pattern you were following.

Tatskaari (approved):

src/query/somepath.go
peterebden: Yeah, it's potentially important to avoid following every possible path downwards through the graph for every target. Reduces the runtime from O(lots) to O(less).

PR #1271, opened 2020-10-02
fix: use $TOOLS_GO in cover test_cmd of go_test

sfirmery:
Allow using defined GoTool in go_test when using "plz cover".


Tatskaari (approved):
Ah nice catch! Thanks for that!

Issue #532, opened 2019-01-14
Support for go build flags...


sean-:
This is pretty simple but I couldn't figure it out:

```go
// +build foo

package main

func main() {
}
```

How do I pass in the `foo` build tag when building a Go library/binary, etc?

peterebden:
Hmmm good spot, we do seem to be missing that. How silly, we have the tool and it supports taking extra tags, they're just not plumbed through :) 
Should be easy to add.

sean-:
It seemed curious and like I must have been missing something or it isn't supported. :)

sean-:
Ick.  Hrm.  Could you add a visitor callback to the filegroup struct that reevaluates the filegroup everytime the list of srcs is queried?

This isn't a critical bug for us - I have removed all of our build tags for the time being and am relying on logic in `BUILD` instead.  This only flares up as an issue when someone `cd foo/bar && go build` and we have files with different build tags.  Currently we have different `BUILD` targets but we want to be able to `plz export` and have that workflow "just work."

numbsafari:
With the move to the go plugins, would it be appropriate to move this issue to the [please-build/go-rules](https://github.com/please-build/go-rules) repo?

In particular, I am trying to use the `go_repo` rules and have need to set build flags/tags (e.g. `noasm`) for certain modules. I'd be happy to open a fresh/specific issue there and simply refer back to this one if that makes more sense.

samwestmoreland:
Yeap I think it makes sense for this to be moved to go-rules if it hasn't been already. This may have already been implemented... or rather I made an attempt to implement it (in the go-rules repo), but saw something recently that suggested it wasn't working very well. I'll look into this.

PR #1280, opened 2020-10-06
Make please at least compile on windows

Tatskaari:


Tatskaari:
Can't get the build to pass until #532 is addressed

Tatskaari:
Closing this for now. Will pick this back up when we come back to this.

advancedwebdeveloper:
Good, I will test once there would be any improvements

hartmannr76:
I just want to make sure I understand the problem, so feel free to correct this if it's wrong:
Please uses itself to build please. Since the tool (not go compiler itself) doesn't correctly propagate go build flags for the compiler to pick up, it doesn't know that you want to include/exclude certain files based on those flags.

Does that sound right?

Tatskaari:
> I just want to make sure I understand the problem, so feel free to correct this if it's wrong:
> Please uses itself to build please. Since the tool (not go compiler itself) doesn't correctly propagate go build flags for the compiler to pick up, it doesn't know that you want to include/exclude certain files based on those flags.
> 
> Does that sound right?

It's able to pick up the correct files based on platform tags. The issue is more that Please just doesn't work on Windows right now because it assumes a lot about the way the underlying OS works. For example, we assume a single filesystem root, which breaks a few things. For example, when we set the path to be `c:/bin`, Please doesn't understand that. There's also some issues around process management. 

I think this is all very solvable however I'm worried it's going to open up a lot of surface area. Ideally I'd like to see some buy in from developers who're motivated to use Please on Windows in anger before trying to support it.

hartmannr76:
For sure and I think that's completely valid. In my mind, dev's who'd use a tool like this would likely not be building something specifically for Windows and would likely be able to get away with using this in WSL2 locally, and a Linux build system for their CI environments. 

I was tinkering with this on my Windows machine and really just wanted to make sure I understand the problem space before getting too deep. It had looked like the go build flags issue was a root problem but after that explanation, it sounds more like there need to be interfaces set between managing OS constructs.

Thank you for the explanation!

Tatskaari:
If you make even a little progress on this, we'd be very happy! We have long term plans to support windows as we believe that windows C++ developers are a very soft market for something like Please. Like I said, we're just a little bit worried about the overhead of providing first party support for a bunch of very different use cases. None of us really know how windows works to any real extent. 

With that being said, if you're motivated to give this a good go, I'd be more than happy to help out where I can :smile: Feel free to jump on our [gitter](https://gitter.im/please-build/Lobby), where I can try and guide you through our code base, and maybe help figure out what needs doing!

PR #1281, opened 2020-10-06
Fix update and add milestone message

Tatskaari:


src/update/update.go
```diff
@@ -48,6 +52,9 @@ var httpClient *retryablehttp.Client
 // forceUpdate indicates whether the user passed --force on the command line, in which case we
 // will always update even if the version exists.
 func CheckAndUpdate(config *core.Configuration, updatesEnabled, updateCommand, forceUpdate, verify bool, progress bool) {
+	httpClient = retryablehttp.NewClient()
+	httpClient.Logger = &utils.HTTPLogWrapper{Logger: log}
```
Tatskaari, line 56: This was being initted in the wrong place which meant `shouldUpdate()` was nil pointering

peterebden (approved):

PR #1282, opened 2020-10-07
Added milestone folder to docs website

Tatskaari:


docs/template.html
```diff
@@ -1,25 +1,25 @@
 <html>
   <head>
     <title>{{ .Title }}</title>
-    <link rel="stylesheet" type="text/css" href="styles.css">
```
Tatskaari, line 4: Had to make all these abs so this template works from `/milestones/*`

Tatskaari:
I've changed the margins to be absolute and added some max width so things look a little better on high res screens, It was hard top read the docs when the lines are the full width of a 4k screen. For the most part it looks the same at 1080p except the left margin is a little smaller. 

Index

Before: 
image

After:
image


Other pages:

Before:
image

After:
image

peterebden (approved):

PR #1283, opened 2020-10-07
Turn gzip compression off by default

peterebden:
Goodness knows why I defaulted this to true. It's a bad default because many servers don't support it.

This isn't a breaking change since all gRPC servers support no compression.


Tatskaari (approved):

peterebden:
Ignoring the failure, it's more Java toolchain stuff

PR #1290, opened 2020-10-09
fix: export to root directory

sagikazarmark:
Exporting outputs fails when the target directory is the repository root.

### DIFF
--- a/src/fs/fs.go
+++ b/src/fs/fs.go
@@ -90,8 +90,10 @@ func WriteFile(fromFile io.Reader, to string, mode os.FileMode) error {
 		return err
 	}
 	dir, file := path.Split(to)
-	if err := os.MkdirAll(dir, DirPermissions); err != nil {
-		return err
+	if dir != "" {
+		if err := os.MkdirAll(dir, DirPermissions); err != nil {
+			return err
+		}
 	}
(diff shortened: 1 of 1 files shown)


sagikazarmark:
This might not be a perfect fix (thinking about the temp file creation), but it's a fix and demonstrates the problem.

src/fs/fs.go
```diff
@@ -90,8 +90,10 @@ func WriteFile(fromFile io.Reader, to string, mode os.FileMode) error {
 		return err
 	}
 	dir, file := path.Split(to)
-	if err := os.MkdirAll(dir, DirPermissions); err != nil {
-		return err
+	if dir != "" {
```
Tatskaari, line 93: Interesting that filepath.Dir("thing") returns "." whereas this returns ""

Tatskaari:
Thanks Mark!

src/fs/fs.go
sagikazarmark: Yeah, it's strange, but at this point the value of `to` is actually just `filename` (given the command invoked is `plz export outputs -o . :filename`). I think `path.Split` would return `.` for `./filename`.

This is why I'm not convinced this is the perfect solution

PR #1292, opened 2020-10-09
Fix more typos

sagikazarmark:
Actually, it's rarely called "golang SDK", it's rather called Go (and stdlib).

Also, fixed a copy-paste typo


Tatskaari:
Ah yeah fair eough

PR #1293, opened 2020-10-09
Reinit wrapper script

### LINKED ISSUE #1260: Add a command for updating pleasew
The title is self-explanatory in my opinion. `pleasew` might change and users might want to keep updating them (especially when improvements around running and downloading please are added to it)

sagikazarmark:
Fixes #1260

It might make sense to add an interactive question before overwriting an existing file, but I can reason about the current behaviour too.


sagikazarmark:
Breaking test is unrelated to this PR, assets need to be regenerated.

Tatskaari (approved):

Issue #1215, opened 2020-09-13
Handling spaces in SRCS


sagikazarmark:
Spaces still cause issues when working with files in the `$SRCS` variable:

`SRCS=path to/file path to/other_file`

Shell will always handle the above as four separate paths instead of two.

There isn't an obvious solution I can think of, but the biggest problem in my opinion is not knowing the cause of the problem. For example, all you could see in such case is this error from `cp`:

```
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
```

Previous issues related to spaces: #338 #849

noxxious:
Isn't $SRCS built from exact filegroup/glob entries ? So it should be fairly easy to escape the spaces in each item, before joining them into $SRCS for invocation.

Tatskaari:
@noxxious You forget bash was designed by pre-humans before we developed a working frontal cortex. 

There's no way (to my knowledge) to export a variable with a space separated list where the items themselves can have spaces. Bash unhelpfully strips out your escaped spaces and replaces with unescaped ones:
```bash
export TEST="space folder/main.go" && cat $TEST # fails to find space and folder/main.go
export TEST=\"space folder/main.go\" # this one throws a syntax error
export TEST="\"space folder/main.go\"" && cat $TEST # same as first
export TEST=space\ folder/main.go && cat $TEST # same as first
```

The best I can really do is throw an exception if any of the file paths have spaces in them...

sagikazarmark:
Don't think this has been resolved

sagikazarmark:
Bump

PR #1302, opened 2020-10-15
Panic when srcs have spaces

Tatskaari:
Fixes: #1215

### DIFF
--- a/src/core/build_env.go
+++ b/src/core/build_env.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"encoding/base64"
+	"fmt"
 	"os"
 	"path"
 	"runtime"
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Doesn't this make quoting paths in build rules entirely useless? Chances are source files and output directories are in the same file tree.

I'm not arguing against this as a solution, but life would be easier if quoting was dropped everywhere and spaces in paths declared as unsupported. It's not even a best-effort thing because of the above reason.

Am I missing something?

Tatskaari:
> Doesn't this make quoting paths in build rules entirely useless? Chances are source files and output directories are in the same file tree.
> 
> I'm not arguing against this as a solution, but life would be easier if quoting was dropped everywhere and spaces in paths declared as unsupported. It's not even a best-effort thing because of the above reason.
> 
> Am I missing something?

Agreed. It needs more thought. Like perhaps we can do some sort of command replacement rather than using env variables.

sagikazarmark:
Bump

sagikazarmark:
Bump

PR #1305, opened 2020-10-16
Make glob() include hidden files in bazel compat mode

### LINKED ISSUE #1297: Document glob() compatibility with Bazel
**Edit**: I missed `hidden` parameter is available on glob(). Changed the aim for this issue. Fixing behavior in Bazel compatible mode is still an option.

During hacking with Please on a JavaScript project, I found all of `glob(['*'])`, `glob(['**'])` and even `glob(['.*'])` don't match dot-prefixed files like `.babelrc` and `.eslintrc`.

~~I'm not sure if it is intended or a bug, but it is undocumented and~~ differs from [Bazel implementation](https://docs.bazel.build/versions/master/be/functions.html#glob).

Changing this behavior can lead to problems like `.cache` folders to be included in deployment images or even to security incidents in case people save credentials as `.secret` file, so introducing a fix with a sort of opt-in (project wide flag in .plzconfig or explicit `.*` in patterns) can be good.

~~But if the fix would complicate things too much, I would still be happy with the current behavior if it is documented.~~

peterebden:
Also extend documentation to be a bit more explicit about the `hidden` argument.

I suppose this could technically be breaking so we might want to submit it in v16.

Fixes #1297

### DIFF
--- a/docs/config.html
+++ b/docs/config.html
@@ -630,7 +630,7 @@
     <p>Currently the only attribute here is <code>compatibility</code>, when that is enabled
       some aspects of the parser behave a little differently; some of the rule names and flags
       change, <code>//visibility:public</code> is interpreted specially and WORKSPACE files are
-      parsed to find external dependencies. The <a href="/lexicon.html#glob"><code>glob</code></a>
+      parsed to find external dependencies. The <a href="lexicon.html#glob"><code>glob</code></a>
       builtin also changes to include hidden files by default.<p>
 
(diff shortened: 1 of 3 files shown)


Tatskaari:
I don't think we need to v16 this imo. Should be pretty harmless.

Tatskaari (approved):

peterebden:
Cool, I'll yeet it in.

PR #1306, opened 2020-10-17
fix: match all conditions in config_setting

sagikazarmark:
According to the documentation above the config_setting rule,
it should match all conditions in values.

Bazel works the same way:
https://docs.bazel.build/versions/master/configurable-attributes.html


peterebden (approved):
Good spot, thanks!

PR #1309, opened 2020-10-18
Make the test output timings a bit clearer

peterebden:
Previously it can print things like
```
593 test targets and 5960 tests run in 9h46m2.957s; 5947 passed, 1 failed, 12 skipped. Total time 16m15.1s.
```
which I've heard comment is a bit confusing since it clearly didn't take 9 hours; that figure is more of a "total" than the one labelled "Total time".
This will print something like
```
593 test targets and 5960 tests run using 9h46m2.957s total compute time; 5947 passed, 1 failed, 12 skipped. Total real time 16m15.1s.
```
I think this is clearer as to what it means but a bit clunky still since it's wordier. Thoughts?

### DIFF
--- a/src/output/shell_output.go
+++ b/src/output/shell_output.go
@@ -275,12 +275,12 @@ func printTestResults(state *core.BuildState, failedTargets []core.BuildLabel, f
 			aggregate.Duration += target.Results.Duration
 			if len(target.Results.TestCases) > 0 {
 				if target.Results.Errors() > 0 {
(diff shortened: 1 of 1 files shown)


sagikazarmark:
How about:

```
593 test targets and 5960 tests run; 5947 passed, 1 failed, 12 skipped.
Total time: 16m15.1s real 9h46m2.957s compute.
```

When reading the summary, I'm more interested in the test results first, then additional info, like how much time it took.

peterebden:
Thanks - I like that better.

Tatskaari (approved):

PR #1310, opened 2020-10-19
Fix Cgo library import path

sagikazarmark:
This PR fixes the `cgo_library` tool: import path is now the correct package import path.

I also refactored the import path resolution logic into a separate rule for reuse.


Tatskaari (approved):

Tatskaari:
Ah nice refactor! Thanks Mark!

PR #1311, opened 2020-10-19
Refactor the docs heading a bit

Tatskaari:


Tatskaari:
The main thing I've done here is move some of the headings around and rename a few pages:

image

docs/dependencies.html
```diff
@@ -104,3 +104,84 @@ <h2>Comparison to other systems</h2>
   tool, but again only outputs a single package at a time. Writing up the dependencies can be
   eased by using something like <code>go list -f '{{.Deps}}' &lt;package&gt;</code> to discover
   all the dependencies for the package in question.</p>
+
+<h2>Verification</h2>
```
Tatskaari, line 108: I know you can add hashes to all rules so it's not specifically related to third party stuff however these seem closely related. Feels like a better home for this docco.

Tatskaari:
And the advanced topic section:

image

docs/lexicon.html
```diff
@@ -14,8 +14,9 @@ <h1>The Lexicon of Please</h1>
       <li><a href="#subrepo">Subrepo rules</a></li>
     </ul>
 
-    <h2><a name="builtins">Builtin functions</a></h2>
-
+    <h2><a name="builtins">Built in functions</a></h2>
+    <p>In addition to <a href="/language.html">the build language</a> built in functions, Please also has built
```
peterebden, line 18: "built in functions" is doubled, feels awkward.

maybe "Please also has a set of built in rules"?

peterebden (approved):

docs/cache.html
```diff
@@ -36,3 +38,8 @@ <h2>The HTTP cache</h2>
 
     <p>Thanks to Diana Costea who implemented the original version of this as part of her internship
       with us, and prodded us into getting on and actually deploying it for our CI servers.</p>
+
+    <p>A reference implementation of the http cache can be found
+      <a href="https://github.com/thought-machine/please/tree/master/tools/http_cache">here</a> however it should be
+      possible to use any off the shelf http server.</p>
```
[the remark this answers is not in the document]
Tatskaari: Ah yeah it explains that 2 paragraphs above. I guess I can say something like "however it should be
      possible to use any off the shelf http server with a little configuration, as described above."

PR #1314, opened 2020-10-19
Implement multiplication and division of integers

peterebden:
And multiplication of strings and lists with ints.


docs/grammar.txt
```diff
@@ -39,4 +39,4 @@ slice = "[" [ expression ] [ ":" expression ] "]";
 lambda = "lambda" [ lambda_arg { "," lambda_arg } ] ":" expression;
 lambda_arg = Ident [ "=" expression ];
 operator = ("+" | "-" | "%" | "<" | ">" | "and" | "or" | "is" | "is" "not" |
-            "in" | "not" "in" | "==" | "!=" | ">=" | "<=" | "|");
+            "in" | "not" "in" | "==" | "!=" | ">=" | "<=" | "|" | "*");
```
Manicben, line 42: Would it make more sense for this to be grouped with the other arithmetic operators?

peterebden: good point, done

src/parse/asp/objects.go
```diff
@@ -183,30 +183,43 @@ func (i pyInt) Property(name string) pyObject {
...
+		case GreaterThanOrEqual:
+			return newPyBool(i >= o)
+		case Modulo:
+			return i % o
+		case In:
+			panic("bad operator: 'in' int")
+		case Multiply:
+			return i * o
+		case Divide:
+			return i / o
```
Manicben, line 208: lel wonder if worth reordering this case statement too to be in line with grammar.txt/go

src/parse/asp/objects.go
```diff
@@ -272,6 +285,12 @@ func (s pyString) Operator(operator Operator, operand pyObject) pyObject {
 		return newPyBool(!strings.Contains(string(s), string(s2)))
 	case Index:
 		return pyString(s[pyIndex(s, operand, false)])
+	case Multiply:
```
Manicben, line 288: could also reorg here and below, but doesn't matter really

src/parse/asp/objects.go
```diff
@@ -183,30 +183,43 @@ func (i pyInt) Property(name string) pyObject {
...
+		case Modulo:
+			return i % o
+		case In:
+			panic("bad operator: 'in' int")
+		case Multiply:
+			return i * o
+		case Divide:
+			return i / o
+		}
+		panic("unknown operator")
```
jakechorleytm, line 210: Nit: _U_nknown operator

src/parse/asp/objects.go
peterebden: Sure - they don't really have an ordering as such, but why not

src/parse/asp/objects.go
peterebden: done

Tatskaari (approved):

Tatskaari:
Did we really not have this already?

src/parse/asp/objects.go
Manicben: There's seems to be a mix of capitalised panic messages and uncapitalised in this file btw. Would be good to stick to one or the other

src/parse/asp/objects.go
```diff
@@ -183,30 +183,43 @@ func (i pyInt) Property(name string) pyObject {
...
+		case LessThan:
+			return newPyBool(i < o)
+		case GreaterThan:
+			return newPyBool(i > o)
+		case LessThanOrEqual:
+			return newPyBool(i <= o)
+		case GreaterThanOrEqual:
+			return newPyBool(i >= o)
+		case Modulo:
+			return i % o
```
Manicben, line 206: :trollface: I would've put Modulo right below Divide, but not fussed.

Manicben (approved):

peterebden:
> Did we really not have this already?

Turns out we did not!

src/parse/asp/objects.go
peterebden: Given that the order doesn't matter in a functional sense I'm not going to move this one... :)

src/parse/asp/grammar.go
```diff
@@ -267,6 +267,10 @@ const (
 	Add Operator = '+'
 	// Subtract implements binary - (only works on integers)
 	Subtract = '-'
+	// Multiply implements multiplication between two types
+	Multiply = '×'
```
Manicben, line 271: wait a sec, shouldn't this be `'*'`?

peterebden: Nope these aren't used that way - they are just arbitrary constants, and for mild performance reasons it's a rune rather than a string. The map below provides a mapping of the written form to the enum, where the written form could be multiple characters (e.g. `>=`); if you look at the rest of the constants they use single-character versions (like `≥`).

PR #1315, opened 2020-10-19
Add language selector to show examples for 4 main langs

Tatskaari:


Tatskaari:
This is how it looks: 
image

docs/example-select.js
```diff
@@ -0,0 +1,19 @@
+function switchTab(event) {
+    let tab = event.target
```
peterebden, line 2: I guess `let` is widely enough supported now that it's fine for  us to use?

docs/example-select.js
```diff
@@ -0,0 +1,19 @@
...
+
+    let lang = tab.dataset.lang
```
peterebden, line 4: semicolons?

peterebden (approved):

docs/example-select.js
tabazevedo, line 2: nit: probably want to go with `const` unless you'll be reassigning these

docs/example-select.js
```diff
@@ -0,0 +1,19 @@
...
+
+    examples.querySelector('.tabs .selected').classList.remove("selected")
+    tab.classList.add("selected")
+
+    examples.querySelector('.examples .selected').classList.remove("selected")
+    examples.querySelector(`.examples > [data-lang="${lang}"]`).classList.add("selected")
+}
+
+document.addEventListener("DOMContentLoaded", function(event) {
+    document.querySelectorAll(".tabs > div").forEach(tab => {
```
tabazevedo, line 16: super nit: probably would want to give the children a `tab` class to make this selector less brittle but i imagine the markup for this won't change very frequently anyway. below you could then also switch to `> .tab` `> .tab:first-child` etc

docs/example-select.js
jakechorleytm: Widely supported at this point https://caniuse.com/?search=let const will be the same

docs/example-select.js
jakechorleytm: Plenty of style guides actively remove them in situations like these

docs/example-select.js
```diff
@@ -0,0 +1,19 @@
+function switchTab(event) {
```
tiagovtristao, line 1: This is highly coupled with this use case. I guess it's fine if this isn't supposed to be reused.

Tatskaari: I suppose i could actually name this stuff as "tabs" and have it be super generic?

docs/example-select.js
Tatskaari: Ah yeah, I guess I'm being lazy.

docs/example-select.js
Tatskaari: Ah we use them in action.js. I'll be consistent.

PR #1318, opened 2020-10-20
Reworked quickstart to be more guided

Tatskaari:


Tatskaari:
Nav thingy looks like
image

peterebden (approved):

PR #1321, opened 2020-10-21
fix: typo in go_toolchain rule

sfirmery:
Fix wrong var name in go_toolchain when using custom url.


Tatskaari:
Ah good catch! Thanks for this. Hope you're finding go_toolchains useful!

sfirmery:
Yes, really useful!

PR #1322, opened 2020-10-21
Cgo without cgo

sagikazarmark:
CGO is often not required for all platforms. This PR improves cross-compilation support by allowing to build regular go libraries when no cgo sources are provided (for example by using `select` to return an empty list on platforms not requiring cgo)


sagikazarmark:
Breaking test seems unrelated

Tatskaari (approved):
I guess creating a cgo library without any c sources is a bit weird but not a problem. Thanks for that!

Tatskaari:
> Breaking test seems unrelated

Yeah I think cirrus just hates us...

PR #1328, opened 2020-10-22
go_get: support multiple patches

chrisnovakovic:
Bring `go_get` into line with `pip_library` and `python_wheel` by allowing the `patch` parameter to define multiple patches in a list as well as a single patch in a string.


Tatskaari (approved):

Tatskaari:
Just need to bump xcode for the darwin build, can you merge master in once that done?

PR #1330, opened 2020-10-22
Fix a nil pointer when failing to parse test results

Tatskaari:
When we get a read error, we were returning a nil metadata which was subsequently blowing up.

### DIFF
--- a/src/test/test_step.go
+++ b/src/test/test_step.go
@@ -366,6 +366,9 @@ func doTestResults(tid int, state *core.BuildState, target *core.BuildTarget, ru
 
 	if runRemotely {
 		metadata, err = state.RemoteClient.Test(tid, target, run)
+		if metadata == nil {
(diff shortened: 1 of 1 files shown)


src/test/test_step.go
```diff
@@ -366,6 +366,9 @@ func doTestResults(tid int, state *core.BuildState, target *core.BuildTarget, ru
 
 	if runRemotely {
 		metadata, err = state.RemoteClient.Test(tid, target, run)
+		if metadata == nil {
```
peterebden, line 369: Should we just return a non-nil metadata from `Test`?

goddenrich (approved):

src/test/test_step.go
Tatskaari: Using the return value of something that errored seems like an anti-pattern no?

peterebden (approved):

PR #1333, opened 2020-10-22
go_get: allow per-target patches to be defined

### LINKED ISSUE #1332: go_get: multiple-target form applies patches to all targets
When the `patch` parameter is used with the multi-`get` version of `go_get`, Please attempts to apply the patch(es) listed in `patch` to each of the targets defined in `get`, which is guaranteed to fail because each target is a separate codebase:

```python
go_get(
    name = "go-test",
    get = [
        "github.com/a/b",
        "github.com/c/d",
    ],
    patch = "a_b.patch",    # Should only be applied to github.com/a/b
)
```

```
Build stopped after 120ms. 1 target failed:
    //third_party/go:_go-test#github.com_c_d
Error building target //third_party/go:_go-test#github.com_c_d: exit status 1
can't find file to patch at input line 16
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
<Contents of a_b.patch>
--------------------------
File to patch:
Skip this patch? [y]
Skipping patch.
2 out of 2 hunks ignored
```

chrisnovakovic:
In the `go_get` rule, allow the value of the `patches` parameter to be a dict whose keys are target names given in the `get` parameter and whose values are the patch(es) that should be applied to that target. Backwards compatibility is maintained: if `get` is a list and `patches` is a string or list, the patch(es) are applied to all targets in `get`.

Closes #1332.

### DIFF
--- a/rules/go_rules.build_defs
+++ b/rules/go_rules.build_defs
@@ -803,7 +803,7 @@ def cgo_test(name:str, srcs:list, data:list=None, deps:list=None, visibility:lis
 
 
 def go_get(name:str, get:str|list, repo:str='', deps:list=[], exported_deps:list=None,
-           visibility:list=None, patch:str|list=None, binary:bool=False, test_only:bool&testonly=False,
+           visibility:list=None, patch:str|list|dict=None, binary:bool=False, test_only:bool&testonly=False,
            install:list=None, revision:str|list=None, strip:list=None, hashes:list=None,
            extra_outs:list=[], licences:list=None, module_major_version:str|list=''):
     """Defines a dependency on a third-party Go library.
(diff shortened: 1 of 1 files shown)


rules/go_rules.build_defs
```diff
@@ -845,6 +847,26 @@ def go_get(name:str, get:str|list, repo:str='', deps:list=[], exported_deps:list
...
+        elif isinstance(value, dict):
+            v = value.get(key)
+            if not v:
+                return []
+            if isinstance(v, str):
+                return [v]
+            elif isinstance(v, list):
+                return v
+            else:
+                error(f"get_param_as_list: unexpected type %s" % type(v))
```
Tatskaari, line 866: Can you not recurse here? I guess not as it's a lamda?

Tatskaari (approved):

Issue #1340, opened 2020-10-26
Make `go_get()` more specific about its outputs


Tatskaari:
Imagine we have a rule:

```python
go_get(
   get = "github.com/me/my_module",
   install = ["foo", "bar/..."],
)
```

Currently this will produce `my_module` as an output however we can do better. Based on the install list we know it will produce a `foo.a` file and a `bar` folder i.e. this rule could have `my_module/foo.a` and `my_module/bar` as outs. 

This will allow us to have two rules that `get` from `my_module` and install different things. This would probably help resolve cyclic dependencies between modules.

We can make `go_get` inspect its install list. If the install ends in `...`, it should add `f"{name}/{install}"` as an out to the rule, otherwise it should add `f"{name}/{install}.a"`.

PR #1341, opened 2020-10-26
Make go_get() more specific about it's outs

Tatskaari:
Fixes: #1340


third_party/go/BUILD
```diff
@@ -661,6 +661,7 @@ go_get(
         "huff0",
         "snappy",
         "zstd/...",
+        "zstd",
```
Tatskaari, line 664: Small regression here. @peterebden I wonder if this means it should be a v16 thing?

peterebden: I suppose it should be.

Does feel a bit weird to have to specify both too. You wouldn't normally have to do that with `go get`. I can't see a good way around though and the added flexibility might be worth the tradeoff (especially if we had a way to generate the `go_get` invocations)

rules/go_rules.build_defs
```diff
@@ -897,10 +898,23 @@ def go_get(name:str, get:str|list, repo:str='', deps:list=[], exported_deps:list
             licences = licences,
             module_major_version = module_major_version,
         )
-        outs += [getroot]
+        get_roots += [getroot]
```
Tatskaari, line 901: This probably needs a better name... This has all the locations where we have `thing/...` because in that case we might have `thing.a` AND a folder `thing` containing additional outs.

third_party/go/BUILD
Tatskaari: I found a way arolund it. ;P

rules/go_rules.build_defs
```diff
@@ -913,10 +906,15 @@ def go_get(name:str, get:str|list, repo:str='', deps:list=[], exported_deps:list
             outs += [getroot]
         else:
             for i in installone:
-                if i.startswith(getroot):
-                    outs += [_install_output(getroot, i.removeprefix(getroot))]
+                if not i:
+                    outs += [getroot + ".a"]
+                elif i.endswith("..."):
```
chrisnovakovic, line 911: `i.endswith("/...")`, perhaps? Although it shouldn't really make any difference...

peterebden (approved):
Nice!

chrisnovakovic (approved):
🎉

Issue #1323, opened 2020-10-21
Platform specfic targets


sagikazarmark:
Certain targets are only required (and can only built on) for certain platforms. For example: certain dependencies might only be required by linux.

My current approach is an if, combined with a custom `select_config` rule:

```
if select_config([{
    "os": "linux",
    "cpu": "amd64",
}]):
    go_library(name = "logrus", srcs = [":_logrus#go_source"], cover = False, import_path = "github.com/sirupsen/logrus", visibility = ["PUBLIC"], deps = select({"//example/third_party/go/__config:darwin_amd64": [], "//example/third_party/go/__config:linux_amd64": ["//example/third_party/go/golang.org/x/sys/unix"], "default": []}))
```

The problem is: it's more generated code, and the buildifier package formats the target itself on a single line.

Is there a better way to represent platform specific targets? (It's not the end of the world, but it's hard to read and more changes in a VCS)

sagikazarmark:
Maybe something like this could work:

```
genrule(
    name = "platformtest",
    cmd = "echo 'asd'",
    labels = ["os:linux"],
)
```

Alternatively, I guess the `manual` label can be used to prevent it from being built automatically.

peterebden:
You could also write it like
```
if CONFIG.OS == "linux":
    go_library(
        ....
    )
```

sagikazarmark:
That's probably not enough as I also use the cpu/arch as a discriminator.

peterebden:
```
if CONFIG.OS == "linux" and CONFIG.ARCH == "amd64":
```
?

sagikazarmark:
And then the possibility of having several of these conditions (like build on all platforms, but linux_386). 😄 

I should probably add that this is generated code. The `select_config` rule I created actually works well, but the formatter formats the code like in the above example and overall, it's not really readable. Using labels would be a nice solution IMHO.

But if there isn't a better or already existing solution, I guess `select_config` is fine.

Tatskaari:
We could have a `platform()` rule that can be used to match a specific platform:

```
def platform(os : str =  CONFIG.OS, arch: str = CONFIG.ARCH, ...) -> bool: 
    return os == CONFIG.OS && arch == CONFIG.ARCH ....

if platform(os = "linux"):
   genrule(...)
```

Otherwise, we could label rules:

```
labels = ['os:linux', 'arch:amd64']
```

Then please could exclude them. Perhaps both are useful?

sagikazarmark:
[...]

PR #1342, opened 2020-10-27
Added is_platform() misc rule

Tatskaari:
A start on #1323

### DIFF
--- a/rules/misc_rules.build_defs
+++ b/rules/misc_rules.build_defs
@@ -567,8 +567,8 @@ def exports_files(srcs, name='exported_files', visibility=['PUBLIC'], licenses=N
     )
 
 
-def is_platform(os:str=CONFIG.OS, arch:str=CONFIG.ARCH, host_os:str=CONFIG.HOSTOS,
-                host_arch:str=CONFIG.HOSTARCH) -> bool:
+def is_platform(os:str|list=CONFIG.OS, arch:str|list=CONFIG.ARCH, host_os:str|list=CONFIG.HOSTOS,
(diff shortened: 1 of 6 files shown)


test/java_rules/java_toolchain/src/main/java/net/thoughtmachine/BUILD
```diff
@@ -1,4 +1,4 @@
-if CONFIG.OS == "linux" or CONFIG.OS == "darwin":
+if is_platform(os = "linux") or is_platform(os = "darwin"):
```
peterebden, line 1:
```
if is_platform(os = ["linux", "darwin"]):
```
WDYT?

Tatskaari: Sounds like it'd be more useful that way :D

rules/misc_rules.build_defs
```diff
@@ -565,3 +565,30 @@ def exports_files(srcs, name='exported_files', visibility=['PUBLIC'], licenses=N
...
+
+    https://github.com/golang/go/blob/master/src/go/build/syslist.go
+
+    Args:
+        os (str): The target operating system we are compiling for.
+        arch (str): The target CPU architecture we are compiling for.
+        host_os (str): The host operating system we are running on.
+        host_arch (str): The host cpu architecture we are running on.
+    """
+    def as_list(i):
```
peterebden, line 584: you could oneline all of this into a lambda ;-)

peterebden (approved):

Issue #1320, opened 2020-10-21
Syntax error on concatenating f-strings


suhr:
`error: unexpected token f"--set-version={major_version} ", expected )`

Context:

```
    cmd = (
        f"perl genpki.pl "
        "--include-path=src/include "
        f"--set-version={major_version} "
        f"{bki_sources}"
    ),
```

PR #1344, opened 2020-10-27
Fix implicit string concat

Tatskaari:
Fixes: #1320


peterebden (approved):
Nice, I like this being in the parser better. Thanks!

PR #1348, opened 2020-10-29
go_get: fix 'strip' parameter type regression

chrisnovakovic:
Commit ed17c353 accidentally reverted the change to `go_get`'s type signature that was made in commit 8d06fc86, preventing a dict from being passed in the `strip` parameter. Correct the `strip` parameter's type to once again allow dicts to be passed.


Tatskaari (approved):

Tatskaari:
Going to need to update rules.

PR #1352, opened 2020-10-30
infer goroot from install list when get repository is empty

namtzigla:
Ref:  https://github.com/thought-machine/please/commit/b76da168a0bb36600dc55818c0349bf2e2c283c8 

This patch will allow this type of go_get pattern where we will download a lib only once but we will split the installation into multiple go_get modules: 
```
go_get(
    name = "genproto",
    get = "google.golang.org/genproto",
    install = [
        "protobuf/...",
    ],
    repo = "github.com/google/go-genproto",
    revision = "8ce4113da6f7a70fe564ce27f8fd9f6149b76120",
    deps = [
        "//third_party/go/google.golang.org:protobuf",
        "//third_party/go/github.com/golang:protobuf",
    ],
)

go_get(
    name = "rpc",
    get = [],
    install = [
        "google.golang.org/genproto/googleapis/rpc/...",
    ],
    deps = [
        ":genproto",
    ],
)
```
This was previously worked only for binaries. 
The drawback is the install rule uses first install statement as getroot so it must be the most comprehensive one

### DIFF
--- a/rules/go_rules.build_defs
+++ b/rules/go_rules.build_defs
@@ -933,10 +933,8 @@ def go_get(name:str, get:str|list, repo:str='', deps:list=[], exported_deps:list
     if srcs == [] and isinstance(get, list) and get == [] and install != [] and binary == False:
         # we are going to infer the getroot name from the 1st install dependency
         getroot = install[0][:-4] if install[0].endswith('/...') else install[0]
-        if len(install) == 1 and not install[0].endswith('/...'):
-            outs += [getroot + ".a"]
-        else:
-            outs += [getroot]
+        # NOTE: for this case in the install list I'm expecting full package names 
+        outs += [i[:-4] if i.endswith('/...') else i+".a" for i in install]
         provides = {'go': go_rule} # I can't find the sources 
 
     # Now compile it in a separate step.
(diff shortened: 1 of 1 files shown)


rules/go_rules.build_defs
```diff
@@ -928,6 +929,14 @@ def go_get(name:str, get:str|list, repo:str='', deps:list=[], exported_deps:list
         else:
             all_installs += [getone]
 
+    # in case of installing packages using other package sources 
+    if srcs == [] and isinstance(get, list) and get == [] and install != [] and binary == False:
```
Tatskaari, line 933: Could we simplify this to: 
```
if not get:
   if not install:
      raise ...
   ...
```
`get` by this point will be converted as such: `get = [get]` unless it's an empty list. I'm not sure what the significance of srcs and binary is? 

I think we'd want to try to do this sort of inference whenever they provide `get = []` i.e. get is falsy?

namtzigla: I'm checking for `binary==False` and `srcs==[]`  just to make sure I'm not going to interfere with outs that might be added in the code block that's above in the file.
I will do the other suggested change

PR #1365, opened 2020-11-06
go_get: don't overcompute outputs when multiple gets/installs are defined

### LINKED ISSUE #1362: go_get overstates which packages to install in multi-get mode
`go_get` computes the list of packages that should be `go install`ed to be the Cartesian product of the values of the `get` and `install` parameters. This is problematic when `get` is a list:

```
go_get(
    name = "test",
    get = [
        "github.com/a/b",
        "github.com/c/d",
    ],
    install = [
        "github.com/a/b/pkg1",
        "github.com/a/b/pkg2",
        "github.com/c/d/pkg3",
        "github.com/c/d/pkg4",
    ],
    # ...
)
```

This passes the following list of packages to `go install`, which naturally fails since some of them are bogus:

```sh
github.com/a/b/pkg1
github.com/a/b/pkg2
github.com/a/b/github.com/c/d/pkg3  # non-existent
github.com/a/b/github.com/c/d/pkg4  # non-existent
github.com/c/d/github.com/a/b/pkg1  # non-existent
github.com/c/d/github.com/a/b/pkg2  # non-existent
github.com/c/d/pkg3
github.com/c/d/pkg4
```

chrisnovakovic:
When a `go_get` target defines multiple gets and installs, only add a particular get/install combination to `OUTS` (and the list of packages to pass to `go install`) if the install path is equal to or a subdirectory of the get path.

Closes #1362.

### DIFF
--- a/rules/go_rules.build_defs
+++ b/rules/go_rules.build_defs
@@ -908,6 +908,17 @@ def go_get(name:str, get:str|list, repo:str='', deps:list=[], exported_deps:list
     get_roots = []
     provides = {}
     srcs = []
+
+    # True if multiple gets and installs were specified, False otherwise
+    multi_get_install = len(get) > 1 and isinstance(install, list) and len(install) != 0
+
+    # Returns the element in "get" that contains the given package, or None if none of them do
+    def get_for_install(pkg):
+        for g in get:
+            if pkg == g or (pkg + "/").startswith(g + "/"):
+                return g
+        return None
+
     for getone, revision, tag, module_major_version in zip(get, revision, tags, module_major_version):
         get_rule, getroot = _go_get_download(
             name = name,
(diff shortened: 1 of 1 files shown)


chrisnovakovic:
A side-effect of this is that the following target will fail to install anything (whereas it would install `<get>/pkg1` and `<get>/pkg2` if only one `get` were defined):

```
go_get(
    name = "test",
    get = [
        "github.com/a/b",
        "github.com/c/d",
    ],
    install = [
        "pkg1",  # github.com/a/b/pkg1
        "pkg2",  # github.com/c/d/pkg2
    ],
)
```

although I'm not convinced that's worse than trying to install the Cartesian product of `get` and `install`, which is always guaranteed to fail.

If the plan for Please v16 is to require the dict form of `install` to be used when more than one `get` is listed, this can be removed as part of that.

Tatskaari:
Let's merge this in v16. I've worked around this issue for v15.

PR #1379, opened 2020-11-17
Fix printing named fields

Tatskaari:
Used to nil pointer for test_tools. Also made it print as a map when named.

### DIFF
--- a/src/query/print.go
+++ b/src/query/print.go
@@ -39,6 +39,7 @@ func Print(graph *core.BuildGraph, targets []core.BuildLabel, fields, labels []s
 	}
 }
 
+
 // specialFields is a mapping of field name -> any special casing relating to how to print it.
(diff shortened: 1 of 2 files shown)


src/query/print.go
```diff
@@ -61,10 +61,22 @@ var specialFields = map[string]func(*printer) (string, bool){
 		return p.genericPrint(reflect.ValueOf(p.target.Visibility))
 	},
 	"tools": func(p *printer) (string, bool) {
+		if len(p.target.AllNamedTools()) > 0 {
```
peterebden, line 64: nit: stash the variable
```
if tools := p.target.AllNamedTools(); len(tools) > 0 {
    return p.genericPrint(reflect.ValueOf(tools))
}
```

peterebden (approved):

Issue #1378, opened 2020-11-16
Please hangs for subinclude in certain directories


sagikazarmark:
I've experienced this issue in the near past quite a few times: Please hangs for building/querying certain targets (created using rules from subrepos) for no apparent reason:

```
❯ plz query deps -v debug --show_all_output //cmd/todo:bindata
22:38:22.582   DEBUG: Found repo root at /Users/mark/Projects/sagikazarmark/todobackend-go-kit
22:38:22.582   DEBUG: Attempting to read config from /etc/please/plzconfig...
22:38:22.582   DEBUG: Attempting to read config from /Users/mark/.config/please/plzconfig...
22:38:22.582   DEBUG: Read config from /Users/mark/.config/please/plzconfig
22:38:22.583   DEBUG: Attempting to read config from /Users/mark/Projects/sagikazarmark/todobackend-go-kit/.plzconfig...
22:38:22.584   DEBUG: Read config from /Users/mark/Projects/sagikazarmark/todobackend-go-kit/.plzconfig
22:38:22.584   DEBUG: Attempting to read config from /Users/mark/Projects/sagikazarmark/todobackend-go-kit/.plzconfig_darwin_amd64...
22:38:22.584   DEBUG: Attempting to read config from /Users/mark/Projects/sagikazarmark/todobackend-go-kit/.plzconfig.local...
22:38:22.584   DEBUG: Read config from /Users/mark/Projects/sagikazarmark/todobackend-go-kit/.plzconfig.local
22:38:22.584   DEBUG: Loading built-in build rules...
22:38:22.601   DEBUG: Parser initialised
22:38:22.601   DEBUG: Original target scan complete
22:38:22.602   DEBUG: Registering subrepo pleasings2 in package //:all
22:38:32.586  NOTICE: Build running for 10s, 1 / 3 tasks done, 0 workers busy
22:38:42.586  NOTICE: Build running for 20s, 1 / 3 tasks done, 0 workers busy
22:38:52.586  NOTICE: Build running for 30s, 1 / 3 tasks done, 0 workers busy
22:39:02.587  NOTICE: Build running for 40s, 1 / 3 tasks done, 0 workers busy
22:39:12.589  NOTICE: Build running for 50s, 1 / 3 tasks done, 0 workers busy
22:39:22.586  NOTICE: Build running for 1m0s, 1 / 3 tasks done, 0 workers busy
[...]

PR #1380, opened 2020-11-17
Demonstrate subrepo include problem

Tatskaari:
This avoids hangs in situations like #1378 because we initiate the parse but don't leverage the channel mechanism for detecting when that parse has completed. 

The problem was:
Thread 1: Parse //cmd:all
    1) subinclude("//pleasings2//go")
    2) register pending target "///pleasings//go"
    3) queue pending parse for ///pleaseings//go
    4) wait for close on ///pleasings//go pending target channel 

Thread 2: parse ///pleasings//go
    1) check subrepo
    2) register pending target "//:pleasings2"
    3) parse //:all
    4) subinclude("///pleasings2//go")
    5) See that we've already queued ///pleasings2//go
    6) wait for close on ///pleasings//go pending target channel  

At this points both threads expect the other to finish building ///pleasings2//go

In my solution, thread 1 just asks for a parse and doesn't register a pending parse. This means that thread 2 doesn't block and carries on to fully parse itself. Thread 1 then polls the graph and picks up when this is done. When it comes to wait for ///pleasings//go to be built, it finds it already has.

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -44,6 +44,9 @@ type pendingTask struct {
 	Dependent BuildLabel // The target that depended on it (only for parse tasks)
 	Run       int        // The run number of this task (only for tests)
 	Type      taskType
+
+	// This channel will be closed when the task is completed successfully or the error will be sent on failure.
+	NotifyChan chan error
 }
 
 func (t pendingTask) Compare(that queue.Item) int {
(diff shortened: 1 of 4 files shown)


Tatskaari:
Fixed this elegantly in #1424

PR #1382, opened 2020-11-18
Don't download dependencies of targets that are needed for subinclude

peterebden:
Currently we do it to arbitrary depth; we only need the proximate target that has been requested.


VERSION
```diff
@@ -1 +1 @@
-15.8.2
+15.8.3
```
Tatskaari, line 1: Can we make this a preview release if we're the only ones interested in it?

src/core/state.go
```diff
@@ -749,6 +749,10 @@ func (state *BuildState) ensureDownloaded(target *BuildTarget) {
 
 // QueueTarget adds a single target to the build queue.
 func (state *BuildState) QueueTarget(label, dependent BuildLabel, rescan, forceBuild bool) error {
+	return state.queueTarget(label, dependent, rescan, forceBuild, forceBuild)
```
Tatskaari, line 752: Why are we passing forceBuild for both here? Shouldn't this be tartget.NeededForSubinclude?

peterebden: No, because this is the thing that sets NeededForSubinclude. Breaking it apart this way means only the "outside" call gets this set, not any of its dependencies.

VERSION
peterebden: sure, can do. I do need a release internally but agreed it's not such a big deal for others.

src/core/state.go
Tatskaari: Well it can also be set to true based on `dependent.IsAllTargets()` which seems odd but i guess that's a way of detecting if it's a subinclude?

Tatskaari (approved):

PR #1387, opened 2020-11-20
Ignore secrets in config hash

Tatskaari:
And a couple misc changes


.circleci/release_s3.sh
```diff
@@ -4,6 +4,10 @@ set -eu
 
 VERSION=$(cat VERSION)
 
+echo "Releasing docs website"
```
Tatskaari, line 7: Make docs release even when we're not doing a full please release

peterebden (approved):

PR #1394, opened 2020-11-23
FIx for protoc_binary rule when cross compile 

namtzigla:
This fixes the existing `protoc_binary` rule for cross compilation.


Tatskaari (approved):

peterebden:
Thanks!

There are a couple of 'alternate' variables set too (XARCH and XOS) but I don't think they extend to HOSTARCH etc. Might bear a slightly more general think - these things are annoyingly variable as you've found & fixed here - `amd64` vs. `x86_64`, `darwin` vs. `osx` (vs. `macos`?), etc.

namtzigla:
@peterebden agree but as a side node every single repo that produce binaries has it's own notation, look at nodejs and java, so having HOSTXOS and HOSTXARCH  would not save us everywhere. 

Thanks for quick MR @Tatskaari !

PR #1396, opened 2020-11-25
Use go mod download to implement go_module() rule

Tatskaari:
This is heavily inspired by the excellent work @sagikazarmark did over at: https://github.com/sagikazarmark/please-go-modules

This is slightly less ambitious as it still uses `go install` to build the third_party deps. 

This solves a few major issues with go_get:

1) it handles all vanity imports automatically as it's using `go mod download` and copying the srcs from the module cache.
2) it handles the module major versions correctly as they form part of the module. We no longer attempt to heuristically determine the module from the get parameter. 
3) Cyclic dependencies can be resolved by pulling the download out of the `go_module()` rule into a `go_mod_download()` rule. `go_module()` can then take this in instead of a module and version and install specific packages from it. 

It's interoperable with `go_get()` so people can gradually migrate over to it.


sagikazarmark:
I tried this approach and the problem is it will download all dependencies again in GOPATH. The larger issue is different modules might depend on different versions of the same dependency which will be a problem when the final binary is linked.

Tatskaari:
> I tried this approach and the problem is it will download all dependencies again in GOPATH. The larger issue is different modules might depend on different versions of the same dependency which will be a problem when the final binary is linked.

Having different versions of the same dependency in the same build graph is strongly discouraged here at TM. Do you find that this has been a problem you've actually encountered? We've got over 1m LOC in golang with no* duplicated third_party dependencies. 

* some devs have done some bad things but there's no good reason we can't resolve this with some effort

sagikazarmark:
> Having different versions of the same dependency in the same build graph is strongly discouraged here at TM.

The problem is the build graph is not actually the same. When you hit `go install` in a dependency, it might choose a different version that another dependency (or the root module) specifies. The Go tool will resolve dependencies in the context of the actual module.

sagikazarmark:
I took a closer look: It looks like you turn off module mode for `go install` and fall back to gopath mode. I guess that could work, but it's the same hack as the current go get implementation. Gopath is essentially deprecated, so I wouldn't build on it in the long term.

Tatskaari:
> I took a closer look: It looks like you turn off module mode for `go install` and fall back to gopath mode. I guess that could work, but it's the same hack as the current go get implementation. Gopath is essentially deprecated, so I wouldn't build on it in the long term.

One idea I have is to move away from go install to writing our own go builder that uses `go list --json` to figure out how to compile a go module. This would probably be the correct solution to this problem as Please shouldn't really use modules or gopaths to find packages. It should rely entirely on the import config.

sagikazarmark:
That's exactly what I did in the repo you linked above. The main problem with that approach though is it's almost impossible calculate dependencies runtime which is why I generate rules.

Tatskaari:
> That's exactly what I did in the repo you linked above. The main problem with that approach though is it's almost impossible calculate dependencies runtime which is why I generate rules.

I'm suggesting that we write a tool that is very similar to `go install` but understands the dependencies between packages (via go list) and is then able to install them. Effectively the biggest difference is that it will take in import configs like `go tool compile` to find packages rather than go paths or go module paths.

rules/go_rules.build_defs
```diff
@@ -809,6 +809,143 @@ def cgo_test(name:str, srcs:list, data:list=None, deps:list=None, visibility:lis
...
+        licences = licences,
+    )
+
+def go_mod_download(name:str, module:str, version:str, test_only:bool=False, visibility:list=None,
+                    licences:list=None, hashes:list=None, labels:list=[],_tag:str=''):
+    cmds = [
+        "export GO111MODULE=on && export TMPDIR=$TMPDIR/tmp && mkdir -p $TMPDIR",
+        "$TOOL mod init dummy", # Create a dummy module
+        # TODO(jpoole): I don't think this is portable beyond modern GNU versions of grep. Might need to switch to sed/awk for this
+        f"export MOD_DIR=$($TOOL mod download -x -modcacherw -json {module}@{version} | grep -Po '\"Dir\": \"\K([^\"]+)')",
```
chrisnovakovic, line 864: Could use `awk -F\" '/"Dir": / { print $4 }'` or `grep -F '"Dir": ' | cut -d\" -f4` here instead of the PCRE regex in `grep` - both are POSIX-compliant.

Tatskaari: Ah brilliant thanks for looking into that for me :)

rules/go_rules.build_defs
```diff
@@ -809,6 +809,156 @@ def cgo_test(name:str, srcs:list, data:list=None, deps:list=None, visibility:lis
...
+        labels = labels + ['link:plz-out/go'],
+        hashes = hashes,
+        sandbox = False,
+        licences = licences,
+        visibility = visibility,
+    )
+
+def go_module(name:str='', module:str='', version:str='', download:str='', deps:list=[], exported_deps:list=None,
+              visibility:list=None, test_only:bool=False, binary:bool=False, install:list=[], hashes:list=None, licences:list=None):
+    if (module or version) and download:
```
peterebden, line 885: maybe `assert` instead?

I sort of feel like one day we should get rid of `raise` and the exception types. Maybe I should add an issue.

test/go_modules/test_repo/pleasew
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bash
```
peterebden, line 1: do you need this file? I assume we wouldn't actually use it for the test?

Tatskaari: Nope! Sorry, i used `plz init` to create this repo!

PR #1405, opened 2020-11-28
feat: better colour display with light colours schemes like Solarized Light

sfirmery:
With light colours schemes like "Solarized Light", things displayed in white are not easily readable.

Add a `Display.ColoursScheme` config parameter.
When set to `light`, white and grey colours are inverted in interactive display.

### DIFF
--- a/docs/config.html
+++ b/docs/config.html
@@ -124,6 +124,10 @@
       <li><b>SystemStats</b> (bool)<br/>
 	Whether or not to show basic system resource usage in the interactive display.
         Defaults to <code>False</code>.</li>
+
+      <li><b>ColourScheme</b> (string)<br/>
+	Shell colour scheme mode, dark or light. Adapts interactive display colours to the colour scheme mode.
+        Defaults to <code>dark</code>.</li>
     </ul>
 
     <h3 id="build">[Build]</h3>
--- a/src/core/config.go
+++ b/src/core/config.go
@@ -269,6 +269,7 @@ func DefaultConfiguration() *Configuration {
 	config.Test.Timeout = cli.Duration(10 * time.Minute)
 	config.Display.SystemStats = true
 	config.Display.MaxWorkers = 40
+	config.Display.ColourScheme = "dark"
 	config.Remote.NumExecutors = 20 // kind of arbitrary
 	config.Remote.HomeDir = "~"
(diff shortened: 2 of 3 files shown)


src/core/config.go
```diff
@@ -359,9 +360,10 @@ type Configuration struct {
...
 	} `help:"The [parse] section in the config contains settings specific to parsing files."`
 	Display struct {
-		UpdateTitle bool `help:"Updates the title bar of the shell window Please is running in as the build progresses. This isn't on by default because not everyone's shell is configured to reset it again after and we don't want to alter it forever."`
-		SystemStats bool `help:"Whether or not to show basic system resource usage in the interactive display. Has no effect without that configured."`
-		MaxWorkers  int  `help:"Maximum number of worker rows to display at any one time."`
+		UpdateTitle   bool   `help:"Updates the title bar of the shell window Please is running in as the build progresses. This isn't on by default because not everyone's shell is configured to reset it again after and we don't want to alter it forever."`
+		SystemStats   bool   `help:"Whether or not to show basic system resource usage in the interactive display. Has no effect without that configured."`
+		MaxWorkers    int    `help:"Maximum number of worker rows to display at any one time."`
+		ColoursScheme string `help:"Shell colours scheme mode, dark or light. Defaults to dark"`
```
Tatskaari, line 366: nit: `ColourScheme`

docs/config.html
```diff
@@ -124,6 +124,10 @@ <h3>[Display]</h3>
       <li><b>SystemStats</b> (bool)<br/>
 	Whether or not to show basic system resource usage in the interactive display.
         Defaults to <code>False</code>.</li>
+
+      <li><b>ColoursScheme</b> (bool)<br/>
```
Tatskaari, line 128: nit: (bool) should be (string)

Tatskaari (commented):
I find it a little weird that the light theme seems darker and works on a light background. The dark theme seems lighter and works on a dark background. I wonder if we can find a better name for these? If not, it does actually make some sense if you think about it.

sfirmery:
Thanks for the review.

A better naming could be `BackgroundTone`, what do you think?

Tatskaari:
Thanks @sfirmery, I think the naming is probably clear enough as is. People should figure it out.

PR #1414, opened 2020-12-02
Pre-loaded subincludes

Tatskaari:
For some reason, parsing locks up when subinclludng pleasings from the root package...


src/parse/init.go
```diff
@@ -58,7 +58,25 @@ func newAspParser(state *core.BuildState) *asp.Parser {
 }
 
 func (p *aspParser) ParseFile(state *core.BuildState, pkg *core.Package, filename string) error {
-	return p.asp.ParseFile(pkg, filename)
+	if pkg.Name == "" {
+		return p.asp.ParseFile(pkg, filename, "")
+	}
+
+	subincludes := make([]string, 0, len(state.Config.Parse.PreloadSubincludes))
```
peterebden, line 65: can you pull all this logic out into the "constructor"? feels like we shouldn't need to do it every time

PR #1427, opened 2020-12-09
Move to fork of go-flags

peterebden:
Cleans up some of the code in core where we don't need build tags or patches any more.


docs/template.go
```diff
@@ -2,14 +2,15 @@ package main
...
-	"github.com/peterebden/go-cli-init"
 	"hash/adler32"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
 	"text/template"
+
+	"github.com/peterebden/go-cli-init/v3"
```
Tatskaari, line 13: What does this do that go-flags doesn't do? Maybe we can merge these?

Tatskaari (approved):

docs/template.go
peterebden: It has a wrapper for flag parsing which we could maybe roll in. Mostly it's more about tying up logging with it for a bunch of niceties that we can share between multiple projects.

PR #1430, opened 2020-12-10
Update python_intro.md, correcting reference to go_get()

joshcoales:
Correcting a reference to go_get(), to refer to pip_library(), because this is the python intro afterall


Tatskaari:
Thanks for the fix! I hope you found the codelabs useful!

PR #1433, opened 2020-12-10
Allow to specify the shell to execute remote actions in via config

peterebden:
We'd like to be able to just say "use this path" on machines that might not have it in the same place as the remote workers.


src/remote/remote.go
```diff
@@ -206,16 +207,18 @@ func (c *Client) initExec() error {
...
 	}
-	// We have to run everything through bash since our commands are arbitrary.
-	// Unfortunately we can't just say "bash", we need an absolute path which is
-	// a bit weird since it assumes that our absolute path is the same as the
-	// remote one (which is probably OK on the same OS, but not between say Linux and
-	// FreeBSD where bash is not idiomatically in the same place).
-	bash, err := core.LookBuildPath("bash", c.state.Config)
-	if err != nil {
-		return fmt.Errorf("Failed to set path for bash: %w", err)
+	if c.state.Config.Remote.Shell != "" {
```
Tatskaari, line 210: Shouldn't this be `c.shellPath == ""`?

peterebden: Yep good spot, I changed that later rather than have an if/else but forgot to change this too.

Tatskaari (approved):

PR #1439, opened 2020-12-11
Make wheel_name_scheme repeated

peterebden:
Because if someone had nefarious schemes they might want more than one of them.


Tatskaari (approved):
Going to need to generate rules

peterebden:
Bring on Go 1.16!

rules/python_rules.build_defs
```diff
@@ -493,23 +494,25 @@ def python_wheel(name:str, version:str, hashes:list=None, package_name:str=None,
...
-                                                        version=version)]
+        urls += [scheme.format(url_base=url_base, package_name=package_name, initial=initial, version=version)
+                 for scheme in CONFIG.PYTHON_WHEEL_NAME_SCHEME]
     else:
         # Populate urls using a reasonable set of possible wheel naming schemes.
         # Look for an arch-specific wheel first; in some cases there can be both (e.g. protobuf
         # has optional arch-specific bits) and we prefer the one with the cool stuff.
         urls += ['{url_base}/{package_name}-{version}-${{OS}}-${{ARCH}}.whl'.format(url_base=url_base,
                                                                                     package_name=package_name,
+                                                                                    initial=initial,
```
Tatskaari, line 508: It's no immediately obvious what initial is or why it's useful. Looks like it ends up as the first character of package_name?

Tatskaari (commented):
I don't see the breaking change. Can this be landed for 15.12.0?

src/core/config.go
```diff
@@ -447,18 +447,18 @@ type Configuration struct {
...
+		DefaultPipRepo      cli.URL  `help:"Defines a location for a pip repo to download wheels from.\nBy default pip_library uses PyPI (although see below on that) but you may well want to use this define another location to upload your own wheels to.\nIs overridden by the repo argument to pip_library." var:"PYTHON_DEFAULT_PIP_REPO"`
+		WheelRepo           cli.URL  `help:"Defines a location for a remote repo that python_wheel rules will download from. See python_wheel for more information." var:"PYTHON_WHEEL_REPO"`
+		UsePyPI             bool     `help:"Whether or not to use PyPI for pip_library rules or not. Defaults to true, if you disable this you will presumably want to set DefaultPipRepo to use one of your own.\nIs overridden by the use_pypi argument to pip_library." var:"USE_PYPI"`
+		WheelNameScheme     []string `help:"Defines a custom templatized wheel naming scheme. Templatized variables should be surrounded in curly braces, and the available options are: url_base, package_name, and version. The default search pattern is '{url_base}/{package_name}-{version}-${{OS}}-${{ARCH}}.whl' along with a few common variants." var:"PYTHON_WHEEL_NAME_SCHEME"`
```
Tatskaari, line 460: Should this help string include initial

rules/python_rules.build_defs
peterebden: That is exactly what it is. There are wheels stored on PyPI in a predictable scheme but it involves the first initial, so we have to make that available to the scheme template.

src/core/config.go
peterebden: Have added

peterebden:
Ergh lost track of this.

> I don't see the breaking change. Can this be landed for 15.12.0?

Yep it's not breaking. Making a singular config entry repeated is fine.

PR #1440, opened 2020-12-14
Update upgrade/downgrade message

janhancic:
Change it to a, subjectively, more pleasing message that reads more naturally.

Appreciate this is up for personal preference so nw if this gets rejected 👍


janhancic:
I hope this change receives the appropriate level of hype in your next release email @Tatskaari!

PR #1441, opened 2020-12-14
Make output dirs rename outputs rather than copying them

### LINKED ISSUE #1407: Please does not preserve file modes
I have the following rule:

```python
def folder_files(name, srcs, directory_name, mode = "644", visibility = None):
    genrule(
        name = name,
        srcs = srcs,
        output_dirs = ["_out"],
        cmd = [
            f"mkdir -p _out/{directory_name}",
            f"install -m {mode} $SRCS _out/{directory_name}",
        ],
        visibility = visibility,
    )
```

With `mode = "0775"` it produces files with the following modes:

```
-rwxr-xr-x 1 suhr users     90984 Nov 30 13:23 clusterdb
-rwxr-xr-x 1 suhr users     99712 Nov 30 13:23 createuser
-rwxr-xr-x 1 suhr users     90216 Nov 30 13:23 dropuser
-rwxr-xr-x 1 suhr users    163784 Nov 30 13:23 initdb
```

But Please output produces files with this modes:

```
.r--r--r--  90k suhr 30 ноя 13:22 clusterdb
.r--r--r--  99k suhr 30 ноя 13:22 createuser
.r--r--r--  90k suhr 30 ноя 13:22 dropuser
.r--r--r-- 163k suhr 30 ноя 13:22 initdb
```

Tatskaari:
Fixes #1407

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -563,7 +563,7 @@ func addOutputDirectoryToBuildOutput(target *core.BuildTarget, dir core.OutputDi
 			target.AddOutput(f.Name())
 			outs = append(outs, f.Name())
 
-			if err := fs.RecursiveCopy(from, to, target.OutMode()); err != nil {
+			if err := os.Rename(from, to); err != nil {
 				return nil, err
 			}
 		}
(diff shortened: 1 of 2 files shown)


peterebden:
There are other paths that can happen here too - e.g. caching.

Tatskaari:
> There are other paths that can happen here too - e.g. caching.

This code is moving the files within the output directory into the temp directory root. The files will then be moved by the standard `moveOutputs()` code. Beyond that point it will follow the same code path as normal outputs so I don't forsee any other problems.

Issue #1446, opened 2020-12-16
filegroup is slow


suhr:
```
% plz build //:files
Build finished; total time 10.16s, incrementality 100.0%. Outputs:
//:files:
  plz-out/gen/local-target/release/353e1ae81a4e6e4fba39a0e4.a
  plz-out/gen/local-target/release/5630951b01a34696e611eda6.a
  plz-out/gen/local-target/release/685b912717f022ea80c706e4.a
  plz-out/gen/local-target/release/6ab540305f20588a3fc51c53.a
  plz-out/gen/local-target/release/6ba81f94ba2d4493b9ecd8ac.so
  plz-out/gen/local-target/release/877c5114f7d26d7ef14651ab.so
  plz-out/gen/local-target/release/93e96c550eed32cbb56280a6.a
  plz-out/gen/local-target/release/af26d9898610ac3a0a3aaec6.so
  plz-out/gen/local-target/release/b6bec3157f581134e9ed1e63.so
  plz-out/gen/local-target/release/cb4824544643bc1cc1f1ba1c.so
  plz-out/gen/local-target/release/d7144ec667b5db8d01f1f021.so
  plz-out/gen/local-target/release/f703a6a749b7c45044808705.so
  plz-out/gen/local-target/release/fbc904654a20257810ba6bc7.so
  plz-out/gen/local-target/release/ff4eea0090a1dfa90be34a4c.so
  plz-out/gen/local-target/release/libs/gost-r-3412-15-lib/625e5e490f4b0c6573ec37bc.a
  plz-out/gen/local-target/release/libs/gost-r-3412-15-lib/625e5e490f4b0c6573ec37bc.so
```

This is much slower than hashing:

```
% hyperfine 'sha1sum local-target/release/{*.a,*.so} local-target/release/libs/gost-r-3412-15-lib/{*.a,*.so}'
Benchmark #1: sha1sum local-target/release/{*.a,*.so} local-target/release/libs/gost-r-3412-15-lib/{*.a,*.so}
  Time (mean ± σ):      2.396 s ±  0.024 s    [User: 2.258 s, System: 0.138 s]
  Range (min … max):    2.365 s …  2.440 s    10 runs
```

By the way, SHA-1 is a rather slow hash for a build system. Consider using [BLAKE3](https://github.com/BLAKE3-team/BLAKE3) instead. Or even some non-cryptographic hash, since they are orders of magnitude faster than cryptographic ones.

Tatskaari:
[...]

PR #1447, opened 2020-12-18
Added new hash funcs

Tatskaari:
Helps with #1446 but still need to figure out what please is doing with the other few seconds...


peterebden (approved):
Thanks!

PR #1448, opened 2020-12-21
Optimise constant fstrings

peterebden:
Lets us optimise constant fstrings (i.e. `f'something'`) into constant strings.
Have observed a couple of cases where these come up, often when there's a vertical list (e.g. inside a `.join()` call) and not all entries actually contain variables.


src/parse/asp/interpreter_test.go
```diff
@@ -351,3 +351,14 @@ func TestDivide(t *testing.T) {
 	assert.EqualValues(t, 7, s.Lookup("j"))
 	assert.EqualValues(t, -2, s.Lookup("k"))
 }
+
+func TestFStringOptimisation(t *testing.T) {
+	s, stmts, err := parseFileToStatements("src/parse/asp/test_data/interpreter/fstring_optimisation.build")
+	require.NoError(t, err)
+	assert.EqualValues(t, s.Lookup("x"), "test")
+	// Check that it's been optimised to something
+	ass := stmts[0].Ident.Action.Assign
```
Tatskaari, line 360: ass?

Tatskaari (approved):

src/parse/asp/interpreter_test.go
peterebden: short for assign

PR #1452, opened 2020-12-30
fix: allow configurable pip to be a target again

noxxious:
The configurable Pip tool, got broken if it was set to a target. This fixes it back.


noxxious:
@Tatskaari this is ready for your review.

Tatskaari (approved):
Thanks Noxx! Sorry if this caused you any issues.

PR #1455, opened 2021-01-04
Add codelab for GitHub Actions

sagikazarmark:
This PR adds a codelab for running Please on GitHub Actions, featuring my [setup-please](https://github.com/sagikazarmark/setup-please-action) action.

It explains in simple terms how GitHub Actions works, but requires a basic understanding of Please and GitHub.

It's a first version, so it can probably be improved (a lot).

### DIFF
--- a/docs/codelabs/github_actions.md
+++ b/docs/codelabs/github_actions.md
@@ -1,0 +1,177 @@
+summary: Running Please on GitHub Actions
+description: GitHub Actions is an extensible CI/CD platform provided by GitHub
+id: github_actions
+categories: intermediate
+tags: medium
+status: Published
+authors: Márk Sági-Kazár
+Feedback Link: https://github.com/thought-machine/please
+
+# Running Please on GitHub Actions
+## Overview
+Duration: 2
+
+### Prerequisites
+- A repository on [GitHub](https://github.com)
+- A project with Please initialized in that repository
+
+### What You’ll Learn
+- How to setup [GitHub Actions](https://github.com/features/actions)
+- How to use Please in a GitHub Actions build
+- How to use the [setup-please](https://github.com/sagikazarmark/setup-please-action) action for better integration
+
+### What if I get stuck?
+If you get stuck with GitHub Actions, check out the [official documentation](https://docs.github.com/en/free-pro-team@latest/actions).
+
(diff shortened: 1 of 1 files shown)


docs/codelabs/github_actions.md
```diff
@@ -0,0 +1,177 @@
...
+Shell scripts (no matter how awesome they are) are not always the right tool for the job. Complex build steps might require a more expressive language which takes us to the second type of workflow steps, called _actions_:
+
+```yaml
+- name: Checkout code
+  uses: actions/checkout@v2
+```
+
+An _action_ can be written in any language (distributed as Docker images), but JavaScript is supported natively.
+
+### Why not just use `./pleasew`?
```
Tatskaari, line 127: This doesn't render to HTML very well with the backticks.

sagikazarmark: Should I just remove it? I left it there because it still makes sense in markdown.

docs/codelabs/github_actions.md
```diff
@@ -0,0 +1,177 @@
...
+The [setup-please](https://github.com/sagikazarmark/setup-please-action) action provides better integration for Please.
+
+### What is an _action_?
+
+As you've seen in the previous examples, workflows consist of _steps_.
+A workflow step can be as simple as a shell script:
+
+```yaml
+- name: Test
+  run: ./pleasew build //...
```
Tatskaari, line 115: Should this be `./pleasew cover //...`? 

I also wonder if we should recommend something like `plz query changes --since master --include_dependencies=transitive`? One of Please's selling points is that it can be used to avoid building/testing stuff that didn't actually get changed.

sagikazarmark: I haven't actually used that before. Maybe it could be part of this code lab, but I was kinda planning to write another one for advanced use cases, like caching.

docs/codelabs/github_actions.md
Tatskaari: Yeah I think the website is more important than the markdown.

docs/codelabs/github_actions.md
Tatskaari: Well I think it should at least be `plz test` or `plz cover`

sagikazarmark: Changed it to test. `plz query changes --since master --include_dependencies=transitive | plz cover -` doesn't work for me: `unknown flag `include_dependencies'`

docs/codelabs/github_actions.md
sagikazarmark: Changed

PR #1462, opened 2021-01-05
Allow go_get() to sandbox install step

Tatskaari:
We need to allow network connections for the download, but the actual install can be sandboxed.


peterebden (approved):
Seems legit, I assume we would know pretty quick if it was denying access to something required

peterebden:
FWIW I think maybe we should have a general feature to have more flexible sandboxing - like rules could opt out of network sandboxing but still get directory / pid sandboxing etc. Not quite certain how it should look.

PR #1464, opened 2021-01-05
Added proper WaitForPackage semantics and used them for subincludes

Tatskaari:
This avoid the lockup. Before the parsing of the subinclude package would also trigger a call to `WaitForTarget` which would hang because the original subinclude had already called that function:

old: 
1) //:all subincludes ///foo/bar//bazz
2) //:all queues ///foo/bar//bazz for parse and waits for it to be built (adding the channel to the pendingTargets map) 
3) parse of ///foo/bar//bazz checks for subrepos and parses the //foo package
4) //foo subincludes  ///foo/bar//bazz
5) //foo believes that something has already queued ///foo/bar//bazz to be built so blocks waiting on the channel from #2

At this point, we lock up essentially because as far as each thread is concerned, the other has committed to building ///foo/bar//bazz.

The new approach tries to avoid this by triggering a parse of the subrepos package before queuing the actual subinclude target:
 
1) //:all subincludes ///foo/bar//bazz
2) //:all queues //foo for parse and waits for the parse (adding the channel to a new non committal packageWaits map) 
3) We must then encounter the subrepo first in the package and add //foo:bar to the build graph
4) //foo subincludes ///foo/bar//bazz
5) We see that the subrepo is in the same package so skip queuing the parse again
6) //foo queues ///foo/bar//bazz to be pares (then built) 
7) Parsing ///foo/bar//bazz checks the subrepo and sees it's already been added so activates the target and waits for it to build
8) At this point the normal build/parse workflow plays out eventually releasing the locks once a successful build/parse build result gets logged.
9) The suninclude from step 1 continues execution, returning immediately when waiting for ///foo/bar//bazz as it was already built

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -226,7 +226,9 @@ type stateProgress struct {
 	pendingTargets     map[BuildLabel]chan struct{}
 	pendingTargetMutex sync.Mutex
 	// Used to track general package parsing requests.
-	pendingPackages     map[packageKey]chan struct{}
+	pendingPackages map[packageKey]chan struct{}
+	// similar to pendingPackages but consumers haven't committed to parsing the package
+	packageWaits        map[packageKey]chan struct{}
 	pendingPackageMutex sync.Mutex
 	// The set of known states
 	allStates []*BuildState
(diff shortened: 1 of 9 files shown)


src/core/state.go
```diff
@@ -669,6 +674,37 @@ func (state *BuildState) WaitForPackage(label BuildLabel) *Package {
 	return state.Graph.PackageByLabel(label) // Important to check again; it's possible to race against this whole lot.
 }
 
+// WaitForPackage is similar to WaitForBuiltTarget however
```
peterebden, line 677: what a cliffhanger!

src/parse/asp/interpreter.go
```diff
@@ -100,6 +101,7 @@ func (i *interpreter) interpretStatements(s *scope, statements []*Statement) (re
 			} else {
 				err = fmt.Errorf("%s", r)
 			}
+			log.Debug(string(debug.Stack()))
```
peterebden, line 104: `log.Debug("%s", debug.Stack())` would be nicer (if you're intending to keep this)

peterebden (approved):

src/core/state.go
Tatskaari: You'll just have to wait for the rest of this message

PR #1466, opened 2021-01-07
fix: fallback to copy in case os.Rename fails due to different filesystems

noxxious:
```os.Rename``` does not work (due to trying to hardlink src to dest) across filesystems, which in this case the source is /tmp, usually a tmpfs on Linux.
Also the removal of the destination file is not required and actually results in target file deleted in case of error. 

The hard issue I found with this was with ```plz fmt``` failing and actually removing the source BUILD file, so even the fallback case in multiple filesystems, this results in removal of precious source BUILD files from the source tree.

### DIFF
--- a/src/fs/fs.go
+++ b/src/fs/fs.go
@@ -129,7 +129,7 @@ func IsPackage(buildFileNames []string, name string) bool {
 	return false
 }
 
-// Try to gracefuly rename the file as the os.Rename does not work accross
+// Try to gracefully rename the file as the os.Rename does not work across
 // filesystems and on most Linux systems /tmp is mounted as tmpfs
 func renameFile(from, to string) (err error) {
 	err = os.Rename(from, to)
--- a/src/fs/fs.go
+++ b/src/fs/fs.go
@@ -86,9 +86,6 @@ func CopyFile(from string, to string, mode os.FileMode) error {
 // WriteFile writes data from a reader to the file named 'to', with an attempt to perform
 // a copy & rename to avoid chaos if anything goes wrong partway.
 func WriteFile(fromFile io.Reader, to string, mode os.FileMode) error {
-	if err := os.RemoveAll(to); err != nil {
-		return err
-	}
 	dir, file := path.Split(to)
 	if dir != "" {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/fs/fs.go [hunk 1]
```diff
@@ -131,3 +128,60 @@ func IsPackage(buildFileNames []string, name string) bool {
...
+
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return err
+	}
+
+	err = out.Sync()
+	if err != nil {
+		return err
+	}
```
Tatskaari, line 175: I don't think we need this if we close `in` and `out`?

src/fs/fs.go
```diff
@@ -131,3 +128,60 @@ func IsPackage(buildFileNames []string, name string) bool {
 	}
 	return false
 }
+
+// Try to gracefully rename the file as the os.Rename does not work across
+// filesystems and on most Linux systems /tmp is mounted as tmpfs
+func renameFile(from, to string) (err error) {
+	err = os.Rename(from, to)
+	if err == nil {
+		return err
```
Tatskaari, line 137: nit: reads a bit better if we return nil here. I thought the if statement was `err != nil` like it normally would be. Returning nil would make it clear this is the happy path.

Tatskaari (approved):
Looks good to me. @pebers, would be nice to have your eyes on this too though.,

src/fs/fs.go
```diff
@@ -131,3 +128,60 @@ func IsPackage(buildFileNames []string, name string) bool {
...
+			err = e
+		}
+	}()
+
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return err
+	}
+
+	err = out.Sync()
```
peterebden, line 172: Is this necessary? Given that you are doing `Close()` above and checking its error I would think that's sufficient?

src/fs/fs.go
noxxious: removed it.

src/fs/fs.go
noxxious: done.

PR #1467, opened 2021-01-07
Update lexicon.html

binhonglee:
It doesn't seem to highlight the right tab when selected. (Always highlight python-builtins when the user clicks on either python-builtins or please-builtins.) Not sure if this will fix it but I don't have a good way to test this.

### DIFF
--- a/docs/lexicon.html
+++ b/docs/lexicon.html
@@ -3,8 +3,8 @@
     <p>This is the reference for the complete set of builtin rules & functions.</p>
 
     <ul>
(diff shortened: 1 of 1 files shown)


Tatskaari:
Hey, you can test this by running `plz run //docs:view` which should start a webserver and pop open a web browser if your OS has a sensiblebrowser alias. Otherwise run `cd plz-out/gen/docs && python3 -m http.server 8080`

Tatskaari:
I believe this got fixed up as part of the big website refactor. Thanks for raising it though, appreciate the effort. :)

Issue #1450, opened 2020-12-25
Invalid pip option "--system" when using Pip/Python installed via Pyenv on Ubuntu/Debian


xinkecf35:
**Description**
While using `pip_library` build rules with the Python interpreter set to one installed by Pyenv, the build fails with the following error: "no such option: --system". As far as I can tell, this is due to the addition of the flag to counter the fact that Debian distributions of `python-pip` have it such that `--user` is enabled by default.

Issue was produced on a Ubuntu 20.04.1 LTS, Pyenv version 1.2.21 with Python version 3.9.1. Also reproducible in Debian 10.7 with Pyenv version 1.2.18 with Python version 3.8.3. Note that issue does not occur when using Pyenv on macOS version 10.15.6. Please version 15.11.0 was tested, but the issue effects versions 15.8 and up. 

**Reproduction steps**
- Install latest pyenv with [pyenv-installer](https://github.com/pyenv/pyenv-installer)
- Follow setup instructions for Pyenv to be initialized in user's shell and home directory
- Set `Path` under `build` section of `.plzconfig` to point at pyenv shims and versions, e.g. `/home/<user>/.pyenv/shims`, 
 `/home/<user/.pyenv/bin`
- Add any pip package and build said package

Sample build log is as follows [here](https://gist.github.com/xinkecf35/e72cbf9d4090f42bc3cd1d17501209a7)

**Possible resolution**
Easiest thing might be an explicit check to see if Pyenv is being used and therefore not use `--system` as part of the pip command, but that feels brittle and suboptimal.

peterebden:
Ugh. The `--system` thing is already a hack to counteract breakage Debian introduced earlier; we are essentially detecting that it's a Debian system and hence we think we need to add that flag.

I guess options here include either (1) just remove the flag in some future version and stop supporting 16.04 / 18.04 or (2) have some explicit config property to add it. I think (2) is probably nicer, although obviously undesirable for us to have to work around the foibles of pip.

xinkecf35:
For what’s it worth, I feel option two is probably the most practical option at this given moment. Option one would be nice, but I don’t know how people will feel about the lack of support for at least for 18.04 given it’s EOL is sometime in 2023

xinkecf35:
@peterebden To follow up on this, if a config property were to be added (assuming that's the desire approach), what would be config property do and look like it? It would be a simple boolean flag to turn off some of `SYS_FLAG` guessing? Or would it be more of a string property where one could specific additional flags?

[...]

PR #1470, opened 2021-01-10
Add Config Option to Disable Vendor Specific Flags for Pip in pip_library

xinkecf35:
This PR adds a new configuration flag for the Python that allow an end user to turn off the system guessing/patch of the pip command. The change attempts to address #1450 in cases which a user is using pip/python not from the distribution and thus may not have those pip options available.

The further changes were completed:
- Adding of the boolean `DisableVendorFlags` to the Python config
- Modifies `pip_library` to not add the vendors flags if `DisableVendorFlags` is true
- Updates `config.html` in the docs to mention the option

Changes were tested by:
- Adding the `DisableVendorFlags` into `.plzconfig.local` of a test project with the path set to Pyenv
- Above done as well in the Please project itself.

Guidance/Suggestions on how to add automated tests for the configuration flag would be greatly appreciated. I am not entirely sure about the best way of doing that seeing as the only reliable way I have of doing so requires a dependency on Pyenv


src/core/config.go
```diff
@@ -282,6 +282,7 @@ func DefaultConfiguration() *Configuration {
 	config.Go.InstallTool = "please_go_install"
 	config.Python.PexTool = "please_pex"
 	config.Python.DefaultInterpreter = "python3"
+	config.Python.DisableVendorFlags = false
```
Tatskaari, line 285: Not strictly necessary as this is the zero value anyway

Tatskaari (approved):

Tatskaari:
Ah brilliant! This was on my list of things to do. Thanks for that. :)

Any reason this is a draft PR? I'd love to get it into tomorrows release.

xinkecf35:
> Ah brilliant! This was on my list of things to do. Thanks for that. :)
> 
> Any reason this is a draft PR? I'd love to get it into tomorrows release.
> 
> Re testing: you could add an e2e test although I'm not sure how much value that would add given the nature of the change. We'd really want to have a couple different linux envs to test it on. I'd be happy to merge as is.

It’s a draft because of the testing question. I had an idea of maybe pulling down Pyenv and setting up pip that way but that feels a little heavy handed.

Tatskaari:
Okay well if it's all good with you, I'll land this change :)

xinkecf35:
> Okay well if it's all good with you, I'll land this change :)
> 
> Edit: looks like you need to resolve some conflicts

Resolved the conflicts with the config.html file. And yes I am good with that.

PR #1471, opened 2021-01-11
effected -> affected on homepage

ojcm:
effected means "cause (something) to happen; bring about"
affected means "influenced or touched by an external factor"


Tatskaari:
Ah many thanks. I keep getting this one wrong...

PR #1477, opened 2021-01-13
Allow `repo` argument on go_get to take local paths

peterebden:
This is equivalent to Go's `replace` directive, ala `replace example.com/original/import/path => /your/forked/import/path` (similar to how you can use `repo` to replace the import with a different github repo etc)


Tatskaari:
I've opened a new pull request that allows vendoring here: #1487

PR #1482, opened 2021-01-14
Fix cgo for go_module() with please install

Tatskaari:
installing the sdk for arm or freebsd seems to break when cgo is enabled


rules/go_rules.build_defs
```diff
@@ -1309,7 +1315,7 @@ def _set_go_env():
         go_root = f'$(dirname $(dirname $TOOLS_GO))'
     else:
         go_root = '$("$TOOLS_GO" env GOROOT)'
-    return f'export GOPATH=$TMP_DIR && export GOROOT={go_root}'
+    return f'export CGO_ENABLED=1 && export GOPATH=$TMP_DIR && export GOROOT={go_root}'
```
peterebden, line 1318: should this not be based on the config flag?

rules/go_rules.build_defs
```diff
@@ -37,11 +37,14 @@ def go_toolchain(name:str, url:str|dict = '', version:str = '', hashes:list = []
     )
 
     cmd = 'tar -xf $SRCS && mv go $OUT && chmod +x $OUT/bin/*'
-
     # If we're targeting another platform, build the std lib for that. We can't use CONFIG.OS as this is always set to
     # the host OS for tools.
     if CONFIG.TARGET_OS != CONFIG.HOSTOS or CONFIG.TARGET_ARCH != CONFIG.HOSTARCH:
-        cmd += f' && (export GOOS={CONFIG.TARGET_OS} && export GOARCH={CONFIG.TARGET_ARCH} && $OUT/bin/go install std)'
+        if CONFIG.CGO_ENABLED == "1":
```
peterebden, line 43: why is this "1"? is that because it's what Go expects?

src/core/config.go
```diff
@@ -442,6 +443,7 @@ type Configuration struct {
 		GoPath           string `help:"If set, will set the GOPATH environment variable appropriately during build actions." var:"GOPATH"`
 		ImportPath       string `help:"Sets the default Go import path at the root of this repository.\nFor example, in the Please repo, we might set it to github.com/thought-machine/please to allow imports from that package within the repo." var:"GO_IMPORT_PATH"`
 		CgoCCTool        string `help:"Sets the location of CC while building cgo_library and cgo_test rules. Defaults to gcc" var:"CGO_CC_TOOL"`
+		CgoEnabled       string `help:"Sets the CGO_ENABLED which controls whether the cgo build flag is set. Defaults to 0 (off)" var:"CGO_ENABLED"`
```
peterebden, line 446: Is defaulting it to off correct? Go defaults it to 1 doesn't it?

Tatskaari: I don't think so. For example, if you don't set it manually, `go install std` doesn't compile `runtime/cgo`. `go build` sets it and compile `runtime/cgo` etc. on the fly like how it compiles the standard library when you cross compile or do `-race`. It's hard for us to do that with our approach though.

PR #1483, opened 2021-01-15
Fix required module parameter in go_module

sagikazarmark:
When using the `go_mod_download` rule together with the `go_module`, the module parameter should always be passed to the `go_module` rule as it is used outside of the download rule.

Maybe separate tests for the `go_mod_download` would make sense.

### DIFF
--- a/rules/go_rules.build_defs
+++ b/rules/go_rules.build_defs
@@ -934,14 +934,14 @@ def go_mod_download(name:str, module:str, version:str, test_only:bool=False, vis
         visibility = visibility,
     )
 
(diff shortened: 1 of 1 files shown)


Tatskaari:
Would be good to demonstrate a circular dep resolved by this as a test. Do you have an example of some modules that are circular?

Tatskaari (approved):
A thanks a lot!

sagikazarmark:
I think grpc and protobuf (or at least some versions of them) are circularly dependent on each other. But we don't need circular dependencies for testing. Any module can be downloaded separately.

PR #1484, opened 2021-01-15
feat: add asm support to please_go_install

sagikazarmark:
This is the first working version after a couple hours of hacking.

I think we should reuse some parts of the Go build tool: https://github.com/golang/go/blob/e125ccd10ea191101dbc31f0dd39a98f9d3ab929/src/cmd/go/internal/work/gc.go#L325

As you can see, it handles a bunch of edge cases.

A couple thins I noticed while I made this work:

- please_go_install uses the gopath as a workspace (instead of its tmpdir) which causes some trouble for packages (eg. `github.com/golang/snappy` uses `pkg/darwin_amd64/github.com/golang/` as a workspace which could overlap with other packages in `github.com/golang`.) Using $TMP_DIR and moving the final archive at the end would be better IMHO. That's actually what the go build tool does.


tools/please_go_install/main.go
```diff
@@ -156,32 +168,61 @@ func (g *pkgGraph) compile(from []string, target string) {
...
+		objectFiles = append(objectFiles, ofile)
+		fmt.Printf("%s tool asm -p %s -I %s -I %s/pkg/include -D GOOS_%s -D GOARCH_%s -o %s %s\n", opts.GoTool, target, pkgDir, build.Default.GOROOT, build.Default.GOOS, build.Default.GOARCH, ofile, src)
+	}
+
+	// repack archive
+	if len(objectFiles) > 1 {
+		// pack r $WORK/b001/_pkg_.a $WORK/b001/decode_amd64.o $WORK/b001/encode_amd64.o # internal
+		fmt.Printf("%s tool pack r %s\n", opts.GoTool, strings.Join(objectFiles, " "))
+	}
+
```
Tatskaari, line 225: We probably need to do this for cgo as well right?

Tatskaari (commented):
Looks good so far! Thanks a lot for this. 

We should probably add to //test/go_modules to test:
1) a module using just asm sources 
2) a module using both cgo and asm sources

Tatskaari:
Re the working directory: sounds very sensible. I assume it's just a case of telling `go tool cgo/compile/link` to spit out the generated files elsewhere.

sagikazarmark:
> Re the working directory: sounds very sensible. I assume it's just a case of telling `go tool cgo/compile/link` to spit out the generated files elsewhere and moving the sources?

Yes, I believe that's correct.

> We probably need to do this for cgo as well right?

The go build tool actually doesn't differentiate between go and cgo libraries. It compiles everything into object files (go, cgo, asm) and packs everything together at the end. I think that's what we should probably do.

I wouldn't recommend merging or even moving forward with this PR. It doesn't even work for all asm libraries. It's just an example. Looking at the go build tool, there are much more cases to cover to support the most common scenarios, so I'd probably copy as much from the go build tool as possible.

Here is the interesting part for generating the relevant tool commands: https://github.com/golang/go/tree/master/src/cmd/go/internal/work

I also found an old tool from one of the Go authors here: https://github.com/constabulary/gb
It is quite similar to the go build tool, but I'm not sure which inspired which.

sagikazarmark:
How should we proceed here?

I really think the best way forward is to try to replicate the behavior of the go build tool as much as possible (without the unnecessary/unwanted features, like caching, buildid, etc).

The obvious benefit of such build tool (other than the fact that we can support a wide range of use cases) is that keeping track of the changes, porting patches become much easier.

I've started building a POC: https://github.com/sagikazarmark/gb
It could become a `please_go_build` tool (feels like install doesn't really work here), but keeping it an out-of-tree tool would also be great IMHO. It could be useful for other build systems as well. (It might also be better to keep it in a separate repo for licensing reasons)

But it's much more complicated than it sounds. The code itself is not really built in a way that supports copying very well. It uses internal packages a lot.

At the same time, the provided fix resolves the immediate issue of not supporting asm builds, so I think it might make sense to merge this PR after all.

sagikazarmark:
@Tatskaari what do you think? Should we merge this? Or should go for a compile tool outlined above?

Tatskaari:
> @Tatskaari what do you think? Should we merge this? Or should we go for a compile tool outlined above?

Sorry, I missed this message. I've not had access to emails for a few days. 

Just spend the afternoon trying to get your PoC to work. I'm just having to pull more and more code form the SDK and I feel it's getting out of hand a bit. It seems very possible to get this working but merging upstream changes down will be v. hard. I also don't feel like I have a good grasp on how the code works. With enough effort we might be able to make that approach work but I don't see the light at the end of the tunnel right now. 

Our current approach might not cover all cases, but I think as it matures, we will cover more and more cases. 


Also yes, the name sucks. I should probably change it.

sagikazarmark:
> I'm just having to pull more and more code form the SDK and I feel it's getting out of hand a bit.

As far as I can see, we only need the cgo bit figured out, although  that could be complicated enough not to use this approach. I don't know.

I still think following the builtin go tool's structure make sense. For example, instead of having a cgo and a non-cgo branch, we should collect object files in the link/pack step. That way we don't have to worry about having to add things to separate places.

The way the toolchain is abstracted is also a nice touch in the go build tool (although it uses a lot of global variables which kinda defeats the purpose).

So how about this: let's try to move forward the POC and just shove in what please_go_install has now for cgo today and improve it later.

I'd also add a whole lot more test cases to cover all the weird edge cases. That will also help if you decide to rewrite this tool from scratch.

Having said all that, should I just rebase this PR to get the changes merged?

Tatskaari:
Yeah we should have a few functions that wrap these steps:

tc.goCompile(): compile a list of `*.go` files to a working directory 
tc.cgo(): run `tool cgo` on a list of `.go` cgo file and output into a working directory 
tc.cc(): run the CC_TOOL on a list of `.c` files and output into a working directory 
tc.asm(): compile asm sources to working dir
tc.pack(): pack all .o files into the .a file (in the working dir)

and wrap this up:

```
func compile(target string, pkg *build.Package, tc toolchain) {
   wd := filepath.Join(opts.WorkDir, pkg.Dir)
  
   for _ src := range pkg.GoSrcs {
     // link srcs here
   }

   if cgo {
       tc.cgo(pkg.Dir, pkg.cogSrcs, wd)
   }
   tc.go(...)
   tc.asm(...)
   tc.cc(...)
   tc.pack(...)
}
```

Tatskaari:
Actually let me see if I can refactor this stuff quickly

sagikazarmark:
Recreated the branch on top of the latest master, but I don't fully understand the paths yet.

PR #1486, opened 2021-01-18
Fix please.build logo and codelabs font

Tatskaari:


docs/template.html
```diff
@@ -52,13 +52,13 @@
...
-      <header>
-        <a class="flex no-underline f2-ns f3 white calibre" href="/index.html"
+      <header class="cf pa3">
+        <a class="flex no-underline f2-ns f3 white calibre fl" href="/index.html"
           >please<span class="violet">.build</span></a
         >
         <button
                 id="menu-button"
-                class="white bg-transparent b--none f3 pointer dn-l"
+                class="white bg-transparent b--none f3 pointer dn-l fr"
```
tiagovtristao, line 61: I would use flex here since it's easier to reason about and has less quirks than floats.

`flex justify-between` to the header class should be enough to get the same result

Issue #1485, opened 2021-01-18
Support grpc_library for newer version?


linuxerwang:
The official protobuf support for go has change from github.com/golang/protobuf to google.golang.org/protobuf, and grpc from protoc-gen-go to protoc-gen-go-grpc.

See the stackoverflow discussion:
https://stackoverflow.com/questions/60578892/protoc-gen-go-grpc-program-not-found-or-is-not-executable

Shouldn't Please also be updated to the new verison (or support both)?

linuxerwang:
I created a demo implementation, PR #1488. It works with the following section in //third_party/go/BUILD.plz:

```
go_get(
    name="gce_metadata",
    get="cloud.google.com/go/compute/metadata/...",
    revision="v0.75.0",
)

go_get(
    name="genproto_rpccode",
    get="google.golang.org/genproto/googleapis/rpc/code/...",
    revision="6edceaf6022f0286b66b688465262f636c7f7f43",
    deps=[
        ":protobuf",
    ],
)

go_get(
    name="x_oauth2",
    get="golang.org/x/oauth2/...",
    revision="d3ed898aa8a312e7e3290a8ac5a123149de68011",
    deps=[
        ":gce_metadata",
        ":x_net",
    ],
)

go_get(
    name="x_crypto",
    get="golang.org/x/crypto/...",
    revision="4d3f4d9ffa16a13f451c3b2999e9c49e9750bf06",
    deps=[":x_sys"],
)

go_get(
    name="x_xerrors",
    get="golang.org/x/xerrors/...",
    revision="5ec99f83aff198f5fbd629d6c8d8eb38a04218ca",
)

go_get(
    name="x_net",
    get="golang.org/x/net/...",
    revision="74dc4d7220e7acc4e100824340f3e66577424772",
    deps=[
        ":x_crypto",
        ":x_text",
    ],
)

go_get(
    name="x_text",
    get="golang.org/x/text/...",
    revision="4e4a3210bb54bb31f6ab2cdca2edcc0b50c420c1",
    deps=[
        ":x_tools",
    ],
)

go_get(
    name="x_tools",
    get="golang.org/x/tools",
    install=[
        "cover",
    ],
)

go_get(
    name="protobuf-go",
    get="google.golang.org/protobuf/...",
    revision="v1.25.0",
    deps=[
        ":go-cmp",
    ],
)

go_get(
    name="protobuf",
    get="github.com/golang/protobuf/...",
    revision="v1.4.3",
    strip=[
[...]

PR #1488, opened 2021-01-19
Demonstrate how to make grpc_library use google.golang.org/protobuf

linuxerwang:
instead of github.com/golang/protubuf.

Not aimed for merge, because it's not backward compatible.

See issue #1485 for the context.


Tatskaari:
Amazing! Thanks for this! I think to make this backwards compatible, could you introduce a feature flag and only add the new plugin tools and flags if it has been set? See the bottom of `type Configuration struct {}` in `config.go`. 

If you don't have time, I can take over.

linuxerwang:
@Tatskaari Please take it over. I only have time at weekend, and I am not really familiar with Please's internals. This PR is just a demo.  Thank you!

src/core/config.go
```diff
@@ -525,6 +526,7 @@ type Configuration struct {
 		ProtocGoPlugin   string   `help:"The binary passed to protoc as a plugin to generate Go code. Defaults to protoc-gen-go.\nWe've found this easier to manage with a go_get rule instead though, so you can also pass a build label here. See the Please repo for an example." var:"PROTOC_GO_PLUGIN"`
 		GrpcPythonPlugin string   `help:"The plugin invoked to compile Python code for grpc_library.\nDefaults to protoc-gen-grpc-python." var:"GRPC_PYTHON_PLUGIN"`
 		GrpcJavaPlugin   string   `help:"The plugin invoked to compile Java code for grpc_library.\nDefaults to protoc-gen-grpc-java." var:"GRPC_JAVA_PLUGIN"`
+		GrpcGoPlugin     string   `help:"The plugin invoked to compile Go code for grpc_library.\nDefaults to protoc-gen-go-grpc." var:"GRPC_GO_PLUGIN"`
```
peterebden, line 529: Should we skip the feature flag and just operate on this config value? Default it to empty, but use it if it's set?

Tatskaari: Yeah I guess I wanted all the config options here to behave the same (i.e. look for the plugin on the path) but go seems to behave differently to the other languages anyway.

PR #1492, opened 2021-01-20
Fix pip_library in newer versions of Debian/python-pip

VJftw:
In 2017, this [change](https://github.com/thought-machine/please/commit/1fee49aa7b1f0123b092acc99b683730271610e5) was added to resolve this bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=830892. Following the bug thread down it has been resolved upstream since December 2018.

---

Without this removed, on Ubuntu 20.04 and running `./bootstrap.sh` in this repository, I get:
```    
...
//third_party/python:_grpcio#srcs
Error building target //third_party/python:_grpcio#srcs: exit status 2

Usage:   
  /home/vjftw/.asdf/installs/python/3.6.12/bin/python3 -m pip install [options] <requirement specifier> [package-index-options] ...
  /home/vjftw/.asdf/installs/python/3.6.12/bin/python3 -m pip install [options] -r <requirements file> [package-index-options] ...
  /home/vjftw/.asdf/installs/python/3.6.12/bin/python3 -m pip install [options] [-e] <vcs project url> ...
  /home/vjftw/.asdf/installs/python/3.6.12/bin/python3 -m pip install [options] [-e] <local project path> ...
  /home/vjftw/.asdf/installs/python/3.6.12/bin/python3 -m pip install [options] <archive url/path> ...

no such option: --system
```

the `--system` option no longer exists in the `python-pip` Debian package as it was only added as a workaround for this bug.

### DIFF
--- a/rules/python_rules.build_defs
+++ b/rules/python_rules.build_defs
@@ -392,11 +392,6 @@ def pip_library(name:str, version:str, hashes:list=None, package_name:str=None, 
         # but that prevents downloading binary wheels. This is more fiddly but seems to work.
         # Unfortunately it does *not* work similarly on the Debian problem :(
         pip_cmd = 'echo "[install]\nprefix=" > setup.cfg; ' + pip_cmd
-    if CONFIG.OS == 'linux' and not CONFIG.DISABLE_VENDOR_FLAGS:
-        # Fix for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=830892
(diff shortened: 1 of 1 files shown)


Tatskaari:
Sorry, Debian really dropped the ball on this one and caused huge problems with forwards and backward compatibility for us. Fixing it like this will break it for debian systems that still need the flag. There's a config option for this that you can set in your `.plzconfig`:

```
[python]
DisableVendorFlags = true
```

VJftw:
Ah perfect, thanks!

PR #1496, opened 2021-01-21
Enable the please_go_install feature flag

Tatskaari:


Tatskaari:
Seems mixing and matching between `go install` and `please_go_install` causes fingerprint errors. We'll need to migrate these all at once.

PR #1497, opened 2021-01-21
Allow depending on //:pleasings even if there's no BUILD file for //:all

Tatskaari:


src/parse/parse_step.go
```diff
@@ -194,7 +194,12 @@ func parsePackage(state *core.BuildState, label, dependent core.BuildLabel, subr
 		}
 	} else {
 		filename, dir := buildFileName(state, label.PackageName, subrepo)
-		if filename == "" {
+		if filename != "" {
+			pkg.Filename = filename
+			if err := state.Parser.ParseFile(state, pkg, pkg.Filename); err != nil {
+				return nil, err
+			}
+		} else if packageName != "" || state.Config.FeatureFlags.RemovePleasings {
```
Tatskaari, line 202: If it's not the root package, or if it is the root package but we don't add the psudo :pleasings rule there, print an error.

peterebden (approved):

PR #1499, opened 2021-01-23
Flatten paths for the generated *.pb.go files.

linuxerwang:


rules/proto_rules.build_defs
```diff
@@ -395,10 +395,15 @@ def _protoc_rule(name, srcs, deps, language, plugin, protoc_flags, root_dir, pre
         cmd += ' ${SRCS}'
 
     cmd = f'(export OUT_DIR=$TMP_DIR/{language_out_dir} && {cmd})'
+    cmd += f' && (mv -f {language_out_dir}/${{PKG_DIR}}/* {out_dir}; true)'
 
-    cmd += f' && (mv -f {language_out_dir}/${{PKG_DIR}}/* {out_dir}; true) && (mv -f {language_out_dir}/* {out_dir}; true)'
+    if language_out_dir == 'go':
+        # Uses find+xargs+mv to flatten the paths and move the *.pb.go files.
+        cmd += f' && (find {language_out_dir} -name "*.pb.go" | xargs -i mv -f {{} {out_dir}; true)'
```
Tatskaari, line 402: We can probably just set `out_dir` to the package directory? 

If protoc outputs `api.pb.go` and `api_grpc.pb.go` to `go/test.com/src/proto/api`, setting  `output_dirs = ["go/test.com/src/proto/api/**"]` would work right?

Tatskaari:
Does this work as it is? I wonder how we figure out the correct package to put in the `.importconfig` file.

linuxerwang:
> Does this work as it is? I wonder how we figure out the correct package to put in the `.importconfig` file.

It works for https://github.com/linuxerwang/goplz, because it simply maps plz-out/gen/foo/bar/api.pb.go to test.com/foo/bar/api.pb.go. If the generate file is at plz-out/gen/foo/bar/test.com/foo/bar/api.pb.go, it will be hard to map.

However, it does have problem. What if the proto specifies:
```
// $TOP/foo/bar/api/api.proto

option go_package = "example.com/foo/bar/api";
```
The generated file is at plz-out/gen/foo/bar/example.com/foo/bar/api.pb.go while the config.go.importpath is "test.com". A user of this package will depend with //foo/bar/api:api, but import will be "example.com/foo/bar/api". 

I am not sure if .importconfig can solve it, but IDEs will not understand .importconfig, right?

rules/proto_rules.build_defs
linuxerwang: Tried it, your code doesn't work. The reason is that my config.Go.ImportPath=**example.com**, and the proto has
```
option go_package = "example.com/foo/bar/api";
```
Now protoc will generate files like plz-out/tmp/foo/bar/api/_grpc_go#protoc_go._build/go/**example.com**/foo/bar/api/api_grpc.pb.go, while {pkg_name} only matches "foo/bar/api" without prefix "example.com", so the files will not be copied.

Tatskaari:
I think this has been fixed now. The issue was with us passing source relative in a way that was no longer being respected. 

See #1705

PR #1505, opened 2021-01-26
Make please_go_install leverage build.ImportDir()

Tatskaari:
While `build.ImportDir()` uses `build.Import()` which in turn can call out to `go list` for non-local imports, `build.ImportDir()` always calls `build.Import()` with `path = "."`, which means that the import path is always local.

### DIFF
--- a/tools/please_go_install/main.go
+++ b/tools/please_go_install/main.go
@@ -3,10 +3,7 @@ package main
 import (
 	"bufio"
 	"fmt"
-	"go/ast"
 	"go/build"
-	"go/parser"
-	"go/token"
 	"os"
 	"path/filepath"
 	"strings"
@@ -20,7 +17,7 @@ var opts = struct {
 	ImportConfig string   `short:"i" long:"importcfg" description:"the import config for the modules dependencies"`
 	GoTool       string   `short:"g" long:"go" description:"The location of the go binary"`
 	CCTool       string   `short:"g" long:"go" description:"The c compiler to use"`
-	Packages     []string `short:"p" long:"packages" description:"The target packages to list dependencies for" default:"."`
+	Packages     []string `short:"p" long:"packages" description:"The target packages to list dependencies for"`
 	Out          string
 }{
 	Usage: `
@@ -41,8 +38,6 @@ This tool determines the dependencies between packages and output a commands in 
 	Packages:     os.Args[7:],
 }
 
-var fileSet = token.NewFileSet()
-
 type pkgGraph struct {
 	pkgs map[string]bool
 }
@@ -51,6 +46,7 @@ func main() {
 	pkgs := &pkgGraph{
 		pkgs: map[string]bool{
 			"unsafe": true, // Not sure how many other packages like this I need to handle
+			"C":      true, // Pseudo-package for cgo symbols
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


tools/please_go_install/main.go
```diff
@@ -108,114 +104,70 @@ func (g *pkgGraph) compile(from []string, target string) {
...
 	pkgDir := filepath.Join(opts.SrcRoot, target)
-
 	// The package name can differ from the directory it lives in, in which case the parent directory is the one we want
 	if _, err := os.Lstat(pkgDir); os.IsNotExist(err) {
 		pkgDir = filepath.Dir(pkgDir)
 	}
 
-	packageASTs, err := parser.ParseDir(fileSet, pkgDir, nil, 0)
+	// TODO(jpoole): is import vendor the correct thing to do here?
+	pkg, err := build.ImportDir(pkgDir, build.ImportComment)
```
sagikazarmark, line 113: You could just pass `0` here. Do you need the comments for something (I believe the relevant cgo comments are parsed anyway, but I could be wrong)

tools/please_go_install/main.go
```diff
@@ -108,114 +104,70 @@ func (g *pkgGraph) compile(from []string, target string) {
...
-	generateCGO := fmt.Sprintf("%s tool cgo %s", opts.GoTool, strings.Join(cgoSrcs, " "))
+	cdPkgDir := fmt.Sprintf("cd %s", pkg.Dir)
+	generateCGO := fmt.Sprintf("%s tool cgo %s", opts.GoTool, strings.Join(pkg.CgoFiles, " "))
 	compileGo := fmt.Sprintf("%s tool compile -pack -importcfg $OLDPWD/%s -o out.a _obj/*.go", opts.GoTool, opts.ImportConfig)
+	// TODO(jpoole): We can use pkg to determine the correct cgo flags to pass to the compiler here
 	compileCGO := fmt.Sprintf("%s -Wno-error -ldl -Wno-unused-parameter -c -I _obj -I . _obj/_cgo_export.c _obj/*.cgo2.c", opts.CCTool)
 	compileC := fmt.Sprintf("%s -Wno-error -ldl -Wno-unused-parameter -c -I _obj -I . *.c", opts.CCTool)
 	mergeArchive := fmt.Sprintf("%s tool pack r out.a *.o ", opts.GoTool)
-	moveArchive := fmt.Sprintf("cd $OLDPWD && mv %s/out.a %s", pkgDir, out)
+	moveArchive := fmt.Sprintf("cd $OLDPWD && mv %s/out.a %s", pkg.Dir, out) //TODO(jpoole): can we just output this in the right place to begin with?
```
sagikazarmark, line 161: Good question. AFAIK the go build tool changes directory into the obj dir and runs all build tasks there.

peterebden: Yep pretty sure it does.

peterebden (approved):
OK this seems neater, as discussed careful that go/build doesn't decide to import the whole world but it looks legit.

tools/please_go_install/main.go
Tatskaari: Defo some room for improvement but i don't think this code is long for the world. We should be using your toolchain approach here.

tools/please_go_install/main.go
Tatskaari: Ah, I was wondering if this was responsible for parsing the linker flags and such. Will have to investigate.

PR #1515, opened 2021-02-01
Remove sqlite3 from cross compile test

Tatskaari:
Turns out passing the c flags to gcc causes us to run into this issue: https://github.com/mattn/go-sqlite3/issues/532

Presumably this code branch was getting ifdef'd out before. 

Might need to invest some time into getting a proper cross compiling toolchain on the CI agent.


peterebden (approved):

peterebden:
It's probably not easy to do that on the Mac workers. Maybe it's sufficient to just do it from one of the Linux ones and conditionally enable that test.

Issue #1517, opened 2021-02-02
Too many open files


sagikazarmark:
I'm hitting this issue lately on mac when using the new `go_module` rule:

```
08:27:21.034 WARNING: Failed to remove temporary directory for //third_party/go:_github.com__aws__aws-sdk-go_
_internal__sdkio#a_rule: readdirnames plz-out/tmp/third_party/go/_github.com__aws__aws-sdk-go__internal__sdkio#a_r
ule._build/third_party/go/src/github.com/aws/aws-sdk-go/models/apis/route53resolver: fcntl: too many open files
08:27:21.205 WARNING: Failed to remove temporary directory for //third_party/go:_github.com__aws__aws-sdk-go_
_aws__client__metadata#a_rule: readdirnames plz-out/tmp/third_party/go/_github.com__aws__aws-sdk-go__aws__client__
metadata#a_rule._build/third_party/go/src/github.com/aws/aws-sdk-go/models/apis/neptune/2014-10-31: fcntl: too man
y open files
```

Tatskaari:
Quite a lot of things have changed in this area. Are you still getting these issues or can I close this issue?

PR #1519, opened 2021-02-02
Close importcfg after parsing

Tatskaari:
Might help #1517

### DIFF
--- a/tools/please_go_install/main.go
+++ b/tools/please_go_install/main.go
@@ -45,27 +45,8 @@ func main() {
 		panic(err)
 	}
 
-	pkgs := &pkgGraph{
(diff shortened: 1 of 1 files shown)


peterebden (commented):
This isn't really going to change much, it won't close until main() returns at which point the process is exiting anyway?
It is a better thing to close it, I agree, but maybe you should move the body of that `if` to its own function?

Tatskaari:
> This isn't really going to change much, it won't close until main() returns at which point the process is exiting anyway?
> It is a better thing to close it, I agree, but maybe you should move the body of that `if` to its own function?

Ah good point... I'll keep hunting then.

Issue #1508, opened 2021-01-27
Source code debug information is not generated on MacOS


idemura:
I'm testing please build on a small project now. It is written in C, I use MacOS and clang 12:
```
Apple clang version 12.0.0 (clang-1200.0.32.2)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```
When I try to use LLDB on generated binary, I only see assembly. While the same binary works perfectly and I'm able to run it. This is my BUILD file:
```
c_library(
    name = "adt",
    srcs = glob(["adt/*.c"]),
    hdrs = glob(["adt/*.h"]),
    deps = [":types"],
)

c_binary(
    name = "adt_test",
    srcs = glob(["adt/tests/*.c"]),
    hdrs = glob(["adt/tests/*.h"]),
    deps = [
        ":adt",
        ":testing",
    ],
)

c_library(
    name = "testing",
    srcs = glob(["testing/*.c"]),
    hdrs = glob(["testing/*.h"]),
    deps = [":types"],
)

c_library(
    name = "types",
    srcs = glob(["types/*.c"]),
    hdrs = glob(["types/*.h"]),
)
```
and plz_config:
```
[Cpp]
CCTool = clang
DefaultOptCflags = --std=c17 -Wall -Werror -march=native -O2 -DNDEBUG
DefaultDbgCflags = --std=c17 -Wall -Werror -march=native -O0 -g
```
I build with `-c dbg`. I also use `-v debug` to make sure command lines correct. As I was suggesting on gitter, I add logs for 
`plz build --shell -c dbg` but it unfortunately fails:
```
igor@mbp owl_llvm % plz build --shell -c dbg
Build stopped after 30ms. 1 target failed:
    //src:types
can't prepare temporary directory for //src:types; filegroups don't have temporary directories
```
But works ok without --shell.

idemura:
PS: Please was installed with Homebrew

Tatskaari:
@peterebden  Any idea what might be going on here? I've not set up clang/gcc with debug flags before. 

Re: --shell not working there: unfortunately c_library() and c_binary() return multiple types of results (headers, and compiled artifacts) so these are grouped up as a filegroup. It might make more sense if you `plz query print //src:all` to see what's going on. 

Essentially, if you want to use `--shell`, you have to target the internal rule directly: `plz build -c dbg --shell //src:_types#lib`

[...]

PR #1522, opened 2021-02-02
Added .dSYM support for c debug symbols on macOS

Tatskaari:
Fixes: #1508


src/core/config.go
```diff
@@ -549,6 +549,7 @@ type Configuration struct {
 		Coverage           bool       `help:"If true (the default), coverage will be available for C and C++ build rules.\nThis is still a little experimental but should work for GCC. Right now it does not work for Clang (it likely will in Clang 4.0 which will likely support --fprofile-dir) and so this can be useful to disable it.\nIt's also useful in some cases for CI systems etc if you'd prefer to avoid the overhead, since the tests have to be compiled with extra instrumentation and without optimisation." var:"CPP_COVERAGE"`
 		TestMain           BuildLabel `help:"The build target to use for the default main for C++ test rules." example:"///pleasings//cc:unittest_main" var:"CC_TEST_MAIN"`
 		ClangModules       bool       `help:"Uses Clang-style arguments for compiling cc_module rules. If disabled gcc-style arguments will be used instead. Experimental, expected to be removed at some point once module compilation methods are more consistent." var:"CC_MODULES_CLANG"`
+		DsymTool           string     `help:"Set this to dsymutil or equivalent on macOS to use this tool to generate xcode symbol information for debug builds." var:"DSYM_TOOL"`
```
peterebden, line 552: Should we default this to something on OSX? or no?

peterebden (approved):

src/core/config.go
Tatskaari: It's only if you're using the clang that comes with xcode so I don't think it's necessarily the right thing to do.

PR #1532, opened 2021-02-04
Migrate some of the go_get rules to go_module()

Tatskaari:
Need to implement `patch` in order to finish this off.


tools/please_go_install/BUILD
```diff
@@ -1,11 +1,28 @@
+#TODO(jpoole): once bugs are fixed, make this use the please_go_install in ~/.please
+package(FF_PLEASE_GO_INSTALL = False)
```
Tatskaari, line 2: Actually these should probably use go_get() otherwise it will break the bootstrap story

rules/go_rules.build_defs
```diff
@@ -943,13 +943,14 @@ def go_mod_download(name:str, module:str, version:str, test_only:bool=False, vis
...
         cmd = ' && '.join(cmds),
-        requires = ['go'],
+        requires = ['go_src'],
         test_only = test_only,
         labels = labels + ['link:plz-out/go'],
         hashes = hashes,
         sandbox = False,
         licences = licences,
         visibility = visibility,
+        deps = deps,
```
Tatskaari, line 953: This allows two download rules to output to the same path as they become dependent on each-other avoiding race conditions

PR #1536, opened 2021-02-05
Add env to build_rule() and go_get()

Tatskaari:
This allows the user to set arbitrary env variables on their build targets. Beats having to export them manually. 

I've added these to go_get() and go_module() because i foresee needing to set the CGO env variables like this. 

I should probably add these elsewhere too but I will do that in a followup PR. Can also probs clean up a lot of `export FOO=bar` stuff too.


peterebden (approved):
Seems legit. I've thought a few times before about having something like this, but obviously never did it.

PR #1541, opened 2021-02-06
Adding a python toolchain 

namtzigla:
In this PR I have a function python_toolchain that's not necessary to be included into the standard python rules, you could leave it to the user to implement that function.
The rest of the changes are there to allow for python rules to consume a toolchain.


rules/python_rules.build_defs
```diff
@@ -129,9 +216,8 @@ def python_binary(name:str, main:str, srcs:list=[], resources:list=[], out:str=N
                      determine something appropriate for the given interpreter.
       labels (list): Labels to apply to this rule.
     """
-    shebang = shebang or interpreter or CONFIG.DEFAULT_PYTHON_INTERPRETER
-    cmd = '$TOOLS_PEX -s "%s" -m "%s" --zip_safe --interpreter_options="%s" --stamp="$STAMP"' % (
-        shebang, CONFIG.PYTHON_MODULE_DIR, CONFIG.PYTHON_INTERPRETER_OPTIONS)
+    cmd = '$TOOLS_PEX -s "$TOOLS_INTERPRETER" -m "%s" --zip_safe --interpreter_options="%s" --stamp="$STAMP"' % (
```
namtzigla, line 219: This command create a pex file with the shebang like this:
```
#!/home/tom/dev/tcn/m/neo/plz-out/bin/tools/python/python/bin/python -S
```
IMO we should have a shebang like this: 
```
#!/usr/bin/env /home/tom/dev/tcn/m/neo/plz-out/bin/tools/python/python/bin/python -S || /usr/bin/env python3 -S
```
to ensure the resulted pex is portable. No matter how I tried I wasn't able to generate that.
If you are interested in that I could use a hand here.

rules/python_rules.build_defs
```diff
@@ -237,10 +323,10 @@ def python_test(name:str, srcs:list, data:list|dict=[], resources:list=[], deps:
       test_runner (str): Specify which Python test runner to use for these tests. One of
                          `unittest`, `pytest`, or a custom test runner entry point.
     """
-    interpreter = interpreter or CONFIG.DEFAULT_PYTHON_INTERPRETER
+    # interpreter = interpreter or CONFIG.DEFAULT_PYTHON_INTERPRETER
     test_runner = test_runner or CONFIG.PYTHON_TEST_RUNNER
-    cmd = '$TOOLS_PEX -t -s "%s" -m "%s" -r "%s" --zip_safe --add_test_runner_deps --interpreter_options="%s" --stamp="$STAMP"' % (
-        interpreter, CONFIG.PYTHON_MODULE_DIR, test_runner,
+    cmd = '$TOOLS_PEX -t -s "$TOOLS_INTERPRETER" -m "%s" -r "%s" --zip_safe --add_test_runner_deps --interpreter_options="%s" --stamp="$STAMP"' % (
```
namtzigla, line 328: see my above comment

PR #1542, opened 2021-02-08
feat: add labels to go_binary

sagikazarmark:
Would it make sense to add `pass_env` as well?

I'd use it in a setup like this:

```starlark
go_binary(
    name = "todo",
    srcs = glob(["*.go"], exclude = ["*_test.go"]),
    definitions = {
        "main.version": "${VERSION:-" + git_branch() + "}",
        "main.commitHash": git_commit()[0:8],
        "main.buildDate": f'$(date -u -d "@{timestamp}" "{date_fmt}" 2>/dev/null || date -u -r "{timestamp}" "{date_fmt}" 2>/dev/null || date -u "{date_fmt}")',
    },
    #labels = ["binary"],
    #pass_env = ["VERSION"],
    #trimpath = True,
    visibility = ["PUBLIC"],
)
```

### DIFF
--- a/rules/go_rules.build_defs
+++ b/rules/go_rules.build_defs
@@ -453,7 +453,7 @@ def _merge_cgo_obj(name, a_rule, o_rule=None, visibility=None, test_only=False, 
 
 
 def go_binary(name:str, srcs:list=[], asm_srcs:list=[], out:str=None, deps:list=[], data:list=None,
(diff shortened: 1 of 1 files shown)


Tatskaari (approved):

Tatskaari:
I think the reason we didn't was mostly because we didn't see the use case. I'm not against adding more of these things to go_binary if they're useful.

PR #1548, opened 2021-02-09
Fix expand stdin for plz run

Tatskaari:


src/please.go
```diff
@@ -440,7 +440,17 @@ var buildFunctions = map[string]func() int{
 				opts.Run.Args.Target.Annotation = opts.Run.EntryPoint
 			}
 
-			run.Run(state, opts.Run.Args.Target, opts.Run.Args.Args.AsStrings(), opts.Run.Remote, opts.Run.Env, dir)
+			annotatedOutputLabels := []core.AnnotatedOutputLabel{opts.Run.Args.Target}
+			if utils.ReadingStdinAnnnotated(annotatedOutputLabels) {
```
peterebden, line 444: I'm not sure why we have to  do it only if we're reading stdin. Aren't there still cases where we need to expand `/...` etc as well?

Tatskaari: Yeah that happens on line 448 below in either case.

PR #1551, opened 2021-02-09
Link codegen back into the repo

Tatskaari:
This adds a new command `plz generate` that will build any rules with the `codegen` label. It is assumed that these rules will also have a label `link:{package_name()}` so that their sources will be linked back to the source directory. 

To help keep the `.gitignore` up to date, this sub-command also has an `--update_gitignore path/to/.gitignore` argument. When provided, Please will add any generated files under that directory to that gitignore.  These files are added under a warning message. Please will not clobber any entries above this message enabling users to ignore other files not generated by Please too. 

I've created an example monorepo project to demonstrate how this might work: https://github.com/Tatskaari/please-codegen-example


rules/misc_rules.build_defs
```diff
@@ -602,3 +602,10 @@ def is_platform(os:str|list=CONFIG.OS, arch:str|list=CONFIG.ARCH, host_os:str|li
     host_arch = as_list(host_arch)
 
     return CONFIG.OS in os and CONFIG.ARCH in arch and CONFIG.HOSTOS in host_os and CONFIG.HOSTARCH in host_arch
+
+def link_generated_outs(name, output):
+    if CONFIG.LINK_GEN_SOURCES:
+        pkg = package_name()
+        add_label(name, f"link:{pkg}")
+        for out in get_outs(name):
+            add_label(name, f"gitignore:/{pkg}/{out}")
```
Tatskaari, line 611: I guess `generated-file:/{pkg}/{out}` might make more sense.

Tatskaari:
An alternative to this approach would be to do away with the `hlink:` labels and just have `plz codegen` do the linking for us. This might prove simpler than using the post build function.

sagikazarmark:
Will it be possible to commit the generated files in the repo? (I guess not with linking and not copying files)

Another concern: what if the generated assets (and as a result the generated gitignore) change (eg. a new file is added, an old one is removed). That could cause problems. I'm not saying Please should resolve that situation (it's probably an edge case), but it's probably good to know.

Tatskaari:
> Will it be possible to commit the generated files in the repo? (I guess not with linking and not copying files)
Do you have a usecase for this? (actually I wonder if that's something we want to do with Please itself instead of having `//rules:generate`)

> Another concern: what if the generated assets (and as a result the generated gitignore) change (eg. a new file is added, an old one is removed). That could cause problems. I'm not saying Please should resolve that situation (it's probably an edge case), but it's probably good to know.
Can't you just run `//:update_gitignore` again?

sagikazarmark:
> Do you have a usecase for this?

Yes, Please 😄 

> Can't you just run //:update_gitignore again?

The concern is not the gitignore, but the links/files left in the source tree that might show up in the stage if it's removed from gitignore.

Tatskaari:
> > Do you have a usecase for this?
> 
> Yes, Please 
> 
> > Can't you just run //:update_gitignore again?
> 
> The concern is not the gitignore, but the links/files left in the source tree that might show up in the stage if it's removed from gitignore.

It's possible to write a test to make sure the output of `//:update_gitignore` and `.gitignore` are in sync.

sagikazarmark:
> It's possible to write a test to make sure the output of

The problem is not .gitignore being in sync. Let's say a file is removed. You update .gitignore, it gets removed from .gitignore, but the link/file remains in the source tree. You have to make sure not to add it to git (accidentally).

Tatskaari:
> > It's possible to write a test to make sure the output of
> 
> The problem is not .gitignore being in sync. Let's say a file is removed. You update .gitignore, it gets removed from .gitignore, but the link/file remains in the source tree. You have to make sure not to add it to git (accidentally).

Yeah... I guess that just needs to be caught in code review :(

sagikazarmark:
I was thinking of some sort of warning based on the diff with the current and the new gitignore. But that can probably be scripted easily.

go.mod
```diff
@@ -1,5 +1,9 @@
 module github.com/thought-machine/please
 
+replace (
+    github.com/bazelbuild/remote-apis-sdks => github.com/peterebden/remote-apis-sdks master
```
peterebden, line 4: eeeeeh we sure we want to do this? there is some... experimentashone happening in there

rules/misc_rules.build_defs
```diff
@@ -602,3 +602,13 @@ def is_platform(os:str|list=CONFIG.OS, arch:str|list=CONFIG.ARCH, host_os:str|li
     host_arch = as_list(host_arch)
 
     return CONFIG.OS in os and CONFIG.ARCH in arch and CONFIG.HOSTOS in host_os and CONFIG.HOSTARCH in host_arch
+
+def link_generated_outs(gen_srcs=CONFIG.LINK_GEN_SOURCES):
```
peterebden, line 606: We sure we want to do this in post-build functions? why can't it be done statically?

src/please.go
```diff
@@ -367,6 +368,12 @@ var opts struct {
 			} `positional-args:"true"`
 		} `command:"filter" description:"Filter the given set of targets according to some rules"`
 	} `command:"query" description:"Queries information about the build graph"`
+	Codegen struct {
+		Gitignore string `long:"update_gitignore" description:"The gitignore file to write the generated sources to"`
+		Args   struct {
+			Targets []core.BuildLabel `positional-arg-name:"targets" description:"Targets to filter"`
+		} `positional-args:"true"`
+	} `command:"codegen" description:"Builds all code generation targets in the repository and prints the generated files."`
```
peterebden, line 376: I sort of feel like this should really be `plz gencode` so it's a verb but `codegen` is more common...

src/codegen/codegen.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/thought-machine/please/src/core"
+)
+
+func UpdateGitignore(graph *core.BuildGraph, labels []core.BuildLabel, gitignore string) error {
```
peterebden, line 12: think this should probably come from the `scm` package and be abstracted by that?

go.mod
Tatskaari: Eeeep sorry I didn't realise this ended up here! I was just trying to get `go install` to work! This was only meant to exist on tatskaari/master

rules/misc_rules.build_defs
Tatskaari: We don't know the outputs until it's built as it's using output directories.

src/please.go
Tatskaari: Yeah good point. We could steal go's nomenclature: `go generate`

PR #1562, opened 2021-02-15
Enable build-rule authors to access a stable directory for global files

linuxerwang:
Introduce a global directory plz-out/global. Build rule authors can use
this directory for files not related to any specific build target. At build-
time the directory can be accessed through environment variable
GLOBAL_DIR.

One use case of global directory is the dart pub cache
(https://dart.dev/tools/pub/cmd/pub-cache):

Dart tools heavily rely on the pub cache, but the cache itself is
location sensitive thus will not work if it is put in TMP_DIR. When dart
tools run, absolute file paths to the pub cache are written to certain
config files, so that a config file generated in plz-out/tmp will not
work in plz-out/gen.

Another issue is that "pub global activate <exe>" compiles a special
snapshot file as an executable (it is binary file but not real
executable files, only the dart interpreter can run it), which is also
location sensitive.

These issues can be solved nicely with GLOBAL_DIR, for example, by
introducing a folder plz-out/global/dart/pub_cache, without breaking
Please's hermeticity.

### DIFF
--- a/src/core/build_env.go
+++ b/src/core/build_env.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"sync"
@@ -31,6 +32,7 @@ func GeneralBuildEnvironment(state *BuildState) BuildEnv {
 		// It's easier to just make these available for Go-based rules.
 		"GOARCH=" + state.Arch.GoArch(),
 		"GOOS=" + state.Arch.OS,
+		"GLOBAL_DIR=" + filepath.Join(RepoRoot, GlobalDir),
 	}
 	if state.Config.Cpp.PkgConfigPath != "" {
 		env = append(env, "PKG_CONFIG_PATH="+state.Config.Cpp.PkgConfigPath)
--- a/src/core/build_target.go
+++ b/src/core/build_target.go
@@ -2,7 +2,6 @@ package core
 
 import (
 	"fmt"
-	"github.com/thought-machine/please/src/fs"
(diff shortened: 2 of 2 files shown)


peterebden:
I don't think this is a good idea. Surely it _does_ break hermeticity; rules could now access things from this global directory that they haven't depended on; targets won't get rebuilt correctly as it changes, there could be issues with concurrent reads/writes, etc.

linuxerwang:
> I don't think this is a good idea. Surely it _does_ break hermeticity; rules could now access things from this global directory that they haven't depended on; targets won't get rebuilt correctly as it changes, there could be issues with concurrent reads/writes, etc.

I see. Good to know. So what's the best way to solve this issue? Thanks.

Tatskaari:
I think I agree with Peter. This begins to undermine some of the core tenants of Please so isn't something I'd want first class support for. I'm not sure how the dart cache works, but it seems to be inherently non-hermetic. Is dart effectively acting as its own build system? Maybe there's another level you can hook in at. For example, we don't use `go build` for our go rules, we use `go tool compile/link`. Does dart have something similar?

linuxerwang:
> I think I agree with Peter. This begins to undermine some of the core tenants of Please so isn't something I'd want first class support for. I'm not sure how the dart cache works, but it seems to be inherently non-hermetic. Is dart effectively acting as its own build system? Maybe there's another level you can hook in at. For example, we don't use `go build` for our go rules, we use `go tool compile/link`. Does dart have something similar?

I think I need to know more about the dart vm build mechanisms. Will abandon this PR.

PR #1564, opened 2021-02-15
Fix remote cross compile

peterebden:
There is a really nasty and complex interaction here.

- Remote execution needs to get the right state object for cross-compiling. Right now it doesn't do that properly.
- However it _also_ has a modified state object to set the remote home dir. Getting the cross-compiled one loses that information and so things break.
- The ultimate solution is to use PleaseDownloadTools & the correct state object which fixes everything.
- However to have a sensible transition to get there we need to check that flag for now.

This seems as good as I can do for now. Definitely looking forward to v16 to simplify some of this code.

### DIFF
--- a/src/remote/action.go
+++ b/src/remote/action.go
@@ -75,10 +75,11 @@ func (c *Client) buildAction(target *core.BuildTarget, isTest, stamp bool) (*pb.
 
 // buildCommand builds the command for a single target.
 func (c *Client) buildCommand(target *core.BuildTarget, inputRoot *pb.Directory, isTest, isRun, stamp bool) (*pb.Command, error) {
+	state := c.state.ForTarget(target)
 	if isTest {
-		return c.buildTestCommand(target)
+		return c.buildTestCommand(state, target)
 	} else if isRun {
-		return c.buildRunCommand(target)
+		return c.buildRunCommand(state, target)
 	}
 	// We can't predict what variables like this should be so we sneakily bung something on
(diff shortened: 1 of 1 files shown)


src/remote/action.go
```diff
@@ -73,12 +73,24 @@ func (c *Client) buildAction(target *core.BuildTarget, isTest, stamp bool) (*pb.
...
 
+// stateForTarget returns an appropriate state for the current target.
+// TODO(peterebden): This is very much a limitation of the current setup; there is a complex interaction between how we set
+//                   HOME and how we get the correct state object for cross-compiling.
+//                   When we release v16 and make this the default we should be able to significantly simplify this.
+func (c *Client) stateForTarget(target *core.BuildTarget) *core.BuildState {
+	if !c.state.Config.FeatureFlags.PleaseDownloadTools {
+		return c.state
+	}
+	return c.state.ForTarget(target)
```
Tatskaari, line 84: Can't we just do:
```
s := c.state.ForTarget(target)
s.Config.HomeDir = s.Config.Remote.HomeDir
return s
```

Tatskaari (approved):
Seems okay. Might be able to improve it a bit if I'm not mistaken? Otherwise go ahead and land.

src/remote/action.go
peterebden: No, because we don't want to change it for anything that executes locally using that same state object.

PR #1569, opened 2021-02-16
Improve the error messages returned from please_go_install

Tatskaari:


peterebden (approved):
Test failure seems worth a look but looks good in general.

Tatskaari:
> Test failure seems worth a look but looks good in general.

Bah my empty dir didn't get checked in for obvious reasons which was changing the error messge...

PR #1570, opened 2021-02-16
Update Please to use go 1.16

Tatskaari:
Should probably build all the tools but right now it's okay because the main .plzconfig is set to build targets


peterebden (approved):

src/core/config.go
```diff
@@ -111,6 +111,32 @@ func defaultConfigFiles() []string {
...
+			config.Please.Location = "~/.please"
+		} else {
+			config.Please.Location = path.Dir(deref)
+		}
+	}
+
+	config.PleaseLocation = fs.ExpandHomePathTo(config.Please.Location, config.HomeDir)
+	if !filepath.IsAbs(config.PleaseLocation) {
+		config.PleaseLocation = filepath.Join(MustFindRepoRoot(), config.PleaseLocation)
+	}
```
Tatskaari, line 137: This is new

PR #1571, opened 2021-02-16
Support Go 1.16 file embedding

peterebden:
This will fail various builds, need #1570 to go in first I think.


tools/please_go_embed/embed/embed.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Parse parses the given files and returns the embed information in them.
+func Parse(gofiles []string) (*EmbedCfg, error) {
+	cfg := &EmbedCfg{
+		Patterns: map[string][]string{},
+		Files:    map[string]string{},
+	}
+	for _, dir := range dirs(gofiles) {
+		pkg, err := build.ImportDir(dir, build.ImportComment)
```
Tatskaari, line 23: Seems like there's a lot of duplicated logic between the build rules, please_go_install, please_go_filter, and now this. I wonder if the build rules could use this new `please_go` tool we're looking at instead of `go tool compile`? This tool could use the same toolchain struct that "please_go_install" uses and they could share all the asm and cgo logic for compiling a single package. It would also remove the need for please_go_filter.

peterebden: Yup I agree the end state should be one tool for all of these things. Wanted to get on with this now though to see what would be involved & prove it doesn't need any magic we can't provide.

Wrapping `go tool compile` is a little orthogonal I think, might be easier in terms of flags though.

PR #1576, opened 2021-02-18
Add v15 as a release branch

Tatskaari:


peterebden (approved):
I think this is all we have to do. 

The release scripts won't necessarily do anything sensible with latest_version though... might be OK as long as master is a beta?

Tatskaari:
> I think this is all we have to do.
> 
> The release scripts won't necessarily do anything sensible with latest_version though... might be OK as long as master is a beta?

Yeah, that's what I believe will be the case too.

PR #1582, opened 2021-02-19
Remove dl tools ff and make update just download plz

Tatskaari:


src/core/config.go
```diff
@@ -443,6 +441,7 @@ type Configuration struct {
 		HashFunction         string       `help:"The hash function to use internally for build actions." options:"sha1,sha256"`
 		ExitOnError          bool         `help:"True to have build actions automatically fail on error (essentially passing -e to the shell they run in)." var:"EXIT_ON_ERROR"`
 		LinkGeneratedSources bool         `help:"If set, supported build definitions will link generated sources back into the source tree. The list of generated files can be generated for the .gitignore through 'plz query print --label gitignore: //...'. Defaults to false." var:"LINK_GEN_SOURCES"`
+		UseToolsFromPath     bool         `help:"If set, Please will use Please will look for its tools in it's executables directory, and on the path, rather than downloading them. Defaults to False."`
```
peterebden, line 444: Do we even want to support this at all in v16?

Tatskaari: As discussed, not all users will want to download the tools however they can just set their `.plzconfig` accordingly. Will remove.

peterebden (approved):
Looks good but I think it needs a rebase/merge

Issue #1364, opened 2020-11-06
Allow single items for TOOLS


Tatskaari:
It's almost never useful to provide a list of tools to the tools argument. We should make tools accept strings:

```
tools = CONFIG.MY_TOOL,
```
and 
```
tools = {
   "MY_TOOL": CONFIG.MY_TOOL,
},
```

Tatskaari:
@PeterBocan This is a fairly easy one to pick up if you were looking for a different ticket :)

PeterBocan:
sign me up, @Tatskaari

PR #1583, opened 2021-02-20
Add support for single strings in the TOOLS argument

PeterBocan:
This change adds support for strings addresses the #1364


Tatskaari (approved):
Thanks Peter! Nice work.

PR #1584, opened 2021-02-22
Attach metadata to outgoing remote execution requests

peterebden:
Discovered this in the proto recently and I think we could definitely make use of it. Right now I'm only attaching it to execution requests - it could be sent with others too but I don't think we would do much with it so haven't bothered for now.

This also introduces a build id as the "correlated invocations id"; for now it is scoped to remote only. Dunno if we want to make this a more general concept (I feel it should be treated with some care).

There's also a "tool invocation id" field, which I think is roughly aimed at actions that contribute towards a single final target, but we don't really have any concept of that.


Tatskaari (approved):
Oh nice! Does this show up in BB at all?

peterebden:
> Oh nice! Does this show up in BB at all?

I don't think so. I can't see any references to the message in bb-browser.

TBH right now it can't since it doesn't get persisted anywhere. I think that is a bit of a drawback to the metadata-based approach (but also means it isn't part of the serialisation so altering it doesn't require anything rebuilding, which I imagine is why it's done the way it is)

PR #1585, opened 2021-02-22
Implemented a native file rule

Tatskaari:
Needs testing


rules/misc_rules.build_defs
```diff
@@ -456,6 +456,43 @@ def remote_file(name:str, url:str|list, hashes:list=None, out:str=None, binary:b
         entry_points = entry_points,
     )
 
+def text_file(name:str, content:str, expand_content:bool=True, hashes:list=None, out:str=None, binary:bool=False,
```
sagikazarmark, line 459: Wouldn't a builtin rule be better?

Tatskaari: This is how filegroup() and remote_file() are implemented. It's essentially a built in rule but it shares most of the params with `build_rule()` so we leverage that.

sagikazarmark: Asking because this requires to add an extra parameter to `build_rule`.

Tatskaari: Yeah I share your concern and I did actually talk with Peter about this. It's engineered this way because everything that generates a build target should be a build rule. Build rules are the primitive and introducing a new primitive probably needs moderate consideration that I'm not that excited to get into right now.

sagikazarmark: I see, makes sense.

PR #1586, opened 2021-02-23
Update remote-apis-sdks to latest

peterebden:
This rolls in a bunch of breaking changes from upstream.

All the stuff on my fork that we want has been upstreamed too so we can get off that \o/

### DIFF
--- a/src/remote/action.go
+++ b/src/remote/action.go
@@ -11,7 +11,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/bazelbuild/remote-apis-sdks/go/pkg/chunker"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/digest"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/filemetadata"
 	"github.com/bazelbuild/remote-apis-sdks/go/pkg/uploadinfo"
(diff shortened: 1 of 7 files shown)


src/remote/remote.go
```diff
@@ -170,13 +170,14 @@ func (c *Client) initExec() error {
 	if err != nil {
 		return err
 	}
+	client.DefaultRPCTimeouts["default"] = time.Duration(c.state.Config.Remote.Timeout)
```
peterebden, line 173: I don't know if we can do something more sensible here. Or just lose the config option completely maybe?

src/remote/action.go
```diff
@@ -27,27 +27,27 @@ import (
 func (c *Client) uploadAction(target *core.BuildTarget, isTest, isRun bool) (*pb.Command, *pb.Digest, error) {
 	var command *pb.Command
 	var digest *pb.Digest
-	err := c.uploadBlobs(func(ch chan<- *chunker.Chunker) error {
+	err := c.uploadBlobs(func(ch chan<- *uploadinfo.Entry) error {
```
Tatskaari, line 30: I don't understand this new nomenclature but whatever

src/remote/action.go
```diff
@@ -273,14 +273,14 @@ func (c *Client) uploadInputDir(ch chan<- *chunker.Chunker, target *core.BuildTa
 	}
 	if !isTest && target.Stamp {
 		stamp := core.StampFile(target)
-		chomk := chunker.NewFromBlob(stamp, int(c.client.ChunkMaxSize))
+		entry := uploadinfo.EntryFromBlob(stamp)
 		if ch != nil {
-			ch <- chomk
+			ch <- entry
```
Tatskaari, line 278: nit: emtry

src/remote/remote.go
Tatskaari: What's this used for? General timeouts for all RPC calls? I can't imagine many people setting this. The only reason I can think to have it configurable is if we need to increase/decrease based on the SLAs you've specced your environment for.

Tatskaari (approved):
Looks good to me. Thanks for that.

src/remote/action.go
peterebden: I think an `uploadinfo.Entry` is a more abstract thing you're going to upload, a `Chunker` breaks one of those into pieces to push it.

In the new version the former is more relevant here (you can't ask the Chunker for its digest, for example)

src/remote/action.go
peterebden: I miss `chomk`

PR #1591, opened 2021-02-24
Remove pleasings

Tatskaari:
I have a new vision for adding new languages to the repo so I've removed the prompt to add pleasings when you create a new project. 

We talked about trying to formalise what a language "module" or "plugin" is. I don't see Pleasings being very central. Instead each language will have their own repo (and the core languages will be moved out to these repos). In this world, Pleasings will effectively just become a collection of incubating build rules that can help anybody get started with that language. As they mature they will get promoted to their own repo. I don't see them being used very much.

I also removed the prompt to add go support during `plz init`. Are we going to do this for every language? That would be annoying.


src/core/config.go
```diff
@@ -550,7 +540,7 @@ type Configuration struct {
 		DefaultLdflags     string     `help:"Linker flags passed to all C++ rules.\nBy default this is empty." var:"DEFAULT_LDFLAGS"`
 		PkgConfigPath      string     `help:"Custom PKG_CONFIG_PATH for pkg-config.\nBy default this is empty." var:"PKG_CONFIG_PATH"`
 		Coverage           bool       `help:"If true (the default), coverage will be available for C and C++ build rules.\nThis is still a little experimental but should work for GCC. Right now it does not work for Clang (it likely will in Clang 4.0 which will likely support --fprofile-dir) and so this can be useful to disable it.\nIt's also useful in some cases for CI systems etc if you'd prefer to avoid the overhead, since the tests have to be compiled with extra instrumentation and without optimisation." var:"CPP_COVERAGE"`
-		TestMain           BuildLabel `help:"The build target to use for the default main for C++ test rules." example:"///pleasings//cc:unittest_main" var:"CC_TEST_MAIN"`
+		TestMain           string `help:"The build target to use for the default main for C++ test rules." example:"///pleasings//cc:unittest_main" var:"CC_TEST_MAIN"`
```
peterebden, line 543: Why isn't this a build label any more? I don't think there is a use case for this being anything else?

Tatskaari: `BuildLabel{}` was being treated as `//` in the build language. I suppose it should be a *BuildLabel. I didn't think of that.

PR #1593, opened 2021-02-25
Stop depending on go sources for installs

Tatskaari:


peterebden:
I guess if everything builds it's fine. FWIW it was hard to get rid of these in go_get but hopefully we are better off here.

PR #1594, opened 2021-02-25
Ensure test inputs are downloaded before opening the shell

peterebden:
Most of this is commonising other bits of code.

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -970,6 +970,44 @@ func (state *BuildState) ForConfig(config ...string) *BuildState {
 	return s
 }
 
+// DownloadInputsIfNeeded downloads all the inputs (or runtime files) for a target if we are building remotely.
+func (state *BuildState) DownloadInputsIfNeeded(tid int, target *BuildTarget, runtime bool) error {
+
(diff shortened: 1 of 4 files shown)


src/core/state.go
```diff
@@ -970,6 +970,43 @@ func (state *BuildState) ForConfig(config ...string) *BuildState {
...
+		for _, datum := range target.AllData() {
+			ch <- datum
+		}
+		for _, datum := range target.TestTools() {
+			ch <- datum
+		}
+		close(ch)
+	}()
+	return ch
+}
```
Tatskaari, line 1008: Do we not also need to download the output of the rule i.e. the test binary?

Tatskaari (approved):
Looks good, just one query.

src/core/state.go
peterebden: I lifted this from the existing implementation in remote/action.go...

Good point but it does seem to be working fairly consistently 🤔

peterebden: I think we implicitly download the test binary at the build step... 😬

PR #1595, opened 2021-02-25
Skip dirs named 'testdata' when walking

peterebden:
I believe this is what `go build` does; there are heaps of directories in `x/tools` that are not correctly buildable, for example.


Tatskaari (approved):
Ah nice one I was struggling with this and didn't realise that's what go build did

PR #1599, opened 2021-02-27
Remove obsolete remote homedir stuff

peterebden:


Tatskaari (approved):
Don't really understand the history here but this looks harmless

peterebden:
> Don't really understand the history here but this looks harmless

Before we downloaded all the tools we had to assume they were on the PATH somewhere, either globally or installed under $HOME on the remote agent as they would be locally.
This gets rid of the last remnant of that, and good riddance too, it has always been yuck.

PR #1600, opened 2021-03-01
Move please_go_install into go_tool

Tatskaari:


tools/please_go/please_go.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/thought-machine/please/tools/please_go/install"
+)
+
+var opts = struct {
+	Usage string
+
+	PleaseGoInstall struct {
+		SrcRoot      string `short:"r" long:"src_root" description:"The src root of the module to inspect" default:"."`
+		ModuleName   string `short:"n" long:"module_name" description:"The name of the module"`
```
peterebden, line 17: are any of these required?

peterebden (approved):

tools/please_go/please_go.go
Tatskaari: Yeah all of them except SrcRoot. Ah I see they're optional by default. Let me fix that.

PR #1601, opened 2021-03-01
Use go 1.16 embed feature to embed builtin rules

PeterBocan:
This feature replaces generated `rules.go` file with Go 1.16 embedding system.

Requires Go 1.16+. 

Merge with care.


peterebden (commented):
Thanks for this!

This does change the behaviour a bit; it used to precompile the builtin rules and load that, now it doesn't. Might be interesting to investigate what the difference is there, given that the preloaded rules need to be loaded initially before anything else can be done.

I also intended to wait a bit before requiring Go 1.16 since there are platforms where it may not yet be readily available. I think Homebrew is now there which is one of the major ones I had my eye on.

PeterBocan:
Yup, will investigate!

Tatskaari:
I've battled the parser quite a bit in the past because I've changed something about the AST and ended up in a tough position where I can't regenerate the rules. Is parsing the built in rules really that much of a performance hit? Given that we're considering moving them out into their own repos, pre-compiling them seems like a bit of a premature optimisation anyway. I think I'm okay with them being un-compiled.

peterebden:
From a bit of investigation I don't think it makes a huge difference. I thought I had numbers back in the day indicating it was more of a win, but whatever there was has been long lost to the mists of time.

That's fine if that's what we want to do and are making a deliberate decision; there is some code we can then clean up as a result which would be nice. Just want to make sure we know we're doing that (the original PR description doesn't mention it, so it's not clear if it was a deliberate choice or not).

Tatskaari:
Okay cool. Homebrew have managed to get 1.16 out. Think we can probably move forward with this. @PeterBocan are you up for getting this finished or should i take over?

PeterBocan:
I can finish this off tomorrow if that's fine with you. If you are in rush however, feel free to take over.

peterebden:
We're not in a particular rush (other than being annoyed by merge conflicts such as you now have!).

Happy to do cleanup in a subsequent PR too, think we can just bite the bullet on this.

PeterBocan:
Ok then, let me deal with this tomorrow.

PR #1602, opened 2021-03-02
Fix IterInputs() to not mark srcs of deps as done early

Tatskaari:


src/core/utils.go
```diff
@@ -161,15 +169,15 @@ func IterInputs(graph *BuildGraph, target *BuildTarget, includeTools, sourcesOnl
 			for _, dep := range dependency.BuildDependencies() {
 				for _, dep2 := range recursivelyProvideFor(graph, target, dependency, dep.Label) {
 					if !done[dep2] && !dependency.IsTool(dep2) {
-						inner(graph.TargetOrDie(dep2))
+						inner(graph.TargetOrDie(dep2), copyDone(done))
```
peterebden, line 172: I'm not totally sure about this. This means it won't deduplicate if we hit the same input twice on two unrelated branches, won't it?

Tatskaari: That's how it worked before. Done only contains the sources of the dependencies we've visited. We will send deps multiple times but `IterSources()` will dedupe them.

peterebden (approved):
OK, you touched it last now ;-)

Issue #1579, opened 2021-02-18
Go toolchain cross-compile does not work with targets containing arch


sagikazarmark:
When using `go_toolchain` with targets containing arch (eg. `@linux_amd64//:target`), cross-compilation fails.

```
    ///linux_amd64//third_party/go:_github.com__bazelbuild__buildtools__tables#a_rule
Error building target ///linux_amd64//third_party/go:_github.com__bazelbuild__buildtools__tables#a_rule: exit status 1
find: /Users/marksk/Projects/sagikazarmark/please-go-modules/plz-out/bin/tools/go_toolchain/pkg/linux_amd64: No such file or directory
```

Example: https://github.com/sagikazarmark/please-go-modules/pull/28

PR #1605, opened 2021-03-03
Add an architectures arg to go_toolchain()

Tatskaari:
Fixes:  #1579


rules/go_rules.build_defs
```diff
@@ -20,6 +21,9 @@ def go_toolchain(name:str, url:str|dict = '', version:str = '', hashes:list = []
                      but not both.
       hashes (list): A list of possible hashes for the downloaded archive. Optional.
       visibility (list): Visibility specification. Defaults to public.
+      architectures (dict): Any additional architectures to install in go architecture format e.g. linux_amd64. This
```
peterebden, line 24: says `dict` but should be `list`?

Tatskaari: Ah good catch

sagikazarmark:
Is this a potential fix for #1579 ? Or an additional feature?

Tatskaari:
> Is this a potential fix for #1579 ? Or an additional feature?

Oh yeah I think so. If you're doing `///{os}_{arch}//some_target:test` and not doing `--arch {os}_{arch}`, then go_toolchain() won't install the SDK for that arch, and it will fail as you've described in your issue.

PR #1607, opened 2021-03-03
levels introduced to changes

goddenrich:
This allows the user to specify a level for query changes. 

-1 is for all transitive dependencies. -2 implies not set. This is important to be able to distinguish between not set and 0 for compatibility reasons.

For backwards compatibility we continue allowing the `--include_dependees` and level to be specified however its worth noting it possible to specify your needs for include_dependees with level ie. none=0, direct=1, transitive=-1. Because of this if both `--include_dependees` and `--level` is set then we warn the user and take level over include_dependees.


src/please.go
```diff
@@ -350,6 +350,7 @@ var opts struct {
 		Changes struct {
 			Since            string `short:"s" long:"since" default:"origin/master" description:"Revision to compare against"`
 			IncludeDependees string `long:"include_dependees" default:"none" choice:"none" choice:"direct" choice:"transitive" description:"Include direct or transitive dependees of changed targets."`
+			Level            int    `long:"level" default:"0" description:"Levels of the dependencies of changed targets."`
```
Tatskaari, line 353: Can you document `-1` here? 

Also we're currently in a major release cycle and I think I'd prefer not to have 2 flags that do basically the same thing. We were burnt last time we outright removed a flag so we should probably keep the flag here but have it do nothing. 

I think it makes sense to have something that looks like: `plz query change --dependency_depth=-1`

Tatskaari (commented):
Nice! I think this makes sense. Let's consider if we want to make the breaking change now rather than waiting for v17.

src/please.go
Tatskaari: Okay as discussed, I think removing this flag right now will cause a lot of issues because this command is used across commit versions. If we remove it, we won't be able to query changes for the commit that upgrades Please...

Tatskaari (approved):
Nice thanks Richard!

PR #1609, opened 2021-03-03
Avoid transitive deps in go

Tatskaari:
A little grungy but `plz build //src/... --rebuild` takes roughly 1m5 before and 25s after.


rules/go_rules.build_defs
```diff
@@ -100,12 +100,23 @@ def go_library(name:str, srcs:list, resources:list=None, asm_srcs:list=None, hdr
...
     src_labels = []
     private = out.startswith('_')
 
+    # We could just depend on go_src for all our deps but that would mean bringing in srcs for targets outside this
+    # package like third party sources which is especially slow on systems with slow syscall performance (macOS)
+    if _all_srcs:
+        deps += [filegroup(
+            name = name,
+            tag = "package_srcs",
+            exported_deps = [dep for dep in deps if dep.startswith(":")],
```
Tatskaari, line 116: I wonder if we should implement some built in functions to help with this sort of thing. I know build rules ideally wouldn't have to worry about this stuff but it's evident to me that it's useful in some cases.

rules/go_rules.build_defs
```diff
@@ -100,12 +100,23 @@ def go_library(name:str, srcs:list, resources:list=None, asm_srcs:list=None, hdr
...
     src_labels = []
     private = out.startswith('_')
 
+    # We could just depend on go_src for all our deps but that would mean bringing in srcs for targets outside this
+    # package like third party sources which is especially slow on systems with slow syscall performance (macOS)
+    if _all_srcs:
+        deps += [filegroup(
+            name = name,
+            tag = "package_srcs",
+            exported_deps = [dep for dep in deps if canonicalise(dep).startswith("//" + package_name() + ":")],
```
peterebden, line 116: Might be worth pulling the `"//" + package_name() + ":"` bit out of the loop? We don't do any hoisting of loop invariants (maybe we should, although `package_name()` is a bit awkwardly tricky for that too)

peterebden (approved):
Bears mentioning that the original concept was that the build config wouldn't be known at parse time (this was back in a more civilised age). Worth revisiting, especially given that when cross-compiling you _are_ allowed to know the architecture which seems pretty equivalent.

PR #1610, opened 2021-03-04
Backport(v15): Avoid transitive deps in go

sagikazarmark:
* Reduce the amount of stuff go rules depend on

* Replace local rule check with something more robust

Signed-off-by: Mark Sagi-Kazar <mark.sagikazar@gmail.com>


sagikazarmark:
Breaking build seems unrelated (java download :\)

Tatskaari:
I was hoping to not push these sort of changes back to v15. I just tagged a pre-release. Do you think I can convince you to use that instead?

Tatskaari:
Looks like we were able to get you on the prerelease. This is backwards incompatible anyway so I'd really not want to merge it into v15,

PR #1611, opened 2021-03-04
Get git SCM commands to report the stdout from git

Tatskaari:


peterebden (approved):
LGTM, but the linter is obviously correct, plz appease it :)

Tatskaari:
> LGTM, but the linter is obviously correct, plz appease it :)

In this case it's got a point but my god it's been hard to get it to leave the test data alone...

PR #1616, opened 2021-03-04
Update prerelease object on prerelease

### LINKED ISSUE #1615: latest_version set to un-released version
Currently https://get.please.build/latest_version returns `16.0.0-prerelease.4, which is an un-released version`. This api is used to know when to pull the new version of please. I suspect a bug has been introduced in the release process which now allows this to be set to prerelease versions.

For context this api is used to know when to update the [Arch Linux AUR package](https://aur.archlinux.org/packages/please-bin/) with [this update script](https://aur.archlinux.org/cgit/aur.git/tree/update-version.sh?h=please-bin).

steakunderscore:
Fix release script so that for prerelease versions, update prerelease s3 object.

Fixes  #1615


peterebden (approved):
I think we should ideally have something more robust here, but this is great for now, thanks!

PR #1618, opened 2021-03-05
Sandbox improvements

peterebden:
Upstreaming some stuff from TM internal; main functional changes add PID namespacing and remount / as read-only.

Adds a new, currently unused binary that leaves net open which could also be useful.

Depends on #1617.


Tatskaari:
Nice! I can look into `sandbox = "all"` `sandbox = "filesystem"` and `sandbox = "network"` towards the end of v16 I guess.

Tatskaari (approved):

peterebden:
> Nice! I can look into `sandbox = "all"` `sandbox = "filesystem"` and `sandbox = "network"` towards the end of v16 I guess.

Yeah probably worth us talking about it a bit; I would like some finer-grained control in the filesystem one for example (which would let us totally upstream all our internal stuff)

PR #1620, opened 2021-03-05
Fix up test_only for the new filegroup in go_library()

Tatskaari:


rules/go_rules.build_defs
```diff
@@ -115,6 +115,8 @@ def go_library(name:str, srcs:list, resources:list=None, asm_srcs:list=None, hdr
             tag = "package_srcs",
             exported_deps = [dep for dep in deps if canonicalise(dep).startswith("//" + package_name() + ":")],
             requires = ["go_src"],
+            test_only = test_only,
+            visibility = visibility,
```
peterebden, line 119: think this is probably unnecessary?

PR #1622, opened 2021-03-06
Make individual sandbox dirs configurable

peterebden:
Also restructures the sandbox config settings into their own section, which I think is a bit more logical (e.g. it's clear that the tool is global rather than `test` pulling in the one from `build`).
The old ones are deprecated, with the old settings carrying over for now; I assume the migration shouldn't be too hard tho.

### DIFF
--- a/docs/config.html
+++ b/docs/config.html
@@ -431,7 +431,8 @@
         <p>
           Activates sandboxing for individual build actions, which isolates them
           from network access, IPC and some aspects of the filesystem. Currently
-          only works on Linux. Defaults to <code class="code">False</code>.
+          only works on Linux. Defaults to <code class="code">False</code>.<br />
+          <span class="red">Deprecated from v16: use <a href="#sandbox">sandbox.build</a> instead.
         </p>
       </div>
     </li>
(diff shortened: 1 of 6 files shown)


docs/config.html
```diff
@@ -484,6 +485,65 @@ <h2 id="build-env" class="title-2">[BuildEnv]</h2>
...
+      </div>
+    </li>
+    <li>
+      <div>
+        <h3 class="mt1 f6 lh-title">
+          Test <span class="normal">(bool)</span>
+        </h3>
+
+        <p>
+          Activates sandboxing for individual test actions, which isolates them
```
Tatskaari, line 525: What's the significance of individual here (and above)? Maybe just "Activates sandboxing for test actions,"? Or am I missing something?

docs/config.html
```diff
@@ -484,6 +485,65 @@ <h2 id="build-env" class="title-2">[BuildEnv]</h2>
...
+      <div>
+        <h3 class="mt1 f6 lh-title">
+          Tool
+        </h3>
+
+        <p>
+          The binary to use for sandboxing build/test actions. Typically this
+          is <code>please_sandbox</code>.<br />
+          It receives the program to execute and all its arguments in its <code>argv</code>,
+          it is expected to do whatever it needs to then <code>exec</code> the real action.
```
Tatskaari, line 501: Worth mentioning that `please_sandbox` only works on linux?

docs/config.html
```diff
@@ -484,6 +485,65 @@ <h2 id="build-env" class="title-2">[BuildEnv]</h2>
...
+    <li>
+      <div>
+        <h3 class="mt1 f6 lh-title">
+          Dir <span class="normal">(repeated string)</span>
+        </h3>
+
+        <p>
+          Directories that the sandbox tool should mask (it's unspecified how, although mounting an
+          empty filesystem over the top is the current implementation).<br />
+          These are passed to the tool itself in the <code>SANDBOX_DIRS</code> env var as a comma-separated list.
```
Tatskaari, line 540: Should we not whitelist directories? We can probably determine this from the PATH, SECRETS and TOOLS variables the rule requires. Maybe that can come in a future release.

Tatskaari (approved):
Nice! Would be interested to hear your opinion on the whitelisting idea.

docs/config.html
peterebden: It's been copied & pasted a bit now, but I imagine what I was originally thinking was that it sandboxes them from one another (as opposed to sandboxing all build actions together in one place).
I'll remove it, it's not really adding anything.

docs/config.html
peterebden: yeah, fair

docs/config.html
peterebden: Those variables aren't sufficient to determine the set of directories you need - for example you might have `/usr/bin` on the PATH, but what do you need for things in there to run? Probably `/usr`, `/lib`, `/lib64`, `/bin` but you might also need some subsets of `/etc` or `/var` depending on what they're doing.

We can definitely think about extending this in future, but this ability to blacklist directories has definitely been useful internally.

PR #1623, opened 2021-03-07
Make please_go handle relative paths to tools

peterebden:
Think this is causing some issues for remote execution; if the path to GoTool is e.g. `third_party/go/bin/go` then we need to make that absolute in order to exec it.

This doesn't come up for us internally; I think our servers still do a little magic to make this work, but will be needed for us to interoperate with Buildbarn / Buildgrid etc.

### DIFF
--- a/tools/please_go/please_go.go
+++ b/tools/please_go/please_go.go
@@ -1,12 +1,14 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


Tatskaari (approved):
Seems harmless to me

PR #1624, opened 2021-03-07
Stop sending output files & dirs

peterebden:
Been wanting to get rid of this for ages.

At this point it is just making things worse as well as being ugly; I think most major REAPI servers support `output_paths` (at least Buildgrid and Buildbarn do), but while Buildgrid is coping OK, Buildbarn doesn't quite follow it right (they iterate all three fields whereas the API says you should ignore Files and Dirs if Paths is set). I might send them a PR for that too.

### DIFF
--- a/src/core/build_target.go
+++ b/src/core/build_target.go
@@ -606,6 +606,19 @@ func (target *BuildTarget) FullOutputs() []string {
 	return outs
 }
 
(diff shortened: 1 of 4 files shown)


Tatskaari (approved):
Beautiful :D

PR #1629, opened 2021-03-09
Ensure GOROOT is absolute

peterebden:
Also a small fix for targets with output directories

### DIFF
--- a/rules/go_rules.build_defs
+++ b/rules/go_rules.build_defs
@@ -1466,7 +1466,7 @@ def _set_go_env():
         go_root = CONFIG.GOROOT
     elif CONFIG.GO_TOOL.startswith(":") or CONFIG.GO_TOOL.startswith("//"):
         label, _, _ = CONFIG.GO_TOOL.rpartition('|')
-        go_root = f'$(realpath $(dirname $(dirname $TOOLS_GO)))'
+        go_root = f'$($TOOLS_GO env GOROOT)'
     else:
         go_root = '$("$TOOLS_GO" env GOROOT)'
     cmd = f'export GOPATH=$TMP_DIR && export GOROOT={go_root}'
--- a/rules/go_rules.build_defs
+++ b/rules/go_rules.build_defs
@@ -1466,7 +1466,7 @@ def _set_go_env():
         go_root = CONFIG.GOROOT
     elif CONFIG.GO_TOOL.startswith(":") or CONFIG.GO_TOOL.startswith("//"):
         label, _, _ = CONFIG.GO_TOOL.rpartition('|')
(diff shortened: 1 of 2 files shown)


rules/go_rules.build_defs
```diff
@@ -1454,7 +1454,7 @@ def _set_go_env():
         go_root = CONFIG.GOROOT
     elif CONFIG.GO_TOOL.startswith(":") or CONFIG.GO_TOOL.startswith("//"):
         label, _, _ = CONFIG.GO_TOOL.rpartition('|')
-        go_root = f'$(dirname $(dirname $TOOLS_GO))'
+        go_root = f'$(realpath $(dirname $(dirname $TOOLS_GO)))'
```
Tatskaari, line 1457: There must be a better way... perhaps we can add some labels to `go_toolchain()` and set the env up in a pre_build function or something. Fine with this for now though.

Tatskaari (approved):

rules/go_rules.build_defs
peterebden: That seems like quite a bit more work just to avoid this one extra call? Unless you mean getting rid of the `dirname` calls too?

Tatskaari: > That seems like quite a bit more work just to avoid this one extra call? Unless you mean getting rid of the `dirname` calls too?

I do. It might be better to have the go_toolchain() rule control the GOROOT rather than effectively guessing at it here. I'd need to think he we'd do that same when `GO_TOOL` is just on the path though.

sagikazarmark: `go env GOROOT` should return the correct GOROOT (if not set as an environment variable). Would that work here?

peterebden: That looks reasonable to me. @Tatskaari any reason you can think of that wouldn't work?

Tatskaari: I tried that at one point. I think the issue was that on FreeBSD it was following hard links when it shouldn't. I've fixed that though so it might work now.

peterebden: OK seems like there's enough subtlety here that I'm going to leave this as is for now; we can come back and look at `go env GOROOT` later.

peterebden: Haha no looks like `realpath` is a no-go on OSX so I will do this now - thanks for the suggestion, I'd have been stuck here without it :D

peterebden: ....but indeed this doesn't seem to work on FreeBSD :(

peterebden: OK I've ended up special-casing it to work around FreeBSD's crazy semantics of KERN_PROC_PATHNAME etc.

PR #1632, opened 2021-03-10
Rewrite argv to new location in sandbox

peterebden:
Previously it was still pointing through to the old workdir, so if things used `argv[0]` to look up their own location, they'd get a readonly copy.

### DIFF
--- a/tools/sandbox/BUILD
+++ b/tools/sandbox/BUILD
@@ -19,3 +19,11 @@ c_binary(
     visibility = ["PUBLIC"],
     deps = [":sandbox"],
 )
+
+cc_test(
+    name = "sandbox_test",
+    srcs = ["sandbox_test.cc"],
+    deps = [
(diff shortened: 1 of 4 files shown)


peterebden:
> I think I understand. Is this useful to remap tools into the sandbox directory?

No, so what happens now is we create everything in `plz-out/tmp/target/build`, let's say the test executable is `plz-out/tmp/target/build/test.bin`. We then create the sandbox on `/tmp/plz_sandbox` and bind-mount it into there.
However when we _run_ it we still execute `plz-out/tmp/target/build/test.bin` but we set `/tmp/plz_sandbox` to be the working directory. Hence anything that looks at `os.Args[0]` will still find the old location (which is now read-only).

After this change, `os.Args[0]` will be `/tmp/plz_sandbox/test.bin` and so things won't accidentally discover the wrong location.

Issue #1598, opened 2021-02-27
Deprecate `raise`


peterebden:
`raise` in asp is kind of weird; there is no `try` (we obviously listened to Master Yoda) so it's really just an immediate fail. You can raise different "types" of exceptions but they're really all the same thing and just trigger a parse failure (so `raise ValueError(...)` is the same as `raise ParseError(...)`).
I put it in originally because it was fairly easy and we already had enough incompatibilities to work out in v12. I think the time has come to revisit that decision.

Skylark has a `fail` builtin IIRC,  I think `fail('you passed the wrong args')` is pretty reasonable as an alternative. If we agree that is a good alternative, we should offer that sometime in v16 and remove `raise` for v17 (after a period of it generating a warning, plus a feature flag etc).

PR #1634, opened 2021-03-10
Deprecate raise

peterebden:
Adds explicit warnings on creating exception types or the `raise` statement.

Starts on #1598 although I don't think we should finish it off until v17.


Tatskaari (approved):
Probably need to generate rules. Is it time for embed?

Tatskaari:
Might want to update the grammar

PR #1635, opened 2021-03-11
Fix some bugs with format()

peterebden:
If it's called with just positional arguments (as in the test) it blows up.

### DIFF
--- a/src/parse/asp/interpreter_test.go
+++ b/src/parse/asp/interpreter_test.go
@@ -367,4 +367,5 @@ func TestFormat(t *testing.T) {
 	s, err := parseFile("src/parse/asp/test_data/interpreter/format.build")
 	assert.NoError(t, err)
 	assert.EqualValues(t, `LLVM_NATIVE_ARCH=\"x86\"`, s.Lookup("arch"))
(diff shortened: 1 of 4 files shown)


src/parse/asp/interpreter_test.go
```diff
@@ -362,3 +362,9 @@ func TestFStringOptimisation(t *testing.T) {
 	assert.NotNil(t, assign.Optimised.Constant)
 	assert.EqualValues(t, "test", assign.Optimised.Constant)
 }
+
+func TestFormat(t *testing.T) {
+	s, err := parseFile("src/parse/asp/test_data/interpreter/format.build")
+	assert.NoError(t, err)
+	assert.EqualValues(t, `LLVM_NATIVE_ARCH=\"x86\"`, s.Lookup("arch"))
```
Tatskaari, line 369: Should `'{}_{}'.format('foo', 'bar')` produce `'foo_bar'`? It doesn't look like it will to me. Could we add a second `{}` to the test if ti should.

Tatskaari (approved):
Test for multiple {} might be nice, but looks good.

peterebden:
Yup added one. Good idea, initially I was just trying to get it not die on any number of args, but worth checking for.

PR #1636, opened 2021-03-11
Save file walking results in glob()

peterebden:
It can be very slow to do repeated `glob()` calls into a large directory; I am testing with some work on LLVM and it takes approx 15 seconds to get plz to sort itself out and get moving.

Basically all the time is spent walking it over and over again. Main issue is that we walk the entire root dir every time, which can be (and in this case is) large, even if the glob expressions are fairly targeted.

This essentially memoises those walks; I think there is more we could do about (possibly?) more efficient matching, or getting really smart about walking bits of the tree as needed, but this is pretty good (and simple) for now.

Reduces time from ~15s -> ~2.7s in my use case.


Tatskaari (approved):
Nice! Looks good to me :D

PR #1647, opened 2021-03-13
fix root_dir in proto_library

### LINKED ISSUE #1646: root_dir in proto_library rule doesn't work
When using `root_dir` in `proto_library` rule with protobuf `3.15.1` code generation (Go) fails with the following error:

```
File does not reside within any path specified using --proto_path (or -I).  You must specify a --proto_path which encompasses this file.  Note that the proto_path must be an exact prefix of the .proto file names -- protoc is too dumb to figure out when two paths (e.g. absolute and relative) are equivalent (it's harder than you think).
```

Looks like the implementation is incomplete: the script changes the directory to the one defined in `root_dir`, but the current directory is not added as a search path.

There are two (trivial) fixes for the problem, one of them will potentially fix an other issue as well:

The most trivial solution is adding `-I.` to the protoc command. Since the script changes the directory to `root_dir`, all imports (pointing to the same package) should work. The problem with this approach is that a bogus (non-existent) `root_dir` will break at changing a directory resulting in an error like this:

```
bash: cd: NOPE: No such file or directory
todo/v1/todo.proto: File not found.
```

The other solution is getting rid of changing directories entirely and adding `-I"$TMP_DIR/root_dir"` to the protoc command. This would result in an error like this:

```
/Users/marksk/Projects/sagikazarmark/todobackend-go-kit/plz-out/tmp/api/todo/v1/_proto#protoc_go._build/apis: warning: directory does not exist.
todo/v1/todo.proto: File not found.
```

I find this option slightly better.

sagikazarmark:
Fixes #1646

### DIFF
--- a/test/proto_rules/root_dir/BUILD
+++ b/test/proto_rules/root_dir/BUILD
@@ -1,0 +1,16 @@
+# Tests on the proto rules.
+
+proto_library(
+    name = "root_dir",
+    srcs = ["test.proto", "other_test.proto"],
+    languages = ["go"],
+    deps = [":protoc"],
+    root_dir = "test\/proto_rules",
+)
+
+# Required for wkt
+protoc_binary(
+    name = "protoc",
+    version = "3.15.1",
+    visibility = ["PUBLIC"],
+)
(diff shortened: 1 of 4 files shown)


Tatskaari:
#1653 addresses this issue

PR #1649, opened 2021-03-14
WIP glob implementation

Tatskaari:
Need to upgrade to 1.16 on linux alt for this to work. It uses the new ReadDir stuff that's more efficient.


peterebden:
This is not super fast :(

```
real	0m11.026s
user	0m20.988s
sys	0m0.259s
```
vs #1645 
```
real	0m0.583s
user	0m0.833s
sys	0m0.079s
```
I can push my branch on one of the internal repos if it'll help you. Could also try to separate it out into a public test repo if people care, but it'll take a bit more work.

> Need to upgrade to 1.16 on linux alt for this to work. It uses the new ReadDir stuff that's more efficient.

We're gonna have to do that anyway for #1601 so shouldn't be a big problem.

Tatskaari:
Need to figure out how to port the prefix optimisation. Will come back to this later.

PR #1653, opened 2021-03-16
Add -I. to protoc

Tatskaari:
Normally, the current wd is added to the include path. If other proto paths are provided (because you're using protoc_binary()) then this isn't the case and you get errors like the one in the linked issue.

I think when we provide `-I` we always want to augment the import path not replace it. This should make that so.  

Fixes: #1646


peterebden (approved):
Makes sense!

Issue #1641, opened 2021-03-13
Well-known types are missing


sagikazarmark:
When compiling proto descriptors relying on well-known types, the process fails with the following:

```
Error building target //api/todo/v1:_proto#protoc_go: exit status 1
google/protobuf/wrappers.proto: File not found.
todo/v1/todo_list.proto, line 14, column 1: Import "google/protobuf/wrappers.proto" was not found or had errors.
todo/v1/todo_list.proto, line 70, column 3: "google.protobuf.StringValue" is not defined.
todo/v1/todo_list.proto, line 71, column 3: "google.protobuf.BoolValue" is not defined.
todo/v1/todo_list.proto, line 72, column 3: "google.protobuf.Int32Value" is not defined.
```

Looks like those files [are actually downloaded](https://github.com/thought-machine/please/blob/2b76379b8fe51415dbdc1f215ac708be2e0cda6c/rules/proto_rules.build_defs#L456), but not required in proto targets. Is that intentional?

sagikazarmark:
Apparently the problem is that the `protoc` target that I use is in a subrepo. If it's in the same repo, it works fine.

Tatskaari:
Yeah looks like it might just work on most of the machines because they have the well known types installed. I guess we need to add `PROTOC_TOOL` to the deps list if it looks like a build label. 

I feel like this would've cropped up a long time ago though. Have we really been depending on the system wide well known types this whole time?

sagikazarmark:
Sadly still can't make this work. WKT don't show up in the workspace.

PR #1655, opened 2021-03-16
Depend on protoc to include wkt when it looks like a label

Tatskaari:
Fixes: #1641

### DIFF
--- a/rules/proto_rules.build_defs
+++ b/rules/proto_rules.build_defs
@@ -402,7 +402,7 @@ def _protoc_rule(name, srcs, deps, language, plugin, protoc_flags, root_dir, pre
 
     # If the protoc tool is a build rule, depend on it as it's probably a protoc_binary() which provides the well known
     # types
-    if CONFIG.PROTOC_TOOL.startswith("//"):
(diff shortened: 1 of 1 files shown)


sagikazarmark:
I was wondering if it would make sense to make the WKT rule a configurable rule in Please?

Or we could just get rid of that entirely? `proto_libraries` relying on WKT could add a dependency without relying on some automatic magic behind the scenes.

Tatskaari:
> I was wondering if it would make sense to make the WKT rule a configurable rule in Please?
> 
> Or we could just get rid of that entirely? `proto_libraries` relying on WKT could add a dependency without relying on some automatic magic behind the scenes.

I think it's reasonable to assume that well known types will be handled automatically. A standard protoc install would put them under `/usr/local/include/google/protobuf`. Ideally `protoc_binary()` would behave similarly.

sagikazarmark:
Would it make sense then to make it a separate rule and configuration? (eg. for whatever reason I want to use a custom version).

Tatskaari:
> Would it make sense then to make it a separate rule and configuration? (eg. for whatever reason I want to use a custom version).

Seems pretty niche. You could write your own rule that outputs `protoc` and provides the well known types so it's possible if that became a use case.

peterebden (approved):

PR #1656, opened 2021-03-16
Don't force remote to skip cache lookup

peterebden:
Arguably it is nice as a performance hint, but may be useful for testing buildgrid to allow it to do it itself.


Tatskaari (approved):
Seems harmless to me

PR #1657, opened 2021-03-16
Update test main to handle import paths with a repeated ending part

peterebden:
Currently seems to have an incorrect ending, e.g. `github.com/thought-machine/gonduit/gonduit` when it should just be `github.com/thought-machine/gonduit`.
I think maybe that only happens at the repo root?


Tatskaari (approved):
Looks reasonable. I will be careful to test this against core3

PR #1662, opened 2021-03-16
Embed assets

Tatskaari:


src/assets/pleasew
```diff
@@ -0,0 +1 @@
+dummy
```
peterebden, line 1: To make sure I understand... these are the files you'd get with 'normal' go embed, but when you build with plz it sets up the real ones instead?

peterebden (approved):

src/assets/pleasew
Tatskaari: Yup!

PR #1665, opened 2021-03-17
Fix proto paths when using root_dir

Tatskaari:
I don't think source relative is meant to be provided like that. It doesn't seem to make a difference. The docs suggest providing it as such:
```
--go_opt=paths=source_relative
```

The rest of the PR is hopefully fairly non-controversial.

I've done a `plz build --include proto` in core3 and they all build with this change.


peterebden:
Possibly this relates to what version of protoc-gen-go is being used? I am pretty certain this used to have some kind of effect (although it is all pretty obscure and badly documented)

peterebden (approved):

peterebden (approved):

PR #1667, opened 2021-03-17
Fix test main root import

Tatskaari:


peterebden (approved):
plz appease the linter but looks good!

PR #1668, opened 2021-03-17
Allow compression when uploading test results

fische:
Adds option `test.gzip` to gzip test results before uploading them to the given URL. By default, it doesn't do it to keep things backward-compatible (even if most HTTP servers now probably support gzip compression).

### DIFF
--- a/src/core/config.go
+++ b/src/core/config.go
@@ -443,6 +443,7 @@ type Configuration struct {
 		Sandbox         bool         `help:"Deprecated, use sandbox.test instead."`
 		DisableCoverage []string     `help:"Disables coverage for tests that have any of these labels spcified."`
 		Upload          cli.URL      `help:"URL to upload test results to (in XML format)"`
+		Gzip            bool         `help:"True to upload the test results gzipped."`
 	} `help:"A config section describing settings related to testing in general."`
 	Sandbox struct {
 		Tool  string   `help:"The location of the tool to use for sandboxing (typically please_sandbox)."`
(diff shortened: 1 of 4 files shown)


src/test/xml_results.go
```diff
@@ -434,9 +436,26 @@ func SerialiseResultsToXML(target *core.BuildTarget, indent bool) []byte {
...
+			return fmt.Errorf("Failed to flush gzip writer: %s", err)
+		}
+		r = &buf
+	}
+	req, err := http.NewRequest(http.MethodPost, url, r)
+	if err != nil {
+		return fmt.Errorf("Failed to create HTTP request: %s", err)
+	}
+	req.Header["Content-Type"] = []string{"application/xml"}
+	req.Header["Content-Encoding"] = []string{"gzip"}
```
goddenrich, line 457: should this be in the if gzipped section?

src/core/config.go
```diff
@@ -443,6 +443,7 @@ type Configuration struct {
 		Sandbox         bool         `help:"Deprecated, use sandbox.test instead."`
 		DisableCoverage []string     `help:"Disables coverage for tests that have any of these labels spcified."`
 		Upload          cli.URL      `help:"URL to upload test results to (in XML format)"`
+		Gzip            bool         `help:"True to upload the test results gzipped."`
```
goddenrich, line 446: Do we want to group these two as they are related?

src/test/xml_results.go
fische: 🤦 Yes it should, my bad.

src/core/config.go
fische: If we do group them, it wont be backward compatible, so I don't think that's what we want.

peterebden: Might be tempted to call this `UploadGzipped` or something so it's clear that it relates to the upload (as opposed to just `test.gzip` which could mean other things)

peterebden (approved):
Thanks!

src/core/config.go
fische: Ah yup good point!

PR #1672, opened 2021-03-18
Make the $TEST var absolute in rex

peterebden:
We typically run the test via `$TEST ...` which doesn't work if it is a relative path (Bash tries a path lookup instead).

We don't get this internally because Mettle makes it absolute; eventually we might try to move away from that.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,5 @@
 run:
+  timeout: 5m
   skip-dirs:
     - test_data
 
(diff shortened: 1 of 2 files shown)


Tatskaari:
I guess we should make sure this isn't a problem locally but that will come up in my root test I'm working on now.

Tatskaari (approved):

peterebden:
This only happens in `src/remote` so it won't affect anything locally.

PR #1673, opened 2021-03-18
Handle relative sandbox paths

peterebden:


peterebden:
Not certain this is correct.

PR #1674, opened 2021-03-18
not setting stdout or stderr on success

goddenrich:
This change stops us from setting stdout or stderr on success. We continue to set stdout and stderr for the flaky failure.


Tatskaari:
What's the motivation for this? Is this taking up a lot of space? I would've thought there's some minimal value in saving this for some scenarios. Maybe the test passed but it was suspiciously fast.

peterebden (approved):

Tatskaari (approved):
Looks good! Can you merge master in as it has a fix for the linter timing out.

PR #1676, opened 2021-03-18
Exclude tools by default in jarcat

peterebden:
Stops them from automatically turning up when we download them during remote execution.


Tatskaari (approved):
Nice! Many thanks for that. We could probably handle named tools without too much effort. Might have a look tomorrow.

peterebden:
Named tools should get handled by this too (they are also part of `$TOOLS`)

PR #1680, opened 2021-03-19
Stop allowing foo/bar/bar as an import for foo/bar

Tatskaari:


peterebden (approved):
Let's do it :breaking_change_parrot:

PR #1683, opened 2021-03-19
Stop writing action results to remote

peterebden:
I don't think most of these are required. The only one that makes a "real" difference is `remote_file` and I think we should assume that the remote server will efficiently cache that for us (it appears this is more what is done elsewhere).

Leaving one for stamped actions at this point. That seems tricky and needs more thought.


Tatskaari (approved):
I assume these are uploaded to the CAS then these are made part of the input root of any rule that depends on them, even if we don't have an action for them? If it still works then i guess this is fine.

peterebden:
Yeah so we still handle them internally in the same way.
The assumption is that the remote asset server is efficient about not redownloading things; that seems to be the case for bb-remote-asset for example.

PR #1685, opened 2021-03-22
Add config option to emit stdout and stderr on successful test runs

Tatskaari:


goddenrich (commented):
This seems like a lot of passing config through the layers of functions... one thing we could do is pass a struct which contains all the config needed ie. gzipped and output on success. that way every layer doesn't have to change each time a new option is added.

Tatskaari:
> This seems like a lot of passing config through the layers of functions... one thing we could do is pass a struct which contains all the config needed ie. gzipped and output on success. that way every layer doesn't have to change each time a new option is added.

We could just pass the state or config struct through but I'm not super happy with how much of a god object those things are becoming. Right now it's just one or two parameters which is totally manageable.

peterebden (approved):

PR #1687, opened 2021-03-22
Remove trailing slashes from filegrups

Tatskaari:


peterebden (approved):
Seems legit but as discussed maybe better to do this to the input

peterebden (approved):

PR #1660, opened 2021-03-16
Fix zipsafe flag

Tatskaari:
Fixes: #1597

PR #1688, opened 2021-03-23
Fix zip-safe flag more

peterebden:
#1660 inadvertently changed the default of the zip-safe flag from "determine from deps" to "unsafe".

This should change it so if it's not passed it determines automatically, and if it's explicitly set either way it just forces that behaviour.


Tatskaari (approved):
Tristate logic strike again! Thanks for that.

PR #1692, opened 2021-03-23
Query graph: use encode to print to stdout in one pass

Tatskaari:


src/query/graph.go
```diff
@@ -15,13 +15,15 @@ import (
 func Graph(state *core.BuildState, targets []core.BuildLabel) {
 	log.Notice("Generating graph...")
 	g := makeJSONGraph(state, targets)
-	log.Notice("Marshalling...")
-	b, err := json.MarshalIndent(g, "", "    ")
-	if err != nil {
+
+	encoder := json.NewEncoder(os.Stdout)
+	encoder.SetIndent("", "    ")
+	encoder.SetEscapeHTML(false)
```
Tatskaari, line 21: This is probably a marginal performance gain. Didn't seem to make any real difference in practice.

peterebden (approved):

PR #1699, opened 2021-03-25
Skip cache lookup for rerun and rebuild

Tatskaari:


src/remote/remote.go
```diff
@@ -588,12 +588,12 @@ func (c *Client) execute(tid int, target *core.BuildTarget, command *pb.Command,
 	} else if target.IsTextFile {
 		return c.buildTextFile(target, command, digest)
 	}
-	return c.reallyExecute(tid, target, command, digest, needStdout, isTest)
+	return c.reallyExecute(tid, target, command, digest, needStdout, isTest, (c.state.ForceRerun && isTest) || (!isTest && c.state.ForceRebuild))
```
peterebden, line 591: Don't think this is quite right - think the build one also needs an `IsOriginalTarget` (c.f. code in `src/build`)

src/remote/remote.go
```diff
@@ -588,12 +588,18 @@ func (c *Client) execute(tid int, target *core.BuildTarget, command *pb.Command,
 	} else if target.IsTextFile {
 		return c.buildTextFile(target, command, digest)
 	}
-	return c.reallyExecute(tid, target, command, digest, needStdout, isTest)
+
+	// We should skip the cache lookup (and override any existing action result) if we --rebuild, or --rerun and this is
+	// one fo the targets we're testing or building.
+	skipCacheLookup := (isTest && c.state.ForceRerun) || (!isTest && c.state.ForceRebuild)
+	skipCacheLookup = skipCacheLookup && c.state.IsOriginalTarget(target)
```
peterebden, line 595: Technically you don't have to bother with `IsOriginalTarget` if it's a test (we only ever do run tests that are the 'original' targets). Probably doesn't matter much though.

peterebden (approved):

PR #1701, opened 2021-03-26
Remove MavenJar FF: download as -sources.jar so tools understand them

Tatskaari:
This allows vscode and intellij to  understand these as the sources for the main jar.


peterebden (approved):
Remember to add this one to the release notes!

PR #1702, opened 2021-03-26
Fix incorrect results from query changes and revdeps. 

Tatskaari:
Essentially the problem is with our depth first approach, certain targets can be removed from the open set of targets to explore prematurely, for example: 

a -> b -> c  
a -> c -> e

Once evaluating the first branch, done will contain a, b, and c. When evaluating the second branch, we will skip c -> e because c is already done. If we do a, c, and e first, we will add e. 

NB: reverse dependencies appears to work slightly differently to query changes. It seems that it's more likely to exclude internal labels because it adds dependencies with this logic:

```
	for _, label := range labels {
		for _, child := range graph.PackageOrDie(label).AllChildren(graph.TargetOrDie(label)) {
			for _, target := range graph.ReverseDependencies(child) {
				if parent := target.Parent(graph); parent != nil {
					uniqueTargets[parent] = struct{}{}
				} else {
					uniqueTargets[target] = struct{}{}
				}
			}
		}
	}
```

I think this still makes sense for query changes and it would be best if these two followed the same logic. 

Apart from replacing some hidden targets with their parent targets, I think the old query changes pretty much returns a subset of what it does now.

### DIFF
--- a/src/query/reverse_deps.go
+++ b/src/query/reverse_deps.go
@@ -2,28 +2,30 @@ package query
 
 import (
 	"fmt"
-	"sort"
-
 	"github.com/thought-machine/please/src/core"
 )
 
 // ReverseDeps finds all transitive targets that depend on the set of input labels.
 func ReverseDeps(state *core.BuildState, labels []core.BuildLabel, level int, hidden bool) {
-	for _, target := range getRevDepTransitiveLabels(state, labels, map[core.BuildLabel]int{}, level) {
+	targets := make(map[core.BuildLabel]int, 100)
+	getRevDepTransitiveLabels(state, labels, targets, level)
+	for target, _ := range targets {
 		if hidden || target.Name[0] != '_' {
 			fmt.Printf("%s\n", target)
 		}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


src/query/reverse_deps.go
```diff
@@ -9,20 +9,20 @@ import (
...
 
-func getRevDepTransitiveLabels(state *core.BuildState, labels []core.BuildLabel, done map[core.BuildLabel]struct{}, level int) core.BuildLabels {
+func getRevDepTransitiveLabels(state *core.BuildState, labels []core.BuildLabel, done map[core.BuildLabel]int, level int) core.BuildLabels {
 	if level == 0 {
 		return nil
 	}
 	for _, l := range getRevDepsLabels(state, labels) {
-		if _, present := done[l]; !present {
-			done[l] = struct{}{}
+		if doneLevel, present := done[l]; !present || doneLevel > level {
```
Tatskaari, line 24: Essentially this is checking if we're currently on a shorter path to the target. If we are we need to reprocess it.

peterebden (approved):

peterebden (approved):

PR #1704, opened 2021-03-27
Add semver_check builtin function

### LINKED ISSUE #1312: Add semver helper functions
I recently had to download a binary that changed the URL pattern after a specific version. I was able to handle it by using multiple download URLs in `remote_file`, but I was wondering if it would make sense to implement native semver helper functions to do that more efficiently.

For example a `version_compare(constraint, version)` would have been useful in my case.

I'd be happy to work on a PR, but wanted to ask your opinion first.

https://github.com/Masterminds/semver seems like a good semver library.

sagikazarmark:
Closes #1312

Note: this PR adds another semver library to the project. The reasons are listed in #1312, but in a nutshell:

- it's better maintained
- comes with more features
- has a more natural API

We can replace the original semver package in a subsequent PR.

Use case for this builtin function: tools sometimes change their versioning and download URL patterns ([example](https://github.com/sagikazarmark/mypleasings/blob/4b566b24e90ee3ac3899faa762a8895f7e0e6ef4/misc/tools.build_defs#L132-L134)). This function helps tool download build rules to introduce conditional logic based on the tool version to download tools from the right URL.

### DIFF
--- a/docs/lexicon.html
+++ b/docs/lexicon.html
@@ -228,6 +228,17 @@
         <code class="code">--plain_output</code> flag if intending to use this.
       </span>
     </li>
+    <li>
+      <span>
+        <code class="code"
+          ><span class="fn-name">semver_check</span><span class="fn-p">(</span
+          ><span class="fn-arg">version</span>, <span class="fn-arg">constraint</span>
+          <span class="fn-p">)</span></code
+        >
+        - checks if a <a href="https://semver.org/">semantic version</a> meets a constraint.
+        Supported version and constraint formats are listed on <a href="https://github.com/Masterminds/semver">this</a> site.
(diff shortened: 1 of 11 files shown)


Tatskaari:
Ah okay so masterminds one lets us use wildcards to compare versions? That could be useful to replace this nonsense in flags.go:

```
// A Version is an extension to semver.Version extending it with the ability to
// recognise >= prefixes.
type Version struct {
	semver.Version
	IsGTE bool
	IsSet bool
}
```

sagikazarmark:
> Ah okay so masterminds one lets us use wildcards to compare versions

It does more than that: it allows all kinds of constraints (see details in the repo).

PR #1705, opened 2021-03-29
Make imports source relative when the import path is provided

Tatskaari:
The old way of applying source relative (on the --go_out flag) no longer worked. It seemed like it wasn't needed as we attempt to collapse the package path when we move outputs to the output directory, so I removed it. Turns out it becomes relevant when applying `option go_package = ...;` as the package path and the import path don't necessarily line up. 

This PR adds the source relative opt back in via the correct flag.

### DIFF
--- a/test/proto_rules/go_grpc_ff/proto/service.proto
+++ b/test/proto_rules/go_grpc_ff/proto/service.proto
@@ -1,5 +1,7 @@
 syntax = "proto3";
 
+option go_package = "github.com/thought-machine/please/test/proto_fules/go_frpc_ff/proto/service";
+
 service Service {
   rpc MakeRequest(Request) returns (Response) {};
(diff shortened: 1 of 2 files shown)


test/proto_rules/go_grpc_ff/proto/service.proto
```diff
@@ -1,5 +1,7 @@
 syntax = "proto3";
 
+option go_package = "github.com/thought-machine/please/test/proto_fules/go_frpc_ff/proto/service";
```
Tatskaari, line 3: This actually ends up as `github.com/thought-machine/please/test/proto_fules/go_frpc_ff/proto` because that's what the name of the grpc_libarary() is.

peterebden (approved):

PR #1707, opened 2021-03-30
Use a non-deprecated openpgp

peterebden:
Post https://github.com/golang/go/issues/44226 it seems we should not be using `x/crypto/openpgp`. https://github.com/ProtonMail/go-crypto seems like the best alternative.

I've built this locally and run an update which it decodes fine. I assume it will also be compatible for any older versions updating to it.

### DIFF
--- a/third_party/go/BUILD
+++ b/third_party/go/BUILD
@@ -19,14 +19,31 @@ go_module(
     install = [
         "ssh/terminal",
         "cast5",
-
-        # TODO(jpoole): consider if requiring this to be explicit is a good idea or if we should do the
-        # `openpgp/openpgp.a` "hack" here
-        "openpgp/...",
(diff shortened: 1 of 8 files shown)


third_party/go/BUILD
```diff
@@ -19,16 +19,33 @@ go_module(
...
+    install = [
+        "bitcurves",
+        "brainpool",
+        "cast5",
+        "curve25519",
+        "eax",
+        "ed25519",
+        "internal/byteutil",
+        "internal/randutil",
+        "internal/syscall/unix",
```
Tatskaari, line 39: This probably wont work for freebsd

peterebden:
Nope, not related, just saw the announcement that x/crypto/openpgp was deprecated and realised we were using it still.

PR #1710, opened 2021-03-30
Fix maven jar hashes being applied to sources instead of main jar

Tatskaari:


rules/java_rules.build_defs
```diff
@@ -368,7 +368,8 @@ def maven_jar(name:str, id:str, repository:str|list=None, hash:str=None, hashes:
       id (str): Maven id of the artifact (eg. org.junit:junit:4.1.0)
       repository (str | list): Maven repositories to fetch deps from.
       hash (str): Hash for produced rule.
-      hashes (list): List of hashes for produced rule.
+      hashes (list): List of hashes downloaded classes jar.
+      sources_hashes (list): List of hashes for the downloaded sources jar.
```
peterebden, line 372: source_hashes? The double plural sounds weird to me.

PR #1714, opened 2021-03-31
Adding Prometheus counter for 'downloading tree digest' errors

isobelormiston:
Testing-wise: I ran some builds against my local Please & checked that without a URL set, it logs and returns. I'm unable to test the actual Prometheus pushes without setting up a pushgateway instance. 

Remaining work will be to set the URL in our plzconfig, either by pointing to an existing pushgateway or setting up a new one.


src/remote/utils.go
```diff
@@ -35,6 +37,24 @@ import (
 // xattrName is the name we use to record attributes on files.
 const xattrName = "user.plz_hash_remote"
 
+func (c *Client) pushTreeDigestDownloadError() {
+	if c.state.Config.Remote.PrometheusGatewayURL == "" {
+		log.Infof("No Prometheus pushgateway URL to push Digest Download error to")
```
Tatskaari, line 42: nit: Could we downgrade this to debug or remove it?

src/remote/utils.go
```diff
@@ -35,6 +37,24 @@ import (
...
+		return
+	}
+	downloadErrorCounter := prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "please",
+		Name:      "tree_digest_download_eof_error",
+	})
```
Tatskaari, line 48: Is this not meant to be long lived? Creating a new counter every time we push will result in the count always being 1 no?

src/remote/utils.go
```diff
@@ -35,6 +37,24 @@ import (
...
+	if err := push.New(
+		c.state.Config.Remote.PrometheusGatewayURL, "tree_digest_download",
+	).Collector(downloadErrorCounter).Push(); err != nil {
+		log.Warningf("Error pushing to Prometheus pushgateway: %s", err)
+		return
+	}
+	log.Infof("Incremented error counter via pushgateway")
```
Tatskaari, line 55: Again not sure this is info worthy. I feel info is more about letting you know what state the system is in mostly during setup e.g. "starting connection pool",  "listening on port 8080" and stuff. We don't want to log for business as usual things.

Tatskaari (commented):
Nice one Izzy! That go.sum update is gunna put you right up there on lines contributed haha

Just got a couple of suggestions/queries but otherwise looks good.

src/core/config.go
```diff
@@ -454,22 +454,23 @@ type Configuration struct {
...
-		CacheDuration cli.Duration `help:"Length of time before we re-check locally cached build actions. Default is unlimited."`
-		BuildID       string       `help:"ID of the build action that's being run, to attach to remote requests."`
+		URL                  string       `help:"URL for the remote server."`
+		CASURL               string       `help:"URL for the CAS service, if it is different to the main one."`
+		AssetURL             string       `help:"URL for the remote asset server, if it is different to the main one."`
+		NumExecutors         int          `help:"Maximum number of remote executors to use simultaneously."`
+		Instance             string       `help:"Remote instance name to request; depending on the server this may be required."`
+		Name                 string       `help:"A name for this worker instance. This is attached to artifacts uploaded to remote storage." example:"agent-001"`
+		DisplayURL           string       `help:"A URL to browse the remote server with (e.g. using buildbarn-browser). Only used when printing hashes."`
+		PrometheusGatewayURL string       `help:"The gateway URL to push prometheus updates to."`
```
peterebden, line 464: I think we should move this to its own section rather than being part of `remote` - can imagine we will want metrics for other parts of plz.

peterebden (commented):
Thanks!

src/core/config.go
Tatskaari:
```
[metrics]
PrometheusGatewayULR = 
```

probably makes the most sense.

peterebden:
FYI I removed the last metrics in 5bbebff982ef8262478f122b050f415cb148a85b, that may be of interest if you want to try to resurrect any parts of it.

src/remote/utils.go
isobelormiston: I've tested this some more locally and checked with observability, and it doesn't reset the counter on the gateway. I have moved it though so it's only called when the remote client is initialised, so that we're not calling it for every counter increment.

isobelormiston:
Ah thanks, I've moved the new metric code to remote/metrics.go but maybe somewhere more general will make sense if we're going to be adding metrics elsewhere.

peterebden:
Yeah, I think we should set this up as a new package in `src/metrics` so it's independent (maybe we define rex-specific metrics in `src/remote` still?).

Sorry, I guess it's turning into a more general thing than you intended originally :)

isobelormiston:
No problem, best to get it set up properly now and then it'll be easier to add future metrics :) I will rearrange a bit

src/metrics/prometheus.go
```diff
@@ -0,0 +1,42 @@
...
+		Help: "Number of times the Unexpected EOF error has been seen during a tree digest download",
+	})
+
+	return &RemoteMetrics{
+		downloadErrorCounter: downloadErrorCounter,
+	}
+}
+
+// DownloadErrorCounterInc increments the 'tree digest download error' counter
+func (r *RemoteMetrics) DownloadErrorCounterInc(gatewayURL string) {
```
Tatskaari, line 31: Might make more sense to store the gateway url as a string in the struct and pass it in to NewRemoteMetrics()

Tatskaari (approved):
Thanks Izzy! Left a suggestion but otherwise looks good. Let me know when it's ready to land.

src/metrics/prometheus.go
isobelormiston: Yep that is neater (I've also made it just a metrics singleton now, which can be used for any future metrics), thanks Jon :)

peterebden (approved):

PR #1716, opened 2021-04-06
Fix well known types not being included because protoc is a tool

### LINKED ISSUE #1703: Well-known types are still missing
Following #1655 and #1641, I did some tests with but I was unable to make it works.

It seems that the dependency is not added because it is already a tools of the rule.

Here is what I did:

With `//third_party/proto:protoc` as `ProtocTool`, only the `//third_party/proto:protoc` rule is returned as a dependency of the `proto_library()` by `plz query deps --hidden`.

Same with `//third_party/proto:protoc` as `ProtocTool` and an explicit dependency on `//third_party/proto:protoc` in the `proto_library()`.

But without any `ProtocTool` and with `//third_party/proto:protoc`  as dependency of the `proto_library()` rule, `plz query deps --hidden` returns `//third_party/proto:_protoc#wkt`.

Tatskaari:
Fixes #1703

### DIFF
--- a/rules/proto_rules.build_defs
+++ b/rules/proto_rules.build_defs
@@ -53,6 +53,19 @@ def proto_library(name:str, srcs:list, deps:list=[], visibility:list=None, label
     pre_build_functions = [plugin['pre_build'] for plugin in plugins if plugin['pre_build']]
     pre_build_functions += [_collect_transitive_labels]
     pre_build = lambda rule: [fn(rule) for fn in pre_build_functions]
+
+    # If the protoc tool is a build rule, depend on it as it's probably a protoc_binary() which provides the well known
+    # types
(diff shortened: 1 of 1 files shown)


rules/proto_rules.build_defs
```diff
@@ -483,7 +491,6 @@ def protoc_binary(name, url:str|dict = '', version:str = '', hashes=None, deps=N
         cmd = '$TOOL x $SRCS --strip_prefix include',
         labels = [
             'protoc:-I"$TMP_DIR"/' + package_name(),
-            'protoc:-I"$TMP_DIR"',
```
Tatskaari, line 486: We add `-I.` to all protoc invocations so this isn't needed and was actually causing issues.

peterebden (approved):

PR #1718, opened 2021-04-06
Implement revdeps without lookahead

Tatskaari:
Before we were doing this:
```
for _, child := range graph.PackageOrDie(label).AllChildren(graph.TargetOrDie(label)) {
	for _, target := range graph.ReverseDependencies(child) {
		if parent := target.Parent(graph); parent != nil {
			uniqueTargets[parent] = struct{}{}
		} else {
			uniqueTargets[target] = struct{}{}
		}
	}
}
```
which was quite slow for `query changes` which would pass in many more targets and evaluate a much larger graph. 

This implements a more traditional breadth first approach which increments the depth only when passing boundaries between rules (ignoring hidden targets of the same rule).

### DIFF
--- a/src/core/build_label.go
+++ b/src/core/build_label.go
@@ -354,6 +354,11 @@ func (label BuildLabel) Parent() BuildLabel {
 	return label
 }
 
+// IsHidden return whether the target is an intermediate target created by the build definition.
+func (label BuildLabel) IsHidden() bool {
+	return label.Name[0] == '_'
+}
+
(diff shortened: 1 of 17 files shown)


Tatskaari:
prerelease.15 takes over 15 minutes (I gave up waiting) to query changes for my PR whereas this approach takes 39s which is mostly taken up by creating the build graph.

peterebden (approved):

PR #1721, opened 2021-04-07
Fix go_get attribute in dependencies docs

leonore:
https://please.build/dependencies.html has the following code for Go:
```
    go_get(
        name = "my_library",
        get = "github.com/me/my_library",
        version = "1.0.0",
    )
```
however go_get uses the revision attribute, not version


Tatskaari (approved):
Opps thanks for that

PR #1724, opened 2021-04-08
Don't set outputs on metadata retrieval

peterebden:
Some investigation shows that we're doing this twice in most cases (it's also done in `Build`).
It is incorrect to do it for tests though; we end up passing the test outputs into future test runs (can reproduce by doing `plz test -n 50 //something` more than once; the second and subsequent ones see an arbitrary number of those runs fail).


Tatskaari (approved):
Thanks for that. Surprised we didn't notice this earlier.

PR #1725, opened 2021-04-08
Improve user experience around ... in go_module()

Tatskaari:
This makes `foo/...` install `foo` as well as everything under `foo`. This avoids having to have:
```
go_module(
   ...
   install = ["foo", "foo/..."],
)
```


peterebden (approved):

Tatskaari:
Going to merge this closer to when we cut a release as it's a pain when switching between go versions.

PR #1726, opened 2021-04-08
plz cover --shell implies --nocoverage_report

peterebden:
Because when you exit the shell, the test is treated as failed, and you don't want to see screeds of files with zero coverage.


Tatskaari (approved):
Maybe do this for `--prepare_only` too?

peterebden:
`--prepare` only exists on `plz build` so it doesn't really apply here.

Honestly don't think it's that useful (I forget but I think maybe it is a weak predecessor to `--shell`) but wevs, `--build` is not drowning in flags anyway.

PR #1731, opened 2021-04-10
feat: retrieve optional outs from cache when using wildcard

sfirmery:
This enhance #1727 to also retrieve optional outs that use wildcard.

To do this, exact optional outputs names are added to the target metadata.

What do you think about adding exact optional output filenames in the target metadata?


Tatskaari (approved):
Nice one! Thanks for that.

PR #1733, opened 2021-04-12
Add --in_tmp_dir to run in a hermetic tmp dir

Tatskaari:


rules/go_rules.build_defs
```diff
@@ -40,7 +40,7 @@ def go_toolchain(name:str, url:str|dict = '', version:str = '', hashes:list = []
         hashes = hashes,
     )
 
-    cmd = 'tar -xf $SRCS && mv go $OUT && chmod +x $OUT/bin/*'
+    cmd = 'tar -xf $SRCS && mv go $OUT && chmod +x $OUT/bin/*; rm -rf $OUT/test'
```
Tatskaari, line 43: Unrelated but this reduces the size of the go_toolchain() quite a bit

src/run/run_step.go
```diff
@@ -55,22 +61,30 @@ func Parallel(ctx context.Context, state *core.BuildState, labels []core.Annotat
...
 			log.Error("%s", err)
 			return err.(*exitError).code
 		}
 	}
 	return 0
 }
 
+func prepareRun(dir string, inTmp bool) {
+	if inTmp && dir != "" {
+		log.Fatal("Can't use both --in_temp_dir and --in_wd at the same time")
```
peterebden, line 78: This should happen in `please.go` - we shouldn't need to refer to flags here.
(Also mildly nicer to error out immediately rather than do all the building then fail...)

src/run/run_step.go
```diff
@@ -55,22 +61,27 @@ func Parallel(ctx context.Context, state *core.BuildState, labels []core.Annotat
...
+		if err := run(state, label, args, true, quiet, remote, env, false, inTmp, dir); err != nil {
 			log.Error("%s", err)
 			return err.(*exitError).code
 		}
 	}
 	return 0
 }
 
+func prepareRun(dir string, inTmp bool) {
+	os.RemoveAll("plz-out/run")
```
peterebden, line 77: Do you need to do anything if this doesn't work?

peterebden (approved):

Issue #797, opened 2019-12-26
go_binary requires gcc


peterebden:
We attach gcc (or whatever's configured) as a tool to pass to `-extld` in case it's linking in external mode. However if we're not it always fails fast if gcc isn't installed.

Not very obvious how we fix since it's hard for us to know which link mode will be needed. Could have a config option but it seems fiddly.

Tatskaari:
Seems reasonable that if your project requires an external linker you must add it to the go config:
```
[go]
extld = clang
```
We then only pass in the link too if this is set. 

@peterebden thoughts?

peterebden:
I guess that is reasonable. It seems slightly fiddly to have `go.extld` as well as `cpp.ldtool` but I guess it is more flexible and explicit.

Tatskaari:
Seems like we can unify all the go CC config into a single `go.cctool` as described in #1736. Going to push this back to v17 though.

Tatskaari:
Duplicate of https://github.com/thought-machine/please-issues/issues/40

PR #1736, opened 2021-04-13
Add extld config for go

Tatskaari:
Fixes: #797


.plzconfig
```diff
@@ -19,6 +19,7 @@ pleasegotool = //tools/please_go
 embedtool = //tools/please_go_embed
 importpath = github.com/thought-machine/please
 gotool = //third_party/go:toolchain|go
+extldtool = cc
```
peterebden, line 22: Why do we need this? Do we need an extld for Please?

peterebden (commented):
Looks good but a little curious why we need to keep this at all?

Tatskaari:
> Looks good but a little curious why we need to keep this at all?

The go tools will look for gcc on the path anyway. I could put this in .plzconfig.freebsd but I feel this controls that a bit better and allows Please to fail fast rather than waiting for the go tools to fail.

.plzconfig
Tatskaari: The go tools will look for gcc on the path anyway. I could put this in .plzconfig.freebsd but I feel this controls that a bit better and allows Please to fail fast rather than waiting for the go tools to fail.

peterebden: Oh ok so we basically always need a C compiler for Go? I guess we inevitably pick up some cgo along the line...

peterebden (approved):

.plzconfig
Tatskaari: > Oh ok so we basically always need a C compiler for Go? I guess we inevitably pick up some cgo along the line...

Oh my apologies, it's for the link static and cgo tests that we need it. Not all go code. If we don't set this, we can still build Please. We just can't run the tests.

Tatskaari:
Seems like the right thing to do would be combine the cgo CC, cgo enabled, and extld tool configs into one `go.LDTool` however that has consequences because there's no easy way to pass ld flags and breaks some musl/glibc stuff for us at Thought Machine. Instead of rushing in this new config option into v16, I'm going to push this back to v17.

PR #1742, opened 2021-04-15
Fix the http cache for post build functions

Tatskaari:
Seems that we can end up failing to write the cache results back into plz-out when we have a post build function, and the target is binary:

Given the initial state:
1. Stuff in plz-out is out of date
2. The dir cache is empty
3. The http cache is populated 
4. The target is binary

Please will:
1) Need to rebuild as plz-out is out of date
2) Restore the metadata file from the http cache into plz-out. This also restores all the other files as the http cache doesn't partially restore a cached item. 
3) Run the post-build function
4) Restore the artefacts from the http cache
5) Fail to override the files as they are read only

This has always been a problem however will become more of an issue as we introduced OptionalOutputs which will end up fetching from the cache again.


src/cache/http_cache.go
```diff
@@ -195,6 +195,16 @@ func (cache *httpCache) retrieve(target *core.BuildTarget, key []byte) (bool, er
 	}
 }
 
+func openFile(header *tar.Header) (*os.File, error) {
+	f, err := os.OpenFile(header.Name, os.O_WRONLY|os.O_TRUNC|os.O_CREATE, os.FileMode(header.Mode))
+	if err != nil {
+		// The file might already exist and be ro. If so, remove it.
```
peterebden, line 201: maybe we should check `os.IsPermission`?

PR #1745, opened 2021-04-16
Fix query changes to not follow subincludes like revdeps

Tatskaari:


tools/http_cache/main.go
```diff
@@ -16,7 +16,7 @@ var opts = struct {
 	Usage     string
 	Verbosity cli.Verbosity `short:"v" long:"verbosity" default:"warning" description:"Verbosity of output (higher number = more output)"`
 	CacheDir  string        `short:"d" long:"dir" default:"" description:"The directory to store cached artifacts in."`
-	Port      int           `short:"p" long:"port" description:"The port to run the server on"`
+	Port      int           `short:"p" long:"port" description:"The port to run the server on" default:"8080"`
```
peterebden, line 19: not related to this change?

peterebden (approved):

tools/http_cache/main.go
Tatskaari: Ah sorry, I was doing some work on testing the cache. Probably nice to have a default though...

PR #1758, opened 2021-04-19
Move cc_embed_binary into //test/build_defs

peterebden:


.circleci/rex_test.sh
```diff
@@ -4,12 +4,13 @@ set -eu
 
 trap 'killall elan mettle zeal' SIGINT SIGTERM EXIT
 
-DIR="${1:-/tmp/please}"
+DIR="/tmp/please"
+WORKSPACE_DIR="${1:-/tmp/workspace/linux_amd64}"
```
Tatskaari, line 8: What's the significance of this?

Tatskaari (approved):

.circleci/rex_test.sh
peterebden: Configuring the tmp dir it uses didn't seem very useful. If you want to run it locally, you can now do it via `plz build //package:please_tarball` and pointing it to `plz-out/gen/package`.

PR #1762, opened 2021-04-20
V16 milestone

Tatskaari:


src/please.go
```diff
@@ -386,6 +386,9 @@ var buildFunctions = map[string]func() int{
 		return toExitCode(success, state)
 	},
 	"hash": func() int {
+		if opts.Hash.Update {
+			opts.FeatureFlags.NoHashVerification = true
+		}
```
Tatskaari, line 391: This can be frustrating. It will ignore hash verification for the targets you specify which usually works. In core3 it complains before we have some third party java stuff outside `//third_party/java`

peterebden: Think this makes sense, I've found it annoying before to get things updated because some stuff is failing along the way.

It is a _little_ naughty since it means we're not verifying the dependencies but I think the UX is more relevant (presumably it would be caught later by CI if someone was actually trying to attack that way).

peterebden (approved):

PR #1766, opened 2021-04-21
WIP: move sandbxing into go

Tatskaari:


src/sandbox/sandbox.go
```diff
@@ -0,0 +1,74 @@
...
+		return fmt.Errorf("failed to bring loopback interface up")
+	}
+
+	fmt.Println("Welcome to the sandbox!")
+	return cmd.Run()
+}
+
+func mountTmp() error {
+	tmpDir := os.Getenv("TMP_DIR")
+	dir := "/tmp/plz_sandbox"
```
peterebden, line 36: this should probably use `core.SandboxDir` - or that should use this - until we can unify it all in here.

src/sandbox/sandbox.h
```diff
@@ -0,0 +1 @@
+int lo_up();
```
peterebden, line 1: You should put the function comment here rather than in the .c file; this is the interface that callers would read.

src/core/build_target.go
```diff
@@ -1430,6 +1430,11 @@ func (target *BuildTarget) AddNamedOutput(name, output string) {
 	target.namedOutputs[name] = target.insert(target.namedOutputs[name], output)
 }
 
+// AddOutput adds a new output to the target if it's not already there.
```
peterebden, line 1433: needs update

src/core/build_target.go
```diff
@@ -1430,6 +1430,11 @@ func (target *BuildTarget) AddNamedOutput(name, output string) {
...
+func (target *BuildTarget) ShouldSandbox() bool {
```
peterebden, line 1434: What's the point of this function? It could be useful to hide Sandbox but I'm not sure we need to - also TestSandbox is separate which this doesn't account for.

src/core/state.go
```diff
@@ -1044,7 +1044,7 @@ func NewBuildState(config *Configuration) *BuildState {
 			"crc64":  fs.NewPathHasher(RepoRoot, config.Build.Xattrs, newCRC64, "crc64"),
 			"blake3": fs.NewPathHasher(RepoRoot, config.Build.Xattrs, newBlake3, "blake3"),
 		},
-		ProcessExecutor: process.New(sandboxTool),
+		ProcessExecutor: process.New(sandboxTool != ""),
```
peterebden, line 1047: I think we may still want a way of honouring the sandbox tool if that's set. Maybe it should basically go through the old method if it is - but we make the default `""` from here on and deprecate that field in future.

peterebden (commented):
Uncomfortably excited about this!

src/sandbox/sandbox.go
Tatskaari: I was thinking the benefit of doing something like this would be to make it parse os.Args() or do you think that would be too slow?

src/core/build_target.go
Tatskaari: There's a target interface that the process stuff uses that looks at a subset of the functions on target. I can't expose fields through an interface so needed to add this. I guess we might have this be `ShouldSandboxNet()` and `ShouldSandboxMount()` and have `target.Sandbox` be `all`, `none`, `net`, or `mount`.

src/core/state.go
Tatskaari: Yeah sounds like a plan

src/core/build_target.go
Tatskaari: I don't think this is quite right. We need to check both Sandbox and TestSandbox. Will have to add another param to the exec then. :(

build_defs/multiversion_wheel.build_defs
```diff
@@ -30,7 +30,7 @@ def python_multiversion_wheel(name, version:str, urls:list, licences:list=None,
         name = name,
         srcs = file_rules,
         outs = [name],
-        cmd = 'for SRC in $SRCS; do $TOOL x -o $TMPDIR $SRC; done && rm -rf %s/*.egg-info %s/*.dist-info',
```
Tatskaari, line 33: For some reason the sandbox tool sets $TMPDIR to /tmp. Not sure if that's needed by some core util but it did break this.

src/core/config.go
```diff
@@ -236,6 +236,10 @@ func ReadConfigFiles(filenames []string, profiles []string) (*Configuration, err
 		config.Sandbox.Tool = config.Build.PleaseSandboxTool
 	}
 
+	if config.Sandbox.Namespace == "" && (config.Sandbox.Build || config.Sandbox.Test) {
+		config.Sandbox.Namespace = "true"
```
Tatskaari, line 240: This is kinda lame. We need 3 state logic: Not set, set to false, set to true.

build_defs/multiversion_wheel.build_defs
peterebden: I think it resets TMPDIR and TMP_DIR so they don't point at the old plz-out directory.

Why does this thing care which it is?

src/process/process.go
```diff
@@ -19,24 +19,43 @@ import (
 
 var log = logging.MustGetLogger("progress")
 
+const (
+	NamespaceAlways = "always"
```
peterebden, line 23: Worth making these a custom type too?

src/sandbox/sandbox_linux.go
```diff
@@ -0,0 +1,131 @@
...
+	"strings"
+	"syscall"
+
+	// #include "sandbox.h"
+	"C"
+)
+
+
+// mdLazytime is the bit for lazily flushing disk writes.
+// TODO(jpoole): find out if there's a reason this isn't in syscall
```
peterebden, line 18: I have a vague concept it might be relatively new.

peterebden (approved):

build_defs/multiversion_wheel.build_defs
Tatskaari: Well it's accessing $TMPDIR but that's been set to `/tmp`. I guess it should be `/tmp/please_sandbox` right?

PR #1768, opened 2021-04-23
Add output grouping option to plz run parallel

tiagovtristao:
First iteration on adding output grouping to `plz run parallel`.

- It introduces a new `--output` flag allowing it to be extended in the future
- The current `--quiet` could be deprecated in favour of `--output=quiet`?
- The `group_immediate` option flushes output for a task in one go once it finishes.


src/please.go
```diff
@@ -169,6 +169,7 @@ var opts struct {
 		Parallel   struct {
 			NumTasks       int  `short:"n" long:"num_tasks" default:"10" description:"Maximum number of subtasks to run in parallel"`
 			Quiet          bool `short:"q" long:"quiet" description:"Suppress output from successful subprocesses."`
+			Output         run.ParallelOutput `short:"o" long:"output" description:"Output (default, quiet, group_immediate)" default:"default"`
```
peterebden, line 172: FYI you can get the flags library to handle some of this for you :)
```
choice:"default" choice:"quiet" choice:"group_immediate"
```
It should generate appropriate help text, but will also ensure the value is one of those choices.

src/please.go
```diff
@@ -169,6 +169,7 @@ var opts struct {
 		Parallel   struct {
 			NumTasks       int  `short:"n" long:"num_tasks" default:"10" description:"Maximum number of subtasks to run in parallel"`
 			Quiet          bool `short:"q" long:"quiet" description:"Suppress output from successful subprocesses."`
```
peterebden, line 171: +1 on deprecating this in favour of `--output` since that is more general.

src/run/run_step.go
```diff
@@ -25,6 +25,30 @@ import (
 
 var log = logging.MustGetLogger("run")
 
+type ParallelOutput uint
```
peterebden, line 28: Might be easier to make this guy a string? Then you wouldn't have to worry about unmarshalling it (combined with the `choice` options above)

src/run/run_step.go
```diff
@@ -46,7 +70,16 @@ func Parallel(ctx context.Context, state *core.BuildState, labels []core.Annotat
 		g.Go(func() error {
 			limiter <- struct{}{}
 			defer func() { <-limiter }()
-			return run(state, label, args, true, quiet, remote, env, detach, inTmp, dir)
+
+			quiet := output == Quiet || output == GroupImmediate
+			out, err := run(state, label, args, true, quiet, remote, env, detach, inTmp, dir)
+
+			if output == GroupImmediate {
+				fmt.Println(label)
```
peterebden, line 78: This will only prefix the label at the start of the whole output, and will wait for it to complete before printing anything.

I think you might want to wrap the `Stdout` and `Stderr` of the command with a writer that prefixes them?

peterebden (commented):
Thanks Tiago!

src/please.go
tiagovtristao: Nice!

src/run/run_step.go
tiagovtristao: Agreed, with the `choice` options this gets simplified.

src/run/run_step.go
tiagovtristao: At the moment it is printing the label and output at the same time, and only after the task finishes.

Having the following:
```python
sh_cmd(
    name = "one",
    cmd = [
        "echo one",
        "sleep 3",
    ],
)

sh_cmd(
    name = "two",
    cmd = [
        "echo two",
        "sleep 2"
    ],
)
```
And running `plz run //src:please -- run parallel //docs:one //docs:two --output=group_immediate`

Outputs the following after 2s
```
//docs:two
two
```

and then the following after 1s (2s from the previous)
```
//docs:one
one
```

Unless I'm missing something?

Tatskaari: @pebers I think VJ wanted some way to group the output as a priority. The prefix idea was just something I thought of, but wasn't quite what was requested. We talked about printing the prefixed stuff during execution that so we get some partial output before clearing the output and printing this all this. I'd rather get on to the VS code debugging though.

tiagovtristao: Yes, this does the grouping but not the prefixing on every line. The reason for this was the possibility of long build label prefixes during discussion. Sticking it to the top would help avoid that problem and still help differentiate the output. I guess we could make the output top-line label more pronounced?

src/please.go
```diff
@@ -167,8 +167,9 @@ var opts struct {
 		InTempDir  bool   `long:"in_tmp_dir" description:"Runs in a temp directory, setting env variables and copying in runtime data similar to tests."`
 		EntryPoint string `long:"entry_point" short:"e" description:"The entry point of the target to use." default:""`
 		Parallel   struct {
-			NumTasks       int  `short:"n" long:"num_tasks" default:"10" description:"Maximum number of subtasks to run in parallel"`
-			Quiet          bool `short:"q" long:"quiet" description:"Suppress output from successful subprocesses."`
+			NumTasks       int                `short:"n" long:"num_tasks" default:"10" description:"Maximum number of subtasks to run in parallel"`
+			Quiet          bool               `short:"q" long:"quiet" description:"Suppress output from successful subprocesses."`
```
Tatskaari, line 171: Can we mention that this is deprecated here too? Also we should raise an issue to remove it and add it to the v17 milestone.

src/run/run_step.go
peterebden: Oh OK sorry, I was just thinking of something different then.

peterebden (approved):

src/run/run_step.go
```diff
@@ -46,7 +54,20 @@ func Parallel(ctx context.Context, state *core.BuildState, labels []core.Annotat
...
+
+			quiet := output == Quiet || output == GroupImmediate
+			out, combined, err := run(state, label, args, true, quiet, remote, env, detach, inTmp, dir)
+
+			if output == GroupImmediate {
+				fmt.Println(label)
+				if err != nil {
+					os.Stderr.Write(combined)
+				} else {
+					os.Stdout.Write(out)
```
Tatskaari, line 66: Doesn't this print it twice? I think out is already encoded in `err` below by `toExitError()`.

Tatskaari (approved):
Looking good! Just got a couple comments.

src/please.go
tiagovtristao: Sure. I might go ahead as well and clarify the removal version on the warning below.

src/run/run_step.go
tiagovtristao: It now logs all errors without duplication. Did some refactoring for clarity and easy future enhancements.

Tatskaari (approved):
Look good! Thanks Tiago

PR #1769, opened 2021-04-23
Fix issue with runtime hashes for local tests

Tatskaari:


src/test/test_step.go
```diff
@@ -565,10 +565,16 @@ func verifyHash(state *core.BuildState, filename string, hash []byte) bool {
 }
 
 // runtimeHash returns the runtime hash of a target, or an empty slice if running remotely.
-func runtimeHash(state *core.BuildState, target *core.BuildTarget, runRemotely bool, run int) ([]byte, error) {
+func runtimeHash(tid int, state *core.BuildState, target *core.BuildTarget, runRemotely bool, run int) ([]byte, error) {
 	if runRemotely {
 		return nil, nil
 	}
+	if target.Local {
+		// If the test is marked to run locally, download the inputs as we need these to calculate the rubtime hash.
```
peterebden, line 573: `runtime`

I assume a rubtime hash would be something else... ;-)

PR #1771, opened 2021-04-23
Record xattrs for the hash when retrieving from the http cache

### LINKED ISSUE #1767: Issue with cached data from HTTP cache
Hello, I'm having an issue v16 and HTTP cache.

HTTP cache server is an Apache with mod_dav using container image `bytemark/webdav:2.4`.

With dir cache and HTTP cache defined, and a simple test like:

```
gentest(
    name="test",
    srcs = ["some_file"],
    test_cmd = "exit 0",
    no_test_output=True,
)
```

Results data from HTTP cache are incorrect.

Here is how to reproduce:

```
plz test //:test       # put the result in the HTTP cache
plz clean //:test && plz test //:test    # clean the local dir cache and then get the results from the HTTP cache
plz test //:test      # test is reran instead of use results in local dir cache
```

In the debug of the last test command, there is:

```
DEBUG: Rerunning //:test, results file has incorrect hash
```

Tatskaari:
Fixes #1767

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -462,6 +462,16 @@ func retrieveArtifacts(tid int, state *core.BuildState, target *core.BuildTarget
 			state.LogBuildResult(tid, target.Label, core.TargetCached, "Cached (unchanged)")
 		}
 		buildLinks(state, target)
+
+		// If we could've potentially pulled from the http cache, we need to write the xattrs back as they will be
+		// missing.
+		if state.Config.Cache.HTTPURL != "" {
+			if err := writeRuleHash(state, target); err != nil {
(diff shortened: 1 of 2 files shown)


peterebden (approved):
Ah I see what you meant about this, sucks a bit that we have to handle this stuff here though :(

Tatskaari:
> Ah I see what you meant about this, sucks a bit that we have to handle this stuff here though :(

Maybe we can make the cache use the full hash and then the http cache can write the xattr when we retrieve from it. To be honest there are a few things I'd like to change about the http cache i.e. that it writes all the files to plz-out even if you only asked for a few. I think it should write all files to the dir cache but only the files we asked for to plz-out.

PR #1773, opened 2021-04-26
Allow PLZ_CONFIG_PROFILE to take multiple entries

peterebden:
Separated with a semicolon; thought about using space but we already have prior art for PLZ_OVERRIDES.


Tatskaari (approved):
Nice cleanup

peterebden:
Thanks, I thought the custom type was breaking env-delim at first, but turned out just to be me getting it wrong :facepalm: 
It's fewer lines of code now though so clearly better ;-)

PR #1774, opened 2021-04-26
Implement per-function profiling metrics for asp

peterebden:
Gives some little subdivisions on hot functions about how long we've spent in each one because of a particular build function.

It's a little limited since we only spend just so long in each one, but still seems nice to have (and it's not very hard to do)

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -991,6 +991,9 @@ func readConfigAndSetRoot(forceUpdate bool) *core.Configuration {
 	}
 	config := readConfig(forceUpdate)
 	// Now apply any flags that override this
+	if opts.Profile != "" {
+		config.Profiling = true
+	}
 	if opts.Update.Latest || opts.Update.LatestPrerelease {
 		config.Please.Version.Unset()
 	} else if opts.Update.Version.IsSet {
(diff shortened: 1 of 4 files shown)


src/please.go
```diff
@@ -991,6 +991,7 @@ func readConfigAndSetRoot(forceUpdate bool) *core.Configuration {
 	}
 	config := readConfig(forceUpdate)
 	// Now apply any flags that override this
+	config.Profiling = opts.Profile != ""
```
Tatskaari, line 994: Perhaps we should set the profile file to some default value (`please.prof`?) if `config.Profile` has been set.

Tatskaari (approved):
Nice! Does this hook into the same dot file as the normal please profile or does it produce it's own call graph? I'm concerned it might double count stuff if it's profiling the interpreter interpreting the asp code AND the asp code itself.

src/please.go
Tatskaari: Ah, I see, this is a psudo config value that can't be set in the `.plzconfig`. Think this makes sense.

peterebden:
Yeah it appears in the same profile file - it doesn't count twice, it's more like it subdivides the existing CPU time into these other slices. I think it is just strictly adding information.

Tatskaari:
> Yeah it appears in the same profile file - it doesn't count twice, it's more like it subdivides the existing CPU time into these other slices. I think it is just strictly adding information.

Well that sounds perfect! Nice one :)

tiagovtristao (approved):
Nice!

Issue #1764, opened 2021-04-21
Fix the download script for arm once v16 is cut


Tatskaari:
The script at get.please.build downloads the amd version of Please on arm systems. Once we have v16 out, we can change this to download the arm version.

linuxerwang:
I got tons of repeated errors:

```
4 errors occurred:
        * Get "https://get.please.build/third_party/python/py3/parse-1.15.0-linux_arm.whl": x509: certificate signed by unknown authority
        * Get "https://get.please.build/third_party/python/py3/parse-1.15.0.whl": x509: certificate signed by unknown authority
        * Error retrieving https://files.pythonhosted.org/packages/py3/p/parse/parse-1.15.0-py3-none-any.whl: 404 Not Found
        * Error retrieving https://files.pythonhosted.org/packages/py2.py3/p/parse/parse-1.15.0-py2.py3-none-any.whl: 404 Not Found
```

Tatskaari:
> I got tons of repeated errors:
> 
> ```
> 4 errors occurred:
>         * Get "https://get.please.build/third_party/python/py3/parse-1.15.0-linux_arm.whl": x509: certificate signed by unknown authority
>         * Get "https://get.please.build/third_party/python/py3/parse-1.15.0.whl": x509: certificate signed by unknown authority
>         * Error retrieving https://files.pythonhosted.org/packages/py3/p/parse/parse-1.15.0-py3-none-any.whl: 404 Not Found
>         * Error retrieving https://files.pythonhosted.org/packages/py2.py3/p/parse/parse-1.15.0-py2.py3-none-any.whl: 404 Not Found
> ```

Seems like it can't find a wheel for your arch. Are you trying to bootstrap on arm? I don't know if we're going to support running all the tests on arm any time soon (or ever). You can still bootstrap please but skip running the tests with `./bootstrap.sh --skip_tests`. Perhaps we can update the script to do that automatically on architectures we don't plan on supporting those tests for.

Tatskaari:
p.s. the only thing blocking this change is that I can't seem to get the CPU resource usage library to work. It keeps throwing unimplemented errors.

linuxerwang:
@Tatskaari  ./bootstrap.sh --skip_tests gives the same errors (tried on v16.3.0). Thanks.

Tatskaari:
[...]

PR #1778, opened 2021-04-27
Update the get script to support arm

Tatskaari:
Fixes: #1764

### DIFF
--- a/pleasew
+++ b/pleasew
@@ -42,19 +42,27 @@ fi
 
 # Find the os / arch to download. You can do this quite nicely with go env
 # but we use this script on machines that don't necessarily have Go itself.
-OS=`uname`
-if [ "$OS" = "Linux" ]; then
-    GOOS="linux"
-elif [ "$OS" = "Darwin" ]; then
-    GOOS="darwin"
-else
-    echo -e >&2 "${RED}Unknown operating system $OS${RESET}"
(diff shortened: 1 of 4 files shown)


peterebden (approved):

pleasew
```diff
@@ -42,19 +42,27 @@ fi
...
+    GOOS="linux";;
+  Darwin)
+    GOOS="darwin";;
+  *)
+    echo "Unknown operating system $(uname)"
+    exit 1;;
+esac
+
+case "`uname -m`" in
+  arm*)
```
VJftw, line 56: on a raspberry pi with a 64-bit kernel enabled, we get 
```
$ uname -m
aarch64
$ uname -a
Linux xxxxx 5.10.17-v8+ #1421 SMP PREEMPT Thu May 27 14:01:37 BST 2021 aarch64 aarch64 aarch64 GNU/Linux
```
aarch64 supposedly is the same as arm64 (https://stackoverflow.com/a/47274698) so I think this could be expanded
to

```bash
case "`uname -m`" in
  arm* | aarch*)
     GOARCH="arm64" ;;
...
``` 
if that makes sense to you too :)

peterebden: Yeah +1, it's one of those things like amd64 / x86_64. Naming is hard...

Issue #1779, opened 2021-04-27
`watch --run` not killing original process


tannerntannern:
First, thank you for such a nice build system!  I am new to tools like bazel, buck, pants, please, etc, but please has so far been the easiest to wrap my head around and requires less boilerplate than the others.

I am facing an issue that was possibly covered in #599, but it doesn't seem to be resolved in my testing.  My issue involves running a flask webserver and using `plz watch --run ...` for local development.  It appears that the original process isn't killed before please tries to restart it, so I get the expected "address in use" error.

I have an easy minimal reproducible example available here where I document the steps to reproduce in more detail: https://github.com/tannerntannern/plz-watch-run-flask-reprex

It's using Please 16.0.1 and my testing was done on macOS 10.15.6

Tatskaari:
Glad you're enjoying it so far! Thanks for raising this with an example to reproduce. Looks like we thought about this, and tried to solve it, but didn't pass the context through... 

I just did a Please release today so you'll probably have to wait until next Tuesday for `16.1.0`. Apologies for the delay.

PR #1780, opened 2021-04-27
Fix plz watch --run process cleanup

Tatskaari:
Fixes: #1779


src/process/process.go
```diff
@@ -103,9 +103,9 @@ func (e *Executor) ExecWithTimeout(target Target, dir string, env []string, time
 	select {
 	case err = <-ch:
 		// Do nothing.
-	case <-time.After(timeout):
+	case <-ctx.Done():
```
tiagovtristao, line 106: Cleaner, nice.

tiagovtristao (approved):
Looks good. Funny enough we talked about passing the context down in my previous PR.

PR #1781, opened 2021-04-29
Redo priority queue

peterebden:
Basically just uses channels instead of the previous priority queue.
I don't think we were _really_ prioritising things properly anyway (among other things, some of the channels were buffered so they could be filled up by lower-priority tasks) and honestly I'm not sure it matters that much.

Motivation: bit of simplification, also post #1102 the queue has become the highest source of lock contention.

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -253,18 +253,25 @@ func (state *BuildState) addPendingBuild(target *BuildTarget) {
 }
 
 // AddPendingTest adds a task for a pending test of a target.
-func (state *BuildState) AddPendingTest(target *BuildTarget, run int) {
-	if state.NeedTests {
-		atomic.AddInt64(&state.progress.numPending, 1)
-		task := TestTask{Label: target.Label, Run: run}
-		go func() {
-			if state.anyRemote && !target.Local {
-				state.pendingRemoteTests <- task
-			} else {
-				state.pendingTests <- task
-			}
-		}()
+func (state *BuildState) AddPendingTest(target *BuildTarget) {
+	if state.TestSequentially {
+		state.addPendingTest(target, 1)
+	} else {
+		state.addPendingTest(target, state.NumTestRuns)
 	}
+}
+
+func (state *BuildState) addPendingTest(target *BuildTarget, numRuns int) {
+	atomic.AddInt64(&state.progress.numPending, int64(numRuns))
+	go func() {
+		ch := state.pendingTests
+		if state.anyRemote && !target.Local {
+			ch = state.pendingRemoteTests
+		}
+		for run := 1; run <= numRuns; run++ {
+			ch <- TestTask{Label: target.Label, Run: run}
+		}
+	}()
 }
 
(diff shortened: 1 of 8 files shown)


src/plz/plz.go
```diff
@@ -38,20 +38,31 @@ func Run(targets, preTargets []core.BuildLabel, state *core.BuildState, config *
...
+		// All parses happen async on separate goroutines so we don't have to worry about them blocking.
+		// They manage concurrency control themselves.
+		for task := range parses {
+			go func(task core.ParseTask) {
+				parse.Parse(0, state, task.Label, task.Dependent, task.ForSubinclude)
+				state.TaskDone()
+			}(task)
+		}
+		wg.Done()
+	}()
```
Tatskaari, line 56: nit: Took me a little while to figure out what all the `go ...` things were doing here. Might read better if this was:

```
go doParses(parses)
```

src/core/state.go
```diff
@@ -268,135 +232,94 @@ func (state *BuildState) addActiveTargets(n int) {
...
 }
 
-func (state *BuildState) taskDone(wasSynthetic, wasBuildOrTest bool) {
+func (state *BuildState) taskDone(wasSynthetic bool) {
 	if !wasSynthetic {
 		atomic.AddInt64(&state.progress.numDone, 1)
 	}
-	if wasBuildOrTest {
-		atomic.AddInt64(&state.progress.numRunning, -1)
-	}
```
Tatskaari, line 379: Why do we not decrement this anymore? TaskDone is always called for a build or test now.

src/core/state.go
```diff
@@ -224,10 +188,10 @@ type stateProgress struct {
 	// Used to count the number of currently active/pending targets
 	numActive  int64
 	numPending int64
-	numRunning int64
```
Tatskaari, line 227: Why did this go?

src/core/state.go
```diff
@@ -895,7 +818,7 @@ func (state *BuildState) queueResolvedTarget(target *BuildTarget, dependent Buil
 
 // queueTarget enqueues a target's dependencies and the target itself once they are done.
 func (state *BuildState) queueTargetAsync(target *BuildTarget, dependent BuildLabel, rescan, forceBuild, forSubinclude, building bool) {
-	defer state.taskDone(true, false)
+	defer state.taskDone(true)
```
Tatskaari, line 821: A little confused by these "sythetic" tasks.

Tatskaari (approved):
Looks good to me. Just left a few comments to prove that I actually read it.  :D

src/plz/plz.go
peterebden: I have a bit of an aversion to passing the waitgroup out to another function...

src/core/state.go
peterebden: We increment it and decrement it, but nothing ever actually read it.

src/core/state.go
peterebden: Yeah, it just keeps the numbers in the display from going bonkers when we generate all these additional tasks

PR #1786, opened 2021-04-29
Sign get plz script

peterebden:
Makes it possible to GPG verify it before running.

### DIFF
--- a/docs/faq.html
+++ b/docs/faq.html
@@ -52,7 +52,7 @@
 </section>
 
 <section class="mt4">
-  <h2 class="title-2"><a name="verify">How can I verify the download of Please?</a></h2>
+  <h2 class="title-2" id="verify">How can I verify the download of Please?</h2>
 
   <p>
     All our release artifacts are signed with a GPG key. If you want assurance that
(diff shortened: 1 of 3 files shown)


docs/faq.html
```diff
@@ -51,6 +51,39 @@ <h2 class="title-2">
   </p>
 </section>
 
+<section class="mt4">
+  <h2 class="title-2"><a name="verify">How can I verify the download of Please?</a></h2>
```
tiagovtristao, line 55: nit: `<h2 id="verify" class="title-2">How can I verify the download of Please?</h2>` will make it consistent with similar usages in other places.

docs/quickstart.html
```diff
@@ -58,8 +58,12 @@ <h2 class="title-2">Installing</h2>
     <code class="code">~/.please</code> but you can put it where you want.
   </p>
 
+  <p>See <a href="faq.html#verify">here</a> for information if you'd like to verify
+    signatures of the download before running.</a>
```
tiagovtristao, line 62: `</a>` => `</p>`

tiagovtristao (approved):
Looks good, just a minor tweak. Thanks.

docs/faq.html
peterebden: Ah thanks, I never knew that worked that way!

docs/quickstart.html
peterebden: I do know the difference between `a` and `p`, honestly... :-P

PR #1788, opened 2021-05-01
Extract please_tools files as ourselves

peterebden:
Apparently when you're root tar behaves differently and extracts the files as the owner in the tarball, which we have set to nobody:nogroup, which can cause problems:
```
tar: please_tools/jarcat: Cannot change ownership to uid 65534, gid 65534: Invalid argument
```
This should fix. AFAICT this flag exists in both GNU and BSD tar.

### DIFF
--- a/src/parse/internal.tmpl
+++ b/src/parse/internal.tmpl
@@ -7,7 +7,7 @@ remote_file(
 genrule(
   name = "please_tools",
   srcs = [":download"],
-  cmd = "tar -xf $SRC",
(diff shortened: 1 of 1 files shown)


Tatskaari:
Yeah I encountered this in my sandboxing PR too.

Tatskaari (approved):

peterebden:
> Yeah I encountered this in my sandboxing PR too.

Ah I had a feeling I'd talked with you about it, but was confused by not finding it on master :)

PR #1789, opened 2021-05-04
Add --cmd flag to 'plz run' as an overriding mechanism for the default binary execution

tiagovtristao:
Add `--cmd` flag to `plz run` as an overriding mechanism for the default binary execution

This will be useful in cases where the binary needs to be executed in a different way, such as having it wrapped in a debugger program for instance.


Tatskaari (approved):
Looks good to me! Thanks Tiago

PR #1793, opened 2021-05-05
Replace sequences in plz run's --cmd

tiagovtristao:
* Replace sequences in `--cmd` and run the command in bash
* Add `$OUTS` and `$OUT` to the environment variable list in plz run


Tatskaari (approved):
I think we also want to run the command in bash so we can expand variables

tiagovtristao:
> I think we also want to run the command in bash so we can expand variables

Just put up those changes

src/run/run_step.go
```diff
@@ -150,46 +150,51 @@ func run(ctx context.Context, state *core.BuildState, label core.AnnotatedOutput
...
+			command = strings.Trim(command, "\"")
 		}
+		args = append(strings.Split(command, " "), args...)
 	}
-	arg0 := strings.Trim(command, "\"")
+
 	// Handle targets where $(exe ...) returns something nontrivial
-	splitCmd := strings.Split(arg0, " ")
-	if !strings.Contains(splitCmd[0], "/") {
+	if !strings.Contains(args[0], "/") {
```
Tatskaari, line 183: We could use this instead: https://golang.org/pkg/path/filepath/#IsAbs

src/run/run_step.go
```diff
@@ -150,46 +150,51 @@ func run(ctx context.Context, state *core.BuildState, label core.AnnotatedOutput
...
+		var command string
 		if tmpDir {
 			command = filepath.Join(dir, target.Outputs()[0])
 		} else {
 			command, _ = core.ReplaceSequences(state, target, fmt.Sprintf("$(out_exe %s)", target.Label))
+			command = strings.Trim(command, "\"")
 		}
+		args = append(strings.Split(command, " "), args...)
 	}
-	arg0 := strings.Trim(command, "\"")
```
Tatskaari, line 175: Not sure i understand what this line was for? I'm reasonably sure `$(out_exe )` can't produce anything with a leading or trailing ". Do you have any idea?

src/run/run_step.go
```diff
@@ -150,46 +150,51 @@ func run(ctx context.Context, state *core.BuildState, label core.AnnotatedOutput
 
 	// ReplaceSequences always quotes stuff in case it contains spaces or special characters,
 	// that works fine if we interpret it as a shell but not to pass it as an argument here.
-	command := ""
 	switch {
+	case overrideCmd != "":
+		command, _ := core.ReplaceSequences(state, target, overrideCmd)
+		// We don't care about passed in args when an override command is provided
+		args = process.BashCommand("bash", strings.Trim(command, "\""), true)
```
Tatskaari, line 157: nit: I know it's a bit silly and probably doesn't matter but we could handle args a bit better here:
```
splitCmd := process.BashCommand("bash", strings.Trim(command, "\""), true)
args = append(splitCmd, args...)
```

Tatskaari (approved):
Looks legit :) land when ready

src/run/run_step.go
tiagovtristao: The block wouldn't be executed for relative paths with `/`, would that be a problem?

src/run/run_step.go
tiagovtristao: I've checked and they seem to be ignored either way. For instance running `plz run --cmd="echo 1" //:target arg1 arg2` maps to `bash --noprofile --norc -e -u -o pipefail -c "echo 1" arg1 arg2` which only outputs `1` and ignores the args. Unless you were looking for something else?

Tatskaari: Ah I see. I suppose the args need to be applies to override command. It's fine, people should just be supplying those sort of things in --cmd.

src/run/run_step.go
Tatskaari: Ah nice thanks for looking into that :)

src/run/run_step.go
Tatskaari: Nah this is checking `args[0]` only which is the same as `filepath.IsAbs`

Tatskaari (approved):

src/core/build_env.go
```diff
@@ -232,6 +232,13 @@ func runtimeDataPaths(graph *BuildGraph, data []BuildInput) []string {
 func RunEnvironment(state *BuildState, target *BuildTarget, inTmpDir bool) BuildEnv {
 	env := TargetEnvironment(state, target)
 
+	outEnv := target.GetTmpOutputAll(target.Outputs())
```
peterebden, line 235: I'm not sure `GetTmpOutputAll` is the right thing here; we're going to be using the real outputs, not temp names?

tiagovtristao: Ya, I guess `GetTmpOutputAll` is meant for the build env, not the run env.

peterebden: Yep exactly, it handles some cases where your output has the same name as a top-level source directory.

In hindsight it would have been better to have outputs be in their package directory rather than at the top level, but it's way too hard to change now :(

peterebden (approved):

PR #1794, opened 2021-05-05
Generate a large tree of BUILD files to test parse performance

peterebden:
Generates 100k or so then runs `plz query alltargets` 5 times and chucks the results into S3.

Still haven't found any obvious thing out there that would do this for us so this will do as a low-tech solution for now. Next PR will render them into some kind of graph thing.


tiagovtristao (approved):
Nice, this looks a good first step :-)

PR #1801, opened 2021-05-07
Make --repo_root absolute

peterebden:
It doesn't seem to work if I do `plz --repo_root tree query print //assets/assets` (where "tree" is a directory containing a valid plz repo, from the perftest stuff).

Not 100% certain what's going on - it seems to work with e.g. `--repo_root ../please-servers`, but I can't think what would care about that kind of distinction. Nonetheless, I think it makes sense for `core.RepoRoot` to be absolute.


Tatskaari (approved):
Looks good to me

PR #1803, opened 2021-05-10
Specialise map implementation for targets

peterebden:
This is the most heavily trafficked map. Specialising it helps a bit, especially because we can specialise the hash function rather than it being done reflectively.
Seeing approx 10% performance improvement on parse-only workloads.

The code started out generated by an upstream tool (that comes with the map implementation we're using) but I had to hack that to get it to compile, then fiddle with the generated code to fix some things there, so eventually it's drifted quite far from that (so I don't think we should try to treat it as purely autogenerated, especially since I've taken the opportunity to simplify some of the interfaces)


src/core/cmap_targets.go
```diff
@@ -0,0 +1,108 @@
...
+// Higher shardCount will improve concurrency but will consume more memory.
+const shardCount = 1 << 8
+
+// shardMask is the mask we apply to hash functions below.
+const shardMask = shardCount - 1
+
+// targetMap is a concurrent safe sharded map to scale on multiple cores.
+// It's a fully specialised version of cmap.CMap for our most commonly used types.
+type targetMap struct {
+	shards []*targetLMap
```
Tatskaari, line 23: Is doing it this way significantly faster than the cmap + cast approach?

Tatskaari (approved):
Looks good to me. Have you got any numbers?

src/core/graph.go
```diff
@@ -203,19 +200,15 @@ func (graph *BuildGraph) SubrepoOrDie(name string) *Subrepo {
...
+	targets := graph.targets.Values()
 	sort.Sort(targets)
 	return targets
 }
 
 // PackageMap returns a copy of the graph's internal map of name to package.
 func (graph *BuildGraph) PackageMap() map[string]*Package {
 	packages := map[string]*Package{}
-	graph.packages.ForEach(func(k, v interface{}) bool {
+	graph.packages.ForEachLocked(func(k, v interface{}) bool {
```
Tatskaari, line 211: Why did this have to change?

peterebden:
Timings on this machine. Before:
```
$ plz run //tools/performance:parse_perf_test 
INFO: Run 1 of 5
INFO: Complete in 21.28s
INFO: Run 2 of 5
INFO: Complete in 24.48s
INFO: Run 3 of 5
INFO: Complete in 24.38s
INFO: Run 4 of 5
INFO: Complete in 22.37s
INFO: Run 5 of 5
INFO: Complete in 18.86s
INFO: Complete, median time: 22.37
```

After:
```
$ plz run //tools/performance:parse_perf_test -- --plz plz-out/bin/src/please
INFO: Run 1 of 5
INFO: Complete in 18.87s
INFO: Run 2 of 5
INFO: Complete in 18.29s
INFO: Run 3 of 5
INFO: Complete in 17.85s
INFO: Run 4 of 5
INFO: Complete in 18.91s
INFO: Run 5 of 5
INFO: Complete in 17.74s
INFO: Complete, median time: 18.29
```

src/core/graph.go
peterebden: It didn't have to, I changed it with the other one (then later changed that too).
I think I will put it back for now. Looks like it might be called concurrently (although I'm a bit nervous about the way this gets called now I look at it...).

tiagovtristao (approved):

PR #1804, opened 2021-05-10
Added jump to definition for build labels/files

Tatskaari:


tools/build_langserver/langserver_main.go
```diff
@@ -38,8 +38,10 @@ func main() {
 	if opts.LogFile != "" {
 		cli.InitFileLogging(string(opts.LogFile), opts.Verbosity, false)
 	}
-	if err := serve(lsp.NewHandler()); err != nil {
-		log.Fatalf("fail to start server: %s", err)
+	for {
+		if err := serve(lsp.NewHandler()); err != nil {
```
peterebden, line 42: I'm not sure I understand. If it returns no error, then keep trying forever? Is that really desirable?

peterebden: Oh it was only accepting a single connection? OK that was definitely wrong then.
TBH I don't think I ever used the TCP implementation really...

peterebden (approved):

PR #1809, opened 2021-05-11
Documented defaultrepo plz config value

RichardoC:
Also warned of expected duplication from plz init config.


Tatskaari (approved):
Thanks Richard!

PR #1814, opened 2021-05-17
Make go mod download rule create a go.mod file

Tatskaari:


rules/go_rules.build_defs
```diff
@@ -1034,7 +1034,7 @@ def go_mod_download(name:str, module:str, version:str, test_only:bool=False, vis
     out = name if not go_path_compatibility else f'src/{module}'
     # TODO(jpoole): write a proper please_go tool for this
     cmds = [
-        _set_go_env(),
+        'echo module example.com/module > go.mod', # stops go reading the main go.mod, and downloading all of those too
```
sagikazarmark, line 1037:
```suggestion
        'echo module please_ignore > go.mod', # stops go reading the main go.mod, and downloading all of those too
```

It doesn't have to be a full URL. This is more descriptive IMHO that it's not a test file, it's there on purpose.

rules/go_rules.build_defs
```diff
@@ -1034,7 +1034,7 @@ def go_mod_download(name:str, module:str, version:str, test_only:bool=False, vis
     out = name if not go_path_compatibility else f'src/{module}'
     # TODO(jpoole): write a proper please_go tool for this
     cmds = [
-        _set_go_env(),
```
sagikazarmark, line 1037: Should this really be removed? IMHO every go command should be set up the same way.

Tatskaari: No, but I had to to make this work. `_set_go_env()` will make the root a go path, and you're not allowed a go.mod in the root of your go path.

sagikazarmark: Slightly related, but I think we talked about reusing the machine/user global GOPATH as a go mod cache. There is an env var for that now that allows you to configure a separate mod cache path (defaults to gopath).

PR #1816, opened 2021-05-18
Promote 'outputs files into directory' warnings to errors

tiagovtristao:
Promote 'outputs files into directory' warnings to errors.

At the moment there's only one error message type, so I've simply replaced the log method. We might extend this in the future if verification of outputs consist of both warnings and errors.

### DIFF
--- a/src/core/package.go
+++ b/src/core/package.go
@@ -209,10 +209,10 @@ func (pkg *Package) Label() BuildLabel {
 // VerifyOutputs checks all files output from this package and verifies that they're all OK;
 // notably it checks that if targets that output into a subdirectory, that subdirectory isn't
 // created by another target. That kind of thing can lead to subtle and annoying bugs.
-// It logs detected warnings to stdout.
+// It logs detected errors to stderr.
 func (pkg *Package) VerifyOutputs() {
-	for _, warning := range pkg.verifyOutputs() {
-		log.Warning("%s: %s", pkg.Filename, warning)
+	for _, errMsg := range pkg.verifyOutputs() {
+		log.Error("%s: %s", pkg.Filename, errMsg)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Tatskaari (approved):
Looks good! Just need to appease the linter.

tiagovtristao:
> I think these should become Go errors, not just logged ones; in general we should either log a warning and attempt to continue or log an error and exit with an error.

Agreed, if we are going to mark it as an error to the user we should just stop there.

src/core/package.go
```diff
@@ -134,29 +134,50 @@ func (pkg *Package) HasOutput(output string) bool {
...
+		for f, t := range pkg.Outputs {
+			outputs[f] = t
+		}
+		outputs[fileName] = target
+		for filename, target := range outputs {
+			for dir := path.Dir(filename); dir != "."; dir = path.Dir(dir) {
+				if target2, present := outputs[dir]; present && target2 != target && !target.HasDependency(target2.Label.Parent()) {
+					return fmt.Errorf("Target %s outputs files into the directory %s, which is separately output by %s. You need to add a dependency to fix this", target.Label, dir, target2.Label)
+				}
+			}
```
tiagovtristao, line 169: Not really comfortable with all this logic here, but I couldn't find a better way to abstract it out given that we should lock the pkg when we are trying to access its data, and it would conflict with the existing lock in this function

.plzconfig
```diff
@@ -85,3 +85,4 @@ accept = notice
 
 [featureflags]
 JavaBinaryExecutableByDefault = true
+PackageOutputsStrictness = true
```
Tatskaari, line 88: We should enable this in core3 soonish

src/parse/parse_step.go
```diff
@@ -213,9 +213,11 @@ func parsePackage(state *core.BuildState, label, dependent core.BuildLabel, subr
...
 	}
 
-	// Verify some details of the output files in the background. Don't need to wait for this
-	// since it only issues warnings sometimes.
-	go pkg.VerifyOutputs()
+	if !state.Config.FeatureFlags.PackageOutputsStrictness {
+		// Verify some details of the output files in the background. Don't need to wait for this
+		// since it only issues warnings sometimes.
+		go pkg.VerifyOutputs()
+	}
```
Tatskaari, line 220: The only reason this was done here instead of on registering outputs was for performance wasn't it? Can we can just make this change now considering it's not a breaking change?

Tatskaari (commented):
Thanks Tiago! Looks good. Just wonder if we can make some of these changes now rather than waiting for a major release?

src/parse/parse_step.go
tiagovtristao: I think the reason was also that `pkg.VerifyOutputs` would only output warnings, so it wasn't critical to put the logic where the outputs get registered.

If we make this change now, builds will error out in cases where warnings were dismissed and not handled. Not sure if this is considered a breaking change or patch one.

PR #1820, opened 2021-05-20
Handle require/provide more accurately when queuing targets

peterebden:
We are slightly over-queuing targets for build; I think this logic existed before and got lost somewhere.


Tatskaari (approved):
Seems sensible but I'm a little surprised this worked as well as it did before.

peterebden:
> Seems sensible but I'm a little surprised this worked as well as it did before.

I think it's because we are still doing it properly in other places (e.g. preparing temp dirs), we're just over-building some targets, and most things don't fail because of that.

PR #1829, opened 2021-05-27
WIP: Fix lockup adding deps in post-build

Tatskaari:


src/core/build_target.go
```diff
@@ -491,8 +491,10 @@ func (target *BuildTarget) resolveDependencies(graph *BuildGraph, callback func(
 	return g.Wait()
 }
 
-func (target *BuildTarget) resolveOneDependency(graph *BuildGraph, dep *depInfo) error {
-	t := graph.WaitForTarget(dep.declared)
+func (target *BuildTarget) resolveOneDependency(state *BuildState, dep *depInfo) error {
+	// The target may have been added as part of a post-build and therefore
+	state.WaitForPackage(dep.declared, target.Label)
```
Tatskaari, line 496: This is the line that resolved the lockup. I've added in a `WaitForPackage()` which is normally not needed. Typically the package would've already been queued because we queue up all the declared deps on the target before resolving it's deps:

```go
	for _, dep := range target.DeclaredDependencies() {
		if err := state.queueTarget(dep, target.Label, rescan, forceBuild, false); err != nil {
			state.asyncError(dep, err)
			return
		}
	}
```

The problem is that when we add new declared deps to the rule in a post-build function, we don't queue these new ones up so their package may never be parsed and the next line just hangs forever. 

This helps however I've still had lockups. I think perhaps we need to properly queue the new declared deps.

PR #1831, opened 2021-05-28
Enqueue dependencies when added post-build

peterebden:
Aiming at the same problem as #1829 but in a different way.

### DIFF
--- a/src/parse/asp/builtins.go
+++ b/src/parse/asp/builtins.go
@@ -738,8 +738,11 @@ func addDep(s *scope, args []pyObject) pyObject {
 	dep := core.ParseBuildLabelContext(string(args[1].(pyString)), s.pkg)
 	exported := args[2].IsTruthy()
 	target.AddMaybeExportedDependency(dep, exported, false, false)
-	err := s.state.QueueTarget(dep, target.Label, true, false)
-	s.Assert(err == nil, "%s", err)
+	// Queue this dependency if it'll be needed.
+	if target.State() > core.Inactive {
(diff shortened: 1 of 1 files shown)


src/parse/asp/builtins.go
```diff
@@ -738,6 +738,12 @@ func addDep(s *scope, args []pyObject) pyObject {
 	dep := core.ParseBuildLabelContext(string(args[1].(pyString)), s.pkg)
 	exported := args[2].IsTruthy()
 	target.AddMaybeExportedDependency(dep, exported, false, false)
+	// Queue this dependency if it'll be needed.
+	if target.State() > core.Inactive {
```
Tatskaari, line 742: I guess we can only call this from a post-build function, but we might need to check that this target is to be built? 

I guess target A depends on B which has a post build. We build Target B which adds C as a dep of A. Won't we queue C to be built even though it only needs to build if A needs to be built? That's kinda why I approached this the way I did. 

I trust you've thought about this but would be good to understand this a bit better.

peterebden: Ah yeah - so that's what this line (741) is doing; it's checking that the target is in some kind of active state, i.e. we have tagged it to be built if we needed to.

I had a slightly earlier iteration of this that left that check out but eventually thought of the same thing, that we don't want to queue these targets unnecessarily.

Tatskaari (approved):

src/parse/asp/builtins.go
Tatskaari: Okay I'm satisfied then. Merge when ready. :)

PR #1837, opened 2021-06-02
Add zstd support to jarcat extract and fix file permissions

peterebden:
Having zstd support is useful as a second opinion for something else I'm working on. While I was there I realised the files always get extracted with 0664 permissions rather than whatever the tarball specifies.


Tatskaari (approved):

Tatskaari:
Still think this is the best zipping program on the planet. Perhaps we need to open source it into it's own repo.

tiagovtristao (approved):

peterebden:
Yeah I think that would be logical and fit with our other plans.

The command-line flags are a little bit ad-hoc though, sometimes it can be a bit unintuitive getting it to do what you want...

PR #1848, opened 2021-06-09
Reduce some allocations in parser

peterebden:
Attempts to size scopes a little more accurately, and parses some expressions in-place rather than creating a new struct and copying.

### DIFF
--- a/src/parse/asp/BUILD
+++ b/src/parse/asp/BUILD
@@ -3,6 +3,7 @@ go_library(
     srcs = glob(
         ["*.go"],
         exclude = [
+            "*_benchmark.go",
             "*_test.go",
             "*.bindata.go",
         ],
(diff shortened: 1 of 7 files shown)


src/parse/asp/grammar_parse.go
```diff
@@ -430,41 +428,47 @@ func (p *parser) parseUnconditionalExpressionInPlace(e *Expression) {
 	}
 }
 
-func concatStrings(lhs *ValueExpression, rhs *ValueExpression) *ValueExpression {
-	// If they're both fStrngs
+func concatStrings(lhs *ValueExpression, rhs *ValueExpression) {
```
Tatskaari, line 431: This is doing the same thing just isn't returning a value to be clearer that it's mutating the parameter?

Tatskaari (approved):
I think this looks sensible. Does it make a big difference?

peterebden:
> I think this looks sensible. Does it make a big difference?

Nope, it only makes a small difference. Just trying to collect a few of these for some relatively low-hanging fruit on reducing GC pressure.

peterebden:
OK so I've written a benchmark for this too:
```
BenchmarkParseFile
BenchmarkParseFile-12                260           4462050 ns/op         1632444 B/op      16476 allocs/op
```
vs.
```
BenchmarkParseFile
BenchmarkParseFile-12                274           4039847 ns/op         1571474 B/op      16336 allocs/op
```
(I have run both a few times and the numbers are pretty consistent)
So it's saving about 1% of allocations which isn't massive but easy enough; I've also dropped the parser changes so it's just scope sizing now.

PR #1855, opened 2021-06-14
Fix root package not being picked up in 'plz query changes'

tiagovtristao:
Fix root package not being picked up in 'plz query changes'


peterebden (approved):

tiagovtristao:
Fix root package not being picked up in 'plz query changes'

PR #1857, opened 2021-06-15
Fix target own inputs check

tiagovtristao:
This fixes the check on targets' own inputs where named sources and data were not being considered. This caused named input sources belonging to a different package to be used as sources without an error happening.


peterebden (approved):

Tatskaari (approved):
Nice! Thanks for that Tiago. I think this was one of the first things I added to Please. Nostalgic.

PR #1861, opened 2021-06-16
Make package filenames relative

peterebden:
The original motivation here was fixing asp log messages like
```
09:31:28.240 WARNING: ///home/pebers/git/please/test/go_rules/BUILD: hello
```
to
```
09:31:28.240 WARNING: //test/go_rules/BUILD: hello
```
I think it should be fine generally for these to be relative though? The location of the repo shouldn't need to show up in this stuff.

### DIFF
--- a/src/parse/parse_step.go
+++ b/src/parse/parse_step.go
@@ -235,7 +235,7 @@ func buildFileName(state *core.BuildState, pkgName string, subrepo *core.Subrepo
 		return "WORKSPACE", ""
 	}
 	for _, buildFileName := range config.Parse.BuildFileName {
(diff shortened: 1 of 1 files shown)


Tatskaari:
I think it should either be the file path:
`/home/pebers/git/please/test/go_rules/BUILD: hello`

Or the package label: 
`//please/test/go_rules: hello` (no build file)

Feels like a hybrid of the two right now

Tatskaari (approved):
This is probably a good change. I think i'd prefer for asp to log the build file as then IDEs might allow you to ctr-click that to jump to the file.

Issue #1833, opened 2021-05-30
plz query missing documentation


RichardoC:
https://please.build/commands.html doesn't list the following valid options

- filter
- roots
- whatoutputs

Can documentation on these be added to please/docs/commands.html  ?

Tatskaari:
I've decided not to document roots. It doesn't seem to work as I would expect and nobody has noticed so I'm thinking I'd like to remove it come v17.

PR #1864, opened 2021-06-16
Document missing plz query commands

Tatskaari:
Fixes: #1833


Tatskaari:
There's a huge amount of subcommands missing from the documentation that I don't want to fix right now. We might want to consider generating this page instead.

PR #1867, opened 2021-06-17
Fix duplicates in remote execution directories

peterebden:
As discussed we sometimes produce duplicates (typically because you have two filegroups outputting the same file which are both depended on by something else).
This is invalid according to the API ("Each child in the directory must have a unique path segment (file name)") so we shouldn't do it.

### DIFF
--- a/src/remote/utils.go
+++ b/src/remote/utils.go
@@ -438,9 +438,37 @@ func (b *dirBuilder) walk(name string, ch chan<- *uploadinfo.Entry) *pb.Digest {
 	}
 	// The protocol requires that these are sorted into lexicographic order. Not all servers
 	// necessarily care, but some do, and we should be compliant.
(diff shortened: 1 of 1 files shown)


peterebden:
This would be much nicer with generics :(

src/remote/utils.go
```diff
@@ -438,9 +438,37 @@ func (b *dirBuilder) walk(name string, ch chan<- *uploadinfo.Entry) *pb.Digest {
...
+			last = d.Name
+		}
+	}
+	dir.Symlinks = syms[:0]
+	for _, s := range syms {
+		if s.Name != last {
+			dir.Symlinks = append(dir.Symlinks, s)
+			last = s.Name
+		}
+	}
```
tiagovtristao, line 470: I think resetting `last` before each section will avoid the edge case of that value preventing the append in the following section with the same value.

peterebden: Ah thanks, good spot!

PR #1875, opened 2021-06-21
IterInputs() bench

Tatskaari:
Before we used to do many allocations:

- 1 for each source via `recursivelyProvideSource()`
- A few for named targets in order to sort and aggregate 

Now we don't do any more allocations that necessary on the hot path. 

Before:
```
BenchmarkIterInputsControl
BenchmarkIterInputsControl-12         	  407613	      2563 ns/op	     640 B/op	       6 allocs/op
BenchmarkIterInputsSimple
BenchmarkIterInputsSimple-12          	  157765	      7493 ns/op	    1936 B/op	      32 allocs/op
BenchmarkIterInputsNamedSources
BenchmarkIterInputsNamedSources-12    	  130479	      9777 ns/op	    4544 B/op	      44 allocs/op
```

After:
```
BenchmarkIterInputsControl
BenchmarkIterInputsControl-12         	  525732	      2388 ns/op	     640 B/op	       6 allocs/op
BenchmarkIterInputsSimple
BenchmarkIterInputsSimple-12          	  169854	      6833 ns/op	    1536 B/op	       7 allocs/op
BenchmarkIterInputsNamedSources
BenchmarkIterInputsNamedSources-12    	  140935	      9396 ns/op	    4144 B/op	      19 allocs/op
```


peterebden (approved):
OK, although this will increase the size of BuildTarget somewhat so there is a bit of a tradeoff.

Maybe we can come back and revisit later (e.g. we've discussed having a more efficient scheme for encoding these...)

Tatskaari:
> OK, although this will increase the size of BuildTarget somewhat so there is a bit of a tradeoff.
> 
> Maybe we can come back and revisit later (e.g. we've discussed having a more efficient scheme for encoding these...)

I didn't realise how costly slices were. Most of this optimisation actually comes from the require sources optimisation. Might have a rethink about the others.

peterebden (approved):

PR #1877, opened 2021-06-22
Remove asp compiler gob stuff

peterebden:
We no longer actually use any of this since we moved to Go's native file embedding.


src/parse/asp/interpreter.go
```diff
@@ -66,6 +66,12 @@ func (i *interpreter) LoadBuiltins(filename string, contents []byte, statements
 		return err
 	} else if len(contents) != 0 {
 		stmts, err := i.parser.ParseData(contents, filename)
+		for _, stmt := range stmts {
```
peterebden, line 69: This got lost in the embedding switchover - moving the code to here so we only do this to the actual builtins (to mimic previous behaviour)

Tatskaari (approved):
Sorry I thought I approved this

PR #1878, opened 2021-06-22
Add benchmarks to upload to s3 to be added to the report later on

Tatskaari:
Adds a build definition that links benchmark results in json from to `plz-out/benchmarks`, which is then uploaded to s3. Also moves the results over to the main website bucket so we can avoid XSS issues. 

Once we have these in place, we can start presenting them in the performance report.


build_defs/benchmark.build_defs
```diff
@@ -0,0 +1,28 @@
+def benchmark(name:str, srcs:list, deps:list=[], data:list=[], visibility:list=None, labels=[]):
+    benchmark_rule = go_benchmark(
+        name = f"{name}_benchmark",
+        srcs = srcs,
+        deps = deps,
+        data = data,
+        labels=labels,
```
peterebden, line 7: nit: spacing around = (and below)

peterebden (approved):

PR #1881, opened 2021-06-23
Don't create the stamped environment when asked not to

peterebden:
Found out that we basically always do it for a stamped target in remote execution, but we should honour the request not to.

Will improve cache lookups & no-op execution speeds.


tiagovtristao (approved):

Tatskaari (approved):
LGTM

PR #1883, opened 2021-06-23
Streamline & simplify grammar a bit

peterebden:
Now we've got rid of the gobbing, we don't need the layer of struct indirection (although in a slightly ugly fashion we need a sentinel to check if we really have an int). Also mark the boolean singletons explicitly rather than using their string representation.

### DIFF
--- a/src/parse/asp/grammar.go
+++ b/src/parse/asp/grammar.go
@@ -138,7 +138,6 @@ type ValueExpression struct {
 	// True if the Int field is set; this helps us distinguish values of 0.
 	IsInt    bool
 	Int      int
-	Bool     string
 	List     *List
 	Dict     *Dict
 	Tuple    *List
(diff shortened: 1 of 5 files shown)


src/parse/asp/grammar.go
```diff
@@ -131,9 +131,13 @@ type OpExpression struct {
...
-		Int int
-	} // Should just be *int, but https://github.com/golang/go/issues/23498 :(
+	// These are true if this represents one of the boolean singletons
+	True  bool
+	False bool
+	None  bool
+	// True if the Int field is set; this helps us distinguish values of 0.
+	IsInt    bool
+	Int      int
 	Bool     string
```
Tatskaari, line 141: Do we still need this?

peterebden: need what?

tiagovtristao (approved):

PR #1886, opened 2021-06-25
Remove map rewrite from allBuildInputs()

Tatskaari:


Tatskaari:
Going to do this atomically in #1888

PR #1893, opened 2021-06-28
`github_repo`: implement semver check using `is_semver`

chrisnovakovic:
Use the new `is_semver` builtin to evaluate whether the revision parameter in a `github_repo` target is a semantic version.


Tatskaari (approved):
Nice! Thanks Chris. :)

PR #1895, opened 2021-06-28
`c_binary`: allow `defines` to be a dict

chrisnovakovic:
The documentation for `c_binary` says this is already allowed (and the `cc_binary` rule upon which `c_binary` is based also allows it), but it produces a type error at run-time because of `c_binary`'s type signature.


peterebden (approved):
Aaargh I did this for `cc_binary` in 220e77d92 but missed this one. Thanks!

PR #1899, opened 2021-06-29
Rework prom metrics a bit

peterebden:
This performs a single push at the completion (successful or otherwise) of a build.
It also splits out metric definitions to individual packages in anticipation of us adding more in future.

Metrics have `Namespace: "plz"`, and `Subsystem` whatever we consider relevant (the package name is probably sufficient here) and follow the [naming conventions](https://prometheus.io/docs/practices/naming/). All metrics have a global `version` label appended.

Much of the noise here is me changing `core.PleaseVersion` into a const string and only parsing to a semver on request; `update` is the only thing that needs it to be a semver object. I've also updated the Prometheus library version.


Tatskaari (approved):

src/core/BUILD
```diff
@@ -7,7 +7,7 @@ genrule(
         "version": ["//:version"],
     },
     outs = ["versioned.go"],
-    cmd = "sed \"s/1.0.9999/`cat $SRCS_VERSION`/\" $SRCS_GO > \"$OUT\"",
+    cmd = "sed \"s/1.0.9999/`cat $SRCS_VERSION`/g\" $SRCS_GO > \"$OUT\"",
```
Tatskaari, line 10: Can't we use go definitions to just set this value?

peterebden: Ah yeah good idea. Have done.

Also thought about using embedding which would mean it was accurate even for `go get` but this is a little easier...

PR #1906, opened 2021-07-01
Lock mutex while resolving deps

Tatskaari:


src/core/build_target.go
```diff
@@ -464,10 +464,12 @@ func (target *BuildTarget) AllURLs(state *BuildState) []string {
 // resolveDependencies matches up all declared dependencies to the actual build targets.
 // TODO(peterebden,tatskaari): Work out if we really want to have this and how the suite of *Dependencies functions
 //                             below should behave (preferably nicely).
-// TODO(tatskaari): Work out if we can use a channel instead of a callback.
```
Tatskaari, line 467: Did manage to do this but it wasn't much better because we have to pass the error group back to the caller. Don't think it's really worth changing now.

tiagovtristao (approved):

src/core/build_target.go
```diff
@@ -464,10 +464,12 @@ func (target *BuildTarget) AllURLs(state *BuildState) []string {
...
 func (target *BuildTarget) resolveDependencies(graph *BuildGraph, callback func(*BuildTarget) error) error {
 	var g errgroup.Group
-	target.mutex.RLock()
+
+	target.mutex.Lock()
+	defer target.mutex.Unlock()
```
peterebden, line 471: I'm dubious about holding this while the waitgroup runs. Could this potentially deadlock or just slow things down a lot?

PR #1909, opened 2021-07-02
don't generate debug symbols in pip install 

namtzigla:
Without this the packages like: 

```
pip_library(
    name = "wrapt",
    version = "1.12.1",
)
```

will generate a different hash after each fresh build (`plz clean`).


peterebden:
Thanks, makes sense - I've never really expected `pip_library` to be very deterministic but this is clearly better!

Issue #1891, opened 2021-06-25
Remove package prefix in `pex` file


bmustiata:
I'm trying to create a python build that needs a library build in the same project.

The library build in the same project resides in `//tools/termcolor-util:termcolor_util`
This library in turn uses `termcolor`. As per recommended best-practices `termcolor` lives in `//thirdparty/python:termcolor` as a `pip_library`.

The BUILD definition is:
```
python_library(
    name="termcolor_util",
    srcs=glob(["**/termcolor_util/__init__.py"]),
    deps=["//thirdparty/python:termcolor"],
)

python_binary(
    name="termcolor_util_main",
    deps=[":termcolor_util"],
    main="termcolor_util/__main__.py",
)
```

However, when I try to run it, I get an error, namely that the library is not found:

```
$ plz run //tools/termcolor-util:termcolor_util_main
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "plz-out/bin/tools/termcolor-util/termcolor_util_main.pex/__main__.py", line 385, in <module>
  File "plz-out/bin/tools/termcolor-util/termcolor_util_main.pex/__main__.py", line 377, in run
  File "plz-out/bin/tools/termcolor-util/termcolor_util_main.pex/__main__.py", line 351, in interact
  File "plz-out/bin/tools/termcolor-util/termcolor_util_main.pex/__main__.py", line 360, in main
  File "/usr/lib/python3.8/runpy.py", line 210, in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/raptor/projects/germanium-monorepo/plz-out/bin/tools/termcolor-util/termcolor_util_main.pex/tools/termcolor-util/termcolor_util/__main__.py", line 1, in <module>
ModuleNotFoundError: No module named 'termcolor_util'
```

Looking at the `pez` file I can see why:

```
unzip -l /home/raptor/projects/germanium-monorepo/plz-out/bin/tools/termcolor-util/termcolor_util_main.pex
Archive:  /home/raptor/projects/germanium-monorepo/plz-out/bin/tools/termcolor-util/termcolor_util_main.pex
  Length      Date    Time    Name
[...]

PR #1912, opened 2021-07-02
Added strip_package_name for python rules.

bmustiata:
This addresses #1891. It's not yet complete, still TBD:

- [x] general implementation
- [x] documentation
- [x] ensure the package structure doesn't leak into the pex file
- [x] extra tests?
- [x] fix linting + existing test issues

I hope it's a good starting point for discussion.


Tatskaari:
Hey! I commented on your issue trying to understand the motivation behind this a bit better.

bmustiata:
Closing since I'm not going to use it, and it's not done.

rules/python_rules.build_defs
```diff
@@ -16,7 +16,7 @@ a target which has only had small changes.
 
 def python_library(name:str, srcs:list=[], resources:list=[], deps:list=[], visibility:list=None,
                    test_only:bool&testonly=False, zip_safe:bool=True, labels:list&features&tags=[], interpreter:str=None,
-                   strip:bool=False):
+                   strip:bool=False, strip_package_name:bool=CONFIG.PYTHON_STRIP_PLEASE_PACKAGE_NAMES):
```
Tatskaari, line 19: Are there other use cases for this, or is it specific to trying to produce a python package you can publish to a repo e.g. pypi? 

It might be more descriptive to call it something like `PYTHON_PACKAGE_ROOT` if this is the roots of the python packages we're installing. The fact we're implementing this by stripping the paths in the pex tool is really an implementation detail IMO.

Tatskaari (commented):
I think this makes sense. @pebers what do you think? You're more of a python developer than I am.

rules/python_rules.build_defs
bmustiata: I think something like `PYTHON_USE_PACKAGE_ROOT` would be better since it's a boolean. `PACKAGE_ROOT` sounds like an option to just package the roots.

bmustiata:
I'm done with the changes, and all the checks pass. It would be great if someone could review and/or merge.

bmustiata:
@Tatskaari , @pebers ?

Tatskaari (approved):
Looks good to me! Appologies for the slow review.

rules/python_rules.build_defs
```diff
@@ -447,7 +476,8 @@ def pip_library(name:str, version:str, hashes:list=None, package_name:str=None,
 def python_wheel(name:str, version:str, hashes:list=None, package_name:str=None, outs:list=None,
                  post_install_commands:list=None, patch:str|list=None, licences:list=None,
                  test_only:bool&testonly=False, repo:str=None, zip_safe:bool=True, visibility:list=None,
-                 deps:list=[], name_scheme:str=None, strip:list=['*.pyc', 'tests']):
+                 deps:list=[], name_scheme:str=None, strip:list=['*.pyc', 'tests'],
+                 use_package_root:bool=CONFIG.PYTHON_USE_PACKAGE_ROOT):
```
Tatskaari, line 480: I'm not sure why we need this here? Third party code can already be imported via the root path.

rules/python_rules.build_defs
```diff
@@ -531,12 +562,23 @@ def python_wheel(name:str, version:str, hashes:list=None, package_name:str=None,
     if post_install_commands:
         cmd += post_install_commands
 
-    cmd += ['$TOOL z -d --prefix $PKG -i ' + ' -i '.join(outs or [name])]
+    if use_package_root:
+        jarcat_command = '$TOOL z -d -i ' + ' -i '.join(outs or [name])
+    else:
+        jarcat_command = '$TOOL z -d --prefix $PKG -i ' + ' -i '.join(outs or [name])
+
+    cmd += [jarcat_command]
```
bmustiata, line 570: "I'm not sure why we need this here? Third party code can already be imported via the root path."

Because implicitly the content of the wheel gets added into the zip, being prefixed by the package name where is defined. So in the zip you won't have the folder `flask`, but rather `third_party/python/flask` (note the `--prefix $PKG`). 

The import via the root path is only possible if the `modulesdir` is configured to that package name in the please config file. Since the modulesdir is defined, the custom module loader in the pex also searches that location, so the imports will look like they're done from the rootpath. In the pex file, they will still reside in the folder matching the package name, not in the root of the archive.

It seems more reasonable to have a single switch that configures everything to be in the root, and use regular imports, since it's also closer to how python operates - namely just unpacking all the packages into its lib folder, and read from there.

rules/python_rules.build_defs
bmustiata: I responded where it's used.

PR #1913, opened 2021-07-03
Add some metrics

peterebden:
This covers the `remote` and `build` packages fairly well. Think `test` and `cache` would still be useful in future.

Open to discussion of whether there are other things we want to measure, or if some of these are unnecessary, etc.

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/thought-machine/please/src/core"
 	"github.com/thought-machine/please/src/fs"
+	"github.com/thought-machine/please/src/metrics"
 	"github.com/thought-machine/please/src/process"
 	"github.com/thought-machine/please/src/worker"
 )
@@ -39,6 +40,42 @@ var httpClient http.Client
 var httpClientOnce sync.Once
 
 var magicSourcesWorkerKey = "WORKER"
+
+var targetsBuilt = metrics.NewCounter(
+	"build",
+	"targets_built_total",
+	"Total number of non-filegroup targets built",
+)
+
+var filegroupsBuilt = metrics.NewCounter(
+	"build",
+	"filegroups_built_total",
+	"Total number of filegroups built",
+)
+
+var targetsUnbuilt = metrics.NewCounter(
+	"build",
+	"targets_unbuilt_total",
+	"Total number of targets that we didn't need to rebuild",
+)
+
+var prepareSourceDurations = metrics.NewHistogram(
+	"build",
+	"prepare_source_duration_seconds",
+	"Time to prepare sources for each target",
+)
+
+var buildDurations = metrics.NewHistogram(
+	"build",
+	"build_duration_seconds",
+	"Time to execute the build action for each target",
+)
+
+var collectOutputDurations = metrics.NewHistogram(
+	"build",
+	"collect_output_duration_seconds",
+	"Time to collect the outputs for each target",
+)
 
 // Build implements the core logic for building a single target.
 func Build(tid int, state *core.BuildState, label core.BuildLabel, remote bool) {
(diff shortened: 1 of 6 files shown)


src/build/build_step.go
```diff
@@ -40,6 +41,42 @@ var httpClientOnce sync.Once
...
+	"build",
+	"filegroups_built_total",
+	"Total number of filegroups built",
+)
+
+var targetsUnbuilt = metrics.NewCounter(
+	"build",
+	"targets_unbuilt_total",
+	"Total number of targets that we didn't need to rebuild",
+)
```
Tatskaari, line 60: How about `targetsUnchanged` or `targetsReused` to borrow from the target/logging lingo? 

Also these are only the targets that were already in `plz-out`. Perhaps we also want a metric for those that were retrieved from the cache (L301). Seems like these are included in the built targets at the moment which doesn't seem that useful. Is the idea that you can infer this information by subtracting the cache hits?

src/build/build_step.go
```diff
@@ -240,8 +278,10 @@ func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget, runR
 				state.LogBuildResult(tid, target.Label, core.TargetCached, "Unchanged")
 			}
 			buildLinks(state, target)
+			filegroupsBuilt.Inc()
 			return nil
 		}
+		defer targetsBuilt.Inc()
```
Tatskaari, line 284: Below we do `retrieveArtifacts(...)` to check the cache but we'll still increment this counter. I'm not sure that's what we want?

src/test/test_step.go
```diff
@@ -34,6 +35,30 @@ var numUploadFailures int64
...
+	"test",
+	"targets_tested_total",
+	"Total number of targets tested",
+)
+
+var targetsCached = metrics.NewCounter(
+	"test",
+	"targets_cached_total",
+	"Total number of targets not re-tested due to caching",
+)
```
Tatskaari, line 48: I think this includes both those cached in `plz-out` and in the dir/http caches. Seems difficult to report those differently in the way this code is set out. I think however we report this, we should probably align that with how we report for building.

Tatskaari (commented):
Nice one! Thanks for that. Just a couple comments.

src/build/build_step.go
peterebden: There are cache metrics for how many got retrieved.

Maybe it's better to have targetsBuilt just be the total and then we can subtract those others from it?

Tatskaari: > There are cache metrics for how many got retrieved.
> 
> Maybe it's better to have targetsBuilt just be the total and then we can subtract those others from it?

This is incremented only when we retrieved from `plz-out` though? We retrieve from the cache closer to line 300, at which point we've already hit `defer targetsBuilt.Inc()`:

```golang

		if state.Cache != nil && !runRemotely && !state.ShouldRebuild(target) {
			// Note that ordering here is quite sensitive since the post-build function can modify
			// what we would retrieve from the cache.
			if target.BuildCouldModifyTarget() {
				log.Debug("Checking for build metadata for %s in cache...", target.Label)
				if metadata = retrieveFromCache(state.Cache, target, cacheKey, nil); metadata != nil {
					addOutDirOutsFromMetadata(target, metadata)
					if target.PostBuildFunction != nil && !haveRunPostBuildFunction {
						postBuildOutput = string(metadata.Stdout)
						if err := runPostBuildFunction(tid, state, target, postBuildOutput, ""); err != nil {
							return err
						}
					}
					// Now that we've updated the rule, retrieve the artifacts with the new output hash
					if retrieveArtifacts(tid, state, target, oldOutputHash) {
						return writeRuleHash(state, target)
					}
				}
			} else if retrieveArtifacts(tid, state, target, oldOutputHash) {
				return nil
			}
		}
```

PR #1917, opened 2021-07-06
plz exec

tiagovtristao:
This PR introduces a new `plz exec` command that allows to execute commands/binaries in a sandboxed environment. Mount and network are sandboxed by default, but you can share them (`--share_network`, `--share_mount`) if you want to.
Below you will find some examples of what you can do with it.

#### `plz exec //tools/jarcat -- curl google.com`

`curl: (6) Could not resolve host: google.com`

#### `plz exec --share_network //tools/jarcat -- curl google.com`

```
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="http://www.google.com/">here</A>.
</BODY></HTML>
```
#### `plz exec //tools/jarcat` or `plz exec //tools/jarcat -- \$OUT`

```
Jarcat is a binary shipped with Please that helps it operate on .jar and .zip files.

Its original and most useful feature is performing efficient concatenation of .jar files...
```


src/exec/exec.go
```diff
@@ -0,0 +1,78 @@
...
+	"gopkg.in/op/go-logging.v1"
+
+	"github.com/thought-machine/please/src/core"
+	"github.com/thought-machine/please/src/fs"
+	"github.com/thought-machine/please/src/process"
+)
+
+var log = logging.MustGetLogger("exec")
+
+func Exec(state *core.BuildState, label core.BuildLabel, cmdArgs []string, shareNetwork bool) {
```
peterebden, line 21: Comment plz? Not sure why the linter is not moaning at you about it...

src/exec/exec.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func exec(ctx context.Context, state *core.BuildState, label core.BuildLabel, cmdArgs []string, shareNetwork bool) error {
+	target := state.Graph.TargetOrDie(label)
+
+	if !target.IsBinary && len(cmdArgs) == 0 {
+		log.Fatalf("Target %s cannot be executed; it's not marked as binary", target.Label)
+	}
+
+	var cmd string
```
peterebden, line 32: I'm always a bit dubious about this kind of `var` above an if pattern. Generally would suggest pulling it out to another function so you can use `:=` (ideally ending up with more of a single-assignment form)

src/please.go
```diff
@@ -194,6 +195,14 @@ var opts struct {
 		Remote bool `long:"remote" description:"Send targets to be executed remotely."`
 	} `command:"run" subcommands-optional:"true" description:"Builds and runs a single target"`
 
+	Exec struct {
+		ShareNetwork bool `long:"share_network" description:"Share network namespace"`
```
peterebden, line 199: I don't have a lot of good ideas for it, but I wonder if we want this to be more flexible? For example, you could apply the same logic to user or pid namespaces.

tools/sandbox/main.c
```diff
@@ -16,5 +18,14 @@ int main(int argc, char* argv[]) {
         fputs("Usage: please_sandbox command args...\n", stderr);
         return 1;
     }
-    return contain(&argv[1], true, true);
+
+    // Network namespace is sandboxed by default unless the `SANDBOX_NETWORK=0` env exists
```
peterebden, line 22: I wonder if this would be clearer as `NOSANDBOX_NETWORK`? Setting it to 0 and having that be different from unset I think is a little unintuitive.

peterebden (commented):
Thanks, very cool!

src/exec/exec.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/thought-machine/please/src/process"
+)
+
+var log = logging.MustGetLogger("exec")
+
+func Exec(state *core.BuildState, label core.BuildLabel, cmdArgs []string, shareNetwork bool) {
+	exec(context.Background(), state, label, cmdArgs, shareNetwork)
+}
+
+func exec(ctx context.Context, state *core.BuildState, label core.BuildLabel, cmdArgs []string, shareNetwork bool) error {
```
Tatskaari, line 25: I suppose we might want a `share_mount` flag too at some point

src/exec/exec.go
```diff
@@ -0,0 +1,78 @@
...
+		var err error
+		if cmd, err = core.ReplaceSequences(state, target, strings.Join(cmdArgs, " ")); err != nil {
+			return err
+		}
+	} else {
+		outs := target.Outputs()
+		if len(outs) != 1 {
+			log.Fatalf("Target %s cannot be executed as it has %d outputs", target.Label, len(outs))
+		}
+		cmd = outs[0]
```
Tatskaari, line 43: I think `plz run` does something slightly different where it uses `$(out_exe ...)`. Not sure if there's a good reason for it though.

src/exec/exec.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	dir := filepath.Join(core.OutDir, "exec", target.Label.Subrepo, target.Label.PackageName)
```
Tatskaari, line 46: Do we ever clean up this dir?

src/please.go
Tatskaari: Hmm yeah, I think the idea was be to add `--share_mount` as well. Not sure if there's a better approach

src/process/exec_linux.go
```diff
@@ -31,9 +31,11 @@ func (e *Executor) ExecCommand(sandbox bool, command string, args ...string) *ex
 			// Otherwise exec the sandbox tool
 			args = append([]string{command}, args...)
 			command = e.sandboxTool
+			env = []string{"SANDBOX_NETWORK=" + boolToString(sandbox.Network), "SANDBOX_MOUNT=" + boolToString(sandbox.Mount)}
 		}
 	}
 	cmd := exec.Command(command, args...)
```
Tatskaari, line 37: We could probably just move this above the if statement and get rid of the intermediate `env` variable

src/process/exec_linux.go
```diff
@@ -43,9 +45,15 @@ func (e *Executor) ExecCommand(sandbox bool, command string, args ...string) *ex
...
+		if sandbox != NoSandbox {
+			if sandbox.Network {
+				// Namespace network
+				cmd.SysProcAttr.Cloneflags |= syscall.CLONE_NEWNET
+			}
+			if sandbox.Mount {
+				// Namespace mount
+				cmd.SysProcAttr.Cloneflags |= syscall.CLONE_NEWNS
+			}
 		}
```
Tatskaari, line 57: Do you need this outer if statement ?

tools/sandbox/main.c
Tatskaari: We could have `SHARE_NETWORK` and default to unsharing it if it's unset

Tatskaari (approved):

src/exec/exec.go
tiagovtristao: I wondered the same a while ago... will add.

src/exec/exec.go
tiagovtristao: I guess it can get messy. I like the suggestion.

src/please.go
tiagovtristao: Does the cli tool allow some sort of `--share=mount --share=network`?

tools/sandbox/main.c
tiagovtristao: `SHARE_NETWORK` is what I'm looking for! This was one of the moments where naming things was hard...

src/exec/exec.go
tiagovtristao: Yeah, I need to come back to this

src/exec/exec.go
tiagovtristao: It will be done during preparation. Should it happen other stages?

src/process/exec_linux.go
tiagovtristao: Ya, the PR needs tidying up. Thanks

src/exec/exec.go
```diff
@@ -0,0 +1,94 @@
...
+
+	outs := target.Outputs()
+	if len(outs) != 1 {
+		return "", fmt.Errorf("Target %s cannot be executed as it has %d outputs", target.Label, len(outs))
+	}
+
+	if !sandbox.Mount {
+		return filepath.Join(core.RepoRoot, runtimeDir, outs[0]), nil
+	}
+	return filepath.Join(core.SandboxDir, outs[0]), nil
```
Tatskaari, line 68: Does `./$OUT` not work for us?

src/sandbox/sandbox_linux.go
```diff
@@ -20,6 +21,8 @@ const mdLazytime = 1 << 25
 
 const sandboxDirsVar = "SANDBOX_DIRS"
 
+var sandboxMountDir = core.SandboxDir
```
Tatskaari, line 24: Not sure we need this?

Tatskaari (approved):
Looks good to me! Nice one Tiago!

src/exec/exec.go
Tatskaari: As long as they're not left in `plz-out` indefinitely

src/exec/exec.go
tiagovtristao: It does. I wasn't sure if it was the idiomatic way of doing it - `"./" + outs[0]` - and I couldn't also find any similar instance in the codebase.

src/sandbox/sandbox_linux.go
tiagovtristao: Not really, I guess I was trying to make this not reliant on `core.Sandbox` if the variable name would ever change.

peterebden (approved):

PR #1918, opened 2021-07-06
Maybe this makes things better.

### LINKED ISSUE #1915: Please will hang on updating hashes
Seems that since 16.1.0, `plz hash --update` will hang.

peterebden:
Possibly fixes #1915 ?


Tatskaari (approved):
Just made that change in my PR ;P 

I'm just changing it to keep building in the case of a hash failure. I don't think we want to stop as there might be other hashes to update.

peterebden:
:slowpoke:

PR #1919, opened 2021-07-06
Keep going when we update a hash as there may be more

### LINKED ISSUE #1915: Please will hang on updating hashes

Tatskaari:
Fixes #1915


peterebden (approved):
plus one bazillion

Can you completely remove errStop now?

Tatskaari:
> plus one bazillion
> 
> Can you completely remove errStop now?

errStop is used by `--shell` unfortunately

PR #1921, opened 2021-07-06
Lock mutex while resolving deps (again) 

Tatskaari:
Accidentally landed the last one incorrectly


src/core/build_target.go
```diff
@@ -464,12 +464,11 @@ func (target *BuildTarget) AllURLs(state *BuildState) []string {
 // resolveDependencies matches up all declared dependencies to the actual build targets.
 // TODO(peterebden,tatskaari): Work out if we really want to have this and how the suite of *Dependencies functions
 //                             below should behave (preferably nicely).
-// TODO(tatskaari): Work out if we can use a channel instead of a callback.
 func (target *BuildTarget) resolveDependencies(graph *BuildGraph, callback func(*BuildTarget) error) error {
 	var g errgroup.Group
-	target.mutex.RLock()
```
peterebden, line 470: Think we should keep this in as a Lock / Unlock (since we're iterating over the slice itself which is shared, but the work on it happens within the waitgroup)

peterebden (approved):

PR #1925, opened 2021-07-09
Implement plugins

Tatskaari:
See `//test:plugins` for a simple example. There's also `//test/python_rules/plugin` which works with https://github.com/please-build/python-rules

Probably want something in the way of unit testing but I think it's about ready for review.


src/core/state.go
```diff
@@ -988,6 +990,12 @@ func (state *BuildState) ForConfig(config ...string) *BuildState {
 	return s
 }
 
+func (state *BuildState) ForSubrepo(name string) *BuildState {
+	s := state.ForConfig()
```
peterebden, line 994: is ForConfig still relevant or is it obsoleted by ForSubrepo?

src/core/state.go
```diff
@@ -988,6 +990,12 @@ func (state *BuildState) ForConfig(config ...string) *BuildState {
...
+	s.CurrentSubrepo = name
```
peterebden, line 995: I wonder if we should do some check on whether `CurrentSubrepo == name` and return the existing one if so?

docs/config.html
```diff
@@ -348,6 +357,88 @@ <h3 class="mt1 f6 lh-title" id="parse.gitfunctions">GitFunctions</h3>
   </ul>
 </section>
 
+<section class="mt4">
+  <h2 id="plugindefinition" class="title-2">[PluginDefinition]</h2>
```
peterebden, line 361: maybe let's mark these sections as experimental so we can break them if we need to?

docs/config.html
```diff
@@ -348,6 +357,88 @@ <h3 class="mt1 f6 lh-title" id="parse.gitfunctions">GitFunctions</h3>
...
+
+  <p>The following sets up a plugin with ID foo.</p>
+
+  <pre class="code-container">
+    <!-- prettier-ignore -->
+    <code>
+    [PluginDefinition]
+    Name = foo
+
+    [PluginConfig "FOO_TOOL"]
```
peterebden, line 375: I'm not sure about FOO_TOOL here. I feel like it should be `"foo_tool"` in this context and `CONFIG.FOO_TOOL` in asp?

third_party/go/github_repos/BUILD
```diff
@@ -0,0 +1,7 @@
+package(default_visibility = ["PUBLIC"])
```
peterebden, line 1: why are we doing this in this way rather than as a `go_module`?

src/core/state.go
Tatskaari: I'm just using it here to clone the state object. I suppose I could move that out into something more explicit.

src/core/state.go
Tatskaari: Oh yeah, that's a good point.

third_party/go/github_repos/BUILD
Tatskaari: Ah sorry this should be removed.

docs/config.html
tiagovtristao: Ya, I think this can be made consistent with how the line above `Name = foo` works, where it's referenced via `CONFIG.FOO`

src/core/config.go
```diff
@@ -567,7 +569,16 @@ type Configuration struct {
...
 		Reject []string `help:"Licences that are explicitly rejected in this repository.\nAn astute observer will notice that this is not very different to just not adding it to the accept section, but it does have the advantage of explicitly documenting things that the team aren't allowed to use."`
 	} `help:"Please has some limited support for declaring acceptable licences and detecting them from some libraries. You should not rely on this for complete licence compliance, but it can be a useful check to try to ensure that unacceptable licences do not slip in."`
-	Alias map[string]*Alias `help:"Allows defining alias replacements with more detail than the [aliases] section. Otherwise follows the same process, i.e. performs replacements of command strings."`
+	Alias            map[string]*Alias  `help:"Allows defining alias replacements with more detail than the [aliases] section. Otherwise follows the same process, i.e. performs replacements of command strings."`
+	Plugin           map[string]*Plugin `help:"Allows defining alias replacements with more detail than the [aliases] section. Otherwise follows the same process, i.e. performs replacements of command strings."`
```
tiagovtristao, line 573: This is a copy of the above

test/plugins/test_repo/foo/tools/fooc/fooc.go
```diff
@@ -0,0 +1,44 @@
+package main
+
+import (
+	"crypto/sha1"
+	"flag"
+	"os"
+)
+
+// A dummy compiler. Just hashes the input but is meant to be come sort of language compiler
```
samwestmoreland, line 9: some*

samwestmoreland (approved):
Looks like it needs gofmt'ing but otherwise looks good to me!

PR #1926, opened 2021-07-09
Fix cycle detector

Tatskaari:
This seems to have a bit of a performance impact. Might be a better solution out there. 

The benchmark seems to suggest we're able to process dependencies rather quickly (4k/ms)

### DIFF
--- a/src/core/BUILD
+++ b/src/core/BUILD
@@ -184,7 +184,6 @@ benchmark(
     deps = [":core"],
 )
 
-
 benchmark(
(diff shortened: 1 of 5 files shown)


peterebden (approved):
Think we should go with this as is, we might find a better solution later but it's more important this works for now.

Tatskaari:
Not sure the benchmark is super representative but will land it for now. Can come back to it once we have this tagged and out there.

PR #1929, opened 2021-07-12
Don't add PleaseLocation to the path by default any more

peterebden:
This isn't necessary since we download all the tools as a build action.


Tatskaari:
This seems a bit scary to just push out as a minor release. Should we only do this for the remote env?

peterebden:
Yeah the tests look unhappy (weirdly it had no issues for me locally). Will revisit.

PR #1930, opened 2021-07-12
Strip out parts of PATH that contain the user's home dir for remote execution

peterebden:
This is a bit grungy but should work until we can figure out something nicer.


goddenrich (approved):
LGTM

Tatskaari (approved):
Nice thanks for that :)

PR #1934, opened 2021-07-13
Fix rex test and issue with verifying outputs

Tatskaari:


src/parse/asp/interpreter.go
```diff
@@ -113,7 +113,7 @@ func (i *interpreter) interpretStatements(s *scope, statements []*Statement) (re
 			} else {
 				err = fmt.Errorf("%s", r)
 			}
-			log.Debug("%s", debug.Stack())
+			log.Debug("%v:\n %s", err,  debug.Stack())
```
Tatskaari, line 116: this got me a little confused when reading `build.log`. The err is printed with a log.Error later on but that ends up after the stack trace.

peterebden (approved):

src/remote/action.go
```diff
@@ -404,12 +404,7 @@ func (c *Client) verifyActionResult(target *core.BuildTarget, command *pb.Comman
 			return fmt.Errorf("Remote build action for %s failed to produce output %s%s", target, core.TestResultsFile, c.actionURL(actionDigest, true))
 		}
 	} else {
-		for _, out := range command.OutputFiles {
-			if !outs[out] {
-				return fmt.Errorf("Remote build action for %s failed to produce output %s%s", target, out, c.actionURL(actionDigest, true))
-			}
-		}
-		for _, out := range command.OutputDirectories {
+		for _, out := range command.OutputPaths {
```
Tatskaari, line 407: This comes from the locally constructed command. We don't set OutputFiles or OutputDirectories. I think this is the right thing to do.

PR #1936, opened 2021-07-13
Make completions only complete one level

Tatskaari:
I think the idea for bash completions is that you should return the next set of possible completions. 

In the Please repo, `//sr` is unambiguously meant to be completed to `//src`. The next valid completions are therefore all the build targets in the `//src` package, as well as the immediate packages under `//src`. I've tested this and it seems to work well with bash. 

Conversely, `//src/t` could be `//src/test` or `//src/tool` so we only complete to those two packages. 

In core3:

```
$ plz build //vault/work
```
auto-completes to `//vault/workflows` which then gives:
```
$ plz build //vault/workflows
//vault/workflows:all                   //vault/workflows/e2e                   //vault/workflows/proto                 //vault/workflows/secrets_templates     //vault/workflows/stream_api_processor  //vault/workflows:workflows
//vault/workflows/callback_router       //vault/workflows/engine                //vault/workflows/schedule_processor    //vault/workflows/simulator             //vault/workflows/transform_executor    //vault/workflows/workflows_api
//vault/workflows/configmaps            //vault/workflows/event_poller          //vault/workflows/schemas               //vault/workflows/stream_api_broker     //vault/workflows:update_gitignore
```


peterebden (approved):
Almost certainly a good thing. What happens if you want completions into a deep subdirectory like
```
src/
    BUILD
    main/
        java/
            build/
                please/
                    is/
                         awesome/
                              BUILD
```
Does it jump through all the subdirs or require you to tab through one by one?
(Honestly think the latter is acceptable but the former would be neater)

Tatskaari:
> Almost certainly a good thing. What happens if you want completions into a deep subdirectory like
> 
> ```
> src/
>     BUILD
>     main/
>         java/
>             build/
>                 please/
>                     is/
>                          awesome/
>                               BUILD
> ```
> 
> Does it jump through all the subdirs or require you to tab through one by one?
> (Honestly think the latter is acceptable but the former would be neater)

Ah good point! It should be recursive.

PR #1937, opened 2021-07-14
Add stack track to lsp panics

Tatskaari:


tiagovtristao (approved):

tools/build_langserver/lsp/lsp.go
```diff
@@ -110,6 +112,10 @@ func (h *Handler) handle(method string, params *json.RawMessage) (i interface{},
 		return nil, &jsonrpc2.Error{Code: jsonrpc2.CodeInvalidParams}
 	}
 	ret := m.Func.Call([]reflect.Value{p.Elem()})
+	if len(ret) == 0 {
+		// Some handlers don't return anything e.g. shutdown
+		return nil, nil
+	}
```
Tatskaari, line 118: This is wrong and i don't think this was every nil.

Issue #1802, opened 2021-05-07
Allow junit-directory to be overridden 


thecapdan:
When running behave tests as a pex with the --in_tmp_dir flag, we are unable to retrieve the test results xml (required to be parsed downstream).

It would be nice if we could override the value set by [behave.py](https://github.com/thought-machine/please/blob/master/tools/please_pex/pex/behave.py#L29)

dontm:
@Tatskaari any chance we might get this soon plz? 🙏

Tatskaari:
I've raised a PR that will check the `RESULTS_FILE` env var. Is that good for you? I don't really want to introduce a flag parsing library.

thecapdan:
as long as we can access the results, that's great. thanks

PR #1940, opened 2021-07-15
Add check for env var to behave test runner

Tatskaari:
Fixes: #1802

### DIFF
--- a/tools/please_pex/pex/behave.py
+++ b/tools/please_pex/pex/behave.py
@@ -26,5 +26,5 @@ def run_tests(args=None):
         logging.info('Getting a single feature dir')
         features = get_features_dir()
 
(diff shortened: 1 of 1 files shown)


tools/please_pex/pex/behave.py
```diff
@@ -26,5 +26,5 @@ def run_tests(args=None):
         logging.info('Getting a single feature dir')
         features = get_features_dir()
 
-    args += features + [ '--junit', '--junit-directory', 'test.results']
+    args += features + [ '--junit', '--junit-directory', os.getenv("RESULTS_FILE") or 'test.results']
```
peterebden, line 29: nit: `os.getenv('RESULTS_FILE', 'test.results')` (it already has a default argument)

peterebden (approved):
Nice, it should really have been doing this all along

PR #1943, opened 2021-07-16
Fix lsp server findLabel URI construction

perrydunn:
`TestDefinitionBuildLabel` was failing; now it passes.


Tatskaari (approved):
Thanks Perry!

PR #1945, opened 2021-07-16
This approach might make more sense

Tatskaari:


src/core/cycle_detector.go
```diff
@@ -20,26 +20,35 @@ func (c dependencyChain) String() string {
 }
 
 type cycleDetector struct {
-	deps  map[*BuildLabel][]*BuildLabel
-	queue chan dependencyLink
+	deps        map[*BuildLabel]map[*BuildLabel]struct{}
```
peterebden, line 23: Might be worth noting what exactly this map represents (like what are the keys/values in each one)

peterebden (approved):

Tatskaari:
#1946 is a much smarter approach

PR #1946, opened 2021-07-18
Do cycle checking asynchronously

peterebden:
This moves cycle checking from happening always to only at a point where we have no active work happening for a few seconds. That means there's a little pause before it'll find the cycle but I don't think that's a big problem.

Most of the guts of it is in `forwardResults()` which intercepts all the parse/build/test updates to add this timing logic. The cycle detector can now use a very simple synchronous algorithm since it should never happen on the hot path.

On my machine this reduces parse times from ~10s to ~6.5s (specifically 9.56s, 9.88s, 12.33s -> 6.50s, 6.54s, 6.61) vs. current head of master.

The in-repo cyclic dependency test passes; can give it a go if we have any other tests around this stuff too.


src/core/cycle_detector.go
```diff
@@ -5,108 +5,60 @@ import (
...
-	if tailDeps, ok := c.deps[tail]; ok {
-		for _, dep := range tailDeps {
-			if dep == head {
-				// If the tail has a dependency on the head, we've found a cycle
-				return append(chain, dep)
-			}
 
-			if newChain := c.buildCycle(append(chain, dep), done); newChain != nil {
-				return newChain
+	for _, target := range c.graph.AllTargets() {
```
Tatskaari, line 40: I wonder if we can get the queued targets that are waiting on their deps somehow?

src/core/cycle_detector.go
```diff
@@ -5,108 +5,60 @@ import (
...
+			if cycle, done := visit(dep); cycle != nil {
+				if done || target == cycle[len(cycle)-1] {
+					return cycle, true // This target is already in the cycle
+				}
+				return append([]*BuildTarget{target}, cycle...), false
 			}
 		}
+		delete(temporary, target)
+		permanent[target] = struct{}{}
+		return nil, false
```
Tatskaari, line 37: If these return values are the path so far, and whether we exhausted the tree, this should probably be `return nil, true`. Same as when the target is in `permanent`. It doesn't matter as we don't check this when the path is nil but this tripped me when trying to understand this.

src/core/cycle_detector.go
```diff
@@ -5,108 +5,60 @@ import (
...
-			}
-
-			if c.checkForCycle(head, dep, done) {
-				return true
+// Check runs a single check of the build graph to see if any cycles can be detected.
+// If it finds one an errCycle is returned.
+func (c *cycleDetector) Check() *errCycle {
+	log.Debug("Running cycle detection...")
+	permanent := map[*BuildTarget]struct{}{}
+	temporary := map[*BuildTarget]struct{}{}
```
Tatskaari, line 17: I wonder if `complete` and `partial` would make more sense as names for these. If I understand:

1. Permanent are targets where we've explored their entire dependency tree and found no cycles. They're completely visited. 
2. Temporary are targets where we're currently exploring their deps so we don't need to re-visit them. They're partially visited.

src/core/cycle_detector.go
```diff
@@ -5,108 +5,60 @@ import (
...
+
+	var visit func(target *BuildTarget) ([]*BuildTarget, bool)
+	visit = func(target *BuildTarget) ([]*BuildTarget, bool) {
+		if _, present := permanent[target]; present {
+			return nil, false
+		} else if _, present := temporary[target]; present {
+			return []*BuildTarget{target}, false
+		}
```
Tatskaari, line 25: Is this how we ensure we get the shorter chain? Smart. Would be nice if we could build the chain the other way because appending to a slice is cheaper but I don't have any ideas.

Tatskaari (approved):
Added some comments to prove I read and understood it ;P 

I think it's probably okay to land. I have a concern that if the cycle detector gets behind the 2 second period for checking cycles, we could have a problem however the algo looks linear in time so should be okay.

src/core/cycle_detector.go
peterebden: yeah I took the names from https://en.wikipedia.org/wiki/Topological_sorting which has an example that is basically this algorithm (we're not actually top-sorting but it's pretty much the same thing).

src/core/cycle_detector.go
peterebden: Yeah there's a bit of faffing with the second return value (which is basically to chop it off so a -> b -> c -> d -> b doesn't return a as well).

Good point, I thought the same on the slices, but at the point we're building up the slice we're doomed anyway so I don't think it will matter much.

src/core/cycle_detector.go
peterebden: Maybe it's a little unclear - I put the second value in later when I was writing the tests. Will try to write down what it actually means...

src/core/state.go
```diff
@@ -491,17 +497,72 @@ func (state *BuildState) LogResult(result *BuildResult) {
...
+			select {
+			case result = <-state.progress.internalResults:
+				// This has to be properly managed to prevent hangs.
+				if !t.Stop() {
+					<-t.C
+				}
+			case <-t.C:
+				go state.checkForCycles()
+				// Still need to get a result!
+				result = <-state.progress.internalResults
```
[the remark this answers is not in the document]
peterebden: It will only continue if something does manage to progress - so for us to start overloading the system we'd have to have events emitted more slowly than 1 every 2 seconds, but also fast enough to re-trigger this quickly. That's _theoretically_ possible but seems so unlikely to happen that way in practice that I wasn't really worrying about it.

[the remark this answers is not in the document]
Tatskaari: Ah I see, I had a minor misunderstanding where I thought we could trigger these on builds finishing if we had no newly active targets coming in. Looks like we wait all active targets to finish before starting the timer though.

src/core/cycle_detector.go
Tatskaari: Ah oh yeah, we only return the chain if we find something! That's actually really clever.

PR #1947, opened 2021-07-19
Enable maligned 

peterebden:
Turns the linter on & optimises a few structs it highlights.

Have disabled it for config & flags which are both big with lots of sub-structs, and each singletons so we are unlikely to gain much from optimising them. No particular problem if someone wants to but I don't feel like it right now :)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -42,8 +42,8 @@ linters:
     - prealloc
     - unconvert
     - whitespace
+    - maligned
 
     # To consider enabling in future (once we fix errors etc):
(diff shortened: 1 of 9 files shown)


Tatskaari:
Let's get the benchmark in before this so we can pat ourselves on the back. :P

peterebden:
Makes sense. Not expecting to see a lot of improvement on the current tests though; the only structs that would be repeated a lot are test-related.

Tatskaari (approved):
feel free to land

PR #1951, opened 2021-07-21
Move test fields out into optional struct on BuildTarget

Tatskaari:
This PR moves all the test fields on `BuildTarget{}` out into their own struct. This reduces the `unsafe.SizeOf()` for of a build target from `832` to `640` bytes saving 96 bytes per build targets which is is around 60mb of heap space in the parse perf test

Need to fix how we print targets as that depended on the order of the fields in the struct which is now out of order.


src/build/incrementality_test.go
```diff
@@ -29,7 +29,7 @@ var KnownFields = map[string]bool{
 	"Command":                     true,
 	"Commands":                    true,
 	"TestCommand":                 true,
-	"TestCommands":                true,
+	"Commands":                true,
```
peterebden, line 32: needs gofmt

src/core/build_target.go
```diff
@@ -54,6 +54,26 @@ const TestResultsDirLabel = "test_results_dir"
...
+type TestFields struct {
+	// Shell command to run for test targets.
+	Command string `name:"test_cmd"`
+	// Per-configuration test commands to run.
+	Commands map[string]string `name:"test_cmd"`
+	// True if the test action is sandboxed.
+	Sandbox bool `name:"test_sandbox"`
+	// True if the target is a test and has no output file.
+	// Default is false, meaning all tests must produce test.results as output.
+	NoTestOutput bool `name:"no_test_output"`
```
peterebden, line 66: Should this just be `NoOutput` now?

src/core/build_target.go
```diff
@@ -54,6 +54,26 @@ const TestResultsDirLabel = "test_results_dir"
...
+	// Like tools but available to the test_cmd instead
+	testTools []BuildInput `name:"test_tools"`
```
peterebden, line 68: similarly here, just `tools`?

src/build/incrementality.go
```diff
@@ -211,9 +211,7 @@ func ruleHash(state *core.BuildState, target *core.BuildTarget, runtime bool) []
 	for _, licence := range target.Licences {
 		h.Write([]byte(licence))
 	}
-	for _, output := range target.TestOutputs {
-		h.Write([]byte(output))
-	}
```
Tatskaari, line 216: Pretty sure these should only contribute to the runtime hash.

src/core/build_target.go
```diff
@@ -92,63 +114,11 @@ type BuildTarget struct {
...
-	// Marks that the target was added in a post-build function.
-	AddedPostBuild bool `print:"false"`
-	// If true, the interactive progress display will try to infer the target's progress
-	// via some heuristics on its output.
-	ShowProgress bool `name:"progress"`
+	Test     *TestFields       `name:"test"`
 	// If ShowProgress is true, this is used to store the current progress of the target.
 	Progress float32 `print:"false"`
 	// The results of this test target, if it is one.
 	Results TestSuite `print:"false"`
```
peterebden, line 121: Can you move the Results field into the sub-struct too?

Tatskaari: AH yeah good catch. It seems I missed a couple fields.

peterebden: yeah it just occurred to me when I saw you were moving flakiness, there are a few not prefixed with `Test`.

src/build/incrementality.go
Tatskaari: > Makes sense.
> 
> Note that these changes will mean it will rebuild _everything_. Maybe we just live with that though.

I mean we do that quite a bit anyway because the tools are in the root of the build graph. Might want to think about this because it was better in this sense before we moved tools to remote files.

peterebden: Yeah. I still think breaking out the tools to separate repos (plugins!) would be the best solution there.

src/core/build_target.go
```diff
@@ -1190,22 +1199,22 @@ func (target *BuildTarget) AddTool(tool BuildInput) {
 
 // AddTestTool adds a new test tool to the target.
 func (target *BuildTarget) AddTestTool(tool BuildInput) {
-	target.testTools = append(target.testTools, tool)
+	target.Test.tools = append(target.Test.tools, tool)
```
tiagovtristao, line 1202: Is it guaranteed that `target.Test` is non nil at this stage?

tiagovtristao: Same for the other functions below

Tatskaari: There's only one* place where we should be constructing targets: `src/parse/asp.createTarget(...)`. When we construct the target we guard against calling these functions with an `if target.IsTest() {}` so it should be set. If it is a target, we make sure this field is initialised. 

That being said, don't look at my PR. I'm infallible and there's nothing to see there. ;P

* okay there are a couple others but they're not taking in user input so are safe.

test/BUILD
```diff
@@ -473,3 +473,23 @@ plz_e2e_test(
     cmd = "plz --completion_script",
     expect_output_contains = "plz completion",
 )
+
+# This rule being here makes sure than when we parse this file, none of the test fields being set causes a nil pointer
```
tiagovtristao, line 477: This gives me peace of mind :-D

tiagovtristao (approved):

PR #1953, opened 2021-07-21
Make query revdeps fail faster on targets that don't exist

peterebden:
This should have no effect on the results, but will tend to fail earlier on if an argument is given that doesn't exist.

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -620,8 +620,9 @@ var buildFunctions = map[string]func() int{
 		})
 	},
 	"revdeps": func() int {
-		return runQuery(true, core.WholeGraph, func(state *core.BuildState) {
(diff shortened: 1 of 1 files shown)


Tatskaari (approved):
Nice!

RichardoC:
We could also do the same for the other commands, including deps

RichardoC (approved):
Thanks! Should be a bit faster for my usecase

peterebden:
> We could also do the same for the other commands, including deps

Hmmm no - deps effectively does this already (most commands just parse for the targets they need, revdeps is a bit unusual in that respect).

Tatskaari:
Yeah this only effects commands that operate on the whole graph but take in targets as an arg. I can't think of any others except revdeps that do that.

peterebden:
`s/effects/affects/`

Tatskaari:
Damn it I'm normally good at that one!

PR #1956, opened 2021-07-21
Size deps slice to the precise capacity

peterebden:
Should save a few allocations & a bit of memory. Doesn't seem to be enough to break through the noise of system-level benchmarks but should have no downside.


Tatskaari (approved):
Nice one

PR #1957, opened 2021-07-21
Update golangci-lint

peterebden:
Some things have changed.
golint -> revive
maligned -> govet fieldalignment (although it has some finicky extra stuff so it's off again for now)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,11 +19,13 @@ issues:
         - errcheck
         - dupl
     - path: src/core/config.go # The config struct is big & complex and there's usually only one.
+      text: fieldalignment
       linters:
-        - maligned
+        - govet
     - path: src/please.go # Similarly for flags.
+      text: fieldalignment
       linters:
-        - maligned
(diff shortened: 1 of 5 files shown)


.golangci.yml
```diff
@@ -44,13 +46,22 @@ linters:
...
+
+linters-settings:
+  govet:
+    enable-all: true
+    disable:
+      - deepequalerrors # Only seen this once and it's a false positive
+      - unusedwrite # Similarly
+      - shadow # Probably not useful enough to clean everything up
+      # TODO(peterebden): Should clean up these warnings & enable (although they are pretty finicky)
+      - fieldalignment
```
Tatskaari, line 67: This is `malalign` right?

Tatskaari (approved):

.golangci.yml
peterebden: Yes, it also has an extra check about GC pointer size which is part of the same checker (but quite a bit more subtle)

PR #1964, opened 2021-07-26
Remove pendingTargets map from graph

peterebden:
This unifies the map of pending targets / packages with the target & package maps themselves. Hence adding a target doesn't require modifying a second map & waiting on it is also cleaner.

Does mean we've got a second custom map type but I think we should just suck that up and refactor them into one common thing once we get generics (which is not so far away now).

Benchmark before:
```
BenchmarkAddingTargets
BenchmarkAddingTargets-12        	  757390	      1852 ns/op	     509 B/op	       3 allocs/op
BenchmarkTargetLookup
BenchmarkTargetLookup/Simple
BenchmarkTargetLookup/Simple-12  	 3883336	       301.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkTargetLookup/WaitForTargetFast
BenchmarkTargetLookup/WaitForTargetFast-12         	 4017552	       304.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkWaitForTargetSlow
BenchmarkWaitForTargetSlow-12                      	 1699608	       889.5 ns/op	     604 B/op	       3 allocs/op
```
After:
```
BenchmarkAddingTargets
BenchmarkAddingTargets-12        	 1387692	       729.8 ns/op	     215 B/op	       0 allocs/op
BenchmarkTargetLookup
BenchmarkTargetLookup/Simple
BenchmarkTargetLookup/Simple-12  	 3927500	       298.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkTargetLookup/WaitForTargetFast
BenchmarkTargetLookup/WaitForTargetFast-12         	 3866668	       297.7 ns/op	       0 B/op	       0 allocs/op
BenchmarkWaitForTargetSlow
BenchmarkWaitForTargetSlow-12                      	 2161027	       660.0 ns/op	     283 B/op	       0 allocs/op
```
i.e. addition is 2-3x faster, waiting is a bit faster, and both allocate less.

### DIFF
--- a/src/core/cmap_packages.go
+++ b/src/core/cmap_packages.go
@@ -82,7 +82,7 @@ func (lm *packageLMap) Set(key packageKey, pkg *Package) bool {
 	defer lm.l.Unlock()
 	if existing, present := lm.m[key]; present {
 		if existing.Package != nil {
-			return false  // already added
+			return false // already added
 		}
 		// Hasn't been added, but something is waiting for it to be.
 		lm.m[key] = packagePair{Package: pkg}
--- a/src/core/cmap_packages.go
+++ b/src/core/cmap_packages.go
@@ -1,0 +1,122 @@
+// This is similar to cmap_targets.go.
+// TODO(peterebden, jpoole): When we get generics in 1.18 this would be a prime use case for them.
+
+package core
+
+import (
+	"sync"
+
+	"github.com/OneOfOne/cmap/hashers"
+)
+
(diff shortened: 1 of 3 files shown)


Tatskaari:
Nice! I think working through some of the excess garbage we generate will help with memory fragmentation if we get enough of them in.

Tatskaari (approved):

PR #1966, opened 2021-07-26
Order fields in `plz query print` based on build_rule() in the .build_defs

Tatskaari:
This will help when I move the test fields out otherwise I would have to just print them at the end or something.


peterebden (approved):

src/query/print.go
```diff
@@ -10,12 +10,15 @@ import (
...
+	"github.com/thought-machine/please/src/parse"
 )
 
 // Print produces a Python call which would (hopefully) regenerate the same build rule if run.
 // This is of course not ideal since they were almost certainly created as a java_library
 // or some similar wrapper rule, but we've lost that information by now.
-func Print(graph *core.BuildGraph, targets []core.BuildLabel, fields, labels []string) {
+func Print(state *core.BuildState, targets []core.BuildLabel, fields, labels []string) {
+	graph := state.Graph
+	order := parse.NewAspParser(state).BuildRuleArgOrder()
```
peterebden, line 21: Why do you need to create a new parser? Can't you just use `state.Parser` to answer this question?

peterebden (changes requested):

src/query/print.go
Tatskaari: I'd have to cast it to `parser.aspParser` and then get at the internal `asp.Parser` field on that. We don't expose any of that so I'd have to expose more things. I think you're right, that's probably the better option.

src/core/state.go
```diff
@@ -53,6 +53,7 @@ type Parser interface {
 	RunPreBuildFunction(threadID int, state *BuildState, target *BuildTarget) error
 	// RunPostBuildFunction runs a post-build function for a target.
 	RunPostBuildFunction(threadID int, state *BuildState, target *BuildTarget, output string) error
+	BuildRuleArgOrder() map[string]int
```
peterebden, line 56: nit: comment for this would be nice

peterebden (approved):

PR #1967, opened 2021-07-27
Mod file cleanup

peterebden:
Getting nagged by dependabot about CVE-2020-27813 which is an overflow in gorilla/websocket. I can't find any good reason why we should need that.

This is generated by running `go mod tidy` plus a few manual removals that it didn't seem to get. AFAICT that's legit since it doesn't seem to try to put them back again.


Tatskaari:
Sorry I meant to approve not land... assume this was okay to go though.

peterebden:
Yeah I wasn't going to do anything else with it.

PR #1969, opened 2021-07-27
Don't do the check on built targets if we're not building.

peterebden:
I don't think any of this is very relevant for parsing and it actually takes several hundred ms on a big graph (because it churns through all the packages one by one).

### DIFF
--- a/src/output/shell_output.go
+++ b/src/output/shell_output.go
@@ -90,14 +90,16 @@ func MonitorState(ctx context.Context, state *core.BuildState, plainOutput, deta
 		printFailedBuildResults(failedNonTests, failedTargetMap, duration)
 		return
 	}
-	// Check all the targets we wanted to build actually have been built.
-	for _, label := range state.ExpandOriginalLabels() {
-		if target := state.Graph.Target(label); target == nil {
-			log.Fatalf("Target %s doesn't exist in build graph", label)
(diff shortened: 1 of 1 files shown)


src/output/shell_output.go
```diff
@@ -90,14 +90,16 @@ func MonitorState(ctx context.Context, state *core.BuildState, plainOutput, deta
...
-		} else if state.NeedBuild && target.State() < core.Built && len(failedTargetMap) == 0 && !target.AddedPostBuild {
-			log.Fatalf("Target %s hasn't built but we have no pending tasks left.\n%s", label, unbuiltTargetsMessage(state.Graph))
+	if state.NeedBuild {
+		// Check all the targets we wanted to build actually have been built.
+		for _, label := range state.ExpandOriginalLabels() {
+			if target := state.Graph.Target(label); target == nil {
+				log.Fatalf("Target %s doesn't exist in build graph", label)
+			} else if (state.NeedHashesOnly || state.PrepareOnly || state.PrepareShell) && target.State() == core.Stopped {
+				// Do nothing, we will output about this shortly.
+			} else if state.NeedBuild && target.State() < core.Built && len(failedTargetMap) == 0 && !target.AddedPostBuild {
```
Tatskaari, line 100: nit: can remove sate.NeedBuild here

PR #1971, opened 2021-07-27
Target level locking

tiagovtristao:
This PR implements:

- Target level locking at build and test stages. Running certain commands multiple times in parallel used to break, such as  `plz build //src/... --rebuild --nolock` or `plz test -i e2e --rerun --nolock`. This no longer happened or rarely during my tests. I'm sure there are other combinations where we can make the commands break, but I think this is a good first step for our use cases for now. The only recurring issue popping up was `Failed to store cache file ... no such file or directory ` when running `plz test -i e2e --rerun --nolock` multiple times. I reckon this is related to test result storage happening asynchronously/after the lock has been released.
- I tried to get rid of the repo level lock feature, but it still has its uses (i.e. getting plz to update itself). In order to get around the edge cases where a repo level lock is necessary, where now it's not enabled be default, I leveraged shared locks as well. 
- Separated file locking and previous operation logic.

TODO:
- `--nolock` no longer seems to make sense, at the least at the repo level. How these changes are rolled out needs to be figured out.


src/core/lock.go
```diff
@@ -1,98 +1,126 @@
...
 
-func openLockFile() {
-	if lockFile != nil {
-		return
+func acquireFileLock(file *os.File, how int, content string) error {
+	// Try a non-blocking acquire first so we can warn the user if we're waiting.
+	log.Debug("Attempting to acquire lock for %s...", file.Name())
+	err := syscall.Flock(int(file.Fd()), how|syscall.LOCK_NB)
+	if err != nil {
+		log.Debug("Looks like another thread has already acquired the lock for %s. Waiting for it to finish...", file.Name())
```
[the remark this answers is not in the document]
peterebden: Also is it worth reading the file contents to get the pid of the thing that's locked it? Wonder if this would help people debug when this happens.

src/update/update.go
```diff
@@ -68,10 +69,9 @@ func CheckAndUpdate(config *core.Configuration, updatesEnabled, updateCommand, f
 		fmt.Fprintf(os.Stderr, "%s Please from version %s to %s", word, pleaseVersion(), config.Please.Version.VersionString())
 	}
 
-	// Must lock here so that the update process doesn't race when running two instances
-	// simultaneously.
-	core.AcquireRepoLock()
-	defer core.ReleaseRepoLock()
+	// Must lock exclusively here so that the update process doesn't race when running two instances simultaneously.
+	// We don't want to release the lock yet since other parts of the code might override it as needed.
```
peterebden, line 73: I'm not sure I follow this. I would expect `AcquireExclusiveRepoLock` to always be paired with a deferred release?

peterebden (commented):
Thanks Tiago, uncomfortably excited about this!

src/build/build_step.go
```diff
@@ -239,6 +239,12 @@ func buildTarget(tid int, state *core.BuildState, target *core.BuildTarget, runR
 			buildLinks(state, target)
 			return nil
 		}
+
+		state.LogBuildResult(tid, target, core.TargetBuilding, "Acquiring file lock...")
+		file := core.AcquireExclusiveFileLock(target.BuildLockFile())
+		defer core.ReleaseFileLock(&file)
```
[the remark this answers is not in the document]
tiagovtristao: Agreed, just refactored it

src/core/lock.go
```diff
@@ -1,98 +1,126 @@
-// Contains utility functions for managing an exclusive lock file.
-// Based on flock() underneath so
+// The logic below relies heavily on flock (advisory locks).
+// Race conditions are a possibility given that we need to open files before applying a lock.
```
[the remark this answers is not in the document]
tiagovtristao: I guess I was trying to convey that the logic is not atomic, although it shouldn't case real issues. Have removed it to avoid confusion

src/core/lock.go
```diff
@@ -1,98 +1,126 @@
...
+// Base function that allows to set up different lock modes and facilitate testing.
+func acquireOpenFileLock(filePath string, how int) (*os.File, error) {
+	lockFile, err := openLockFile(filePath)
+	if err != nil {
+		return nil, err
 	}
-	if err := lockFile.Close(); err != nil {
-		log.Errorf("Failed to close lock file: %s", err)
+
+	if err = acquireFileLock(lockFile, how, fmt.Sprint(os.Getpid())); err != nil {
```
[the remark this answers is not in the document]
tiagovtristao: Nice, done.

src/core/lock.go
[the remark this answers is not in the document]
tiagovtristao: Good idea!

src/update/update.go
tiagovtristao: Actually, in other for this to be more correct, I needed to `defer AcquireSharedRepoLock()`. This was working before because `AcquireSharedRepoLock` is called in `runPlease`. Since the repo lock fd is a singleton, replacing the lock mode is enough.

peterebden (approved):
Thanks Tiago! LGTM, would be good for @Tatskaari to check it over too so he's familiar with the brave new world of locking :)

src/please.go
```diff
@@ -958,11 +958,15 @@ func Please(targets []core.BuildLabel, config *core.Configuration, shouldBuild,
...
-	if (state.NeedBuild || state.NeedTests) && !opts.FeatureFlags.NoLock {
-		core.AcquireRepoLock()
-		defer core.ReleaseRepoLock()
-	}
+	// Every plz instance gets a shared repo lock which provides the following:
+	// 1) Multiple plz instances can run concurrently.
+	// 2) If another process tries to obtain an exclusive repo lock, it will have to wait until any existing repo locks are released in other processes.
+	// This is useful for things like when plz tries to download and update itself.
+	// 3) A new plz process will have to wait to acquire its shared repo lock, if there's already an existing process with an exclusive repo lock.
+	core.AcquireSharedRepoLock()
```
Tatskaari, line 966: I'm not sure i fully understand why we need a shared lock. If I understand how this is implemented, a `plz build` would acquire the shared lock, and then a `plz update` would have to wait for the build to finish before downloading the new Please version. Is downloading Please while building problematic?

src/test/test_step.go
```diff
@@ -179,6 +179,11 @@ func test(tid int, state *core.BuildState, label core.BuildLabel, target *core.B
 		}
 	}
 
+	state.LogBuildResult(tid, target, core.TargetTesting, "Acquiring file lock...")
```
Tatskaari, line 182: `Acquiring target lock...` would be more descriptive I think.

src/update/update.go
```diff
@@ -68,10 +69,10 @@ func CheckAndUpdate(config *core.Configuration, updatesEnabled, updateCommand, f
...
 	}
 
-	// Must lock here so that the update process doesn't race when running two instances
-	// simultaneously.
-	core.AcquireRepoLock()
-	defer core.ReleaseRepoLock()
+	// Must lock exclusively here so that the update process doesn't race when running two instances simultaneously.
+	// Once we are done we replace/restore the mode to the shared one.
+	core.AcquireExclusiveRepoLock()
+	defer core.AcquireSharedRepoLock()
```
Tatskaari, line 75: Should this be `core.ReleaseExclusiveRepoLock()` instead ?

Tatskaari (commented):
Looks good! I'm a little confused about the shared locks though.

src/please.go
tiagovtristao: As discussed IRL, I'm not really sure whether it would problematic. The current logic does rely on it so I kept it. Although only `plz update` uses an exclusive lock for now, there might be other code paths in the future that require a exclusive lock, so this logic is already accounting for that.

src/test/test_step.go
tiagovtristao: Agreed

src/core/lock.go
```diff
@@ -1,98 +1,139 @@
...
+type logFunc func(logger *logging.Logger, format string, args ...interface{})
+
+func acquireFileLock(file *os.File, how int, levelLog logFunc) error {
+	// Try a non-blocking acquire first so we can warn the user if we're waiting.
+	log.Debug("Attempting to acquire lock for %s...", file.Name())
+	err := syscall.Flock(int(file.Fd()), how|syscall.LOCK_NB)
+	if err != nil {
+		pid, err := ioutil.ReadFile(file.Name())
+		if err != nil || len(pid) == 0 {
+			return fmt.Errorf("Failed to retrieve PID of process holding the lock for %s", file.Name())
```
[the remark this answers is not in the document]
tiagovtristao: 👍

src/update/update.go
tiagovtristao: No, as it would release the lock making it prone to race conditions with other processes. We want to keep the lock open for the lifetime of the process and just swap between the modes (ie. exclusive and shared) as needed. The shared mode should always be the default one. You can almost think of it as the lock being "released" in one sense, but it brings the benefits described in an earlier comment.

Tatskaari (approved):
Thanks Tiago! Looks good to me. Let's hold off on landing until I get the next Please release out this evening. I'll take some time to test this out locally.

peterebden (approved):
Thanks Tiago! Agreed with Jon, this is a good one for us to dogfood for a bit before unleashing :)

PR #1973, opened 2021-07-28
Add an add_data builtin for adding data to a target

samwestmoreland:
No idea if this is right but let me know your thoughts. Still missing the named data stuff.

Also getting an error when I try to run the new test about add_data not being defined from line 280 in test/BUILD... I'm sure it's something silly!


test/BUILD
```diff
@@ -273,6 +273,21 @@ plz_e2e_test(
...
+genrule(
+    name = "_add_data_gen",
+    cmd = "echo hello > _add_data_gen.txt",
+    pre_build = lambda name: add_data(name, "add_data_gen.txt"),
+)
+
+gentest(
+    name = "_add_data_test",
+    data = [":_add_data_gen"],
+    labels = ["manual"],
```
Tatskaari, line 286: This marks the test as a manual test. This means it won't get run on the CI servers. Not sure that's what we want.

test/BUILD
```diff
@@ -273,6 +273,21 @@ plz_e2e_test(
     expected_failure = True,
 )
 
+# Test add_data functionality
+genrule(
+    name = "_add_data_gen",
+    cmd = "echo hello > _add_data_gen.txt",
+    pre_build = lambda name: add_data(name, "add_data_gen.txt"),
```
Tatskaari, line 280: This isn't going to work unless you're using a Please version that has `add_data()`. As you can see, the tests pass on CI because they build you a new version of Please before running this test. 

You can `plz plz test` this, however it might be better to use a `please_repo_e2e_test()`. These can run the development version of Please in a test repo so you don't have to worry about this sort of stuff. There are plenty of examples but five me a shout if you want a hand.

Tatskaari:
Looks good so far! Nice one. :)

src/parse/asp/builtins.go
```diff
@@ -753,6 +754,41 @@ func addDep(s *scope, args []pyObject) pyObject {
...
+// Add runtime dependencies to target
+func addData(s *scope, args []pyObject) pyObject {
+	s.Assert(s.Callback, "can only be called from a pre- or post-build callback")
+	target := getTargetPost(s, string(args[0].(pyString)))
+
+	// add_data() builtin can take a string, list, or dict
+	if isType(args[1], "string") {
+		addDataToTargetAndMaybeQueue(s, fmt.Sprint(args[1].(pyString)), target)
+	} else if isType(args[1], "list") {
+		for str := range args[1].(pyList) {
```
Tatskaari, line 776: str here is going to be the index of the slice. You probably want:
```
for _, str := range args[1].(pyList) {}
```

src/parse/asp/builtins.go
```diff
@@ -753,6 +754,41 @@ func addDep(s *scope, args []pyObject) pyObject {
...
+			addDataToTargetAndMaybeQueue(s, fmt.Sprint(str), target)
+		}
+	} else if isType(args[1], "dict") {
+		for _, v := range args[1].(pyDict) {
+			for str := range v.(pyList) {
+				addDataToTargetAndMaybeQueue(s, fmt.Sprint(str), target)
```
Tatskaari, line 782: We probably want to add these as named data.

src/parse/asp/builtins.go
samwestmoreland: Thanks for pointing out the issue with str as well. Doesn't make the test pass though - my latest test is just adding a string so it should go into the first condition, it just ends up not adding the data for some reason

src/parse/asp/builtins.go
```diff
@@ -753,6 +754,64 @@ func addDep(s *scope, args []pyObject) pyObject {
...
+	} else if isType(args[1], "dict") {
+		for name, v := range args[1].(pyDict) {
+			for _, str := range v.(pyList) {
+				if bi := ParseBuildInput(s, str, string(label.(pyString)), systemAllowed, tool); bi != nil {
+					addNamedDatumToTargetAndMaybeQueue(s, name, target, str, systemAllowed, tool)
+				}
+			}
+		}
+	} else {
+		log.Error("Unrecognised data type passed to add_data")
```
Tatskaari, line 807: Although we shouldn't hit this line, this should probably be `log.Fatal` otherwise the build will carry on which could lead to some funny situations. `log.Error()` just prints the message to the terminal but doesn't cause the build to terminate.

src/parse/asp/builtins.go
```diff
@@ -753,6 +754,64 @@ func addDep(s *scope, args []pyObject) pyObject {
 	return None
 }
 
+func addDatumToTargetAndMaybeQueue(s *scope, target *core.BuildTarget, obj pyObject, systemAllowed, tool bool) {
+	datum := core.NewFileLabel(string(obj.(pyString)), s.pkg)
+	target.AddDatum(datum)
```
Tatskaari, line 759: This isn't quite right. We should be taking in `datum` as a param from `addData()`

src/parse/asp/builtins.go
```diff
@@ -753,6 +754,64 @@ func addDep(s *scope, args []pyObject) pyObject {
...
+
+	label := args[0]
+	target := getTargetPost(s, string(label.(pyString)))
+
+	systemAllowed := false
+	tool := false
+
+	// add_data() builtin can take a string, list, or dict
+	if isType(args[1], "str") {
+		if bi := ParseBuildInput(s, args[1], string(label.(pyString)), systemAllowed, tool); bi != nil {
```
Tatskaari, line 789: Oops, it looks like this was local to the package anyway so we didn't actually need to make this public. Sorry about that. Given that it takes the interpreter scope, it probably shouldn't be exposed outside of this package. Ideally we'd be able to parse a build input without the need for the interpreter scope though..

test/plz_run/test_repo/src/BUILD_FILE
```diff
@@ -15,4 +15,13 @@ sh_cmd(
...
\ No newline at end of file
+)
+
+# Test add_data functionality
+build_rule(
+    name = "_add_data_rule",
+    test = True,
+    no_test_output = True,
+    test_cmd = "if [ $DATA != $PKG_DIR/dummydata.txt ]; then exit 1; else exit 0; fi;",
+    pre_build = lambda name: add_data(name, "dummydata.txt"),
```
Tatskaari, line 26: Would be nice to see this add a build target as data too.

Tatskaari (commented):
Very close! Just a couple changes and I think we can get this in :D

rules/builtins.build_defs
```diff
@@ -232,6 +232,8 @@ def add_dep(target:str, dep:str, exported:bool=False):
     pass
 def add_exported_dep(target:str, dep:str):
     add_dep(target, dep, True)
+def add_data(target:str, deps:str|list|dict):
```
peterebden, line 235: deps -> data (or datum if you want to get technical) here

src/core/state.go
```diff
@@ -874,6 +874,19 @@ func (state *BuildState) queueTarget(label, dependent BuildLabel, rescan, forceB
 	return nil
 }
 
+func (state *BuildState) QueueTestTarget(target *BuildTarget) {
+	go state.queueTestTarget(target)
```
peterebden: not sure if you need to do something about the count of pending actions or not here? maybe it has already been taken care of?

src/parse/asp/builtins.go
```diff
@@ -753,6 +754,62 @@ func addDep(s *scope, args []pyObject) pyObject {
...
+		err := s.state.QueueTarget(l, target.Label, true, false)
+		s.Assert(err == nil, "%s", err)
+	}
+}
+
+// Add runtime dependencies to target
+func addData(s *scope, args []pyObject) pyObject {
+	s.Assert(s.Callback, "can only be called from a pre- or post-build callback")
+
+	label := args[0]
```
peterebden, line 779: super nit: might do a `datum := args[1]` here since you repeatedly use it below (and IMO it just reads a bit nicer to name them)

src/parse/asp/builtins.go
```diff
@@ -753,6 +754,62 @@ func addDep(s *scope, args []pyObject) pyObject {
...
+				if bi := parseBuildInput(s, str, string(label.(pyString)), systemAllowed, tool); bi != nil {
+					addNamedDatumToTargetAndMaybeQueue(s, name, target, bi, systemAllowed, tool)
+				}
+			}
+		}
+	} else {
+		log.Fatal("Unrecognised data type passed to add_data")
+	}
+
+	// TODO(peterebden): Do we even need the following any more?
```
peterebden, line 808: oof yeah I think I tried to get rid of this a while back... Sort of think we shouldn't but fair TODO :)

test/plz_run/test_repo/src/BUILD_FILE
peterebden: yeah +1 that is a significantly different codepath internally so would be more comfortable if we knew we covered that.

peterebden (commented):
Thanks Sam - nice work!

test/plz_run/test_repo/src/BUILD_FILE
Tatskaari: Looks like there's another test below  that covers that now :)

src/core/state.go
```diff
@@ -874,6 +874,19 @@ func (state *BuildState) queueTarget(label, dependent BuildLabel, rescan, forceB
 	return nil
 }
 
+func (state *BuildState) QueueTestTarget(target *BuildTarget) {
+	state.queueTestTarget(target)
```
Tatskaari, line 878: I guess we should add a TODO (and raise an issue?) to properly look into making this async.

Tatskaari (approved):
Looks good to me! Nice one Sam. Land when ready.

PR #1978, opened 2021-07-29
Cross compile from darwin amd64 -> arm64 instead

Tatskaari:


tools/please_go/install/install.go
```diff
@@ -299,6 +299,10 @@ func (install *PleaseGoInstall) compilePackage(target string, pkg *build.Package
 			}
 		}
 
+		if f := os.Getenv("CFLAGS"); f != "" {
```
peterebden, line 302: Feel like this should probably come into this tool as a flag rather than it just grabbing an env var internally?

peterebden (approved):

PR #1982, opened 2021-07-30
Adding behave to the list of test runners in python_test

joshcoales:
This brings it into line with what's stated on the config page: https://please.build/config.html


Tatskaari:
Thanks :D

PR #1983, opened 2021-07-30
Revert "Implement per-function profiling metrics for asp (#1774)"

peterebden:
This reverts commit 444a626645424258096d9e05aecd6b367a864a59.

I'm not finding this drastically useful; a lot of the time seems to remain unlabelled, and it's probably better not to have profiling-specific paths.

### DIFF
--- a/src/core/config.go
+++ b/src/core/config.go
@@ -578,9 +578,6 @@ type Configuration struct {
 	PleaseLocation string
 	// buildEnvStored is a cached form of BuildEnv.
 	buildEnvStored *storedBuildEnv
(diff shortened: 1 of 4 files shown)


Tatskaari (approved):
Shame. Was this not useful? It seemed like an interesting idea but I guess it's unlikely to provide much insight as build defs are unlikely to be complex calls.

peterebden:
It hasn't really enlightened me a lot. Maybe we could give it a bit longer though, it is nice to be able to make use of this thing.

Tatskaari:
I guess somebody might find a use for it one day and it's not really doing any harm.

PR #1984, opened 2021-07-31
Add cpu-counting code to benchmark

peterebden:
Thought it might help shed a bit of light on some of the ups and downs of the performance benchmarks, and is just useful info to have (especially if e.g. Circle moved from m4s to c5s or smthn and we found it all got globally a bit faster but had no idea why).

### DIFF
--- a/tools/performance/parse_perf_test.py
+++ b/tools/performance/parse_perf_test.py
@@ -45,9 +45,22 @@ def run(i: int):
     log.info('Complete in %0.2fs, using %d KB', duration, mem)
     return duration, mem
 
+
 def parse_time_output(output):
     parts = output.split(" ")
     return float(parts[0].strip()), int(parts[1].strip())
+
+
+def read_cpu_info():
+    """Return the CPU model number & number of CPUs."""
+    try:
+        with open('/proc/cpuinfo') as f:
+            models = [line[line.index(':')+2:] for line in f if line.startswith('model name')]
+        return models[0].strip(), len(models)
+    except:
+        log.exception('Failed to read CPU info')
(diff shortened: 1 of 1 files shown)


Tatskaari (approved):

tools/performance/parse_perf_test.py
```diff
@@ -45,10 +45,23 @@ def run(i: int):
...
+
 def parse_time_output(output):
     parts = output.split(" ")
     return float(parts[0].strip()), int(parts[1].strip())
 
+
+def read_cpu_info():
+    """Return the CPU model number & number of CPUs."""
+    try:
+        with open('/proc/cpuinfo') as f:
```
tiagovtristao, line 57: Do we get virtualised machines on circle ci? I was trying to see if we could get more fine grained information from it but could only find a reference in https://circleci.com/docs/2.0/configuration-reference/, `Note: If you want to confirm how much memory you have been allocated, you can check the cgroup memory hierarchy limit with grep hierarchical_memory_limit /sys/fs/cgroup/memory/memory.stat.` I wonder if the same could be retrieved for the allocated cpu.

tiagovtristao (approved):

tools/performance/parse_perf_test.py
peterebden: Yeah they are virtualised within Docker (you can also have VM runners but we don't).

It's possible the cgroup setup could be relevant too but that should just match the machine type we've requested; the info here should still be useful (e.g. Go will determine GOMAXPROCS this way, I don't think it knows about cgroups).

PR #1989, opened 2021-08-03
Upgrade pytest to 5.4.3

chrisnovakovic:
pytest 5.4.0 removes pytest's dependency on the `parser` module, which was deprecated in Python 3.9.

This doesn't change the minimum version of Python required to run pytest (>= 3.5), nor does it require any newer dependencies than the ones that are already used, so it should be a safe upgrade.


Tatskaari (approved):
Thanks Chris :D

PR #1992, opened 2021-08-06
Plz query somepath --hidden flag

samwestmoreland:
Add a --hidden flag to show internal targets as well when user does a plz query somepath.

Is this kind of what we were going for??


test/BUILD
```diff
@@ -38,6 +38,13 @@ plz_e2e_test(
     expected_output = "query_somepath_nopath_test.txt",
 )
 
+plz_e2e_test(
+    name = "query_somepath_hidden_test",
+    cmd = "HIDDEN=`plz query somepath //tools/build_langserver:build_langserver //tools/build_langserver/lsp:all --hidden | wc -l`; \
```
Tatskaari, line 43: super nit: The back tick stuff is somewhat discouraged now days. We should really be using the `$(plz query ...)` syntax instead.

Tatskaari (approved):
Nice one! A lot easier than the last ticket :)

PR #1995, opened 2021-08-06
Move all docker workloads to /mnt/ramdisk

peterebden:
Just following https://circleci.com/docs/2.0/executor-types/#ram-disks 

It's possible we'll hit memory limits. Think we should just try and see; hopefully it will reduce those poxy ETXTBUSY errors.


peterebden:
Right, looks like we run out of memory _immediately_ -> abort, abort...

PR #1998, opened 2021-08-09
Some more query completions fixes

Tatskaari:


src/query/completions.go
```diff
@@ -135,23 +142,18 @@ func getAllCompletions(config *core.Configuration, currentPackage, prefix, repoR
...
+	if binary && len(labels) == 0 && len(completions.Pkgs) == 1 {
+		return labelsInPackage(graph, completions.Pkgs[0], completions.NamePrefix, binary, test, hidden)
 	}
+	return labels
 }
 
-func printLabelsInPackage(graph *core.BuildGraph, packageName, prefix string, binary, test, hidden bool) int {
-	count := 0
+func labelsInPackage(graph *core.BuildGraph, packageName, prefix string, binary, test, hidden bool) []string {
+	ret := make([]string, 0)
```
peterebden, line 156: super nit: would probably write just `var ret string` if it's going to explicitly be empty initially

src/core/utils.go
```diff
@@ -15,13 +15,13 @@ import (
...
 // some things we need to remember this.
-var initialWorkingDir string
+var InitialWorkingDir string
 
-// initialPackage is the initial subdir of the working directory, ie. what package did we start in.
-// This is similar but not identical to initialWorkingDir.
-var initialPackage string
+// InitialPackagePath is the initial subdir of the working directory, ie. what package did we start in.
+// This is similar but not identical to InitialWorkingDir.
+var InitialPackagePath string
```
Tatskaari, line 24: Had to rename this because `InitialPackage()` already existed.

test/test_completions.txt
```diff
@@ -1,2 +1 @@
-//test/completions:all
```
peterebden, line 1: Ah nice just ran into this yesterday!

peterebden (approved):

PR #1999, opened 2021-08-10
Plz tool langserver quick fix

samwestmoreland:
I'm working on making `plz tool` not re-exec plz but it seems like that's more nice-to-have - Jon and I think the below change fixes this issue with the language server crashing so I'm thinking maybe it's worth landing this first? And then I'll carry on with the plz tool work? Just noticed the issue was being complained about on #please today!


Tatskaari (approved):
Thanks! Will be great to get this in. Looking forward to the refactor!

PR #2002, opened 2021-08-11
Properly singleflight all subinclude() calls

peterebden:
The old implementation let them race, but it turns out that happens too often. Here we enforce that each one is parsed exactly once.

I have another shot at this using x/sync/singleflight but while that helps it doesn't seem to fully deduplicate them. This is slightly more code on our side (although still hoping for generics to fix that since we could share some with the maps in src/core).

This is about 10% faster internally. I don't know if we'll see it on our synthetic benchmark, it might not be so heavy on `subinclude()`

### DIFF
--- a/src/parse/asp/interpreter.go
+++ b/src/parse/asp/interpreter.go
@@ -17,9 +17,8 @@ import (
 type interpreter struct {
 	scope           *scope
 	parser          *Parser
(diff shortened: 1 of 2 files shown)


tiagovtristao (approved):
Looks good!

samwestmoreland (approved):
Looks ok to me!

PR #2003, opened 2021-08-11
Remove duplicate map lookup

peterebden:
I think this is from an older version where it used to have an RWMutex which required it to look up again once it had the writer lock. Eventually got rid of that and so this is completely redundant.


Tatskaari (approved):
This code looked very dodgy before :S

peterebden:
nope, see above; it's a remnant from an implementation that had an RWMutex which was eventually simplified out

PR #2005, opened 2021-08-12
Re-route plz tool through plz run rather than re-exec'ing

samwestmoreland:


src/please.go
```diff
@@ -864,6 +867,61 @@ var buildFunctions = map[string]func() int{
 	},
 }
 
+// Check if tool is given as label or path and then run
+func CheckIsLabelOrPathAndRun(config *core.Configuration, _tool tool.Tool, args []string) {
```
Tatskaari, line 871: Nice! This doesn't need to be public though.

src/please.go
```diff
@@ -864,6 +867,61 @@ var buildFunctions = map[string]func() int{
...
+//		log.Warning("Sam: Code path 1")
+		if !filepath.IsAbs(target) {
+			t, err := core.LookBuildPath(target, config)
+			if err != nil {
+				log.Fatalf("%s", err)
+			}
+			target = t
+		}
+		// Hopefully we have an absolute path now, so let's run it.
+		if success, state := runBuild(label, true, false, false); success {
```
Tatskaari, line 888: We don't need to run a build if the tool isn't a build label. We can just do `tool.Run(config, "build_langserver", opts.Tool.Args.Args.AsStrings())`. `tool.Run()` can assume the tool is either an abs path to a binary, or it's a file on the path. IF we move the above logic out into `too.Run()`, that should be fine.

src/please.go
```diff
@@ -864,6 +867,61 @@ var buildFunctions = map[string]func() int{
 	},
 }
 
+// Check if tool is given as label or path and then run
+func CheckIsLabelOrPathAndRun(config *core.Configuration, _tool tool.Tool, args []string) {
+//	log.Warning("Sam: Hi I'm in CheckIsLabelOrPathAndRun()")
+	tools := tool.MatchingTools(config, string(_tool))
+//	log.Warning("Sam: tools=%s", tools)
+	target := fs.ExpandHomePath(tools[tool.AllToolNames(config, string(_tool))[0]])
+	label := core.ParseBuildLabels([]string{target})
```
Tatskaari, line 876: I think this sort of thing can be handled by `tool.Run()`

src/please.go
```diff
@@ -864,6 +867,61 @@ var buildFunctions = map[string]func() int{
...
+//	log.Warning("Sam: target=%s", target)
+	if !core.LooksLikeABuildLabel(target) {
+//		log.Warning("Sam: Code path 1")
+		if !filepath.IsAbs(target) {
```
Tatskaari, line 880: This can also probably be handled by `tool.Run()`

src/please.go
```diff
@@ -864,6 +867,61 @@ var buildFunctions = map[string]func() int{
...
+	//	err := syscall.Exec(plz, args, os.Environ())
+	//	log.Fatalf("Failed to exec %s run %s: %s", plz, target, err) // Always a failure, exec never returns.
+
+	if success, state := runBuild(label, true, false, false); success {
+		var dir string
+		if opts.Run.InWD {
+			dir = originalWorkingDirectory
+		}
+		annotatedOutputLabels := core.AnnotateLabels(label)
+		run.Run(state, annotatedOutputLabels[0], opts.Tool.Args.Args.AsStrings(), opts.Run.Remote, opts.Run.Env, opts.Run.InTempDir, dir, opts.Run.Cmd)
```
Tatskaari, line 921: We want to pass options from the run sub-command. They won't be populated because we're running the `tool` sub-command here. That's not a problem as we don't actually want to run tools like that. This can probably just be something like:
```
run.Run(state, annotatedOutputLabels[0], opts.Tool.Args.Args.AsStrings(), false,  nil, false, "", "")
```

Tatskaari (commented):
Nice one! Looks like it's getting there :)

src/please.go
```diff
@@ -614,7 +614,7 @@ var buildFunctions = map[string]func() int{
 		return toExitCode(help.Help(string(opts.Help.Args.Topic)), nil)
 	},
 	"tool": func() int {
-		tool.Run(config, opts.Tool.Args.Tool, opts.Tool.Args.Args.AsStrings())
+		runTool(opts.Tool.Args.Tool)
```
Tatskaari, line 617: can we make this `return runTool(...)` and return a non-zero exit code when the build fails?

src/tool/tool.go
```diff
@@ -70,17 +57,26 @@ func matchingTools(config *core.Configuration, prefix string) map[string]string
...
 
-// allToolNames returns the names of all available tools.
-func allToolNames(config *core.Configuration, prefix string) []string {
+func MatchingTool(config *core.Configuration, tool string) (string, bool) {
+	tool, ok := knownTools(config)[tool]
+	return tool, ok
+}
+
+// AllToolNames returns the names of all available tools.
+func AllToolNames(config *core.Configuration, prefix string) []string {
```
Tatskaari, line 79: Does this need to be exported?

Tatskaari (approved):
Looks good! Got a couple minor comments but otherwise ready to land :D

src/tool/tool.go
samwestmoreland: You're right I don't think it does. It was being used for the unknown tool error message before so I've moved the error message back into tool.go

PR #2006, opened 2021-08-12
Remove all the modified targets and rescan stuff

peterebden:
I don't think this is necessary any more. With the newer dependency setup they should be triggered as needed.

Haven't been able to test much locally due to my machine basically just refusing to do things (Eset really is a horrible piece of shit...) but we'll see what CircleCI says.

### DIFF
--- a/src/parse/parse_step_test.go
+++ b/src/parse/parse_step_test.go
@@ -103,7 +103,7 @@ func TestAddDepRescan(t *testing.T) {
 	assertPendingBuilds(t, state) // Note that the earlier call to assertPendingBuilds cleared it.
 
 	// Now running this should activate it
-	rescanDeps(state, map[*core.BuildTarget]struct{}{target1: {}})
+	//rescanDeps(state, map[*core.BuildTarget]struct{}{target1: {}})
 	time.Sleep(time.Millisecond * 100)
 
 	assertPendingBuilds(t, state, "//package1:target4")
--- a/src/parse/parse_step_test.go
+++ b/src/parse/parse_step_test.go
@@ -103,7 +103,7 @@ func TestAddDepRescan(t *testing.T) {
 	assertPendingBuilds(t, state) // Note that the earlier call to assertPendingBuilds cleared it.
 
 	// Now running this should activate it
-	//rescanDeps(state, map[*core.BuildTarget]struct{}{target1: {}})
+	// rescanDeps(state, map[*core.BuildTarget]struct{}{target1: {}})
 	time.Sleep(time.Millisecond * 100)
 
 	assertPendingBuilds(t, state, "//package1:target4")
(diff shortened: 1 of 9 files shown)


samwestmoreland (approved):

src/parse/parse_step_test.go
```diff
@@ -98,12 +98,12 @@ func TestAddDepRescan(t *testing.T) {
...
 	// Fake test: calling this now should have no effect because rescan is not true.
-	state.QueueTarget(buildLabel("//package1:target1"), core.OriginalTarget, false, false)
+	state.QueueTarget(buildLabel("//package1:target1"), core.OriginalTarget, false)
 	assertPendingParses(t, state)
 	assertPendingBuilds(t, state) // Note that the earlier call to assertPendingBuilds cleared it.
 
 	// Now running this should activate it
-	rescanDeps(state, map[*core.BuildTarget]struct{}{target1: {}})
+	// rescanDeps(state, map[*core.BuildTarget]struct{}{target1: {}})
 	time.Sleep(time.Millisecond * 100)
```
tiagovtristao, line 107: Is any of this needed anymore?

tiagovtristao:
Where can I find the newer dependency setup that makes this obsolete?

tiagovtristao (approved):

Tatskaari:
> Where can I find the newer dependency setup that makes this obsolete?

Yeah I'm not sure either. @peterebden are you referring to the queue asyc stuff? Did we used to mark them as modified and queue them synchronously later on or something?  

This all looks sensible and it doesn't seem like it should be necessary seeing as the `add_deps()` and `add_data()` functions queue up the deps and data. I guess the old dependency setup didn't do that.

Tatskaari (approved):
Looks sensible but I'm not sure I understand the history here. Seems like you know what you're doing though.

peterebden:
> Where can I find the newer dependency setup that makes this obsolete?

It's really #1102 that did it, in this loop: https://github.com/thought-machine/please/blob/aebd08e292eb8ff5731dd48931b3087735672d6b/src/core/state.go#L932
Essentially we loop around all the target's dependencies until they all become built, then we can queue it. That means any newly added dependency is automatically picked up and we don't have to explicitly trigger it.

peterebden:
> I am wondering how we make sure that the targets are built in the right order without the rescan thing? Say A has a post-build that adds C as a dep to B. We queue up A, then B, then C because that's a valid order until the post-build runs. Once we add the dep from B to C, this becomes invalid. Do we handle this case somehow?

I don't think we do handle it, and never did. In that case B should have a dep on A in order for it to work out properly. Maybe we could do something to detect when this has been violated?

src/parse/parse_step_test.go
peterebden: Fair point. I've removed the two skipped test cases - there is no plan to bring them back again (I don't think what they are doing is a good enough simulation of reality). TBH I'm a little tempted to drop the whole thing - the rest is a bit similar - but at least they are working for now.

PR #2011, opened 2021-08-17
Set local to True for system_library

samwestmoreland:


rules/misc_rules.build_defs
```diff
@@ -362,6 +362,7 @@ def system_library(name:str, srcs:list, deps:list=None, hashes:list=None,
         visibility = visibility,
         test_only = test_only,
         sandbox = False,
+        local = True,
```
Tatskaari, line 365: I wonder if we can do this more generally by setting local to true automatically in the go code if `system_srcs` are set?

Tatskaari (approved):

Tatskaari:
Thanks Sam! Looks good to me.

PR #2013, opened 2021-08-17
Fix completions on packages that are also prefixes for other packages

Tatskaari:
`plz query completions //src/plz` was matching `//src/plz` exactly and printing its labels. Now it also matches `//src/plzinit` and so completes correctly.


samwestmoreland (approved):
Nice!

samwestmoreland:
> Nice!

Apart from the failing test 🙃

Tatskaari:
> > Nice!
> 
> Apart from the failing test

I have no idea what you're talking about ;P

PR #2014, opened 2021-08-17
Tag v16.5.0

Tatskaari:


samwestmoreland (approved):
👍🏽

ChangeLog
```diff
@@ -1,3 +1,13 @@
+Version 16.5.0
+--------------
+    * Singleflight all subincludes removing any potential for lockups #2002
```
peterebden, line 3: Bit late but this isn't quite accurate; the singleflighting is a performance enhancement, I don't think it affects potential lockups. Although I'd be ecstatic if I found out that it did :D

Issue #2007, opened 2021-08-12
Invalid linker options on macOS with pip_library


xinkecf35:
# Problem
When asking Please to fetch a Pip package for which no pre built binaries are available, pip fails to generate the C extension due to the lack of the `--build-id=none` flag on macOS and a few other compiler and linker flags not being present with `clang`

## Reproduction
- on macOS, use the following `pip_library` rule
```Starlark
pip_library(
    name = "cx-Oracle",
    version = "8.2.1"
)
```
- Build said rule will result with following error: 
```
  ld: unknown option: --build-id=none
  clang: error: linker command failed with exit code 1 (use -v to see invocation)
  error: command 'clang' failed with exit status 1
  ----------------------------------------
  ERROR: Failed building wheel for cx-Oracle
  Running setup.py clean for cx-Oracle
 ```
 
 Expected output is that this succeeds with the `pip_library` being successfully fetched and built.  Said rule was working on 15.17.1

Regression seems to have been introduced with PR #1909 and seems to be similar to a previous encountered issue as resolved  by [this commit](https://github.com/thought-machine/please/commit/6cd64a6bbe868c2e09183827492af1f58f53e3f8)

Reproduced on both macOS Catalina 10.15.7 and macOS 11.5.1 with clang versions 12.0.0 and 12.0.5 respectively.

## Possible resolution:

Like with the `cc_library` rule, add a condition to omit the CFLAGS and LDFLAGs from the pip command if operating system is macOS/darwin.

PR #2015, opened 2021-08-18
Added condition to set no linker flags for macOS for pip_cmd

xinkecf35:
This PR seeks to address #2007 by primarily adding an additional check of the OS for darwin and for the time being removes the CFLAGS and LD_FLAGS just for macOS specifically.

Feedback would be welcome on whether to look into alternative flags that might be used to achieve a similar result as it does on Linux.

Also, feedback welcome on whether additional tests are needed to capture the issue introduced in #2007

### DIFF
--- a/rules/python_rules.build_defs
+++ b/rules/python_rules.build_defs
@@ -373,9 +373,12 @@ def pip_library(name:str, version:str, hashes:list=None, package_name:str=None,
 
     pip_tool = "$TOOLS_PIP" if CONFIG.PIP_TOOL else '$TOOLS_PYTHON -m pip'
 
+    # Unset these compiler and linker flags as macOS doesn't have all these options.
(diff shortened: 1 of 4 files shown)


Tatskaari:
Isn't it just the `--build-id=none` that's a GNU flag? Do the others not work? 

As for tests, it might be nice to add a `pip_library()` rule for `cx-Oracle` and add a test under `//test/python_rules` to make sure we can import it. We do the same for tensorflow.

xinkecf35:
> Amazing! Thanks for investigating this issue. Appreciate the effort you've put in.
> 
> Isn't it just the `--build-id=none` that's a GNU flag? Do the others not work?
> 
> As for tests, it might be nice to add a `pip_library()` rule for `cx-Oracle` and add a test under `//test/python_rules` to make sure we can import it. We do the same for tensorflow.

Sure thing. I'll take a look at including `cx-Oracle` or like library which doesn't have pre-built wheels for macOS as a test.

As for the flags, I'll take a look at least including `-g0` back in. I'll be honest and state I don't normally interact with clang and was having trouble tracking info down. I finally found the correct section of the man page so I should find out shortly.

Tatskaari (approved):
Looks good to me!

PR #2017, opened 2021-08-19
Use the os.SameFile for cross platform support

hartmannr76:
In order to help enable better cross platform support (i.e. Windows) I'm trying to figure out what may already be built into existing libraries to enable that.

inode is not a concept in Windows, and I noticed the `os.SameFile` call seems it should be doing the same thing
```
SameFile reports whether fi1 and fi2 describe the same file. For example, on Unix this means that the device and inode fields of the two underlying structures are identical; on other systems the decision may be based on the path names. SameFile only applies to results returned by this package's Stat. It returns false in other cases.
```

It probably makes more sense to use this instead since it defers OS level checks to the library

### DIFF
--- a/src/fs/fs.go
+++ b/src/fs/fs.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
-	"syscall"
 
 	"github.com/thought-machine/please/src/process"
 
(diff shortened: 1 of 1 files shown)


Tatskaari (approved):
Thanks! I had no idea the go SDK had these now. :D

PR #2018, opened 2021-08-20
Cleanup output code a bit

peterebden:
I have some other ideas in this space but this is enough for one PR.

Basically this encapsulates the interactive / plain display logic in a little interface, and handles all the updates on one goroutine (previously there were two with some slightly shonky data sharing between them).
It simplifies the lifetime of it by having it terminate when the results channel is closed & removing the context (so one less thing to keep track of).

### DIFF
--- a/src/output/interactive_display.go
+++ b/src/output/interactive_display.go
@@ -2,7 +2,6 @@ package output
 
 import (
 	"bytes"
-	"context"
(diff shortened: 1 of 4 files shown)


Tatskaari (approved):
Looks good to me! Thanks Pebers.

PR #2020, opened 2021-08-20
add ExcludeGlob from coverage

namtzigla:
This adds a new option under `[coverage]` that will exclude entire paths from coverage computations.

It is especially helpful when you have java generated code that can't be otherwise excluded.

Thanks


namtzigla:
@Tatskaari you can keep `ExcludeExtension` for backward compatibility until the next major release or longer

Tatskaari (approved):
Nice one! I think I'll just quickly update this to use out double star globber then land it.

PR #2022, opened 2021-08-23
More output refactoring

peterebden:
This does four main things:
 - Encapsulates more of the logic about tracking results into its own type, removing some of the weirdness about passing around pointers to slices
 - Moves the trace writer logic and gets rid of the thing where it can be created but not do anything when no file is given
 - Breaks up the local and remote actions as passed to the output writers.
 - And a small fix to the previous PR where it sometimes eats the line above.

Again, this isn't the main thing I want to get to, but another good step along the way (think I should be ready after this).


tiagovtristao (approved):
Nice refactoring.

samwestmoreland (approved):
Nice!

PR #2023, opened 2021-08-24
Use fs.Match() instead

Tatskaari:


tiagovtristao (commented):
Looks good. The main change here seems to be the addition of `**` to be parseable?

Tatskaari:
> Looks good. The main change here seems to be the addition of `**` to be parseable?

yeah exactly. It uses the same glob code that `glob()` uses.

PR #2024, opened 2021-08-24
Apply release tags based on the go tool

Tatskaari:
Before we were applying these based on the version of go we used to compile `please_go` rather than the version of `go.gotool`


tools/please_go/install/install.go
```diff
@@ -35,13 +37,24 @@ type PleaseGoInstall struct {
...
+
+	version, err := install.tc.GoMinorVersion()
+	if err != nil {
+		log.Fatalf("failed to determine go version: %v", err)
+	}
+
+	install.buildContext.ReleaseTags = []string{}
+	for i := 1; i <= version; i++ {
+		install.buildContext.ReleaseTags = append(install.buildContext.ReleaseTags, "go1."+strconv.Itoa(i))
+	}
```
tiagovtristao, line 52: I'm sure this will bite us when go2 is released, but I guess this is true in other places.

tools/please_go/install/toolchain/toolchain_test.go
```diff
@@ -0,0 +1,16 @@
...
+	"path/filepath"
+	"testing"
+)
+
+func TestGetVersion(t *testing.T) {
+	tc := &Toolchain{GoTool: filepath.Join(os.Getenv("DATA"), "bin/go")}
+
+	ver, err := tc.GoMinorVersion()
+	require.NoError(t, err)
+	require.Equal(t, 17, ver)
```
tiagovtristao, line 15: This will fail with every toolchain upgrade.

tiagovtristao (approved):
Looks good, just a small comment.

tools/please_go/install/toolchain/toolchain_test.go
Tatskaari: Yeah but that's fairly rare and easy to fix :)

tools/please_go/install/install.go
Tatskaari: They're not going to release a 2.0 version unless something very drastic happens. They're a firm believer that fracturing the ecosystem like that is a bad idea.

PR #2026, opened 2021-08-25
Set GOMAXPROCS at startup

peterebden:
We've had some good results from this internally. Would be interested to see what effect it has on our benchmark (where it should indeed set the limit quite differently to the default)


samwestmoreland (approved):
Nice! How did we set the procs before? Or did we not?

peterebden:
We just followed the default Go behaviour (to honour the env var GOMAXPROCS and if not set it uses the number of current logical CPUs). This package is aware of cgroups on Linux so if running within a container will limit itself to the number of accessible CPUs rather than the total number - e.g. in our perf test, we should have 8 vCPUs available, but the machines generally appear to have 36 so we would likely overprovision OS threads.

samwestmoreland:
Ah I see. Thanks!

tiagovtristao (approved):
Looking forward to seeing the results of this!

Issue #2027, opened 2021-08-25
LSP typed nil bug


perrydunn:
Forgive me for I [have sinned](https://github.com/thought-machine/please/pull/1962)...

That commit introduced a typed nil bug by virtue of the `resp != nil` [comparison](https://github.com/thought-machine/please/blob/master/tools/build_langserver/lsp/lsp.go#L60) in `Handle`.

Discovered by inspecting my nvim lsp client logs, eg
[...]

PR #2029, opened 2021-08-26
Fix LSP responding when not intended

perrydunn:
https://github.com/thought-machine/please/issues/2027


Tatskaari (approved):
Thanks Perry! Appreciate it.

PR #2030, opened 2021-08-28
fix the install script to download other architectures then x86_64

namtzigla:
the install script had an amd64 hardcoded


tools/misc/get_plz.sh
```diff
@@ -16,7 +16,17 @@ else
     exit 1
 fi
 
-PLEASE_URL="https://get.please.build/${GOOS}_amd64/${VERSION}/please_${VERSION}.tar.gz"
+ARCH=`uname -m`
+if [ "$ARCH" = "x86_64" ]; then 
+    ARCH="amd64"
+elif [ "$ARCH" = "arm64" ]; then
+    ARCH="amd64"
```
peterebden, line 23: Should be `arm64`, right?

peterebden (commented):
Thanks - not surprised there are some of these scattered around (back when this was written ARM wasn't very relevant to us... how times change!)

tools/misc/get_plz.sh
namtzigla: yep

namtzigla:
@peterebden @Tatskaari I think is this ready

peterebden (approved):
Thanks!

PR #2033, opened 2021-08-30
Re-implement cgo part of sandbox in Go

peterebden:
x/sys/unix has the equivalent functions (think these might be fairly recent, I had to update it to get them).

Simplifies things a bit, and removes the need for having GCC installed during build.


samwestmoreland (approved):

Tatskaari (approved):
LGTM! I had a go at doing this but couldn't find the right library.

PR #2034, opened 2021-08-31
Have this benchmark picked up automatically

peterebden:
Little bit of fiddling needed too since it's the first one that uses `data`.


Tatskaari (approved):
LGTM

PR #2036, opened 2021-08-31
Implement embed config for please_go install

Tatskaari:


tiagovtristao (approved):
Looks good to me. There's a build error might worth checking though as I'm not sure if it's related

PR #2041, opened 2021-09-02
sandbox: ignore non-existent directories in `SANDBOX_DIRS`

### LINKED ISSUE #2040: Sandbox exits fatally if masked directory doesn't exist
A comment in the sandbox code suggests that it's fine for a masked directory specified in the Please configuration file not to exist (in which case the entry should simply be ignored):

https://github.com/thought-machine/please/blob/beefa578e0b5de19f1e7ffd60944bf80c85af5c1/tools/sandbox/sandbox.c#L119-L127

but in fact it seems to be a fatal error for a non-existent directory to be specified as a masked directory:

```
$ grep -FA3 '[sandbox]' .plzconfig
[sandbox]
build = true
test = true
dir = /doesnt/exist

$ plz build [...] --shell
Temp directories prepared, total time 350ms:
  [...]

mount tmpfs: No such file or directory
```

chrisnovakovic:
`mount()` sets `errno` to `ENOENT` when the target doesn't exist, which causes the sandbox to error fatally when it is asked to mask a directory that doesn't exist on the host. Instead, output the existing "Not mounting over <target> since it isn't a directory" warning that is shown when the given target is a non-directory file, and continue.

Fixes #2040.

### DIFF
--- a/tools/sandbox/sandbox.c
+++ b/tools/sandbox/sandbox.c
@@ -117,7 +117,7 @@ int mount_tmp(char** argv0) {
       char *token = strtok(dirs, ",");
       while(token) {
         if (mount("tmpfs", token, "tmpfs", flags | MS_RDONLY, NULL) != 0) {
-          if (errno == ENOTDIR) {
+          if (errno == ENOENT || errno == ENOTDIR) {
             // This isn't fatal, it's OK for them not to exist (in that case we just have nothing to sandbox).
             fprintf(stderr, "Not mounting over %s since it isn't a directory\n", token);
           } else {


chrisnovakovic:
Before:

```
$ grep -FA3 '[sandbox]' .plzconfig
[sandbox]
build = true
test = true
dir = /doesnt/exist

$ plz build [...] --shell
Temp directories prepared, total time 350ms:
  [...]

mount tmpfs: No such file or directory
```

After:

```
$ plz build [...] --shell
Temp directories prepared, total time 260ms:
  [...]

Not mounting over /doesnt/exist since it isn't a directory
bash-5.0$
```

PR #2043, opened 2021-09-03
Use exec instead of a subprocess in sandbox

peterebden:
Came across this while I was looking into its performance; this doesn't make a lot of difference to that but I think it's better not to create an additional process. Also has some mild advantages like signal forwarding and propagating the correct exit code.

### DIFF
--- a/src/sandbox/sandbox_linux.go
+++ b/src/sandbox/sandbox_linux.go
@@ -26,11 +26,12 @@ func Sandbox(args []string) error {
 	if len(args) < 2 {
 		return fmt.Errorf("incorrect number of args to call plz sandbox")
 	}
-	cmd := exec.Command(args[0], args[1:]...)
 
-	cmd.Stdout = os.Stdout
-	cmd.Stdin = os.Stdin
-	cmd.Stderr = os.Stderr
+	env := os.Environ()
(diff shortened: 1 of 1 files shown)


tiagovtristao (approved):
The change makes sense to me, thanks.

src/sandbox/sandbox_linux.go
```diff
@@ -65,22 +66,19 @@ func Sandbox(args []string) error {
 			return fmt.Errorf("Failed to bring loopback interface up: %s", err)
 		}
 	}
-
-	return cmd.Run()
+	err = syscall.Exec(cmd, args, env)
+	return fmt.Errorf("Failed to exec %s: %s", cmd, err)
```
PeterBocan, line 70: do you want to always return an error?

tiagovtristao: An error will only be returned if it failed to execute `cmd`. I think `syscall.Exec` uses `execve` under the hood which replaces the current process, so this function would never return on success.

peterebden: Indeed :D

samwestmoreland (approved):
Looks good to me!

PR #2045, opened 2021-09-03
Allow `Python.DefaultInterpreter` to be a build target name

chrisnovakovic:
Allow the path to the default Python interpreter to be a Please build target name, in which case that target will be built and the full path to its output binary will be embedded in PEX files generated by `python_binary` and `python_test` targets. This allows `python_binary` and `python_test` targets to depend on a Python interpreter that is itself built in the repository without first needing to install the interpreter externally.

### DIFF
--- a/rules/python_rules.build_defs
+++ b/rules/python_rules.build_defs
@@ -37,9 +37,8 @@ def python_library(name:str, srcs:list=[], resources:list=[], deps:list=[], visi
                        (the most obvious reason not is when it contains .so modules).
                        See python_binary for more information.
       labels (list): Labels to apply to this rule.
-      interpreter (str): The Python interpreter to use. Defaults to the config setting
-                         which is normally just 'python', but could be 'python3' or
-                         'pypy' or whatever.
+      interpreter (str): The Python interpreter to use. Defaults to the value of the
+                         Python.DefaultInterpreter setting.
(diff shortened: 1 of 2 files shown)


rules/python_rules.build_defs
```diff
@@ -122,14 +121,14 @@ def python_binary(name:str, main:str, srcs:list=[], resources:list=[], out:str=N
...
-                         which is normally just 'python', but could be 'python3' or
-                         'pypy' or whatever.
+      interpreter (str): The Python interpreter to use. Defaults to the value of the
+                         Python.DefaultInterpreter setting.
       shebang (str): Exact shebang to apply to the generated file. By default we will
                      determine something appropriate for the given interpreter.
       labels (list): Labels to apply to this rule.
     """
-    shebang = shebang or interpreter or CONFIG.DEFAULT_PYTHON_INTERPRETER
+    interpreter = interpreter or CONFIG.DEFAULT_PYTHON_INTERPRETER
```
Tatskaari, line 130: We can just do `interpreter:str=CONFIG.DEFAULT_PYTHON_INTERPRETER` in the function param btw

Tatskaari (approved):

rules/python_rules.build_defs
chrisnovakovic: Good spot, thanks!

PR #2046, opened 2021-09-03
Handle relative paths for config.Please.Location and some clean up

tiagovtristao:
- Handle relative paths for `config.Please.Location`
- Expand `config.Please.Location` to its full path during read, so it's done only once.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,7 +148,7 @@ jobs:
            command: tar -xzf /tmp/workspace/darwin_amd64/please_*.tar.gz
        - run:
            name: Cross-compile
-           command: ./please/please build -p --profile ci --arch darwin_arm64 -o go.cgoenabled:1 -o go.cgocctool:clang -o "go.cflags:-target arm64-apple-macos11" //package:release_files
+           command: ./please/please build -p --profile ci --arch darwin_arm64 -o go.cgoenabled:1 -o go.cgocctool:clang -o "go.cflags:-target arm64-apple-macos11" -o please.location:./please //package:release_files
        - persist_to_workspace:
            root: plz-out/pkg
            paths:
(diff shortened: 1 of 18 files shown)


.circleci/config.yml
```diff
@@ -175,7 +175,7 @@ jobs:
            command: tar -xzf /tmp/workspace/linux_amd64/please_*.tar.gz
        - run:
            name: Cross-compile
-           command: ./please/please build -p --profile ci --arch freebsd_amd64 //package:release_files
+           command: ./please/please build -p --profile ci --arch freebsd_amd64 -o please.location:~/please/please //package:release_files
```
peterebden, line 178: Hmmm, curious why these need to change?

tiagovtristao: I've set the please location config value to `./plz-out/please` in `.plzconfig` to aid development. At the moment we have to set the location to the full path in `.plzconfig.local` which is a bit annoying.

.github/workflows/golangci-lint.yml
```diff
@@ -15,9 +15,9 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
-          go-version: '^1.16'
+          go-version: '^1.17'
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          version: v1.41.1
+          version: v1.42.0
```
tiagovtristao, line 22: Had to bump this for the new testing method `Setenv` in 1.17

src/fs/home.go
```diff
@@ -7,12 +7,11 @@ import (
 	"github.com/peterebden/go-deferred-regex"
 )
 
-var home = os.Getenv("HOME")
 var homeRex = deferredregex.DeferredRegex{Re: "(?:^|:)(~(?:[/:]|$))"}
 
 // ExpandHomePath expands all prefixes of ~ without a user specifier to $HOME.
 func ExpandHomePath(path string) string {
-	return ExpandHomePathTo(path, home)
+	return ExpandHomePathTo(path, os.Getenv("HOME"))
```
tiagovtristao, line 14: Moved this here otherwise the value would be stale. I faced this issue when writing tests, but it seems also reasonable otherwise.

samwestmoreland (approved):
Nice! What's going on with the `//go:build ...` lines? Is `//go:build` the same as `// +build`?

samwestmoreland:
Oh it looks like `//go:build` is what they use on the go website so I suppose it's preferred

Tatskaari (approved):

PR #2047, opened 2021-09-03
Introduce a DeferredRegexp type

peterebden:
This is basically a cleanup of #2044 so we don't have to explicitly manage a sync.Once per regexp (or group of them).

If this looks like a good idea I think it might be worth doing a separate module for it (there are a few other cases I can't use it for since this is sitting in `core` rn)


Tatskaari (approved):
Eh... I've seen better.

PR #2060, opened 2021-09-09
Save outputs from plz exec

Tatskaari:


src/please.go
```diff
@@ -476,7 +480,24 @@ var buildFunctions = map[string]func() int{
...
+			to := filepath.Join(opts.Exec.Output.OutputPath, out)
+
+			if err := fs.EnsureDir(to); err != nil {
+				log.Fatalf("%v", err)
+			}
+
+			if err := fs.RecursiveLink(from, to, 0644); err != nil {
+				log.Fatalf("failed to move output: %v", err)
+			}
+		}
```
tiagovtristao, line 499: Can't decide whether it makes sense for this to be here or inside `exec.Exec`. Perhaps here is the right choice.

tiagovtristao (approved):
LGTM, thanks.

src/please.go
Tatskaari: Yeah I know what you mean. It feels like it could be merged with the logic from `plz test` but it's hard to tell until we actually get `plz test` to use this.

PR #2068, opened 2021-09-15
Use distlib and packaging to get wheel urls

samwestmoreland:


.plzconfig
```diff
@@ -83,8 +83,9 @@ accept = MPL-2.0
 accept = LGPL
 accept = Artistic License
 accept = ISC
-# Not really a licence, but Bazel projects commonly describe things this way.
+# Not really a license, but Bazel projects commonly describe things this way.
```
peterebden, line 86: This was correct before - in British English the noun is `licence`.

(It's very common to see `license` because it's more American-dominated where the noun and verb are the same. I think we aim to accept both in BUILD files.)

src/core/config.go
```diff
@@ -354,6 +354,7 @@ func DefaultConfiguration() *Configuration {
 	config.Go.PleaseGoTool = "//_please:please_go"
 	config.Go.EmbedTool = "//_please:please_go_embed"
 	config.Python.PexTool = "//_please:please_pex"
+	config.Python.WheelResolverTool = "//_please:wheel_resolver"
```
peterebden, line 357: I wonder if we should look at moving the Python stuff to a plugin and doing this there rather than adding to core plz at this point?

peterebden (commented):
I'm only seeing a couple of changed files - has some of this got lost maybe?

.plzconfig
samwestmoreland: Ah fair enough 🤣
I got confused because all the licence names above are spelt with an 's'

src/core/config.go
samwestmoreland: Oh yes good point. I'll speak to @Tatskaari about how best to go about it.

PR #2069, opened 2021-09-15
plz debug

tiagovtristao:
This PR implements the logic underpinning the new `plz debug` command.

- Support for Go, via `go_binary` and `go_test`, has been added. At the moment Delve needs to be available in the host machine and/or set by the user which isn't ideal. We hope this constraint/limitation can be solved once language plugins are finished. 
- Support for Python and other languages will be added next.

Note: Some refactoring and homogenisation around build env vars was also done.


src/core/build_env.go
```diff
@@ -161,51 +148,28 @@ func withUserProvidedEnv(target *BuildTarget, env BuildEnv) BuildEnv {
...
-	env := TargetEnvironment(state, target)
+	env := RuntimeEnvironment(state, target, path.IsAbs(testDir), true)
 	resultsFile := path.Join(testDir, TestResultsFile)
-	abs := path.IsAbs(testDir)
 
 	env = append(env,
 		"TEST_DIR="+testDir,
 		"TMP_DIR="+testDir,
 		"TMPDIR="+testDir,
+		"HOME="+testDir,
```
Tatskaari, line 158: Are these not part of the runtime env?

src/core/build_env.go
```diff
@@ -220,38 +184,87 @@ func TestEnvironment(state *BuildState, target *BuildTarget, testDir string) Bui
...
-			env = append(env, "TOOLS_"+strings.ToUpper(name)+"="+strings.Join(toolPaths(state, tools, true), " "))
+	return withUserProvidedEnv(target, env)
+}
+
+// ExecEnvironment creates the environment variables for a `plz exec`.
+func ExecEnvironment(state *BuildState, target *BuildTarget, execDir string) BuildEnv {
+	env := append(RuntimeEnvironment(state, target, true, true),
+		"TMP_DIR="+execDir,
+		"TMPDIR="+execDir,
+		"HOME="+execDir,
```
Tatskaari, line 206: Sam here

src/core/build_target.go
```diff
@@ -1218,6 +1230,14 @@ func (target *BuildTarget) AddTestTool(tool BuildInput) {
 	}
 }
 
+// AddDebugTool adds a new debug tool to the target.
+func (target *BuildTarget) AddDebugTool(tool BuildInput) {
```
Tatskaari, line 1234: We might need to think about whether this stuff should contribute to the build hash.

Tatskaari (commented):
Looking good! A lot of the env code seems cleaner too!

src/core/build_env.go
tiagovtristao: They are part of a sandboxed runtime environment. At least that's the approach I took which allows me to use `RuntimeEnvironment` for `plz run`

src/core/build_target.go
tiagovtristao: Ya, let's catch up on this

src/core/build_env.go
```diff
@@ -220,38 +184,90 @@ func TestEnvironment(state *BuildState, target *BuildTarget, testDir string) Bui
...
+		"TOOLS=" + strings.Join(toolPaths(state, tools, abs), " "),
+	}
+	if len(tools) == 1 {
+		env = append(env, "TOOL="+toolPath(state, tools[0], abs))
+	}
+	for name, tools := range namedTools {
+		env = append(env, "TOOLS_"+strings.ToUpper(name)+"="+strings.Join(toolPaths(state, tools, abs), " "))
+	}
+	return env
+}
```
tiagovtristao, line 266: This tries to homogenise how tool-related vars are generated for different environments. Both build and test environment would construct both `TOOLS` and `TOOL` differently via a mixture of just relying on unnamed tools or all tools. This approach makes more sense to me and is also aligned with how other vars are currently being handles (i.e. `SRCS`). This might however break build for some users?

samwestmoreland (approved):

src/core/state.go
```diff
@@ -43,6 +43,12 @@ type TestTask struct {
 	Run   int
 }
 
+// Debug is the type for debugging a target
+type Debug struct {
+	Debugger string
```
Tatskaari, line 48: How do we expect this to be used? Will the build rules know about each debugger that we might configure, and how to invoke them? It doesn't seem like we have a use case for this yet so maybe we shouldn't add it until we have a better idea of that use case?

src/build/incrementality_test.go
```diff
@@ -70,6 +70,12 @@ var KnownFields = map[string]bool{
 	"Test.Flakiness": true,
 	"Test.Results":   true, // Recall that unsuccessful test results aren't cached...
 
+	// Debug fields don't contribute to any hash
+	"Debug":            true,
+	"Debug.Command":    true,
+	"Debug.tools":      true,
+	"Debug.namedTools": true,
```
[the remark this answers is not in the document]
tiagovtristao: I made it consistent with `Test` above. I reckon there must have been a reason to keep these fields private?!

rules/go_rules.build_defs
```diff
@@ -568,19 +569,26 @@ def go_binary(name:str, srcs:list=[], resources:list=None, asm_srcs:list=[], out
         import_path="main",
     )
     cmds, tools = _go_binary_cmds(static=static, definitions=definitions)
+
+    # TODO(tiagovtristao): Find a way to include the original deps' sources too
+    debug_cmd, debug_data, debug_tools = _debug_cmd("./$OUT", "", _srcs)
+    _data = data + debug_data if data else debug_data
```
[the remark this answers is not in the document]
tiagovtristao: Ya, I guess tests relying on `$DATA` to be unchanged, even if it just has now more items, might be problematic.

src/core/state.go
tiagovtristao: We (the user) don't have to configure anything. The build rule owner however might want to provide different ways (i.e. debuggers) to debug the targets. Ideally, running, say, `plz help go_binary` the comment would include the keys to different debuggers that might be supported. If only one, the default one, is supported no need to make this visible.

If I'm authoring `go_binary` to support both Delve and GDB, I should take into account `CONFIG.DEBUG.DEBUGGER` that the use might have chosen.

rules/builtins.build_defs
```diff
@@ -2,16 +2,16 @@
 
 
 # Do not change the order of arguments to this function without updating the iota in targets.go to match it.
-def build_rule(name:str, cmd:str|dict='', test_cmd:str|dict='', srcs:list|dict=None, data:list|dict=None, outs:list|dict=None,
-               deps:list=None, exported_deps:list=None, secrets:list|dict=None, tools:str|list|dict=None, test_tools:str|list|dict=None,
-               labels:list=None, visibility:list=CONFIG.DEFAULT_VISIBILITY, hashes:list=None, binary:bool=False, test:bool=False,
+def build_rule(name:str, cmd:str|dict='', test_cmd:str|dict='', debug_cmd:str='', srcs:list|dict=None, data:list|dict=None,
```
tiagovtristao, line 5: Added `debug_cmd`, `debug_data` and `debug_tools`

src/build/incrementality_test.go
```diff
@@ -70,9 +70,17 @@ var KnownFields = map[string]bool{
...
+	"Debug.Command":    true,
+	"Debug.data":       true,
+	"Debug.namedData":  true,
+	"Debug.tools":      true,
+	"Debug.namedTools": true,
+
 	// These only contribute to the runtime hash, not at build time.
 	"Data":              true,
-	"namedData":         true,
+	"NamedData":         true,
```
tiagovtristao, line 83: For consistency

src/core/build_env.go
```diff
@@ -220,74 +184,115 @@ func TestEnvironment(state *BuildState, target *BuildTarget, testDir string) Bui
...
+		env = append(env, prefix+"TOOL="+toolPath(state, allTools[0], abs))
+	}
+	for name, tools := range namedTools {
+		env = append(env, prefix+"TOOLS_"+strings.ToUpper(name)+"="+strings.Join(toolPaths(state, tools, abs), " "))
+	}
+	return env
+}
+
+// Creates data-related env variables
+func dataEnv(state *BuildState, allData []BuildInput, namedData map[string][]BuildInput, prefix string, inTmpDir bool) BuildEnv {
```
tiagovtristao, line 274: Refactored data-related env variables into this function. It also fixes the problems where we were previously using the old `target.AllDataPaths(state.Graph)` if `inTmpDir` was true. This function was resolving requires/provides before returning the paths, which was not in line with how we prepare runtime data - where we don't resolve requires/provides.

src/core/build_env.go
```diff
@@ -220,74 +184,115 @@ func TestEnvironment(state *BuildState, target *BuildTarget, testDir string) Bui
...
+func resolveOut(out string, dir string, sandbox bool) string {
+	// Bit of a hack; ideally we would be unaware of the sandbox here.
+	if sandbox && runtime.GOOS == "linux" && !strings.HasPrefix(RepoRoot, "/tmp/") && dir != "." {
+		return path.Join(SandboxDir, out)
+	}
+	return path.Join(dir, out)
+}
+
+// Creates tool-related env variables
+func toolsEnv(state *BuildState, allTools []BuildInput, namedTools map[string][]BuildInput, prefix string, abs bool) BuildEnv {
```
tiagovtristao, line 260: This refactored tool-related env variables into this function. It now guarantees that they are constructed consistently across different environments. There used to be some mismatches before. This follows the pattern used by `$SRCS`

src/build/build_step.go
```diff
@@ -486,7 +486,7 @@ func runBuildCommand(state *core.BuildState, target *core.BuildTarget, command s
 	}
 	env := core.StampedBuildEnvironment(state, target, inputHash, path.Join(core.RepoRoot, target.TmpDir()), target.Stamp)
 	log.Debug("Building target %s\nENVIRONMENT:\n%s\n%s", target.Label, env, command)
-	out, combined, err := state.ProcessExecutor.ExecWithTimeoutShell(target, target.TmpDir(), env, target.BuildTimeout, state.ShowAllOutput, process.NewSandboxConfig(target.Sandbox, target.Sandbox), command)
+	out, combined, err := state.ProcessExecutor.ExecWithTimeoutShell(target, target.TmpDir(), env, target.BuildTimeout, state.ShowAllOutput, false, process.NewSandboxConfig(target.Sandbox, target.Sandbox), command)
```
samwestmoreland, line 489: Don't know if this is really an issue but would it better for readability to say
```
foreground := false
out, combined, err := state.ProcessExecutor.ExecWithTimeoutShell(..., foreground, ...)
```
?

tiagovtristao: I don't think it's worth it. Jumping quickly to the function tells you exactly what that parameter is about. Also, if I were to make this change here, I would have to create the same variable in all other places where this function call exists.

rules/go_rules.build_defs
```diff
@@ -1667,3 +1694,45 @@ def _get_ldflags_and_pkgconfig(name):
...
+                cmd = "find . -name \"*.go\" | cut -c 3- | xargs -I {} sh -c 'mkdir -p $OUTS/$(dirname {}) && cp {} $OUTS/$(dirname {})'",
+                outs = [debug_srcs_out],
+                deps = deps,
+                test_only = test_only,
+                requires = ["go_src"],
+                needs_transitive_deps = True,
+            )
+        ]
+        # Extract sources before running the debugger.
+        cmd = f"cp -R $PKG_DIR/{debug_srcs_out}/* . && {cmd}"
```
tiagovtristao, line 1730: This can get slow for targets with lots of transitive dependencies. Not a big problem for debugging, but an area for improvement.

tiagovtristao: Replaced this with packing and unpacking of sources, yielding much faster results.

Tatskaari (approved):
I think this makes sense. Probably about time to get this landed so we can start playing with it.

rules/go_rules.build_defs
```diff
@@ -1667,3 +1694,45 @@ def _get_ldflags_and_pkgconfig(name):
...
+                srcs = srcs,
+                # Archive all sources.
+                cmd = f"find . -name \"*.go\" | tar -cf {debug_srcs_tar} -T -",
+                outs = [debug_srcs_tar],
+                deps = deps,
+                test_only = test_only,
+                requires = ["go_src"],
+                needs_transitive_deps = True,
+            )
+        ]
```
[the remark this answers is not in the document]
tiagovtristao: Currently, we are only allowing for local debugging, although remote debugging might be a use case for later on. And IDEs acting as clients will have access to the sources files.

src/build/incrementality_test.go
```diff
@@ -70,9 +70,17 @@ var KnownFields = map[string]bool{
 	"Test.Flakiness": true,
 	"Test.Results":   true, // Recall that unsuccessful test results aren't cached...
 
+	// Debug fields don't contribute to any hash
+	"Debug":            true,
+	"Debug.Command":    true,
+	"Debug.data":       true,
+	"Debug.namedData":  true,
+	"Debug.tools":      true,
+	"Debug.namedTools": true,
```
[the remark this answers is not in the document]
tiagovtristao: Consistency with what I seen with `Test`. There was also no reason for making them public at the moment.

PR #2073, opened 2021-09-19
Don't use %w in log messages

peterebden:
It doesn't print very nicely:
```
13:14:52.596   ERROR: Unable to store current operation to  plz-out/.previous_op: %!w(*fs.PathError=&{write plz-out/.previous_op 28})
```
I think this should be better. Can't test the exact thing so readily now tho.


tiagovtristao (approved):
%s is indeed the right verb, thanks.

PR #2074, opened 2021-09-20
Let `python_wheel` optionally take a tool to fetch wheel urls

samwestmoreland:


samwestmoreland:
We'll push this straight into the plugin

PR #2083, opened 2021-09-21
Allow overriding plugin config with package()

Tatskaari:


samwestmoreland (approved):

src/parse/asp/builtins.go
```diff
@@ -198,8 +198,19 @@ func pkg(s *scope, args []pyObject) pyObject {
...
 		k = strings.ToUpper(k)
-		s.Assert(s.config.Get(k, nil) != nil, "error calling package(): %s is not a known config value", k)
-		s.config.IndexAssign(pyString(k), v)
+		configVal := s.config.Get(k, nil)
+		s.Assert(configVal != nil, "error calling package(): %s is not a known config value", k)
+
+		// Merge the config value together for plugins
+		if plugin := s.state.Config.GetPlugin(k); plugin != nil {
+			dict, ok := v.(pyDict)
+			s.Assert(ok, "error calling package(): %s is a plugin so that argument should be a dict", k)
```
tiagovtristao, line 207: Why does this builtin function require knowledge of plugins, is it to display the type error via the assert?

Tatskaari: I just changed it to merge dictionaries together regardless of if they're plugin config

tiagovtristao: Ya, I think it makes more sense too. Thanks.

tiagovtristao (approved):

PR #2085, opened 2021-09-22
Close off the pending target channel for the architeture psudo-target

### LINKED ISSUE #2084: Cross-compiling with --arch doesn't work when using label wildcards.
Hello I realised a while ago that since the update from please 16.2.1 my cross compile aliases for my docker images are not working anymore. 
The command `plz run parallel -i docker-build --arch linux_amd64 //...` is producing 3 different outputs on 7 minor releases.

1. On version 16.0.0 to 16.2.1 it actually build all the images as expected linux amd64. 
2. On version 16.3.0 to 16.5.1 it only finds one target and then gets stuck.
Screenshot 2021-09-22 at 13 48 52
3. On version 16.6.0 to current (16.9.0) it prints the following:
Screenshot 2021-09-22 at 13 53 25

Tatskaari:
Fixes #2084

### DIFF
--- a/src/remote/remote.go
+++ b/src/remote/remote.go
@@ -683,6 +683,8 @@ func (c *Client) reallyExecute(tid int, target *core.BuildTarget, command *pb.Co
 		ActionDigest:    digest,
 		SkipCacheLookup: skipCacheLookup,
 	}, updateProgress)
+	log.Debug("completed ExecuteAndWaitProgress() for %v", target.Label)
+
 	if err != nil {
 		// Handle timing issues if we try to resume an execution as it fails. If we get a
 		// "not found" we might find that it's already been completed and we can't resume.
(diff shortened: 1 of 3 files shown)


src/remote/remote.go
```diff
@@ -683,6 +683,8 @@ func (c *Client) reallyExecute(tid int, target *core.BuildTarget, command *pb.Co
 		ActionDigest:    digest,
 		SkipCacheLookup: skipCacheLookup,
 	}, updateProgress)
+	log.Debug("completed ExecuteAndWaitProgress() for %v", target.Label)
```
peterebden, line 686: is this deliberate? or testing?

peterebden (approved):

src/remote/remote.go
Tatskaari: I was going to open a PR separately but I think this might help diagnose some issues. Might be helpful to know when Please leaves that function to figure out where it's getting stuck "executing" or "queued".

PR #2086, opened 2021-09-23
Add types to the plugin config definitions

Tatskaari:


docs/config.html
```diff
@@ -448,6 +448,17 @@ <h3 class="mt1 f6 lh-title" id="pluginconfig.repeatable">
         </p>
       </div>
     </li>
+    <li>
+      <div>
+        <h3 class="mt1 f6 lh-title" id="pluginconfig.boolean">
+          Boolean <span class="normal">(bool)</span>
```
peterebden: This is out of line with below, no?

src/parse/asp/objects.go
```diff
@@ -967,6 +968,37 @@ func loadPluginConfig(state *core.BuildState, c pyDict) {
 	c[strings.ToUpper(pluginName)] = pluginNamespace
 }
 
+func toPyObject(key, val, toType string) pyObject {
+	if toType == "" || toType == "str" {
+		return pyString(val)
+	}
+
+	if val == "" {
+		return pyNone{}
```
peterebden, line 977: Should we do this? Or should they become the zero value (0 / False)?

src/parse/asp/objects.go
```diff
@@ -967,6 +968,37 @@ func loadPluginConfig(state *core.BuildState, c pyDict) {
...
+	}
+
+	if toType == "bool" {
+		if val == "true" || val == "True" {
```
peterebden, line 981: The config files support yes/no and on/off as well; should we include that here too?

src/parse/asp/objects.go
Tatskaari: This is the difference between an optional boolean value that hasn't been set, and one that's explicitly been set to False. Perhaps we don't need to model that but it seems more correct.

src/parse/asp/objects.go
Tatskaari: Good point. Done.

docs/config.html
Tatskaari: you were too fest

src/parse/asp/objects.go
Tatskaari: Yeah okay I'll change this.

tiagovtristao (approved):
Lgtm. As discussed offline, I agree with the zero-value consistency that already exists for configuration. I just wonder how the `Optional` field plays with it and whether it might cause some confusion.

Tatskaari:
> Lgtm. As discussed offline, I agree with the zero-value consistency that already exists for configuration. I just wonder how the `Optional` field plays with it and whether it might cause some confusion.

I guess the Optional field now means that the value has to be non-zero i.e. not 0 or "". Not sure it makes sense for bools.

PR #2089, opened 2021-09-23
Fix concurrent map writes in package()

Tatskaari:


peterebden (approved):
Let that be a lesson to you.

PR #2095, opened 2021-09-28
Add `--wd` flag and deprecate `--in_wd`

samwestmoreland:


test/plz_run/BUILD
```diff
@@ -58,3 +58,13 @@ please_repo_e2e_test(
...
     repo = "test_repo",
 )
+
+# Test --wd flag changes directory and outputs there
+please_repo_e2e_test(
+    name = "test_wd_flag",
+    expected_output = {
+        "test_wd/guava.txt": "guava",
+    },
+    plz_command = "mkdir test_wd; plz run --wd test_wd //src:create_file",
```
Tatskaari, line 68: Can we add a test to check that `--wd=.` works the same as the old `--in_wd`?

Tatskaari (approved):
Looks good to me!

test/plz_run/BUILD
samwestmoreland: Added another test. Seems to just work!

PR #2100, opened 2021-09-28
Only strip sources from go_toolchain() when not debugging

Tatskaari:


peterebden (commented):
Can this not be the default for the argument? I _think_ `strip_srcs=CONFIG.BUILD_CONFIG == "dbg"` might work, and IMO it would be more intuitive than this not honouring the argument in some configuratoins.

PR #2102, opened 2021-09-29
Remove concurrent-map dependency

peterebden:
Was noodling around with this and found it's only used in one place and we can just use sync.Map instead.

Ran a `go mod tidy` after which removed a bunch of stuff. Shout if there are any concerns...


tiagovtristao (approved):
Lgtm

PR #2103, opened 2021-09-29
Handle `--wd=.` explicitly

samwestmoreland:


samwestmoreland:
I'll just sort out the relative path stuff

src/please.go
```diff
@@ -925,6 +928,14 @@ func (overrides ConfigOverrides) Complete(match string) []flags.Completion {
 	return core.DefaultConfiguration().Completions(match)
 }
 
+// Get an absolute path from a relative path.
+func getAbsolutePath(path string, here string) string {
+	if path[0] != '/' && path[0] != '~' {
```
Tatskaari, line 933: I think `~` is expanded by bash (or similar)  so the latter check is probably not necessary.

Tatskaari (approved):
LGTM :)

src/please.go
tiagovtristao: I've had issues before where the shell wasn't expanding `~`, so it might be better to take it into account.

src/please.go
```diff
@@ -925,6 +928,14 @@ func (overrides ConfigOverrides) Complete(match string) []flags.Completion {
...
+		return here + "/" + path
```
tiagovtristao, line 934: I feel that we should we should leverage `filepath.IsAbs` and `filepath.Join` here when handling paths. They will handle relative paths like `../../this/dir` for you, among other things, such as providing cross-compatibility between different OSes.

PR #2104, opened 2021-09-29
Have the install script honour the location config override.

peterebden:
If it's in `plzconfig.local` (which is the only file that overrides the one we set in `.plzconfig`) then install to whatever that specifies.


tiagovtristao (approved):
TIL about this script.

PR #2111, opened 2021-10-01
Fix issue with loading plugin config on subinclude

Tatskaari:
This fixes a few issues with loading config from subrepos. 

1. We were doing a shallow copy of the Plugin map. This PR makes sure that we have new Plugin structs when loading in the subrepo config into it's config. 
2. We were applying the config values from the subrepos state when loading plugin config for the host repo's package when doing a subinclude. This change makes sure that we read the plugin definition from the subrepo's config but the values from the host repo's config. This used to mostly work because the subrepo's config is layered ontop of the host repos config. however breaks down when they differ.

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -1020,31 +1020,12 @@ func (state *BuildState) findArch(arch cli.Arch) *BuildState {
 	return nil
 }
 
-func copyPlugin(plugin Plugin) *Plugin {
-	values := map[string][]string{}
-	for k, v := range plugin.ExtraValues {
-		values[k] = v
-	}
-	plugin.ExtraValues = values
-	return &plugin
-}
-
-func copyConfig(config Configuration) *Configuration {
-	config.buildEnvStored = &storedBuildEnv{}
-	plugins := map[string]*Plugin{}
-	for name, plugin := range config.Plugin {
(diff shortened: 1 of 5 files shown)


src/parse/asp/objects.go
```diff
@@ -915,25 +915,25 @@ func newConfig(state *core.BuildState) *pyConfig {
 	c["TARGET_ARCH"] = pyString(state.TargetArch.Arch)
 	c["BUILD_CONFIG"] = pyString(state.Config.Build.Config)
 
-	loadPluginConfig(state, c)
+	loadPluginConfig(state.Config, state, c)
```
Tatskaari, line 918: This means that we load plugin config when working in the plugin repo. In this case, we make sure that `FoocTool` is loaded. 

```
[PluginDefinition]
name = foo

[Plugin "foo"]
FoocTool = fooc
```

src/core/state.go
```diff
@@ -1020,14 +1020,31 @@ func (state *BuildState) findArch(arch cli.Arch) *BuildState {
 	return nil
 }
 
+func copyPlugin(plugin Plugin) *Plugin {
```
peterebden, line 1023: might be mildly nice as a member function?

peterebden (approved):

PR #2114, opened 2021-10-04
Only strip the go sdk sources when not debugging

Tatskaari:
This helps with `plz debug`. Without this we can't debug any of the package from the go sdk.


tiagovtristao (approved):
Thanks, it would be nice to get this released for `plz debug` but it's not urgent.

PR #2127, opened 2021-10-11
Fix issue where protos don't appear in revdeps as they don't technica…

Tatskaari:
…lly depend on eachother


samwestmoreland (approved):
Lgtm!

tiagovtristao:
Can you also add a test for what this PR is fixing?

PR #2128, opened 2021-10-11
Add headers to remote files

samwestmoreland:
So here we drop into bash in any case where any authentication needs to be done, and let `curl` handle it.
I pass this flag `--anyauth` which seems pretty useful. From the docs -

>  Tells curl to figure out authentication method by itself, and use the most secure one the remote site claims to support. This is done by first doing a request and checking the response-headers, thus possibly inducing an extra network round-trip. This is used instead of setting a specific authentication method, which you can do with --basic, --digest, --ntlm, and --negotiate.

> Used together with -u, --user.

This change also uses `curl` if any headers are specified, as we can pass them straight in as arguments. This avoids having to add a bunch of new fields to `build_rule`, which we may or may not want to do in the future.


peterebden:
I'm dubious about using curl for this. That seems like a backwards step (once upon a time we used to but it got moved to a native action so we had more control).

Tatskaari (approved):
I'm happy for this to go in as is but maybe we should have a chat with @peterebden to see if it's worth making this more first class.

PR #2133, opened 2021-10-14
Panic if glob does not match anything

samwestmoreland:
I'm not sure about this - is it sometimes ok for the glob to not match anything? E.g. if we're globbing for `deps`... So if that's the case, we don't want to be panicking in the glob, but rather when we're parsing `srcs`?


rules/go_rules.build_defs
```diff
@@ -123,6 +123,9 @@ def go_library(name:str, srcs:list, resources:list=None, asm_srcs:list=None, hdr
       package (str): The package as it would appear at the top of the go source files. Defaults
                      to name.
     """
+    if CONFIG.FF_FAIL_ON_BAD_GLOB and srcs == []:
```
peterebden, line 126: This shouldn't happen here - it should be a general check in the `glob` implementation in `builtins.go`

peterebden (changes requested):
We might also want to have an argument on `glob` that lets the caller opt out of this behaviour for cases where they know something might not exist?

rules/go_rules.build_defs
samwestmoreland: I was unsure about putting it in the `glob` implementation because I didn't know if we wanted to fail on _any_ bad glob. Sticking it in the go rules means it wouldn't be a breaking change I think as we already fail in that case, just with an unhelpful error message. But I'm happy to stick it in there if you think that's best - Jon is also of the opinion that no one would intentionally write a glob that doesn't match anything. Also agree on the opt-out - would soften the impact of the breaking.

samwestmoreland:
Running with this feature flag enabled in `core3` gives at least 144 errors (I've only built 119,000/300,000 targets so far). The majority of the them are either `glob[*.yaml]` or `glob[OWNER]` globs, which Jon suggested might be legitimate uses, so it looks like we might leave this one.

tiagovtristao:
What was the state on this? I think it was decided to be abandoned given the impact of the changes on existing code bases?

PR #2135, opened 2021-10-20
Remove SingleSHA1Hash

peterebden:
Bit of coding to pass the time on the plane.
Removes the SingleSHA1 fflag and associated code. This simplifies these code paths a bit for the next change (which is the real thing I wanted, but split this out to simplify it a bit).

Should be a v17 change since we never made this the default in v16.


peterebden:
Bad bot, we're waiting to make breaking changes.

PR #2148, opened 2021-10-31
Make the targets containing subrepos output into plz-out/subrepos

peterebden:
There are some subtle errors that can come up with the existing setup; one I have seen is that `glob` can behave differently between runs (for example if something outputs a file that would be caught by the glob, on the first run it doesn't exist, on a later run maybe it does and then the build behaves differently or fails).

This puts the subrepo trees in `plz-out/sub` where they should be less subject to change. I don't think there is a lot of cost to that (we create another name in `plz-out` but it's not like we're about to run out there).

### DIFF
--- a/src/core/build_target.go
+++ b/src/core/build_target.go
@@ -29,7 +29,7 @@ const GenDir string = "plz-out/gen"
 const BinDir string = "plz-out/bin"
 
 // SubrepoDir is the output directory for targets that define subrepos.
-const SubrepoDir = "plz-out/sub"
+const SubrepoDir = "plz-out/subrepos"
 
 // DefaultBuildingDescription is the default description for targets when they're building.
 const DefaultBuildingDescription = "Building..."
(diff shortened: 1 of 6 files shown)


src/core/build_target.go
```diff
@@ -28,6 +28,9 @@ const GenDir string = "plz-out/gen"
 // BinDir is the output directory for binary targets.
 const BinDir string = "plz-out/bin"
 
+// SubrepoDir is the output directory for targets that define subrepos.
+const SubrepoDir = "plz-out/sub"
```
Tatskaari, line 32: I wonder if `repos` is more descriptive?

Tatskaari (approved):
Looks good to me :)

src/core/build_target.go
tiagovtristao: I would join both answers - `subrepos`

tiagovtristao (approved):
Nice, I thought this would be harder to implement.

peterebden:
I sort of liked the idea of it being three letters (like `bin`, `gen` and `tmp`) but you guys win :)

PR #2149, opened 2021-10-31
Remove two unused arguments from build_rule

peterebden:
I don't think we have to wait for anything in particular before getting rid of these?


samwestmoreland (approved):

Tatskaari:
Yeah I don't think we allow positional params for build_rule and even if we did, what lunatic would?

Tatskaari (approved):

PR #2160, opened 2021-11-07
Update the lexicon with some missing functions

robxu9:
I noticed json() and text_file() came through, but I didn't see docs reflect. Also, removed another reference to `raise` and fixed a spacing nit with `is_semver`'s display on the webpage.

(disclaimer: I haven't tested this, but I assume this is probably a small enough fix, unless I missed some other file I had to modify, also.)

### DIFF
--- a/docs/lexicon.html
+++ b/docs/lexicon.html
@@ -231,9 +231,17 @@
     <li>
       <span>
         <code class="code"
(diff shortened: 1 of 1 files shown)


Tatskaari (approved):
Looks good to me! Thanks for that.

Issue #2151, opened 2021-11-01
Plz generate should update nearest gitignore


jakec-github:
If a generated source gets linked from `plz-out/gen` and there is not already a git ignore entry for it then Please should generate one in the nearest `.gitignore` file or the one specified by `update_gitignore`

PR #2163, opened 2021-11-11
Implement automatically updating gitignores

### LINKED ISSUE #2158: Enable hard linking of generated sources for those that want to check them in
Link generated sources currently symlinks files into the source tree. For those that want to check these in, hard links would be preferable. We should allow the following: 

```
LinkGeneratedSources = True ; symlinks
LinkGeneratedSources = soft ; symlinks
LinkGeneratedSources = hard ; hardlinks
```

Tatskaari:
This PR adds `hard` as an option for link generated sources to make it link using hardlinks. It also updates the relevant .gitignores during the build if `UpdateGitignore` is enabled. 

Fixes #2158 and #2151


src/core/config.go
```diff
@@ -424,7 +425,8 @@ type Configuration struct {
 		HTTPProxy            cli.URL      `help:"A URL to use as a proxy server for downloads. Only applies to internal ones - e.g. self-updates or remote_file rules."`
 		HashFunction         string       `help:"The hash function to use internally for build actions." options:"sha1,sha256"`
 		ExitOnError          bool         `help:"True to have build actions automatically fail on error (essentially passing -e to the shell they run in)." var:"EXIT_ON_ERROR"`
-		LinkGeneratedSources bool         `help:"If set, supported build definitions will link generated sources back into the source tree. The list of generated files can be generated for the .gitignore through 'plz query print --label gitignore: //...'. Defaults to false." var:"LINK_GEN_SOURCES"`
+		LinkGeneratedSources string       `help:"If set, supported build definitions will link generated sources back into the source tree. The list of generated files can be generated for the .gitignore through 'plz query print --label gitignore: //...'. Defaults to false. Set to hard for hardlinks and soft or true for symlinks." var:"LINK_GEN_SOURCES"`
```
tiagovtristao, line 428: I think we can make the available options clearer. Something along... `The available options are: "hard" (hardlinks), "soft" (symlinks), "true" (symlinks) and "false" (default)`

src/core/config.go
```diff
@@ -812,9 +814,8 @@ func (config *Configuration) ApplyOverrides(overrides map[string]string) error {
 				field.Set(reflect.ValueOf(v))
 			}
 		case reflect.Bool:
-			v = strings.ToLower(v)
-			// Mimics the set of truthy things gcfg accepts in our config file.
-			field.SetBool(v == "true" || v == "yes" || v == "on" || v == "1")
+			v, _ := gcfgtypes.ParseBool(strings.ToLower(v))
```
tiagovtristao, line 817: nit: `ParseBool` will lowercase it itself.

src/core/config.go
```diff
@@ -943,6 +944,11 @@ func (config *Configuration) NumRemoteExecutors() int {
 	return config.Remote.NumExecutors
 }
 
+func (config *Configuration) ShouldLinkGeneratedSources() bool {
+	isTruthy, _ := gcfgtypes.ParseBool(config.Build.LinkGeneratedSources)
+	return config.Build.LinkGeneratedSources == "hard" || isTruthy
```
tiagovtristao, line 949: Isn't the `soft` option missing here?

tiagovtristao (approved):

PR #2164, opened 2021-11-12
Merge build and test queues

peterebden:
Dreaming a bit about being able to unify more of the behaviour (at least in terms of incrementalism) between the build & test tasks.

Think this is nicer in itself though (particularly the `plz.go` part where it can now just range a single channel).

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -34,10 +34,18 @@ type ParseTask struct {
 	ForSubinclude    bool
 }
 
+// A TaskType identifies whether a task is a build or test action.
+type TaskType uint16
+
+const (
+	BuildTask TaskType = 0
+	TestTask  TaskType = 1
+)
+
 // A Task is the type for the queue of build/test tasks.
 type Task struct {
 	Label BuildLabel
-	Test  bool  // True if this task is a test
+	Type  TaskType
 	Run   int32 // Only present for tests (the run of a build is always zero)
 }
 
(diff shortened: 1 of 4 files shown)


src/core/state.go
```diff
@@ -34,13 +34,19 @@ type ParseTask struct {
 	ForSubinclude    bool
 }
 
-// BuildTask is the type for the build task queue
-type BuildTask = BuildLabel
+// A TaskType identifies whether a task is a build or test action.
+type TaskType uint16
```
tiagovtristao, line 38: nit: `uint8`

src/core/state.go
```diff
@@ -34,13 +34,19 @@ type ParseTask struct {
...
+	BuildTask TaskType = 0
+	TestTask  TaskType = 1
+)
+
+// A Task is the type for the queue of build/test tasks.
+type Task struct {
 	Label BuildLabel
-	Run   int
+	Type  TaskType
+	Run   int32 // Only present for tests (the run of a build is always zero)
```
tiagovtristao, line 49: nit: unsigned variant better describes this?

src/core/state_test.go
```diff
@@ -109,20 +109,20 @@ func TestAddTargetFilegroupPackageOutputs(t *testing.T) {
 
 func TestAddDepsToTarget(t *testing.T) {
 	state := NewDefaultBuildState()
-	_, builds, _, _, _ := state.TaskQueues() //nolint:dogsled
+	_, builds, _ := state.TaskQueues() //nolint:dogsled
```
tiagovtristao, line 112: I think this nolint directive can be removed now.

tiagovtristao (approved):
Me likes. Nice refactoring!

src/core/state.go
peterebden: sure thing - also lets us handle up to 4 billion runs, just in case!

Tatskaari (approved):
Oh nice! That was easy.

PR #2165, opened 2021-11-12
Add option to force mount sandboxing when rules opt out

Tatskaari:


peterebden (commented):
This is very specific to mount namespacing; would we not want to potentially enforce this for e.g. user or pid namespaces too?

Tatskaari:
> This is very specific to mount namespacing; would we not want to potentially enforce this for e.g. user or pid namespaces too?

Perhaps the name is a bit misleading because this does apply the PID and user namespacing. 

I guess the question is, how much control do we want to provide over this. Currently we always namespace user and pid. It seems that we can have 4 sandboxing modes:

1) Sandboxing off 
2) PID+USER (SHARE_NETWORK=1, SHARE_MOUNT=1) - Not accessable with the current setup
3) PID+USER+MOUNT (SHARE_NETWORK=1, SHARE_MOUNT=0) - Not accessable with the current setup
4) PID+USER+MOUNT+NETWORK (SHARE_NETWORK=0, SHARE_MOUNT=0)

I suppose technically there's also the option to share mount but not network but I doubt there's any use for that. 

I wonder if these are good enough or if we want to have more control over this, and how do we communicate this through the .plzconfig? 

We could add in 2 more env vars for PID and user and then have: 
```
BaseIsolationLevel = share_net,share_mount,share_user,share_pid ; default value
SandboxIsolationLevel = ; Everything enabled
```

tiagovtristao:
I understand the current use case for the `namespace` field, but from a user's perspective I have tripped myself over the current options (always, never, sandbox) for it within the `[sandbox]` section, as `sandboxing` and `namespacing` are kinda related to each other.

For instance, I would expect:

``` 
[sandbox]
build = true
test = true
```

To imply namespacing by default. Although sandboxing and namespacing aren't exactly the same thing, I think it would be nice to "wrap" both into the "sandbox" word for the user.
I would imagine something like the following (which would allow to easily extend it to other sandbox/namespace types if needed in the future:

Choose sandbox options (we don't have to support everything now):
```
[sandbox]
build = mount
build = net
```

Just enable sandboxing. Whatever that might mean. Perhaps this is sort of an alias to the union of several other sandbox/namespace options.
```
[sandbox]
build = true
```

Forcing options:
```
[sandbox]
build = true
build = force_mount
```

This is perhaps outside of this PR, but I would find the above easier to reason about.

Tatskaari:
@peterebden @tiagovtristao Would this be overkill:

```
[Sandbox "build"]
User = true
MountTmp = true
Network = true
LoopbackAddress = true ; Might be a little unecessary
SandboxDir = blah

[Sandbox "test"]
...

[Sandbox "custom"]
...
```

And then we can have:
```
build_rule(
    sandbox = True, # Uses the "build" sandbox above
    test_sandbox = True, # Uses the "test" sandbox above
)

build_rule(
    sandbox = "custom", # Uses the custom sandbox above
)
```

PR #2173, opened 2021-11-16
Add `plz query reporoot` command

peterebden:
Discussed in https://github.com/please-build/cc-rules/pull/12 where it seems it would be useful (and we have had other cases too). It's really not much code at all.


tiagovtristao (approved):
Thanks!

PR #2174, opened 2021-11-17
Fixes named fields not showing up when using 'plz query print --field=srcs'

tiagovtristao:
This was happening because only the first field (ie. unnamed sources) was being considered.

### DIFF
--- a/src/query/print.go
+++ b/src/query/print.go
@@ -192,10 +192,10 @@ func (p *printer) findField(field string) reflect.StructField {
 		resIndex := -1
 		for i := 0; i < v.NumField(); i++ {
 			if f := t.Field(i); p.fieldName(f) == fieldName {
-				if resIndex == -1 {
-					resIndex = i
-				} else if v.Field(resIndex).IsZero() && !v.Field(i).IsZero() {
+				if !v.Field(i).IsZero() {
 					return t.Field(i), true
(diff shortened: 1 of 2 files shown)


samwestmoreland (approved):

src/query/print.go
```diff
@@ -182,20 +182,37 @@ func (p *printer) PrintFields(fields []string) bool {
...
+	// perspective. The function below takes that into consideration.
+	findFieldStruct := func(value interface{}, fieldName string) (reflect.StructField, bool) {
+		v := reflect.ValueOf(value).Elem()
+		t := v.Type()
+
+		resIndex := -1
+		for i := 0; i < v.NumField(); i++ {
+			if f := t.Field(i); p.fieldName(f) == fieldName {
+				if resIndex == -1 {
+					resIndex = i
```
Tatskaari, line 196: I wonder if we can return early here if the field value isn't zero and skip the check on the next line?

```
if !v.Field(i).IsZero() {
    return t.Field(i), true
} else if resindex == -1 {
    resIndex = i 
}
```

Tatskaari (approved):
Looks good to me. Just a small optimisation we might be able to make.

src/query/print.go
tiagovtristao: Ya, sounds better

PR #2181, opened 2021-11-23
Devendorise zip

peterebden:
The Go 1.17 archive/zip package has added builtins that are sufficient to do what we need without a copy of our own \o/

We do lose the ability to set alignment. Most stuff doesn't care these days (it's mostly a thing for Android); we could bring it back if upstream added an `Offset()` function to pair with `SetOffset()`.


samwestmoreland (approved):
Nice!

Tatskaari (approved):
Looks good to me! Much cleaner.

tiagovtristao (approved):

PR #2185, opened 2021-11-24
Refactor the variable names in the lexer so it's easier to follow

Tatskaari:


src/parse/asp/lexer.go
```diff
@@ -85,13 +85,17 @@ func newLexer(r io.Reader) *lex {
 
 // A lex is a lexer for a single BUILD file.
 type lex struct {
-	b      []byte
-	i      int
-	line   int
-	col    int
+	// The raw bytes we're lexing
+	bytes []byte
```
Tatskaari, line 89: Might be nice to make this a reader, and read bytes off it at a time to avoid allocations

chrisnovakovic (approved):
LGTM, thanks!

PR #2188, opened 2021-11-24
v17: Remove the short flag for incremental on `plz cover` as it conflicts with --include

Tatskaari:
This can probably land in v17


samwestmoreland (approved):

peterebden:
Technically it is breaking, yes; `plz cover -i` now means something different. Note that in general that is OK (e.g. `plz test -n 5` means something different from `plz -n 5 test`).

Honestly I don't remember much of the motivation for this flag. I'm not sure if it gets used much.

Tatskaari:
> 

Will land in v17. It was causing some confusion as `plz test -i` behaved differently to `plz cover -i` so I think it's worth fixing.

PR #2191, opened 2021-11-24
Switch the lexer to use bufio rather than ReadAllBytes()

Tatskaari:
Not sure how useful this is but I couldn't resist. I think it's a little cleaner even if it doesn't improve performance in any way.

### DIFF
--- a/src/parse/asp/lexer.go
+++ b/src/parse/asp/lexer.go
@@ -64,7 +64,7 @@ func newLexer(r io.Reader) *lex {
 		filename: NameOfReader(r),
 		indents:  []int{0},
 	}
-	l.nextByte, _ = l.reader.ReadByte()
+	l.nextRune, _, _ = l.reader.ReadRune()
 	l.readNextByte()
 	l.Next() // Initial value is zero, this forces it to populate itself.
 	// Discard any leading newlines, they are just an annoyance.
@@ -79,7 +79,7 @@ type lex struct {
 	// The reader we're lexing
 	reader *bufio.Reader
 	// the current and next bytes from pos
-	currentByte, nextByte byte
+	currentRune, nextRune rune
 	// Positional information about where we are in the file
 	pos, line, col int
 	// The current level of indentation we're on in the file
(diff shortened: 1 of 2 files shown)


src/parse/asp/lexer_test.go
```diff
@@ -56,10 +56,10 @@ func TestLexFunction(t *testing.T) {
...
 	l := newLexer(strings.NewReader("懂了吗 你愁脸 有没有"))
 	assertToken(t, l.Next(), Ident, "懂了吗", 1, 1, 1)
-	assertToken(t, l.Next(), Ident, "你愁脸", 1, 11, 11)
-	assertToken(t, l.Next(), Ident, "有没有", 1, 21, 21)
-	assertToken(t, l.Next(), EOL, "", 1, 30, 30)
-	assertToken(t, l.Next(), EOF, "", 2, 1, 31)
+	assertToken(t, l.Next(), Ident, "你愁脸", 1, 5, 5)
+	assertToken(t, l.Next(), Ident, "有没有", 1, 9, 9)
+	assertToken(t, l.Next(), EOL, "", 1, 12, 12)
+	assertToken(t, l.Next(), EOF, "", 2, 1, 13)
```
Tatskaari, line 62: As we don't use pos to index into the buffer, we can now consider a rune a single position regardless of how many bytes it is.

src/parse/asp/lexer.go
```diff
@@ -59,22 +59,13 @@ func NameOfReader(r io.Reader) string {
...
-	if len(b) > 0 && b[len(b)-1] != '\n' {
-		b = append(b, '\n')
-	}
 	l := &lex{
-		bytes:    append(b, 0, 0), // Null-terminating the buffer makes things easier later.
+		reader:   bufio.NewReader(r),
 		filename: NameOfReader(r),
 		indents:  []int{0},
 	}
+	l.nextRune, _, _ = l.reader.ReadRune()
```
tiagovtristao, line 67: We should be handling any potential error here too.

src/parse/asp/lexer.go
```diff
@@ -108,6 +99,34 @@ type lex struct {
...
+			l.nextRune = 0 // This essentially null terminates the file which is useful later on
+		}
+	}
+}
+
+// advance reads moves to the next position in the reader, updating nextRune and currentRune
+func (l *lex) advance() {
+	l.col++
+	l.pos++
+	l.readNextByte()
```
tiagovtristao, line 127: I think `readNextByte` should be named `readNextRune` since it operates on `ReadRune()` underneath.

Also, the size of a rune doesn't always equal 1 single byte. The `ReadRune()` function returns the number of bytes read which should be used when updating `pos`?

Tatskaari: Good catch. I refactored this lexer to be rune based without considering this.

PR #2195, opened 2021-11-25
Download arcat & remove jarcat

peterebden:
This simplifies the bootstrap side of things (we no longer need `jarcat_unzip`) and just fetches it directly in the internal package.

There is some fiddling with please_pex which pulls in part of it as a library; for now I have kept the code in a new location, but I expect this to become irrelevant when we move to the python-rules repo.


Tatskaari (approved):

src/core/config.go
```diff
@@ -370,7 +370,7 @@ func DefaultConfiguration() *Configuration {
 	config.Go.EmbedTool = "//_please:please_go_embed"
 	config.Python.PexTool = "//_please:please_pex"
 	config.Java.JavacWorker = "//_please:javac_worker"
-	config.Java.JarCatTool = "//_please:jarcat"
+	config.Java.JarCatTool = "//_please:arcat"
```
samwestmoreland, line 373: Should line 525 say "Defaults to arcat"?

samwestmoreland (approved):

src/core/config.go
peterebden: Good spot, thanks.

All those comments are actually out of date (they don't default to being in the install directory any more)...

tiagovtristao (approved):

PR #2199, opened 2021-11-27
Add plz exec docs

tiagovtristao:


docs/commands.html
```diff
@@ -656,6 +656,52 @@ <h2 id="run" class="title-2">plz run</h2>
...
 </section>
 
+<section class="mt4">
+  <h2 id="exec" class="title-2">plz exec</h2>
+
+  <p>
+    This command is similar to <code class="code">plz run</code>, but defaults
+    to executing the binary target result in a sandboxed environment if that's
+    the environment where it was built on. This is desirable in situations where
+    the execution side-effects shouldn't propagate to the host system.
```
Tatskaari, line 666: I think more generally, this executes the target in the `hermetic build environment` rather than the host machines shell. If Please is configured to apply sandboxing to the target, that's implied by that statement. 

It might be worth giving some context behind the motivation for this command too. Maybe mention that this is designed for programmatic use e.g. running e2e tests from a CI pipeline that have remote environment state that doesn't fall neatly into Please's caching approach, or debugging/profiling tests with a true to life env.

Tatskaari (approved):

PR #2200, opened 2021-11-28
Fix test fields not being displayed correctly when running 'plz query…

tiagovtristao:
… print --field'

Test fields now live in a separate data structure, but the index found was still being applied to the main build target data structure.


peterebden (approved):
Nice spot!

Tatskaari (approved):

PR #2201, opened 2021-11-29
Set Prometheus pusher to use text format

Hamishpk:
As of Prometheus version 2.0, all processes that expose metrics to Prometheus need to use a text-based format. However the go client defaults to a proto buffer format. This PR sets the pusher to push metrics in a test based format, fixing the Error pushing Prometheus metrics issue.


Tatskaari (approved):
Thanks Hamish! Much appreciated.

PR #2203, opened 2021-11-29
Make C/C++ library output filenames customisable

chrisnovakovic:
There are occasions when a C/C++ library needs to be given a specific filename, e.g. when it needs to be automatically detected by an Autoconf-generated configure script, but doing this using the built-in rules requires either a library target whose name matches the desired library name or an additional `genrule` target to do the renaming if this isn't possible.

Add an `out` parameter to the `cc_library` and `cc_static_library` rules (and their C equivalents) to allow the library filename to be customised. Similarly to the existing behaviour for `cc_shared_object`, a sensible default is chosen if no value of `out` is given.

### DIFF
--- a/rules/c_rules.build_defs
+++ b/rules/c_rules.build_defs
@@ -8,7 +8,7 @@ you must pay attention to interoperability when needed (e.g. 'extern "C"' and so
 """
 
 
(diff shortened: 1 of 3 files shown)


peterebden (approved):
Makes sense, thanks!

Issue #2204, opened 2021-11-29
Running `plz test --debug` allows wildcard targets


windrunner99:
It is possible to run `plz test -d //src/...` or even `plz test -d` which result in an output that is not very friendly to understand.
For example running it in the please repository results in: 
```
00:31:29.607 CRITICAL: Target //docs/test:docs_test hasn't built but we have no pending tasks left.

The following targets have not yet built:
//:pleasings  //docs:acknowledgements_html  //docs:basics_html  //docs:build_rules_html  //docs:cache_html
//docs:codelabs_html  //docs:commands_html  //docs:config_html  //docs:cross_compiling_html  //docs:dependencies_html  
//docs:docs  //docs:faq_html  //docs:index_html  //docs:language_html  //docs:lexicon_html  //docs:milestones_15_9_1_html
//docs:milestones_16_0_0_html  //docs:performance_html  //docs:pleasings_html  //docs:post_build_html
//docs:quickstart_dropoff_html  //docs:quickstart_html  //docs:remote_builds_html  //docs:require_provide_html  //docs:rules 
...
```
Please should also check if a target does contain a wildcard when it is the only one provided to the `plz test` command with the `--debug` flag set.

PR #2205, opened 2021-11-30
Check for wildcard before running `plz test -d`

windrunner99:
My attempt of fixing issue #2204

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -1062,7 +1062,7 @@ func Please(targets []core.BuildLabel, config *core.Configuration, shouldBuild, 
 	}
 
 	// Only one target that is _not_ named "all" or "..." is allowed with debug test.
-	if state.DebugFailingTests && (len(targets) != 1 || (len(targets) == 1 && (targets[0].IsAllTargets() || targets[0].IsAllSubpackages()))) {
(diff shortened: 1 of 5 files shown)


src/please.go
```diff
@@ -1061,7 +1061,7 @@ func Please(targets []core.BuildLabel, config *core.Configuration, shouldBuild,
 		}
 	}
 
-	if state.DebugFailingTests && len(targets) != 1 {
+	if state.DebugFailingTests && (len(targets) != 1 || (len(targets) == 1 && targets[0].Name == "...")) {
```
peterebden, line 1064: This should be `(targets[0].IsAllSubpackages() || targets[0].IsAllTargets())`; that picks up `:all` as well.

We could also add a method to check for either of those two, there are other places this has come up too.

peterebden (commented):
Thanks!

src/please.go
windrunner99: Used the functions liked you said. The logic now looks a little bit messy so I added a comment.
But you are right, adding a Method that checks if either of those two are true to check if the BuildLable potentially refers more than one rule wild clear this up.
Let me think of a descriptive name and I will add it to this PR.

peterebden: `IsPseudoTarget` is the best I can think of to capture both of these cases where it's not actually referring to a real target.

Thanks!

peterebden:
Thanks!

PR #2210, opened 2021-12-01
Add link to generate a virtual env under plz-out

Tatskaari:


tiagovtristao (approved):

samwestmoreland (approved):
How come one's a `dlink` and the other a `link`? Is it something about retaining the directory structure?

Tatskaari:
> How come one's a `dlink` and the other a `link`? Is it something about retaining the directory structure?

The srcs rules output files in slightly different ways. The wheel rules produce `{module}/thing.py` whereas the pip rules produce `{name}/{module}/thing.py`.

The `dlink` flag removed the extra level of nesting.

afterthought:
@Tatskaari, I tested this and it doesn't quite work. Only 1 pip_library gets linked to python/venv and then the remaining ones are skipped because this [line](https://github.com/thought-machine/please/blob/3a04445a810829976518757f8401ae4eb6d4ce23/src/build/build_step.go#L1021) is checking if the destination exists and it does already after the first lib creates python/venv.

PR #2212, opened 2021-12-02
Infer config key if not set

samwestmoreland:
Couldn't seem to test this - thought I might be able to do it with `plz query config` but couldn't get it to print out any of the plugin-specific config values.


src/parse/asp/objects.go
```diff
@@ -964,6 +964,9 @@ func loadPluginConfig(subrepoConfig *core.Configuration, packageState *core.Buil
 	contextPackage := &core.Package{SubrepoName: packageState.CurrentSubrepo}
 	configValueDefinitions := subrepoConfig.PluginConfig
 	for key, definition := range configValueDefinitions {
+		if definition.ConfigKey == "" {
+			definition.ConfigKey = strings.ReplaceAll(key, "_", "")
+		}
```
Tatskaari, line 969: We don't want to write this back to the definition do we? Probably just want to create a local variable for it.

samwestmoreland: Oh sorry thought we did! Updated it now.

Tatskaari (approved):
Looks good to me!

PR #2213, opened 2021-12-04
Simplify debugging logic and some fixes

tiagovtristao:
- Simplify debugging logic by relying less on state and more on config.
- Add logic to be able to explode pex contents for access by vscode.
- Minor fixes.

### DIFF
--- a/rules/builtins.build_defs
+++ b/rules/builtins.build_defs
@@ -263,19 +263,14 @@ def licenses(licences):
     assert CONFIG.BAZEL_COMPATIBILITY, 'licenses() can only be called in Bazel compat mode'
     package(default_licences = licences)
 
+
 def json(value) -> str:
     """Returns a JSON-formatted representation of a plz value"""
     pass
 
+
 def breakpoint():
     """Breaks into an interactive debugging session."""
-    pass
-
-def is_debug_label(name:str) -> bool:
-    """Checks whether a rule's name within a given package matches the label passed to the
-    plz debug command. This is used to modify the build definition logic for that specific
-    label to make it debuggable.
-    """
     pass
 
 def is_semver(s:str) -> bool:
--- a/rules/builtins.build_defs
+++ b/rules/builtins.build_defs
@@ -263,14 +263,19 @@ def licenses(licences):
     assert CONFIG.BAZEL_COMPATIBILITY, 'licenses() can only be called in Bazel compat mode'
     package(default_licences = licences)
 
-
 def json(value) -> str:
     """Returns a JSON-formatted representation of a plz value"""
     pass
 
-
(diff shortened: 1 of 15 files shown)


rules/builtins.build_defs
```diff
@@ -263,16 +263,21 @@ def licenses(licences):
...
 def json(value) -> str:
     """Returns a JSON-formatted representation of a plz value"""
     pass
 
-
 def breakpoint():
     """Breaks into an interactive debugging session."""
     pass
 
+def is_debug_label(name:str) -> bool:
```
tiagovtristao, line 274: This provides a more ergonomic mechanism for rule authors to modify their targets when they match the target passed to `plz debug`, if they want to support debuggable targets.

This also fixes a previous issue where the only check being done was through `CONFIG.get("DEBUG")`, which meant unwanted targets were being modified as well, which was causing a few issues.

src/please.go
```diff
@@ -169,9 +169,10 @@ var opts struct {
 	} `command:"cover" description:"Builds and tests one or more targets, and calculates coverage."`
 
 	Debug struct {
-		Debugger string `short:"d" long:"debugger" description:"Name of supported debugger"`
-		Port     int    `short:"p" long:"port" description:"Debugging server listen port"`
-		Args     struct {
+		Debugger  string `short:"d" long:"debugger" description:"Name of supported debugger"`
+		Port      int    `short:"p" long:"port" description:"Debugging server listen port"`
+		BuildOnly bool   `long:"build_only" description:"Only builds the artifacts skipping the debug session."`
```
tiagovtristao, line 174: This is used by vscode to access third party sources included in the pex. If the pex is executed in a sandboxed environment we can't access them.

samwestmoreland (approved):
Lgtm! With the python rule changes, I was thinking we want to be careful of the Please defs and plugin defs diverging too much, but I guess it's not that much of an issue. We should be able to just drop these changes right in.

Also these pex tool changes will need to be dropped into the plugin. I'm hoping to land https://github.com/please-build/python-rules/pull/23 today so I can look into copying these changes over too.

tiagovtristao:
> Lgtm! With the python rule changes, I was thinking we want to be careful of the Please defs and plugin defs diverging too much, but I guess it's not that much of an issue. We should be able to just drop these changes right in.
> 
> Also these pex tool changes will need to be dropped into the plugin. I'm hoping to land [please-build/python-rules#23](https://github.com/please-build/python-rules/pull/23) today so I can look into copying these changes over too.

Yes, the changes are straightforward to copy. I don't foresee needing to add more changes.

Tatskaari (approved):
Nice one! Looks good to me.

PR #2216, opened 2021-12-08
Stop plugins from inheriting their config from the host repo

Tatskaari:
This PR still isn't quite right but I will describe the problem here. 

# Setup

Imagine a future in which we've create the go plugin. The `go-rules` subrepo has some config like this:

```ini
[Plugin "go"]
ImportPath = github.com/please-build/go-rules
; a go_toolchain() rule downloading the 1.16 go sdk
GoTool = //third_party/go:toolchain|go
; This is a go_binary() rule for the please_go tool 
PleaseGoTool = //tools/please_go:binary 
DefaultDep = //third_party/go:foo
...

[PluginDefinition]
Name = go
Description = ...

[PluginConfig "go_tool"]
ConfigKey = GoTool
DefaultValue = go

[PluginConfig "import_path"]
ConfigKey = ImportPath
Optional = true

[PluginConfig "please_go_tool"]
ConfigKey = PleaseGoTool
; This is a remote file rule to download the tool from the releases page
DefaultValue = //tools/please_go 

; This config value is contrived nonsense for the sake of the example. Don't think about it too much.  
[PluginConfig "default_deps"]
ConfigKey = DefaultDep
Repeatable = true
Optional = true

...
```

Now imagine a repo that then consumes this plugin with the following config:
```ini
[Plugin "go"]
ImportPath = github.com/awsome-user/amazing-project
; a go_toolchain() rule downloading the 1.17 go sdk (different to the plugin)
GoTool = //third_party/go:toolchain|go
DefaultDep = //third_party/go:bar
```

They then have some go code they want to compile in `//src:main`:
```python
subinclude("///go//build_defs:go")

go_binary(
    name = "main",
    srcs = ["main.go"],
    # Not really important what this is but maybe it's a bunch of extensions to the Please SDK to manipulate go build rules. 
    deps = ["///go//lib:sdk"], 
)
```

# Issues with resolving labels

There are two contexts in which we need to load config from the plugin repo: 
1) When we're parsing a package within a subrepo. This is true for subrepos and plugins. 
2) When subincluding from a subrepo that has plugin config. We need to load the plugin definition from the subrepo to properly bind config values to the asp `CONFIG.PLUGIN` section. 

We need to be aware of these two context that we're loading config from especially with regards to build labels.

## Parsing a package
 
Let's look at the default deps config above. Prior to this PR, we would layer the subrepo config on top of the host repo's config. This means that when parsing a package in the subrepo, the `DefaultDeps` would end up as `[]string{"//third_party/go:foo", "//third_party/go:bar"}`.  As `//third_party/go:bar` is defined in the host repo, this should actually be `@//third_party/go:bar`. 

By not loading the config from the host repo, we avoid this problem because we always know the config values are defined in the current subrepo. This isn't perfect though. For example, we want to use the same `GoTool` config between the host and plugin repo otherwise we'd get frame pointer errors. 

A simple but slightly grungy solution to this problem might be to allow the plugin to define the config options that should be inherited from the host repo:
```init
[PluginConfig "go_tool"]
ConfigKey = GoTool
DefaultValue = go
Inherit = true
```

We could then ensure the labels are made relative to the correct repo when copying them across. This can be handled similarly default values when loading plugin config into the package's `CONFIG.PLUGIN` dict. We may want to switch this to default to true though. 

Note that the `loadPluginConfig()` function takes in the package state, and the subrepo state. This is built around subincludes where the subrepo state is the state of the subrepo in the subinclude label. When parse package, we set these two to the same. Going forward, this needs to change. The state should be the parent state that created the subrepo state from. 

## Subincluding a build definition

We have the reverse problem when subincluding build definitions from a plugin repo. Say we have a build defs file like so in `///go//build_defs:go`:
```python
def go_mock(...):
    ...
    tools = ["//third_party/go:mockgen"]
    ...
    return build_rule(...)

```
When working in the plugin repo, this will work as expected, however when working outside of the plugin repo, this label will be resolved relative to the host repo instead. The introduction of a new psudo-subrepo called "self" was designed to help here. The label could be rewritten to `///self//third_party/go:mockgen`, however resolving self in this context is not trivial. 

The way this works currently requires that the label be a config value:
```ini
[PluginConfig "mockgen_tool"]
ConfigKey = MockGenTool
DefaultValue = ///self//third_party/go:mockgen
```

Because the default values for config values are always defined in the plugin, we know how to resolve self here. We can also achieve something similar without self here though. When loading the default config, we could inject the plugin's subrepo into `//third_party/go:mockgen` in much the same way. 

Perhaps this is good enough. Resolving self in the subinclude context is very tricky as we loose information about where the label was defined. The expression is evaluated in the host repo's package so the current package's subrepo doesn't correspond to the subrepo in which the string was defined in. To do this properly we'd need to annotate string expressions in the ast so we know where they were defined.  

I don't think we can get self to work as we intended it to without significant work to the parser. Perhaps we should remove it.

# Traditional subrepos

For subrepos that aren't plugins, I think we want to do the opposite, as in we should load the host repo config by default. The subrepo can then opt out of this is the incorrect thing to do. We should also not load the subrepo's `.plzconfig` unless they opt in. 

The rational here is that plugins can be carefully designed so they only load the correct values from the host repo. Historically subrepo's always just used the config from the host repo so to be backwards compatible, we should continue to do so. Additionally, if we start loading the subrepo's `.plzconfig`, we might inherit a config option like `Go.GoTool`, which would be awkward to undo. 

Doing it this way leaves the option open to provide a specific set of config values to load via the `config` field on `subrepo()`. If we want to load in the subrepo's config file, we can always opt in. This could be done via an additional flag.


rules/subrepo_rules.build_defs
```diff
@@ -270,3 +270,41 @@ def local_repository(name:str, path:str):
...
+        hashes = hashes,
+    )
+    extract_rule = build_rule(
+        name = name,
+        srcs = [remote_rule],
+        tools = [CONFIG.JARCAT_TOOL],
+        outs = [name],
+        cmd = " && ".join([
+          "$TOOL x $SRCS -o _tmp",
+          "mv $(dirname $(find . -name .plzconfig | sort | head -n 1)) $OUT",
```
Tatskaari, line 302: This provides some niceties so that the download just has to contain a please repo somewhere inside it.

src/parse/asp/objects.go
```diff
@@ -944,31 +944,50 @@ func newConfig(state *core.BuildState) *pyConfig {
 		}
 	}
 
-	loadPluginConfig(state.Config, state, c)
+	loadPluginConfig(state, state, c)
```
samwestmoreland, line 947: This took me a while to get my head around but I guess it makes sense. There are two distinct cases here right? Either ones a plugin and the other isn't, in which case we're parsing the plugin config in the host-repo context, or we're in a plugin and are parsing the config in the plugin context? Neat!

samwestmoreland (approved):

src/parse/asp/objects.go
samwestmoreland: Oh sorry - just read the comment. That's pretty much what you were talking about.

src/parse/asp/builtins.go
```diff
@@ -987,11 +994,20 @@ func selectTarget(s *scope, l core.BuildLabel) *core.BuildTarget {
 	return subincludeTarget(s, l)
 }
 
+const (
```
peterebden, line 997: Could chuck these inside the `subrepo` function to be clear that's what they refer to (would also mean you wouldn't need to repeat `Subrepo` in each one)

src/parse/asp/config.go
```diff
@@ -0,0 +1,247 @@
...
+
+func toPyObject(key, val, toType string) pyObject {
+	if toType == "" || toType == "str" {
+		return pyString(val)
+	}
+
+	if toType == "bool" {
+		val = strings.ToLower(val)
+		if val == "true" || val == "yes" || val == "on" {
+			return pyBool(true)
```
peterebden, line 225: these should just be `return True` and `return False` (bools are singletons)

peterebden (approved):
OK this seems like it could be a go to me. Nice!

tiagovtristao (approved):
Finally got around to look into this PR. Looks good! I think storing the `RepoConfig` and the parent-child relationship between different contexts should give us most (if not all) of the information we need to get this right.

Thinking of it and how flexible plugins are with deciding what to load, it almost feels that it is a generalised case for subrepos, where our current `subrepo` rule is a more specific implementation - this if I understood things correctly.

PR #2222, opened 2021-12-10
Release on v16

Tatskaari:


.circleci/config.yml
```diff
@@ -350,7 +350,7 @@ workflows:
             - test-http-cache
           filters:
             branches:
-              only: master
+              only: v16
```
peterebden, line 353: do we not want this to be something like
```
only:
 - master
 -  v16
```
?

Tatskaari: It's only being merged into the v16 branch so it doesn't make any difference AFAIK

Maybe that's more correct. I can change it.

peterebden: Ah right, I didn't quite get that point. Fair enough.

peterebden (approved):

PR #2227, opened 2021-12-16
Don't show line-by-line coverage results if things failed

peterebden:
This has annoyed me a few times; if your tests fail then `-l` gives you a sea of red, and you have to scroll up past it to find out what went wrong.

Given that coverage is always zero on test failure, there seems to be no downside to this (although you could argue maybe we should be interpreting it anyway, in which case it might provide a little info?).


tiagovtristao (approved):
This makes sense to me too. Having to scroll up is a bit annoying when the test fails.

samwestmoreland (approved):
Seems reasonable to me!

PR #2228, opened 2021-12-16
Replaces hard-codegenned maps with a generic implementation

peterebden:
This won't actually build obvs but illustrates what it might look like.


peterebden:
I think this is pretty decent; the main thing I don't like is always having to specialise `New[int, int]` when you call it. I can't think of a sensible way of getting V in as a type param though.

peterebden:
Closing this, was only sharing for reference (although think I'll revisit at some future date when we start using 1.18 in earnest)

PR #2235, opened 2021-12-29
Parallel hashing of directories

peterebden:
This is really a proof-of-concept to show how we can hash big directory trees faster by hashing the files in them in parallel. 

```
goos: linux
goarch: amd64
BenchmarkHashTree
BenchmarkHashTree/1Way
BenchmarkHashTree/1Way-12                      1        1048907728 ns/op         296.96 MB/s           297.0 MB/s/thread
BenchmarkHashTree/2Way
BenchmarkHashTree/2Way-12                      2         558632038 ns/op         557.58 MB/s           278.8 MB/s/thread
BenchmarkHashTree/4Way
BenchmarkHashTree/4Way-12                      4         310531512 ns/op        1003.07 MB/s           250.8 MB/s/thread
BenchmarkHashTree/8Way
BenchmarkHashTree/8Way-12                      5         231179949 ns/op        1347.36 MB/s           269.5 MB/s/thread
PASS
```

(The numbers can be surprisingly wonky, will try to redo this on a machine without corporate nonsense)

Unfortunately it is a complete change to our hash strategy for directories; the previous one writes the contents of all files into a single hash, the new one hashes them all separately them combines them together at the top. The previous one simply cannot be parallelised as is.

Still a bit to do; the parallelism limit should apply to all hashes, not just directories, and some of the other stuff could be simplified (e.g. maybe we could use `x/sync/singleflight` rather than the current mechanism).

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -1143,12 +1143,12 @@ func NewBuildState(config *Configuration) *BuildState {
 		pendingRemoteActions: make(chan Task, 1000),
 		hashers: map[string]*fs.PathHasher{
 			// For compatibility reasons the sha1 hasher has no suffix.
-			"sha1":   fs.NewPathHasher(RepoRoot, config.Build.Xattrs, sha1.New, "sha1"),
-			"sha256": fs.NewPathHasher(RepoRoot, config.Build.Xattrs, sha256.New, "sha256"),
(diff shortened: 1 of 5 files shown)


peterebden:
After discussion we think there's enough stuff planned for v17 that breaking hashes as well probably isn't a good idea.

PR #2242, opened 2022-01-04
Handle aliases to multi-level commands like `run parallel`

peterebden:
Related to #2154, needs to use the same `ActiveFullCommand` logic when handling aliases too.


samwestmoreland (approved):

Tatskaari (approved):

tiagovtristao (approved):
Thanks, nice catch!

PR #2243, opened 2022-01-04
Change `plz run sequential` to have the same output flag as parallel

peterebden:
It would be nicer for people if these two mirror one another.

### DIFF
--- a/src/run/run_step.go
+++ b/src/run/run_step.go
@@ -25,12 +25,12 @@ import (
 
 var log = logging.MustGetLogger("run")
 
-type ParallelOutput string
+type ProcessOutput string
 
 const (
-	Default        ParallelOutput = "default"
-	Quiet          ParallelOutput = "quiet"
-	GroupImmediate ParallelOutput = "group_immediate"
+	Default        ProcessOutput = "default"
+	Quiet          ProcessOutput = "quiet"
+	GroupImmediate ProcessOutput = "group_immediate"
 )
 
 // Run implements the running part of 'plz run'.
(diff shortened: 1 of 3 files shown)


Tatskaari (approved):

src/run/run_step.go
```diff
@@ -89,15 +71,35 @@ func Parallel(ctx context.Context, state *core.BuildState, labels []core.Annotat
 	return 0
 }
 
+// runWithOutput runs a subprocess with the given output mechanism.
+func runWithOutput(ctx context.Context, state *core.BuildState, label core.AnnotatedOutputLabel, args []string, output ProcessOutput, remote, env, detach, inTmp bool, dir string) error {
```
tiagovtristao, line 75: Much better having this logic here.

tiagovtristao (approved):
`GroupImmediate` should yield the same results as `Default` for `sequential`, but I guess this isn't an issue. None of the options are mutual exclusive, so I guess it makes sense for now to have both subcommands aligned. Thanks.

peterebden:
> `GroupImmediate` should yield the same results as `Default` for `sequential`, but I guess this isn't an issue. None of the options are mutual exclusive, so I guess it makes sense for now to have both subcommands aligned. Thanks.

I think `GroupImmediate` adds a heading with the label of each target that's being run? So it's slightly different - but agreed the difference is much less significant than for `parallel`.

samwestmoreland (approved):

tiagovtristao:
> > `GroupImmediate` should yield the same results as `Default` for `sequential`, but I guess this isn't an issue. None of the options are mutual exclusive, so I guess it makes sense for now to have both subcommands aligned. Thanks.
> 
> I think `GroupImmediate` adds a heading with the label of each target that's being run? So it's slightly different - but agreed the difference is much less significant than for `parallel`.

Ah yes, it does that, so you can say they are not exactly the same :-)

PR #2247, opened 2022-01-05
Update gtest

Tatskaari:


third_party/cc/gtest.build
```diff
@@ -0,0 +1,9 @@
+cc_library(
+    name = "gtest",
+    srcs = glob(["googletest/src/*.cc"], exclude=["googletest/src/gtest-all.cc"]),
+    hdrs = glob(["googletest/include/**.h","googletest/src/*.h"]),
```
peterebden, line 4: Think this should be
```
hdrs = glob(["googletest/include/**.h"]),
private_hdrs = glob(["googletest/src/*.h"]),
```
It's a little bit purist but the headers in `src` presumably aren't meant to be consumed by other things.

PR #2249, opened 2022-01-06
fix: allow a token to access private github repos

afterthought:
Hi. I want to start using build defs in private repos. I made this change to the subrepos, but I think I need help testing it. I have my project that uses a few public repos for build_defs already. Is there a page showing me how to get a custom build and test this out?

Thanks...

### DIFF
--- a/rules/subrepo_rules.build_defs
+++ b/rules/subrepo_rules.build_defs
@@ -193,6 +193,12 @@ def github_repo(name:str, repo:str, revision:str, build_file:str=None, hashes:st
 
     prefix = f'{repo}-{prefix}'
 
+    headers = {}
+    if access_token:
+        headers = {
+            'Authorization': f'token {access_token}'
+        }
+
     return new_http_archive(
         name = name,
         urls = [f'https://github.com/{org}/{repo}/archive/{revision}.zip'],
@@ -201,9 +207,7 @@ def github_repo(name:str, repo:str, revision:str, build_file:str=None, hashes:st
         build_file = build_file,
         hashes = hashes,
         config = config,
-        headers = {
-            'Authorization': f'token {access_token}'
-        },
+        headers = headers,
         bazel_compat = bazel_compat,
     )
 
--- a/rules/subrepo_rules.build_defs
+++ b/rules/subrepo_rules.build_defs
@@ -202,7 +202,7 @@ def github_repo(name:str, repo:str, revision:str, build_file:str=None, hashes:st
         hashes = hashes,
         config = config,
         headers = {
-            'Authorization': f"token {access_token}"
+            'Authorization': f'token {access_token}'
         },
         bazel_compat = bazel_compat,
     )
--- a/rules/subrepo_rules.build_defs
+++ b/rules/subrepo_rules.build_defs
@@ -59,7 +59,7 @@ def http_archive(name:str, urls:list, strip_prefix:str=None, hashes:str|list&sha
 def new_http_archive(name:str, urls:list, build_file:str=None, build_file_content:str=None,
                      strip_prefix:str=None, strip_build:bool=False, hashes:str|list&sha256=None,
                      username:str=None, password_file:str=None, headers:dict={}, secret_headers:dict={},
-                     config:str=None, bazel_compat:bool=False, visibility:list=None):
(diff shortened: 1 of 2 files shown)


rules/subrepo_rules.build_defs
```diff
@@ -151,6 +160,7 @@ def new_local_repository(name:str, path:str):
 
 def github_repo(name:str, repo:str, revision:str, build_file:str=None, hashes:str|list=None,
                 strip_prefix:str=None, strip_build:bool=False, config:str=None,
+                access_token:str=None,
```
Tatskaari, line 163: This looks like it's the actual access token as a string. We can't have people checking that into their codebase. We should pass this in via a secret. Doesn't github support basic auth? We could pass through the username and password file fields.

rules/subrepo_rules.build_defs
```diff
@@ -190,6 +201,9 @@ def github_repo(name:str, repo:str, revision:str, build_file:str=None, hashes:st
         build_file = build_file,
         hashes = hashes,
         config = config,
+        headers = {
+            'Authorization': f'token {access_token}'
+        },
```
Tatskaari, line 206: I think we'd need to guard against this otherwise we'll be passing an incomplete authorisation header to public repos

Tatskaari (changes requested):

rules/subrepo_rules.build_defs
afterthought: I was going to pass the string $GITHUB_TOKEN to use the env var present in GitHub Actions and in Codespaces. I'm not against secret files. I just wasn't sure how that would work. Would that require an absolute path? Or is the idea to make it relative to step up out of the git project? 

Re: basic auth on GitHub: https://developer.github.com/changes/2020-02-14-deprecating-password-auth/

rules/subrepo_rules.build_defs
afterthought: done

rules/subrepo_rules.build_defs
afterthought: Thoughts on this @Tatskaari? I'm happy to help fix how you want it. I'm starting to ramp up other teams and need to start sharing build defs in private repos.

Tatskaari (approved):
Thanks for the update :) I think I see where you're going with this. 

I pulled your branch to test if it works and encountered a few issues with remote file... I hope you don't mind but I've pushed those fixes to your branch. I'm happy with this if you are. I'll ask my team to have a quick look too.

Tatskaari:
So we can access private repos like so:
```
github_repo(
    name = "supersecretsauce",
    repo = "thought-machine/supersecretsauce",
    revision = "master",
    access_token = "GIT_TOKEN",
)
```

and `export GIT_TOKEN=asdfasdfa && plz build //:supersecretsauce`

samwestmoreland (approved):
Nice! Looks good to me.

rules/subrepo_rules.build_defs
```diff
@@ -182,6 +196,14 @@ def github_repo(name:str, repo:str, revision:str, build_file:str=None, hashes:st
 
     prefix = f'{repo}-{prefix}'
 
+    headers = {}
+    pass_env = []
+    if access_token:
+        headers = {
+            'Authorization': 'token \$%s' % access_token,
```
tiagovtristao, line 203: Is this meant to be escaped?

afterthought:
Is it possible to start testing this? Was there a prerelease build maybe?

afterthought:
@Tatskaari @tiagovtristao  I tested this and it looks like there's a problem. I had verified that the curl with the auth header would work. However, I must have used a flag to follow redirects and didn't verify that remote_file also does this. Because remote_file isn't following the redirect, the following is downloaded as the zip file. 

    <html><body>You are being <a href="https://codeload.github.com/<org>/<repo>/zip/<myref - redacted>?token=<redacted>">redirected</a>.</body></html>

I'm not comfortable modifying remote_file to support redirect follows. Seems like a risky change. What are your thoughts?

afterthought:
I just made this (#2275) PR to address the problem I found. I think it's safe enough. But you let me know if it's OK.

afterthought:
I just did built plz myself and tested this. I confirmed that it now downloads properly with the -L flag.

PR #2253, opened 2022-01-10
Plz help plugins

samwestmoreland:
https://github.com/thought-machine/please-issues/issues/50


samwestmoreland:
Screenshot 2022-01-20 at 14 36 08

src/help/help.go
```diff
@@ -64,6 +68,103 @@ func help(topic string) string {
...
+	return ""
+}
+
+func helpForPlugin(topic string) string {
+	config, err := core.ReadDefaultConfigFiles(nil)
+	if err != nil {
+		panic("Failed to read config")
+	}
+	if _, ok := config.Plugin[topic]; ok {
+		message := fmt.Sprintf("${BOLD_BLUE}%v${RESET} is a plugin defined in the .plzconfig file.", topic)
```
Tatskaari, line 86: Nice! Get some colour in there!

src/help/help.go
```diff
@@ -64,6 +68,103 @@ func help(topic string) string {
...
+		if val, ok := config.Plugin[topic].ExtraValues["subrepo"]; ok {
```
Tatskaari, line 87: I remember us talking about having some sort of convention where we might try and find the plugin in a default place? I can't remember what we said... we can have a think about that though, as I'd like this to work even if they don't specify the subrepo.

src/help/help.go
```diff
@@ -64,6 +68,103 @@ func help(topic string) string {
...
+			subrepo := val[0]
+			buildLabel := core.BuildLabel{PackageName: "", Name: subrepo, Subrepo: subrepo}
```
Tatskaari, line 89: Doesn't this cause a parse error because `///{subrepo}//:{subrepo}` doesn't exist? I think `name` should be the psudo-label `:all`?

src/core/config.go
```diff
@@ -596,7 +596,10 @@ type Configuration struct {
 	Alias            map[string]*Alias  `help:"Allows defining alias replacements with more detail than the [aliases] section. Otherwise follows the same process, i.e. performs replacements of command strings."`
 	Plugin           map[string]*Plugin `help:"Used to define configuration for a Please plugin."`
 	PluginDefinition struct {
-		Name string `help:"The name of the plugin"`
+		Name          string   `help:"The name of the plugin"`
+		Description   string   `help:"A description of what the plugin does"`
+		BuildDefsDir  []string `help:"Directory to look in when prompted for help topics that aren't known internally" example:"build_defs"`
+		Documentation string   `help:"A link to the documentation for this plugin"`
```
Tatskaari, line 602: nit: might be confused with the literal documentation. Perhaps `DocumentationSite` would be better?

src/help/help.go
```diff
@@ -64,6 +68,103 @@ func help(topic string) string {
...
+			state := newState()
+
+			// Parse the subrepo (Run reads the plugin config into config)
+			plz.Run([]core.BuildLabel{buildLabel}, nil, state, config, state.TargetArch)
+			downloadBuildLabel := core.BuildLabel{PackageName: "", Name: "_" + subrepo + "#download", Subrepo: ""}
```
Tatskaari, line 94: Is this just to try and figure out where it's downloaded from? Might be a little grungy as it depends on the internal structure of the build target. It also seems to assume the target is in the root of the repo. You could use a label though. If the subrepo rule labels itself somehow e.g. `labels = ["plugin_repo:https://github.com/please-build/python-rules"]`, you can fetch that off the target directly without having to delve into the details of the rule.

src/help/help.go
```diff
@@ -64,6 +68,103 @@ func help(topic string) string {
...
+			state = state.Graph.Subrepo(subrepo).State
+			config := state.Config
+			if config.PluginDefinition.Description != "" {
+				message += "\n" + config.PluginDefinition.Description + "\n"
+			}
+			if config.PluginDefinition.Documentation != "" {
+				message += "\n" + config.PluginDefinition.Documentation + "\n"
+			}
+			configOptions := ""
+			for _, v := range config.PluginConfig {
```
Tatskaari, line 114: Might be nice to handle repeated values here, and change `str` to `string` to be inline with the please website:
https://please.build/config.html#parse.buildfilename

src/help/help.go
```diff
@@ -64,6 +68,103 @@ func help(topic string) string {
...
+						arglist += arg.Name + ")"
+					}
+				}
+				buildDefs += fmt.Sprintf("${BLUE}%v${RESET}\n", arglist)
+			}
+			if buildDefs != "" {
+				message += "\n${YELLOW}And provides the following build defs:${RESET}\n" + buildDefs
+			}
+		} else {
+			log.Warningf("To see more information, specify the subrepo field for the plugin %v", topic)
```
Tatskaari, line 163: Might make more sense if we said: 
> specify the `Subrepo` field in the `[Plugin "%v"]` section in your .plzconfig

src/help/help.go
```diff
@@ -64,6 +68,103 @@ func help(topic string) string {
...
+					valueType,
+					optional,
+					v.DefaultValue)
+			}
+			if configOptions != "" {
+				message += "\n${YELLOW}This plugin has the following options:${RESET}\n" + configOptions
+			}
+
+			p := asp.NewParser(state)
+			buildFuncMap := map[string]*asp.Statement{}
```
Tatskaari, line 134: I wonder if we can move some of these things out into their own functions to clean this up a bit e.g. `buildFuncMap = readBuildFunction(state.Config.PluginDefinition.BuildDefsDir)`

Tatskaari (commented):
Looking good! Just a few comments.

src/help/help.go
samwestmoreland: So it looks like it's not super straightforward to get the URL from the remote_file rule because there can be multiple URLs attached to the rule. I think one day/in another PR it might be nice for the remote_file to label itself with what URL it ultimately went with.

docs/config.html
```diff
@@ -402,6 +402,44 @@ <h2 id="plugindefinition" class="title-2">[PluginDefinition]</h2>
...
+  <ul class="bulleted-list">
+    <li>
+      <div>
+        <h3 class="mt1 f6 lh-title" id="plugindefinition.description">
+          Description <span class="normal">(string)</span>
+        </h3>
+
+        <p>
+        A description of the plugin. If set, this can be read and output by
+        <code class="code">plz help <plugin></code>.
```
tiagovtristao, line 415: You might have to escape `<plugin>` since it's not a valid HTML tag - `&lt;plugin&gt;` should do.

You can also run `plz run //docs:server` and go to `localhost:8080` to verify the changes.

docs/config.html
```diff
@@ -402,6 +402,44 @@ <h2 id="plugindefinition" class="title-2">[PluginDefinition]</h2>
...
+    </li>
+    <li>
+      <div>
+        <h3 class="mt1 f6 lh-title" id="plugindefinition.builddefsdir">
+          BuildDefsDir <span class="normal">(repeatable string)</span>
+        </h3>
+
+        <p>
+          Tells Please about any directories in the plugin that contain custom
+          build definitions. These are used by `plz help <plugin>` to provide
```
tiagovtristao, line 427: Same here.

src/help/help.go
```diff
@@ -64,9 +68,123 @@ func help(topic string) string {
...
+	if err != nil {
+		panic("Failed to read config")
+	}
+	if _, ok := config.Plugin[topic]; ok {
+		message := fmt.Sprintf("${BOLD_BLUE}%v${RESET} is a plugin defined in the ${GREEN}.plzconfig${RESET} file.\n", topic)
+
+		// This is if there's a subrepo specified for the plugin. If there isn't, we'll check for the subrepo in the plugins/
+		var subrepoName string
+		buildLabel := core.BuildLabel{Name: "all"}
+		if val, ok := config.Plugin[topic].ExtraValues["subrepo"]; ok {
```
tiagovtristao, line 92: This should now be `config.Plugin[topic].Target` as per the latest master. This new `Target` field is also compulsory (as per https://github.com/thought-machine/please/commit/3d64e323719c5ad284655a2e11b42d404ea72f31#diff-bfaa51a2df379d9e9351315998d3720e62cd16b8ebd972e450e9508dcde68e9dR131), so it might simplify the logic here as well.

src/help/help.go
```diff
@@ -64,9 +68,123 @@ func help(topic string) string {
...
+			valueType,
+			optional,
+			v.DefaultValue)
+	}
+	if configOptions != "" {
+		message += "\n${BOLD_YELLOW}This plugin has the following options:${RESET}\n" + configOptions
+	}
+
+	buildFuncMap := map[string]*asp.Statement{}
+	populatePluginBuildFuncs(buildFuncMap, subrepo)
```
tiagovtristao, line 143: Might be nicer to return `buildFuncMap` from `populatePluginBuildFuncs` since it's meant to be only mutated within it.

tiagovtristao (commented):
Very nice.

PR #2254, opened 2022-01-10
Stop duplicating go_module: labels on go_module() and go_mod_download()

Tatskaari:


rules/go_rules.build_defs
```diff
@@ -1272,7 +1272,7 @@ def go_module(name:str='', module:str, version:str='', download:str='', deps:lis
             "go": f":{name}", # provide the filegroup otherwise exported deps don't work
             "go_src": download,
         },
-        labels = [f"go_package:{i}" for i in install] + [f"go_module:{module}"],
+        labels = [f"go_package:{i}" for i in install],
```
Tatskaari, line 1275: This label was applied in `go_mod_download()` which is used by this rule.

peterebden (approved):

PR #2261, opened 2022-01-15
Show the download rate of files in interactive display

peterebden:
This displays like:
```
=> [ 5.1s] //test/java_rules/java_toolchain/src/main/java/net/thoughtmachine:java11 Fetching... (3.8%, 1.4 MB/s, est 2m9s remaining)
```
Thought it was kind of nice to see (although the calculation is a bit crap, it should really be a rolling average over the last few seconds).

Bonus: does not add any size to BuildTarget because it slotted into an existing 32 bytes of padding.

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -1097,10 +1097,11 @@ func fetchOneRemoteFile(state *core.BuildState, target *core.BuildTarget, url st
 	var r io.Reader = resp.Body
 	if length := resp.Header.Get("Content-Length"); length != "" {
 		if i, err := strconv.Atoi(length); err == nil {
-			r = &progressReader{Reader: resp.Body, Target: target, Total: float32(i)}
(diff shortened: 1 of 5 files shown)


Tatskaari (commented):
LGTM though I think there's still a bit of room for improvement. 

If we made `Progress` an interface with a pointer value, it could store more bespoke information without blowing out the build target struct size. I think there's some slightly grungy code to parse out progress updates from stdout that might benefit from that too i.e. storing a custom regex?

peterebden:
OK sounds reasonable (insofar as it is conceivable that my PR could possibly be improved in any way)

Tatskaari:
> OK sounds reasonable (insofar as it is conceivable that my PR could possibly be improved in any way)

Yeah, mostly just pondering for the future. Might create a ticket.

PR #2263, opened 2022-01-17
updating action digest so data can always be downloaded

tobyfielding1:
I had an issue with data dependencies failing to be downloaded when remote execution is enabled with the following error:
`Failed to retrieve action result for <data-dep-tgt>`
It appears that please was looking at the wrong digest to get the data because the digest added to `unstampedBuildActionDigests` is incorrect. It seems that at least some of the time the uploaded action digest doesn't match the original build action digest, and so we need to update the original digest before storing it in `unstampedBuildActionDigests`.

After this change I have not seen the above error again.

### DIFF
--- a/src/remote/remote.go
+++ b/src/remote/remote.go
@@ -594,7 +594,9 @@ func (c *Client) execute(tid int, target *core.BuildTarget, command *pb.Command,
 		}
 	}
 	// We didn't actually upload the inputs before, so we must do so now.
-	command, digest, err := c.uploadAction(target, isTest, false)
+	command, newDigest, err := c.uploadAction(target, isTest, false)
+	// Update the original digest to ensure it matches the uploaded action.
+	*digest = *newDigest
 	if err != nil {
 		return nil, nil, fmt.Errorf("Failed to upload build action: %s", err)
 	}


src/remote/remote.go
```diff
@@ -594,7 +594,9 @@ func (c *Client) execute(tid int, target *core.BuildTarget, command *pb.Command,
 		}
 	}
 	// We didn't actually upload the inputs before, so we must do so now.
-	command, digest, err := c.uploadAction(target, isTest, false)
+	command, newDigest, err := c.uploadAction(target, isTest, false)
+	// Update the original digest to ensure it matches the uploaded action.
```
peterebden, line 598: How is this different from before? Are you deliberately mutating the passed-in digest for other things to observe it?

tobyfielding1: Yes because there is a map storing the digests which is used later to pull data for a target, 'unstampedBuildActionDigests'. We need to get the digest out of this function. It is probably better to just return it, so I can do that instead.

Tatskaari: I don't think we should be updating the action digest. It should be the same so I think we should nail down why they're different. I have an idea though.

Tatskaari:
I wonder if the reason these digests were different has something to do with the logic for storing the action result. For context:

Stamped targets have access to env variables which contain the SCM information so change with every commit. We omit these from the digest we use to look up action results to improve incrementality. This code is meant to handle that:

This calculates the unstamped digest for stamped targets:
```golang
	var unstampedDigest *pb.Digest
	if target.Stamp {
		command, digest, err := c.buildAction(target, false, false)
                ...
		unstampedDigest = digest
	}
```

Then we go ahead and actually execute the target:
```golang
command, stampedDigest, err := c.buildAction(target, false, true)
```

This is slightly confusing. The `stampedDigest` is actually the unstamped digest for build actions that set `stamp = False`. This means we have the following logic to store the action digest in the `unstampedBuildActionDigests` map:

```golang
	if target.Stamp && err == nil {
		...
		c.unstampedBuildActionDigests.Put(target.Label, unstampedDigest)
	} else {
		c.unstampedBuildActionDigests.Put(target.Label, stampedDigest)
	}
``` 

The problem could be to do with when err is not nill. which would result in us caching via the stamped digest for stamped target.

PR #2265, opened 2022-01-18
Only exit plz fmt with exit code 1 on error or if in quiet mode

marcuscaisey:
This PR changes the exit code of `plz fmt` to only be 1 if either
- there was an error during formatting
- `--quiet` is given and `--write` is not 

This differs from the existing behaviour where running `plz fmt` with no options would exit with a 1 even if the file was successfully formatted and output to stdout. The rationale for the `--quiet` and not `--write` combination comes from the description of the `--quiet` flag:
> Don't print corrections to stdout, simply exit with a code indicating success / failure (for linting etc).

Having both successful formatting and error cases exit with 1 messes with tools (like https://github.com/sbdchd/neoformat) which check the error code before doing something (in this case writing the formatters stdout to the current buffer).

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -637,8 +637,8 @@ var buildFunctions = map[string]func() int{
 	"format": func() int {
 		if changed, err := format.Format(config, opts.Format.Args.Files.AsStrings(), opts.Format.Write, opts.Format.Quiet); err != nil {
 			log.Fatalf("Failed to reformat files: %s", err)
-		} else if changed && !opts.Format.Write {
-			return 0
+		} else if changed && opts.Format.Quiet {
+			return 1
 		}
 		return 0
 	},
--- a/src/please.go
+++ b/src/please.go
@@ -638,7 +638,7 @@ var buildFunctions = map[string]func() int{
 		if changed, err := format.Format(config, opts.Format.Args.Files.AsStrings(), opts.Format.Write, opts.Format.Quiet); err != nil {
 			log.Fatalf("Failed to reformat files: %s", err)
 		} else if changed && !opts.Format.Write {
-			return 1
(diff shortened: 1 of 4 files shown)


src/please.go
```diff
@@ -636,7 +636,7 @@ var buildFunctions = map[string]func() int{
 	"format": func() int {
 		if changed, err := format.Format(config, opts.Format.Args.Files.AsStrings(), opts.Format.Write, opts.Format.Quiet); err != nil {
 			log.Fatalf("Failed to reformat files: %s", err)
-		} else if changed && !opts.Format.Write {
+		} else if changed && !opts.Format.Write && opts.Format.Quiet {
```
Tatskaari, line 639: Hmm I think --quiet implies that we're not writing, and we just want to produce an exit code. Perhaps the correct thing to do would be to consider `plz fmt --write --quiet` invalid and just print usage?

marcuscaisey: that makes sense to me, will change

marcuscaisey:
Modified so that providing both `--write` and `--quiet` logs an error. Similarly to how this is dealt with in other commands like https://github.com/thought-machine/please/blob/721dc413b9aa128d99240edbb2de27cc8116b8d2/src/please.go#L1075-L1081

Also added some basic test coverage under `test/plz_format` using all of the different flags

marcuscaisey:
hey @Tatskaari, could you take another look at this when you get a free moment please? (the failed checks look unrelated to my changes btw)

Tatskaari (approved):
Sorry! Seems like this PR fell through the cracks. Looks good to me. Let's get this merged.

marcuscaisey:
thanks man 👍

PR #2266, opened 2022-01-18
Allow hashing based on timestamp for large files

Tatskaari:


rules/misc_rules.build_defs
```diff
@@ -230,6 +230,28 @@ def gentest(name:str, test_cmd:str|list|dict, labels:list&features&tags=None, cm
         env = env,
     )
 
+def export_large_file(name:str, src:str, visibility:list=None, binary:bool=False, test_only:bool&testonly=False):
+       """The same as export_file() but uses a less robust timestamp based hashing algorithm. This is especially usefule
```
samwestmoreland, line 234: useful*

src/core/build_target.go
```diff
@@ -1172,6 +1172,11 @@ func (target *BuildTarget) UnprefixedHashes() []string {
 	return hashes
 }
 
+// HashLastModified is whether we should has the last modified times for this target
```
samwestmoreland, line 1175: should hash*

samwestmoreland (approved):
Lgtm!

rules/misc_rules.build_defs
```diff
@@ -230,6 +230,28 @@ def gentest(name:str, test_cmd:str|list|dict, labels:list&features&tags=None, cm
         env = env,
     )
 
+def export_large_file(name:str, src:str, visibility:list=None, binary:bool=False, test_only:bool&testonly=False):
```
tiagovtristao, line 233: I think it would be nice to support system sources too, what do you think?

tiagovtristao (approved):

rules/misc_rules.build_defs
Tatskaari: I don't know. I feel like a certain level of awkwardness around system sources is good. I'm on the fence about this rule to be honest. I almost feel like this should be difficult to avoid people doing it without due consideration.

PR #2267, opened 2022-01-19
Allow hash checkers to be configured

tiagovtristao:


src/build/build_step_test.go
```diff
@@ -507,6 +530,25 @@ func TestSha1SingleHash(t *testing.T) {
 	}
 }
 
+func newStateWithHashCheckers(label, hashFunction string, hashCheckers []string) (*core.BuildState, *core.BuildTarget) {
+	config, _ := core.ReadConfigFiles(nil, nil)
+	if len(hashFunction) > 0 {
```
peterebden, line 535: is this really what you meant given this is a `string` (but `hashCheckers` is a `[]string`)?

src/build/build_step_test.go
```diff
@@ -507,6 +530,25 @@ func TestSha1SingleHash(t *testing.T) {
 	}
 }
 
+func newStateWithHashCheckers(label, hashFunction string, hashCheckers []string) (*core.BuildState, *core.BuildTarget) {
```
peterebden, line 533: this might be a good case for a `...string`?

peterebden (approved):

src/build/build_step_test.go
tiagovtristao: Yeah, I guess it a little nicer.

src/build/build_step_test.go
tiagovtristao: Funny enough I thought about that, but the end result should be the same (as opposed to `hashFunction != ""`). I guess, idiomatically speaking and for consistency `!= ""` is more in line with checking empty strings. Happy to change.

Tatskaari (approved):

PR #2270, opened 2022-01-23
Fix resolving of special @self keyword

tiagovtristao:
While working on the `python proto` plugin which relies both on https://github.com/please-build/proto-rules and https://github.com/please-build/python-rules as plugins, I kept getting the following error:

```
Build stopped after 29.72s. 1 target failed:
    //tools/wheel_resolver:wheel_resolver                                                                                                                                                                                    
//third_party/python:_grpc#download depends on //tools/wheel_resolver:wheel_resolver, but the directory tools/wheel_resolver doesn't exist: tools/wheel_resolver
```

The `//tools/wheel_resolver:wheel_resolver` is part of the `python` plugin and it is set as the default wheel tool in https://github.com/please-build/python-rules/blob/master/.plzconfig#L74 which uses the special`@self` keyword. `//third_party/python:_grpc#download` is part of the host `python proto` plugin.

The `@self` keyword was being resolved to nothing instead of the the subrepo where it was defined.

TODO: Add test to cover for `@self`.

### DIFF
--- a/src/parse/asp/config.go
+++ b/src/parse/asp/config.go
@@ -89,9 +89,9 @@ func resolveSelf(values []string, subrepo string) []string {
 	ret := make([]string, len(values))
 	for i, v := range values {
 		if core.LooksLikeABuildLabel(v) {
-			l := core.ParseBuildLabel(v, "")
-			if l.Subrepo == "self" {
-				l.Subrepo = subrepo
+			l, err := core.TryParseBuildLabel(v, "", subrepo)
+			if err != nil {
+				panic(err)
 			}
 			// Force the full build label including empty subrepo so this is portable
 			v = fmt.Sprintf("///%v//%v:%v", l.Subrepo, l.PackageName, l.Name)


peterebden (approved):

Tatskaari:
I don't necessarily think this is wrong, but I'm not sure I understand how it solves your issue. The config value is being loaded from a default value, so even though we're loading it for the package `//third_party/python:_grpc#download`, the `subrepo` parameter to `resolveSelf()` should be set to `python`. 

Shouldn't both approaches work the same here? The only difference is how `//tools/wheel_resolver` will get resolved (without `@self`). Before this would be evaluated relative to the host repo, leaving `@self` to specify that it's relative to the current plugin repo. I was planning on getting rid of `@self` though, sticking with the older `/////tools/wheel_resolver` syntax to distinguish between relative and host repo labels.

Tatskaari:
Ah right i see, `@self` is being resolved to `""` in `ParseBuildLabel()`. I'm happy for this to land but I may want to remove `@self` before we tag a proper release.

Tatskaari (approved):

tiagovtristao:
> Ah right i see, `@self` is being resolved to `""` in `ParseBuildLabel()`. I'm happy for this to land but I may want to remove `@self` before we tag a proper release.

Sure, this will unblock me for now.

PR #2273, opened 2022-01-25
Sort the user provided env for rex

Tatskaari:


Tatskaari:
I think this might also contribute to #2263

tobyfielding1:
noice!

src/sandbox/sandbox_linux.go
```diff
@@ -68,7 +68,10 @@ func Sandbox(args []string) error {
 		}
 	}
 	err = syscall.Exec(cmd, args, env)
-	return fmt.Errorf("Failed to exec %s: %s", cmd, err)
+	if err != nil {
+		return fmt.Errorf("Failed to exec %s: %s", cmd, err)
+	}
+	return nil
```
Tatskaari, line 74: not sure how this got in here :disappointed:

peterebden: Guessing there's a newer linter now that didn't used to check this.

I think it's wrong in this case though, `Exec` only ever returns on failure so the code before was legit.

peterebden (approved):
Can't wait for generics so we can have ordered maps...

src/sandbox/sandbox_linux.go
Tatskaari: I guess the linter doesn't know that. This shuts it up so I think it's fine.

peterebden: Yeah IIRC there are ways of marking this kind of thing in C compilers but it's not something Go knows about. I guess it's not a big deal, although maybe that linter could special-case one or two functions like this.

Tatskaari:
+1 on generics... will be great to have some proper data types.

PR #2278, opened 2022-01-27
Allow outputs to be downloaded transitively in remote when debugging is on

tiagovtristao:
Transitive artifacts were missing when running debugging sessions with remote on. This will only apply to debug sessions, but can eventually be exposed if needed.


Tatskaari (approved):

src/core/state.go
```diff
@@ -50,6 +50,18 @@ type Task struct {
 	Run   uint32 // Only present for tests (the run of a build is always zero)
 }
 
+// A OutputDownloadOption is the option for how outputs should be downloaded.
+type OutputDownloadOption uint8
+
+const (
```
samwestmoreland, line 56: Should we use iota here? Or does it not matter because it's not very many constants?

samwestmoreland (approved):
Lgtm!

src/core/state.go
tiagovtristao: I guess iota is better.

PR #2281, opened 2022-01-27
Handle preloads as a special case in parse_step.go

Tatskaari:
This preloads subincludes in a more robust way. The previous implementation just injected the subincludes into the BUILD files for you. This had a major drawback in that we had to hackily exclude any packages that might be needed to facilitate that subinclude. This fell appart when the preloaded subinclude itself has subincludes. 

This PR has the parse step wait for the preloads to be complete before parsing any packages. In order to allow parses that are required for the preloads, any parse tasks that are for subincludes are ommited from this check. Because we can't parse any packages until we've finished preloading, we know that any subincludes must be for the preloaded subincludes. 

Some things still need to be done:

1. Handle preloads in subrepos. We should probably load the preloaded subincludes from `state.RepoConfig` in some situations. 
2. Consider if we should parse the preloads sequentially so that they can depend on each other.  
3. Consider if we want to have a set of preloads for Bazel repos. This could use useful to provide compatibility layers around the Please builtins.


src/core/state.go
```diff
@@ -545,6 +549,39 @@ func (state *BuildState) logResult(result *BuildResult) {
...
+		wg := sync.WaitGroup{}
+		wg.Add(len(state.Config.Parse.PreloadSubincludes))
+
+		preload := func(inc string) {
+			l, err := TryParseBuildLabel(inc, "", "")
+			if err != nil {
+				log.Fatalf("failed to preload subinclude: %v", err)
+			}
+
+			if err := state.queueTarget(l, OriginalTarget, true, true); err != nil {
```
tiagovtristao, line 564: What's the reason for the force build?

tiagovtristao (approved):
Looks good and cleaner!

src/core/state.go
Tatskaari: Was just a copy paste from elsewhere. I'm using `state.WaitForBuiltTarget()`

src/parse/asp/interpreter.go
```diff
@@ -164,7 +164,9 @@ func (i *interpreter) getConfig(state *core.BuildState) *pyConfig {
 
 	i.configMutex.Lock()
 	defer i.configMutex.Unlock()
-	i.config[state.Config] = c
+	if state.FinishedPreloading {
```
Tatskaari, line 167: If we cache this before we finish pre-loading, we'd end up caching the state without the plugin config.

src/parse/asp/config.go
```diff
@@ -82,6 +82,15 @@ func (i *interpreter) newConfig(state *core.BuildState) *pyConfig {
 	base["BUILD_CONFIG"] = pyString(state.Config.Build.Config)
 	base["DEBUG_PORT"] = pyInt(state.DebugPort)
 
+	// Preloaded subincludes don't technically get subincluded so we need to handle them here
+	for _, preloadedSubinclude := range state.Config.Parse.PreloadSubincludes {
+		target := state.Graph.Target(preloadedSubinclude)
+		if target == nil || target.Subrepo == nil {
```
Tatskaari, line 88: These can be nil while we're preloading as we're parsing the packages they're defined in

tiagovtristao (approved):
Nice adding comments throughout.

PR #2282, opened 2022-01-27
Allow restricting the targets that can opt out of sandboxing

Tatskaari:


src/core/config.go
```diff
@@ -481,11 +481,12 @@ type Configuration struct {
...
+		Dir              []string     `help:"Directories to hide within the sandbox"`
+		Namespace        string       `help:"Set to 'always', to namespace all actions. Set to 'sandbox' to namespace only when sandboxing the build action. Defaults to 'never', under the assumption the sandbox tool will handle its own namespacing. If set, user namespacing will be enabled for all rules. Mount and network will only be enabled if the rule is to be sandboxed."`
+		Build            bool         `help:"True to sandbox individual build actions, which isolates them from network access and some aspects of the filesystem. Currently only works on Linux." var:"BUILD_SANDBOX"`
+		Test             bool         `help:"True to sandbox individual tests, which isolates them from network access, IPC and some aspects of the filesystem. Currently only works on Linux." var:"TEST_SANDBOX"`
+		WhitelistTargets []BuildLabel `help:"If set, only targets that match these wildcards will be allowed to opt out of the sandbox"`
```
peterebden, line 489: This is a bit unintuitive to me - `Sandbox.WhitelistTargets` refers to the set of targets that can _not_ be sandboxed. I'd expect a property with that name to be the set of targets that _can_ be sandboxed.

peterebden: maybe `ExcludeableTargets` or `OptOutTargets` or something?

PR #2286, opened 2022-02-01
pleasew: download binary for correct arch

towe75:
I found that pleasew downloads the wrong binary when running on apple silicon / m1. Subsequent problem is that some build files are also looking at the wrong HOSTARCH etc.

I'm not sure if the proposed solution is valid for all supported operating system and arch combinations. We can, of course, replace it with a more distinct "if darwin" case.

### DIFF
--- a/pleasew
+++ b/pleasew
@@ -9,7 +9,15 @@ RESET="\x1B[0m"
 DEFAULT_URL_BASE="https://get.please.build"
 
 OS="$(uname)"
-ARCH="$(uname -m)"
+
+if [[ $OS == "Darwin" ]]; then
+    # switch between mac amd64/arm64 
+    ARCH="$(uname -m)"
+else
+    # default to amd64 on other operating systems
+    # because we only build intel binaries
+    ARCH="amd64"
+fi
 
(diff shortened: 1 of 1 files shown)


pleasew
```diff
@@ -9,8 +9,7 @@ RESET="\x1B[0m"
 DEFAULT_URL_BASE="https://get.please.build"
 
 OS="$(uname)"
-# Don't have any builds other than amd64 at the moment.
-ARCH="amd64"
+ARCH="$(uname -m)"
```
Tatskaari, line 12: I think this returns x86_64 on linux so we'll need to convert it to `amd64` manually.

towe75: Yes, indeed. Is a "if darwin { arch=uname -m } else { "amd64" }" block ok for you? Or rather a switch+convert over known ARCH names with a fallback to amd64? Second variant might also work for e.g. linux on arm (e.g. AWS instances)

towe75: Oh. Found that there are no linux arm binaries yet ¯\_(ツ)_/¯. 
I will commit the "if darwin" check.

Tatskaari:
Looks good to me :) Thanks a lot for the PR

PR #2290, opened 2022-02-03
plz init plugin <plugin>

samwestmoreland:
Inject config values and add plugin target to build file.

Think this essentially works. I need to make a couple of little changes in the gcfg library: it feels like being able to inject a blank line would be helpful; and if the config file only consists of comments, injectField ought to treat those comments as a header and inject stuff below, rather than above as it does now.

Also not sure about the http request bit. Are we supposed to be unmarshalling the response into a struct like that or is there a better way of doing it?


src/plzinit/plugins.go
```diff
@@ -0,0 +1,285 @@
...
+	return file
+}
+
+// targetExistsInFile checks to see if the plugin target already exists
+// in plugins/BUILD
+func targetExistsInFile(location, plugin string) bool {
+	if _, err := os.Stat(location); err != nil {
+		log.Debug("Target already exists in %s", location)
+		return false
+	}
```
tiagovtristao, line 209: It seems that you are checking whether `plugins/BUILD` exists here? I'm confused by the `os.Stat` usage here, but we have the `fs.FileExists` that provides extra guarantees.

src/plzinit/plugins.go
```diff
@@ -0,0 +1,285 @@
...
+
+	b, err := ioutil.ReadFile(location)
+	if err != nil {
+		panic(err)
+	}
+
+	str := "plugin_repo\\(...name = \"" + plugin + "\""
```
tiagovtristao, line 216: Ideally, we would query the graph to check for things like this, was this not possible?

src/plzinit/plugins.go
```diff
@@ -0,0 +1,285 @@
...
+	return exists
+}
+
+// createTarget writes the plugin target to plugins/BUILD
+func createTarget(location, plugin string) error {
+	if targetExistsInFile(location, plugin) {
+		return nil
+	}
+
+	dir := filepath.Dir("plugins/")
```
tiagovtristao, line 230: `dir := filepath.Dir(location)` will guarantee that it works even if `plugins/BUILD` is replaced with a different path

src/plzinit/plugins.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	if err := injectPluginConfig(plugin); err != nil {
+		return err
+	}
+	return nil
+}
+
+func injectPluginConfig(plugin string) error {
+	configPath := path.Join(core.RepoRoot, ".plzconfig")
+	if _, err := os.Stat(configPath); errors.Is(err, os.ErrNotExist) {
```
tiagovtristao, line 51: Can use our `fs.FileExists` here too.

src/plzinit/plugins.go
```diff
@@ -0,0 +1,285 @@
...
+	file := ast.Read(configFile)
+	switch plugin {
+	case "python":
+		file = writePythonConfigFields(file)
+	case "java":
+		file = writeJavaConfigFields(file)
+	case "cc":
+		file = writeCCConfigFields(file)
+	default:
+		log.Fatalf("failed to initialise plugin. \"%s\" not recognised", plugin)
```
tiagovtristao, line 70: I wonder if this should really fail here. What `injectPluginConfig` is trying to do is actually migrating old config to new config. There will be new plugins that won't require any migration eventually and so the inject logic should just be skipped.

src/plzinit/plugins.go
```diff
@@ -0,0 +1,285 @@
...
+		"defaultdbgcflags":   "DefaultDbgCFlags",
+		"defaultoptcppflags": "DefaultOptCppFlags",
+		"defaultdbgcppflags": "DefaultDbgCppFlags",
+		"defaultldflags":     "DefaultLdFlags",
+		"pkgconfigpath":      "PkgConfigPath",
+		"testmain":           "TestMain",
+		"dsymtool":           "DsymTool",
+	}
+
+	return writeFieldsToConfig("cc", file, configMap)
```
tiagovtristao, line 127: It seems that the cc plugin initialisation and migration is doing more than java and python, is this as expected?

src/plzinit/plugins.go
```diff
@@ -0,0 +1,285 @@
...
+		return "", err
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	var result Response
+	if err := json.Unmarshal(body, &result); err != nil {
```
tiagovtristao, line 280: Yeah, this is fine. That's how it is usually done.

src/plzinit/plugins.go
samwestmoreland: Woah I have no idea what I was doing here. The idea was: if the file doesn't exist, create the file and add the target to it. I'll tidy this up.

src/plzinit/plugins.go
```diff
@@ -0,0 +1,285 @@
...
+	log.Debug("Initialising plugin(s): %v", plugins)
+	for _, p := range plugins {
+		if err := initPlugin(p); err != nil {
+			log.Warningf("could not initialise plugin %s. Got error: %v", p, err)
+		}
+	}
+}
+
+func initPlugin(plugin string) error {
+	if err := createTarget("plugins/BUILD", plugin); err != nil {
```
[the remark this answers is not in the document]
samwestmoreland: This should be dealt with below in targetExistsInFile(), I think. I'll double-check.

src/plzinit/plugins.go
samwestmoreland: No not expected, sorry, just forgot to tidy up the java and python functions!

src/plzinit/plugins.go
tiagovtristao: I think that would be a better approach and more resilient. If you know what needs to be built then it should be straightforward and fast enough since there are only so many plugins.

src/please.go
```diff
@@ -268,6 +268,11 @@ var opts struct {
 		} `command:"pleasings" description:"Initialises the pleasings repo"`
 		Pleasew struct {
 		} `command:"pleasew" description:"Initialises the pleasew wrapper script"`
+		Plugin struct {
+			Args struct {
+				Plugins []string `positional-arg-name:"plugin" required:"true" description:"Plugins to install"`
+			} `positional-args:"true"`
+		} `command:"plugin" description:"Install a plugin and migrate any language-specific config values"`
```
tiagovtristao, line 275: As discussed IRL, it might make sense to hide this until v17. There are also other issues with plugins that need fixing meanwhile. Adding `hidden:"true"` after `command:"plugin"` should do it.

go.sum
```diff
@@ -262,6 +262,12 @@ github.com/pkg/xattr v0.4.3 h1:5Jx4GCg5ABtqWZH8WLzeI4fOtM1HyX4RBawuCoua1es=
 github.com/pkg/xattr v0.4.3/go.mod h1:sBD3RAqlr8Q+RC3FutZcikpT8nyDrIEEBw2J744gVWs=
 github.com/please-build/gcfg v1.6.0 h1:pwVJohRItsBAkWh7tw3uWUFqk4a53cLl+Wu4VW33tP4=
 github.com/please-build/gcfg v1.6.0/go.mod h1:qwx/xbr5eBETgfBLP9eXDgXQ5VzLQM5dIjcseh3pkVk=
+github.com/please-build/gcfg v1.6.1-0.20220121092333-3b730990cafa h1:etgi05fzq4gn03Q9H0DzWUlR7pmAGlmKTZ+Psa3jswA=
+github.com/please-build/gcfg v1.6.1-0.20220121092333-3b730990cafa/go.mod h1:qwx/xbr5eBETgfBLP9eXDgXQ5VzLQM5dIjcseh3pkVk=
+github.com/please-build/gcfg v1.6.1-0.20220303142845-6feafe84fba0 h1:QF0BqiPP+JB3w/YDt4W9uROfwifLLlXX8VxuZ1jQDmc=
+github.com/please-build/gcfg v1.6.1-0.20220303142845-6feafe84fba0/go.mod h1:qwx/xbr5eBETgfBLP9eXDgXQ5VzLQM5dIjcseh3pkVk=
+github.com/please-build/gcfg v1.6.1-0.20220308170049-10ea9d657afb h1:6HKarWPya5mPR1xHqTqRNAN+Xox2jndkqVKJRJqu2V8=
+github.com/please-build/gcfg v1.6.1-0.20220308170049-10ea9d657afb/go.mod h1:qwx/xbr5eBETgfBLP9eXDgXQ5VzLQM5dIjcseh3pkVk=
```
tiagovtristao, line 270: There's quite a big of duplication here. I wonder if `go mod tidy` will get rid of some?

src/plzinit/plugins.go
```diff
@@ -0,0 +1,286 @@
...
+		if err = subrepo.LoadSubrepoConfig(); err != nil {
+			panic(err)
+		}
+		config = subrepo.State.Config
+		for _, v := range config.PluginConfig {
+			if len(v.DefaultValue) == 0 {
+				file = ast.InjectField(file, v.ConfigKey, "", section, plugin, v.Repeatable)
+			} else {
+				// Check for build labels and resolve relative to host
+				val := v.DefaultValue[0]
```
tiagovtristao, line 191: If `v.DefaultValue` is an arrary, it might be better to iterate through all the values as `v.DefaultValue[1]` could also be a build label that needs normalisation?

src/plzinit/plugins.go
```diff
@@ -0,0 +1,286 @@
...
+				if core.LooksLikeABuildLabel(val) {
+					val = "///" + plugin + val
```
tiagovtristao, line 193: There's the `core.TryParseBuildLabel` that handles build label creation. ie. `core.TryParseBuildLabel(val, "", plugin)`

tiagovtristao (approved):

PR #2292, opened 2022-02-03
Allow no_test_output in cc_test

ktf:


Tatskaari (approved):
Ah nice! Thanks for the PR. Looks good to me.

PR #2295, opened 2022-02-07
Remove the special "self" subrepo functionality

tiagovtristao:
Remove the special "self" subrepo functionality, as it's not needed.

### DIFF
--- a/src/parse/asp/builtins.go
+++ b/src/parse/asp/builtins.go
@@ -1008,8 +1008,10 @@ func subrepo(s *scope, args []pyObject) pyObject {
 	s.NAssert(s.pkg == nil, "Cannot create new subrepos in this context")
 	name := string(args[NameArgIdx].(pyString))
 	dep := string(args[DepArgIdx].(pyString))
-	var target *core.BuildTarget
+
+	// Root
 	root := name
+	var target *core.BuildTarget
 	if dep != "" {
 		// N.B. The target must be already registered on this package.
 		target = s.pkg.TargetOrDie(core.ParseBuildLabelContext(dep, s.pkg).Name)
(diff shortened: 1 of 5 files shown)


src/parse/asp/builtins.go
```diff
@@ -1022,17 +1024,23 @@ func subrepo(s *scope, args []pyObject) pyObject {
...
 	}
+
+	// Base name
+	subrepoName := path.Join(s.pkg.Name, name)
+
+	// State
 	var state *core.BuildState
 	if args[ConfigArgIdx] != None {
-		state = s.state.ForSubrepo(name, path.Join(s.pkg.Name, string(args[ConfigArgIdx].(pyString))))
+		state = s.state.ForSubrepo(subrepoName, path.Join(s.pkg.Name, string(args[ConfigArgIdx].(pyString))))
```
tiagovtristao, line 1034: `s.CurrentSubrepo` should include the package as well.

Tatskaari: Nice! I spotted this but wasn't 100% sure how it should work. I think this makes sense though.

src/parse/asp/config.go
```diff
@@ -94,7 +94,7 @@ func (i *interpreter) newConfig(state *core.BuildState) *pyConfig {
 	return &pyConfig{base: base}
 }
 
-func resolveSelf(values []string, subrepo string) []string {
+func resolvePluginValue(values []string, subrepo string) []string {
```
Tatskaari, line 97: I think this needs updating? Isn't it changing `//foo:bar` to `/////foo:bar` for default values? I think we want to do something like:

```
if core.LooksLikeABuildLabel(v) && !strings.HasPrefix(v, "///") && !strings.HasPrefix(v, "@") {
       // convert to ///{subrepo}{v}
}
```

src/parse/asp/config_test.go
```diff
@@ -0,0 +1,21 @@
...
+	assert.Equal(t, []string{"///path/to/subrepo//path/to:target"}, values)
+
+	values = resolvePluginValue([]string{"///path/to/another/plugin//path/to:target"}, "path/to/subrepo")
+	assert.Equal(t, []string{"///path/to/another/plugin//path/to:target"}, values)
+
+	values = resolvePluginValue([]string{"///path/to/another/plugin//path/to:target"}, "")
+	assert.Equal(t, []string{"///path/to/another/plugin//path/to:target"}, values)
+
+	values = resolvePluginValue([]string{"//path/to:target"}, "")
+	assert.Equal(t, []string{"/////path/to:target"}, values)
```
Tatskaari, line 20: Is this what we want?

tiagovtristao: That's how it seems to be meant to work. In general, evaluated labels are resolved fully according to their context (i.e. if they are in a subrepo or not).
So, `resolvePluginValue([]string{"//path/to:target"}, "")` refers to the `//path/to:target` config value in the host repo. Since this is to be used in the plugin, it needs the `///` prefix, otherwise it will get resolved locally to the subrepo.

Example:

Host repo
`.plzconfig` file
```ini
[Plugin "proto"]
LanguageDef = //build_defs:python
```

Proto plugin
`build_defs/BUILD` file
```python
subinclude("//build_defs/sdk")

protoc_plugins(
    name = "proto_languages",
    build_defs = CONFIG.PROTO.DEFINITIONS,
    type = "proto_language",
)
```

If `//build_defs:python` doesn't get the `///` prefix, it will be resolved as something like `///proto//build_defs:python`

src/parse/asp/config.go
tiagovtristao: Default values come from plugin config, which means the `subrepo` argument will be non-empty (i.e. `proto`), yielding `///proto//foo:bar` instead.

Tatskaari (approved):
Thanks for clarifying :) Looks good to me

PR #2307, opened 2022-02-14
Please shim that handles first installation and updates before executing main binary

tiagovtristao:
Please shim that installs plz if it doesn't exist, or tries to update plz before running the main binary.

This shim incorporates:
- The functionality of `pleasew`
- The functionality of `update` package
- Flag completion.

How to test?
- Run `plz build //tools/please_shim`
- Rename `plz-out/bin/tools/please_shim/please_shim` to `plz-out/bin/tools/please_shim/plz` (the renaming is to guarantee flag completion)


samwestmoreland (approved):
Lgtm! So we're assuming that there's always a config, right? If there isn't, this just fails after MustFindRepoRoot(). Is installing without any existing config something we want to support one day? I.e. with some kind of default config?

tiagovtristao:
> Lgtm! So we're assuming that there's always a config, right? If there isn't, this just fails after MustFindRepoRoot(). Is installing without any existing config something we want to support one day? I.e. with some kind of default config?

That's a good point. At the moment there are only a handful of commands that can work without any config (i.e. `plz init`). I guess this tool should handle that case, if we are to make it public.

tools/please_wrapper/main.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	tmpDir := os.TempDir()
+	if tmpDir == "" {
+		log.Fatal("No default directory to use for temporary files found")
+	}
+
+	// Prevent the repo's default `plz-out/log/build.log` file from being overridden
+	logFilePath := filepath.Join(tmpDir, "plz_build.log")
+	cli.InitFileLogging(logFilePath, cli.Verbosity(logging.DEBUG), false)
```
Tatskaari, line 32: Can we not init logging without a file?

tiagovtristao: Where would the logs go?

tools/please_wrapper/main.go
```diff
@@ -0,0 +1,50 @@
...
+	plzPath := filepath.Join(config.Please.Location, "please")
+
+	if fs.FileExists(plzPath) {
+		out, err := exec.Command(plzPath, "--version").Output()
+		if err != nil {
+			log.Fatalf("Error getting Please version: %s", err)
+		}
+
+		core.PleaseVersion = strings.TrimSpace(strings.TrimPrefix(string(out), "Please version "))
+	}
```
Tatskaari, line 43: Can't we just check if `filepath.Join(config.Please.Lcation, version,  "please")` exists? I think we talked about not needing to change the active version of Please, and just executing Please in place?

tiagovtristao: The update logic relies on `core.PleaseVersion` to be set. In order to keep the same behaviour this shim will set `core.PleaseVersion` to the current `filepath.Join(config.Please.Location, "please")` binary. This guarantees that it works the same way as before. Unless I'm missing something?

peterebden:
> > Lgtm! So we're assuming that there's always a config, right? If there isn't, this just fails after MustFindRepoRoot(). Is installing without any existing config something we want to support one day? I.e. with some kind of default config?
> 
> That's a good point. At the moment there are only a handful of commands that can work without any config (i.e. `plz init`). I guess this tool should handle that case, if we are to make it public.

Yeah I think this is a good point; there are things like `plz help` etc that do work outside a repo, it'd be nice for that to continue even if this was `plz` on your path. Maybe it can be permissive and if it doesn't find a repo it can just try to download / execute the latest version?

samwestmoreland (approved):
Nice! 👍🏽

PR #2310, opened 2022-02-18
Download targets that are linkable

tiagovtristao:
Download targets in remote execution with links. This guarantees code generation works as expected.

### DIFF
--- a/src/core/build_target.go
+++ b/src/core/build_target.go
@@ -1773,6 +1773,20 @@ func (target *BuildTarget) GetFileContent(state *BuildState) (string, error) {
 	return ReplaceSequences(state, target, target.FileContent)
 }
 
+// HasLinks returns true if the outputs are meant to be linked somewhere (i.e. via symlinks).
+// This check is useful in deciding whether this target should be downloaded during remote execution or not.
+func (target *BuildTarget) HasLinks(state *BuildState) bool {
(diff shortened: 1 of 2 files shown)


src/core/build_target.go
```diff
@@ -1773,6 +1773,20 @@ func (target *BuildTarget) GetFileContent(state *BuildState) (string, error) {
 	return ReplaceSequences(state, target, target.FileContent)
 }
 
+// HasLinks returns true if the outputs are meant to be linked somewhere (i.e. via symlinks).
+// This check is useful in deciding whether this target should be downloaded during remote execution or not.
+func (target *BuildTarget) HasLinks(state *BuildState) bool {
```
tiagovtristao, line 1778: Not really sure this is the best place for it, but I reckon it makes sense since certain labels have a meaning and functionality attached to them

Tatskaari (approved):

peterebden:
Whoa hang on, this could download a _lot_ of stuff, no? Go sources are probably OK since they're filegroups but it will download every `go_module` (or at least their sources) which is probably undesirable.

tiagovtristao:
> Whoa hang on, this could download a _lot_ of stuff, no? Go sources are probably OK since they're filegroups but it will download every `go_module` (or at least their sources) which is probably undesirable.

Argh, you're right. This should be fixed before it's deployed internally. I'll get this reverted and release another beta version, and come back to it later.

peterebden:
Thanks!

Issue #2309, opened 2022-02-15
python/venv support broken


afterthought:
Ref: #2210 

I'm posting this as a new issue so it's not lost on a closed PR. This is holding back plz adoption on our team.

@Tatskaari, I tested this and it doesn't quite work. Only 1 pip_library gets linked to python/venv and then the remaining ones are skipped because this [line](https://github.com/thought-machine/please/blob/3a04445a810829976518757f8401ae4eb6d4ce23/src/build/build_step.go#L1021) is checking if the destination exists and it does already after the first lib creates python/venv.

afterthought:
I've started looking at this more. It might be OS specific. I'm seeing the issue on Debian. However, I tested inside the [official please ubuntu image](https://github.com/thought-machine/please/blob/master/tools/images/ubuntu/Dockerfile) and it seems to work correctly here.

afterthought:
No one has looked at my fork yet. I'm going to create a PR I guess. Hopefully it's not too far off base.

PR #2315, opened 2022-02-24
fix: python/venv not created properly on debian

afterthought:
See #2309 for details.


tiagovtristao:
Jon's more familiar with this venv support, let's wait for him. He's back next week.

PR #2324, opened 2022-03-04
reorder CONFIGS priority in pleasew

nmpowell:
This raises the priority of the `.plzconfig.<profile>` above the `.plzconfig_<arch>`, as indicated in the docs: https://please.build/config.html

Tested with 16.19.0.

Show that it's currently not behaving as the docs indicate:
```
echo "[buildconfig]\ntoolchain-version = 3.7.3" > .plzconfig.testprofile
cat .plzconfig_linux_amd64
...
[buildconfig]
toolchain-version = 3.5.0
...
```
```
PLZ_CONFIG_PROFILE=testprofile
plz query config buildconfig.toolchain-version
3.5.0
```

### DIFF
--- a/pleasew
+++ b/pleasew
@@ -25,8 +25,8 @@ PROFILE="${PLZ_CONFIG_PROFILE:-$(sed -E 's/.*--profile[= ]([^ ]+).*/\1/g' <<< "$
 # Config files on order of precedence high to low.
 CONFIGS=(
   ".plzconfig.local"
-  ".plzconfig_${OS}_${ARCH}"
   "${PROFILE:+.plzconfig.$PROFILE}"
+  ".plzconfig_${OS}_${ARCH}"
   ".plzconfig"
   "$HOME/.config/please/plzconfig"
   "/etc/please/plzconfig"


nmpowell:
I'd have thought the profile should be able to override the arch, but [this comment](https://github.com/thought-machine/please/pull/2065#discussion_r707318365) seems to contradict.

tiagovtristao:
I don't think the order specified on the site is correct. Looking at the code, config file order is different with regards to profiles i.e. https://github.com/thought-machine/please/blob/master/src/core/config.go#L186-L195.
And taking into account the defaults as in https://github.com/thought-machine/please/blob/master/src/core/config.go#L158-L165, this means that the order will be, lowest to highest:

- /etc/please/plzconfig
- ~/.config/please/plzconfig
- .plzconfig
- .plzconfig_<arch>
- .plzconfig.local

And if you specify `PLZ_CONFIG_PROFILE=testprofile`, it will try the following order from lowest to highest:
- /etc/please/plzconfig
- /etc/please/plzconfig.tesprofile
- ~/.config/please/plzconfig
- ~/.config/please/plzconfig.testprofile
- .plzconfig
- .plzconfig.testprofile
- .plzconfig_<arch>
- .plzconfig_<arch>.testprofile
- .plzconfig.local
- .plzconfig.local.testprofile

The order that you see on the `pleasew` script is a best attempt at this, but only to try to figure out where to find please. Once it finds please it will run the actual `plz query config buildconfig.toolchain-version` which will follow the order above and resolve to `3.5.0`.

Having said that I think your change still makes sense to the `pleasew` script, but the docs need indeed to be updated.

tiagovtristao:
Fixes #2327

Issue #2198, opened 2021-11-27
Fix implementation detail leaking into error message.


tiagovtristao:
Given the build rule below that fails to ensure the specified output:

```python
genrule(
    name = "license",
    srcs = ["LICENSE"],
    outs = ["LICENSE"],
    cmd = "",
)
```

The following error message is shown: `error moving outputs for target //:license: rule //:license failed to create output plz-out/tmp/license._build/LICENSE.out`.

The `LICENSE.out` filename is an implementation detail that shouldn't be shown in the error message, as it might confuse users.

PR #2325, opened 2022-03-07
Ensure action outputs are downloaded under the correct name

peterebden:
I swear there is an issue for this somewhere, but I can't find it...

Anyway, if there is an input named (say) `tcl` and the action also outputs `tcl`, we rename the output to `tcl.out` to avoid clashes. This has to exist on the ActionResult, but when we download it we then need to clean it back off again so they end up in the expected place.


tiagovtristao:
We have https://github.com/thought-machine/please/issues/2198, but it's more about not letting the user know about this internal implementation. That annoying bot decided to close the ticket.

tiagovtristao (approved):

peterebden:
> We have #2198, but it's more about not letting the user know about this internal implementation. That annoying bot decided to close the ticket.

Thx, I was thinking about a different one, but evidence is mounting that I've just imagined it.

PR #2329, opened 2022-03-07
Make pkg_sources correctly aware of subrepos

peterebden:
At some point we have changed `canonicalise` to produce more correct output that includes the subrepo name. However now this disagrees with the home-grown label building here.

Now using `canonicalise` for both so they should stay in sync better (and it's quite a bit fiddlier to handle the subrepo correctly).


tiagovtristao (approved):
Thanks!

PR #2333, opened 2022-03-09
Allow adding of arbitrary labels to third-party build rules

chrisnovakovic:
Some rules that download and/or build third-party code, e.g. `go_mod_download`, already accept a `labels` parameter whose value is passed through to the underlying `build_rule`. Do the same for the remaining third-party build rules that don't already allow this:

* `maven_jar`
* `pip_library`
* `python_wheel`
* `http_archive`
* `new_http_archive`
* `github_repo`
* `gitlab_repo`


tiagovtristao (approved):
Lgtm.

PR #2336, opened 2022-03-10
Do not pass user-supplied lables to nested remote_file

orfeasa:
While querying labels we realised that there are duplicated labels that are caused by the fact that remote file calls another remote file internally and also passes the labels argument.
This diff removes the passing of labels in the nested remote files.


tiagovtristao (approved):
I think this makes sense, it should only be applied to one of the two within the `extract` logic. Thanks.

peterebden (changes requested):
It's not clear to me that there is anything wrong with the existing behaviour - labels can be (and often are) duplicated on multiple targets.

This could also be a breaking change if someone relies on the current behaviour (although I imagine that is unlikely)

tiagovtristao:
> It's not clear to me that there is anything wrong with the existing behaviour - labels can be (and often are) duplicated on multiple targets.
> 
> This could also be a breaking change if someone relies on the current behaviour (although I imagine that is unlikely)

Fair, I thought `plz query` commands were returning intermediate results by default too - causing duplication - but that's not the case in general. Also, after speaking with @orfeasa, he was trying to run `plz query print //third_party/... -l dependency_metadata:` where the behaviour is different and seems to return intermediate results by default.

@orfeasa, looking more into it, there's an additional flag `--omit_hidden` which might do what you are looking for?

orfeasa:
Thanks for the explanations. Unfortunately `--omit_hidden` does not include some build rules that start with an underscore, but we can deduplicate the results of the query so no worries :)

PR #2337, opened 2022-03-11
Allow arbitrary files to be included/excluded in subrepo rules

chrisnovakovic:
Add `files` and `strip_files` parameters to `new_http_archive`:

* `files` is a list of files that the rule should output; the semantics are similar to those of the `install` parameter in `go_module`.
* `strip_files` removes any files matching the given list of patterns from the output; the semantics are identical to those of the `strip` parameters found in other rules, e.g. `go_mod_download`.

The default behaviour is to output all files in the archive, which is backwards-compatible.

Also add these parameters to the convenience wrappers around `new_http_archive`, and pass them through to `new_http_archive`.

### DIFF
--- a/rules/subrepo_rules.build_defs
+++ b/rules/subrepo_rules.build_defs
@@ -24,7 +24,7 @@ def workspace(name:str):
     )
 
 
-def http_archive(name:str, urls:list, strip_prefix:str=None, labels:list=[],
+def http_archive(name:str, urls:list, files:list=[], strip_prefix:str=None, strip_files:list=[], labels:list=[],
                  hashes:str|list&sha256=None, config:str=None, bazel_compat:bool=False,
                  visibility:list=None):
     """Fetches a remote file over HTTP and expands its contents.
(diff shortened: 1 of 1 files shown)


rules/subrepo_rules.build_defs
```diff
@@ -126,7 +139,7 @@ def new_http_archive(name:str, urls:list, build_file:str=None, build_file_conten
             'build': [build_file],
         },
         tools = [CONFIG.JARCAT_TOOL],
-        outs = [name],
+        outs = [f"{name}/{path}" for path in files] if files else [name],
```
tiagovtristao, line 142: I'm not sure about this part, as the out of this rule becomes the root for the subrepo rule below.

If I specify `files = ["path/to/file"]`, then `path/to/file` is considered the root for the subrepo. I.e. https://github.com/thought-machine/please/blob/master/src/parse/asp/builtins.go#L1043 and https://github.com/thought-machine/please/blob/master/src/core/subrepo.go#L57-L58

I think ideally the output would be the single directory with only the files that are specified.

PR #2339, opened 2022-03-11
Add validation for plugin config fields

samwestmoreland:
Addresses https://github.com/thought-machine/please-issues/issues/66

Tested this in core3 and it appears to work. Any suggestions on the wording of the warning are welcome. Could align the warning more with the one for normal config validation but I think this is clearer.


src/core/config.go
```diff
@@ -296,12 +297,25 @@ func ReadConfigFiles(filenames []string, profiles []string) (*Configuration, err
...
 			newExtraValues[strings.ToLower(k)] = v
+			// Can only validate if there's anything to check against. The plugin config
+			// hasn't necessarily been loaded yet.
+			valid := len(config.PluginConfig) == 0
+			for _, val := range config.PluginConfig {
+				if strings.EqualFold(k, val.ConfigKey) {
+					valid = true
+					break
+				}
+			}
```
tiagovtristao, line 315: This logic is being applied to all `config.Plugin`s and I reckon this will cause some issues, as the PluginConfig/PluginDefinition is specific to one single plugin. For instance, if the PluginConfig is related to Python, but you are going over the Java plugin then the config keys of Java won't be found on the PluginConfig for Python

samwestmoreland: Yeah I did think that but it should be fine, right? Because it will eventually loop over the Java config and find something. As in, it always loops over all the config for every plugin you have, so it will eventually find one that matches. At least, I think that's how it works. But anyway even if it does work, it's not super pretty. I'll see if I can improve the logic here.

samwestmoreland: As discussed, you're right it doesn't work the way I thought. I'll get on the fix!

samwestmoreland:
Tested in core3. You get these warnings:
Messages:
16:55:01.226 WARNING: Unrecognised config value "notakey" for plugin "python"
16:55:01.381 WARNING: Unrecognised config value "home" for plugin "java"
16:55:01.381 WARNING: Unrecognised config value "home" for plugin "java"
16:55:01.385 WARNING: Unrecognised config value "notakey" for plugin "python"
16:55:01.386 WARNING: Unrecognised config value "notakey" for plugin "python"
16:55:01.509 WARNING: Unrecognised config value "home" for plugin "java"
16:55:01.509 WARNING: Unrecognised config value "home" for plugin "java"
16:55:01.511 WARNING: Unrecognised config value "notakey" for plugin "python"
16:55:01.512 WARNING: Unrecognised config value "notakey" for plugin "python"

It's a bit annoying that you get each warning multiple times. Not sure how to resolve that as we go through the subinclude code multiple times.

The "home" one is a missing configkey in the java plugin config - I'll update that.

Also maybe it should say "config key".

samwestmoreland:
Abort - this change breaks stuff. Can't build the python plugin repo for some reason.

Tatskaari (approved):
LGTM :)

PR #2289, opened 2022-02-02
Remove context package to allow preloaded subincludes to subinclude other rules

Tatskaari:

PR #2340, opened 2022-03-12
Fix race in nested subincludes

peterebden:
This one took a while to track down...

Say you have a package like:
```
filegroup(
    name = "a",
    srcs = ["a.build_defs"],
)
filegroup(
    name = "b",
    srcs = ["b.build_defs"],
)
```
and `a.build_defs` contains
```
subinclude("//build/defs:b")
```
There is a race condition during parsing when something else does `subinclude("//build/defs:a")`. We go through the package top-down, and as soon as we interpret the first filegroup `a` becomes available to parse. At that point the subinclude is unblocked, and it begins parsing and interpreting `a.build_defs`. Meanwhile the top-level `BUILD` file is still being parsed, but is now racing against `a.build_defs`; if we start interpreting the subinclude in `a.build_defs` before `b` has become available, we blow up with
```
Build stopped after 110ms. 1 target failed:
    //smthn/smthn:all
plz-out/gen/build/defs/a.build_defs:1:1: error: Target :b is not defined in this package; it has to be defined before the subinclude() call

subinclude("//build/defs:b")
^
```
I think the solution is straightforward; that check only needs to apply when parsing the BUILD file itself. The subincludes are parsed separately and the normal mechanisms for waiting for them to become available can apply.

(This is a toy example that illustrates the issue but won't really trigger it in practice; the real one involves a directory with ~50 build_defs files in which offers more time for the races to occur in)

I've tested this locally; fortunately it turns out to be an anti-heisenbug in that adding a bunch of debug logging actually makes it more likely to occur (by stretching out the duration of one side of the race). With that I could pretty reliably trigger it within ~10 runs or so; after this change I can't trigger it again at all.

I'm pretty sure the original cause happens in #2289 so this bug doesn't exist in any non-beta release.

### DIFF
--- a/src/parse/asp/builtins.go
+++ b/src/parse/asp/builtins.go
@@ -314,7 +314,7 @@ func subinclude(s *scope, args []pyObject) pyObject {
 func subincludeTarget(s *scope, l core.BuildLabel) *core.BuildTarget {
 	pkg := s.contextPackage()
 	pkgLabel := pkg.Label()
-	if l.Subrepo == pkgLabel.Subrepo && l.PackageName == pkgLabel.PackageName {
+	if l.Subrepo == pkgLabel.Subrepo && l.PackageName == pkgLabel.PackageName && s.subincludeLabel == nil {
 		// This is a subinclude in the same package, check the target exists.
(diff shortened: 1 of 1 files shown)


tiagovtristao (approved):
Nice, I guess this makes sense. Looking at the PR where this was introduced, the context package logic returned by the scope has changed, which seems to explain this behaviour.

PR #2342, opened 2022-03-16
Update to Go 1.18

peterebden:
We still build on 1.17 as well so this (sadly) doesn't mean we can start using generics right away - we probably want to think when that is (I don't think we need to wait until 1.19 but we shouldn't require everyone to upgrade on day 1 either).


Tatskaari (approved):
LGTM bring on the generics!

PR #2345, opened 2022-03-16
Handle missing sandbox tool

Tatskaari:
Currently we fail with a fairly confusing error message like: `fork/exec no such file`


peterebden (approved):
Nice

samwestmoreland (approved):

PR #2354, opened 2022-03-17
Adding histogram for tracking build times

isobelormiston:
This tracks the time taken to build individual targets, locally and remotely, and records it in histogram buckets pushed to the pushgateway. I've tested this by running local & remote builds and the times are pushed successfully to our TM gateway.


Hamishpk (approved):
LGTM

PR #2355, opened 2022-03-17
Fix race condition when registering arch subrepos

Tatskaari:
This makes the ForArch function work essentially the same way as the ForSubrepo one. These two functions (ForArch and ForSubrepo) are quite similar but I can't seem to find a nice way to unify them.


samwestmoreland (approved):
LGTM. Thanks!

PR #2357, opened 2022-03-17
Fix perf test script

tiagovtristao:
This fixes the performance test script that was failing with `Cannot provide an empty srcs list to go_library` errors due to missing sources.


samwestmoreland (approved):

peterebden (approved):
Thanks Tiago!

PR #2359, opened 2022-03-18
Don't run in a fake root env when using the built in please sandbox

Tatskaari:


peterebden (approved):

src/sandbox/sandbox_linux.go
```diff
@@ -67,6 +69,29 @@ func Sandbox(args []string) error {
...
+		if err != nil {
+			return fmt.Errorf("invalid SANDBOX_UID: %v", user)
+		}
+		execCmd := exec.Command(cmd, args[1:]...)
+		execCmd.Env = env
+		execCmd.Stdout = os.Stdout
+		execCmd.Stdin = os.Stdin
+		execCmd.Stderr = os.Stderr
+		execCmd.SysProcAttr = &syscall.SysProcAttr{
+			Pdeathsig:  syscall.SIGHUP,
```
chrisnovakovic, line 84: Maybe SIGTERM or SIGINT instead?

chrisnovakovic (approved):

src/sandbox/sandbox_linux.go
Tatskaari: I _think_ this is the signal Please sends to its sub-processes when the process is cleaned up by it's process manager. It does seem like the process group is cleaned up correctly.

chrisnovakovic: Right, I misunderstood the `Pdeathsig` documentation. Thanks for clarifying!

PR #2366, opened 2022-03-21
stop using find to find the repo root for plugin_repo()

Tatskaari:
This was sometimes picking up the wrong .plzconfig file which causes errors. This approach assumes more about the repo structure but ultimately works for github.


peterebden (approved):
Guess this only works at the top-level, but I don't think we have a particular need for something else right now.

samwestmoreland (approved):

PR #2377, opened 2022-03-27
Parallelise fmt

peterebden:
Noticed it was kind of slow on the tree generation for the perf test.
On this machine, this reduces the time from 33.5s -> 7.4s (approx 5x faster).


samwestmoreland (approved):
Nice!

Tatskaari (approved):
Nice! Looks good to me.

PR #2379, opened 2022-03-28
Allow annotated labels for plz exec

Tatskaari:


src/please.go
```diff
@@ -1449,6 +1449,8 @@ func execute(command string) int {
 		}()
 	}
 	defer worker.StopAll()
+
+	log.Debugf("plz command: %v", strings.Join(os.Args, " "))
```
peterebden, line 1453: Needed?

peterebden (approved):

samwestmoreland (approved):

src/please.go
Tatskaari: Bit of a bonus change. I think this could be useful. I found it hard to figure out which command the log files belonged to when debugging the cards exporter build.

PR #2381, opened 2022-03-29
FF to exclude go rules

Tatskaari:


src/core/build_env.go
```diff
@@ -28,10 +28,12 @@ func GeneralBuildEnvironment(state *BuildState) BuildEnv {
 		// These are slightly modified forms that are more convenient for some things.
 		"XARCH=" + state.Arch.XArch(),
 		"XOS=" + state.Arch.XOS(),
-		// It's easier to just make these available for Go-based rules.
-		"GOARCH=" + state.Arch.GoArch(),
-		"GOOS=" + state.Arch.OS,
 	}
+
+	if !state.Config.FeatureFlags.ExcludeGoRules {
+		env = append(env, "GOARCH="+state.Arch.GoArch(), "GOOS="+state.Arch.OS)
```
peterebden, line 34: it seems nice to exclude stuff like this, but are you going to be able to handle this OK in the plugin?

Tatskaari: I guess I'll have to amp it up ;P 

But yeah I should just be able to `env = {"GOARCH": CONFIG.OS}` where necessary

PR #2384, opened 2022-03-29
Allow nested fstrings

Tatskaari:


peterebden (approved):
Could be _slightly_ slower since we had a faster lookup on the config before, but I doubt it matters much, and the genericity is nice.

PR #2385, opened 2022-03-30
Fix some data races

peterebden:
The core ones are benign but it's nice not to see them (admittedly at a minor cost of 3 unused bytes in BuildTarget).

The asp one should also be benign but it might still trigger Go's map race detector. Seems like no real cost there (in fact the sync.Once is probably faster then re-setting all those map entries)

Was testing this by temporarily adding `-race` into the bootstrap script. This fixes nearly all of it (there was also some weird stuff in asp but I think it's confused, it was complaining about not being able to restore stacks) but there is still something occasional in the logger. That deserves a separate PR though.

### DIFF
--- a/src/core/atomic_bool.go
+++ b/src/core/atomic_bool.go
@@ -16,3 +16,9 @@ func (b *atomicBool) Set() {
 func (b *atomicBool) IsSet() bool {
 	return atomic.LoadInt32(&b.b) == 1
 }
+
+func (b *atomicBool) Or(set bool) {
+	if set {
+		b.Set()
+	}
+}
--- a/src/core/atomic_bool.go
+++ b/src/core/atomic_bool.go
@@ -1,0 +1,18 @@
+package core
+
+import (
+	"sync/atomic"
+)
+
+// An atomicBool is a type we use as an opaque boolean that doesn't trigger the race detector.
+type atomicBool struct {
+	b int32
+}
+
+func (b *atomicBool) Set() {
+	atomic.StoreInt32(&b.b, 1)
+}
+
+func (b *atomicBool) IsSet() bool {
+	return atomic.LoadInt32(&b.b) == 1
+}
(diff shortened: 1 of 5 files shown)


src/core/state.go
```diff
@@ -995,9 +995,9 @@ func (state *BuildState) queueTargetAsync(target *BuildTarget, forceBuild, build
 		}
 	}
 	for {
-		called := false
+		var called atomicBool
```
tiagovtristao, line 998: I'm surprised -race complained about this one. Can't immediately see where it could go wrong. Unless it tries to be conservative.

src/core/state.go
```diff
@@ -951,7 +951,7 @@ func (state *BuildState) queueTestTarget(target *BuildTarget) {
 
 // queueResolvedTarget is like queueTarget but once we have a resolved target.
 func (state *BuildState) queueResolvedTarget(target *BuildTarget, forceBuild, neededForSubinclude bool) error {
-	target.NeededForSubinclude = target.NeededForSubinclude || neededForSubinclude
+	target.neededForSubinclude.Or(neededForSubinclude)
```
tiagovtristao, line 954: Interesting to see this functional pattern emerge for go sync stuff. Hadn't seen it before.

tiagovtristao (approved):

src/core/state.go
peterebden: Actually now wondering if I should get rid of Or and just do the if here... that might be simpler

src/core/state.go
peterebden: ye it's not very obvious but the function below that does `Set()` can be called from another goroutine.

This whole thing is benign but it's nicer not to have the complaints - eventually I'd like to be "race clean" on this stuff.

peterebden (commented):
Thanks Tiago!

src/core/atomic_bool.go
```diff
@@ -0,0 +1,24 @@
+package core
+
+import (
+	"sync/atomic"
+)
+
+// An atomicBool is a type we use as an opaque boolean that doesn't trigger the race detector.
```
Tatskaari, line 7: I've wanted this before too! Feel like this could be useful in the sdk,

Tatskaari (approved):
Looks legit to me

samwestmoreland (approved):
Looks good!

PR #2386, opened 2022-03-30
Fix data race in the system stats code

peterebden:
This is also getting accessed concurrently. Protect stuff with a mutex.

### DIFF
--- a/src/core/resources.go
+++ b/src/core/resources.go
@@ -15,10 +15,12 @@ var resourceUpdateFrequency = 500 * time.Millisecond
 // UpdateResources continuously updates the resources on this state object.
 // It should probably be run in a goroutine since it never returns.
 func (state *BuildState) UpdateResources() {
+	stats := &state.stats.Stats
+
 	lastTime := time.Now()
 	// Assume this doesn't change through the process lifetime.
 	count, _ := cpu.Counts(true)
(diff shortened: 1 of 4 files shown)


tiagovtristao (approved):

tiagovtristao:
I think I've seen the visibility error before but it seems unrelated to these changes.

src/core/resources.go
```diff
@@ -15,10 +15,12 @@ var resourceUpdateFrequency = 500 * time.Millisecond
 // UpdateResources continuously updates the resources on this state object.
 // It should probably be run in a goroutine since it never returns.
 func (state *BuildState) UpdateResources() {
+	stats := &state.stats.Stats
```
Tatskaari, line 18: Do we need to lock the mutex here?

Tatskaari (approved):

src/core/resources.go
peterebden: Ah nice spot, I was doing that, and removed it for testing then mistakenly committed it 🤦

peterebden: Wait no, this is not correct to lock; the loop below is long-running so we need to lock within there rather than for the length of the function.

samwestmoreland (approved):

Issue #2308, opened 2022-02-14
Validate that the plugin ID matches the subrepo name 


Tatskaari:
When we load a subrepo's config that contains a plugin definition, we should also check that the subrepo name matches the plugin ID. 

https://github.com/thought-machine/please/blob/master/src/parse/parse_step.go#L64

PR #2393, opened 2022-03-31
Move plugin config validation and validate subrepo name

samwestmoreland:
Part 1 moves the config validation from the asp package to the parse package. This should improve things wrt the warning being printed out multiple times as we only go through once here per plugin (or sometimes twice?).

Part 2 adds validation that the plugin ID matches the subrepo name, as per #2308.


src/parse/parse_step.go
```diff
@@ -64,6 +64,27 @@ func parse(tid int, state *core.BuildState, label, dependent core.BuildLabel, fo
 		if err := subrepo.LoadSubrepoConfig(); err != nil {
 			return err
 		}
+
+		// Validate plugin ID is the same as the subrepo name
+		if pluginID := state.RepoConfig.PluginDefinition.Name; pluginID != "" {
+			if !strings.EqualFold(pluginID, subrepo.Name) {
+				log.Errorf("Subrepo name \"%s\" should be the same as the plugin ID \"%s\"", subrepo.Name, pluginID)
```
Tatskaari, line 71: I think this should be `return fmt.Errorf()` so it actually stops the build.

src/parse/parse_step.go
```diff
@@ -64,6 +64,27 @@ func parse(tid int, state *core.BuildState, label, dependent core.BuildLabel, fo
...
+		// Validate the plugin config keys set in the host repo
+		definedKeys := map[string]bool{}
+		for key, definition := range state.RepoConfig.PluginConfig {
+			configKey := getConfigKey(key, definition.ConfigKey)
+			definedKeys[configKey] = true
+		}
+		if plugin := state.Config.Plugin[subrepo.Name]; plugin != nil {
+			for key := range plugin.ExtraValues {
+				if _, ok := definedKeys[strings.ToLower(key)]; !ok {
+					log.Warning("Unrecognised config key \"%v\" for plugin \"%v\"", key, subrepo.Name)
```
Tatskaari, line 84: Want to have a chat about making these mandatory too? I think Please only makes them warnings so it's able to do basic commands with invalid config. For plugins, it's far less problematic so we can probably return an error here too.

samwestmoreland: Yeah makes sense!

src/parse/parse_step.go
```diff
@@ -64,6 +64,31 @@ func parse(tid int, state *core.BuildState, label, dependent core.BuildLabel, fo
...
+			configKey := getConfigKey(key, definition.ConfigKey)
+			definedKeys[configKey] = true
+		}
+		if plugin := state.Config.Plugin[subrepo.Name]; plugin != nil {
+			for key := range plugin.ExtraValues {
+				if _, ok := definedKeys[strings.ToLower(key)]; !ok {
+					return fmt.Errorf("Unrecognised config key %q for plugin %q", key, subrepo.Name)
+				}
+			}
+		}
```
Tatskaari, line 91: Can we move this into it's own function? This top level function is quite big already.

Tatskaari (commented):
Looks good! Just a bit of a readability nit

Tatskaari (approved):
Looks good to me :)

PR #2395, opened 2022-03-31
Use go plugin and simplify bootstrap

Tatskaari:


samwestmoreland (approved):
Lgtm! Some tests failing about being able to find go toolchain it looks like, but otherwise looks good I mean.

.plzconfig
```diff
@@ -97,4 +90,11 @@ JavaBinaryExecutableByDefault = true
...
 PythonWheelHashing = true
 NoIterSourcesMarked = true
-GoDontCollapseImportPath = true
\ No newline at end of file
+GoDontCollapseImportPath = true
+ExcludeGoRules = true
+
+[Plugin "go"]
+Target = //plugins:go
+pleasegotool = @go//tools/please_go:bootstrap
```
samwestmoreland, line 98: nit: Should this be `///` for consistency?

.plzconfig
```diff
@@ -1,6 +1,3 @@
-[please]
-location = ./plz-out/please
```
tiagovtristao, line 2: I remember this being added for ease of development, where you run `./bootstrap` and then `plz` would resolve correctly. It looks like that won't work anymore?

src/help/help_test.go
```diff
@@ -1,11 +1,23 @@
...
 
 import (
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
+func TestMain(t *testing.M) {
+	f, err := os.Create(".plzconfig")
```
peterebden, line 11: You could just use `os.WriteFile` for this which is a couple of lines shorter ;-)

src/help/help_test.go
```diff
@@ -1,11 +1,23 @@
...
+	if err != nil {
+		panic(err)
+	}
+	f.Close()
+
+	t.Run()
```
peterebden, line 17: `os.Exit(t.Run())`

you're losing the exit code

test/go_rules/BUILD
```diff
@@ -79,15 +79,3 @@ gentest(
     test_cmd = "$DATA | tee $RESULTS_FILE",
 )
 
-go_library(
-    name = "package_test_lib",
-    srcs = ["package_lib.go"],
-    package = "package_lib",
-)
-
-go_test(
```
peterebden, line 88: is this just going away?

Tatskaari: Yeah in the plugin the import path is defined by the directory it's in and the package is defined in the source code.

plugins/BUILD
```diff
@@ -0,0 +1,5 @@
+plugin_repo(
+  name = "go",
+  revision = "master",
```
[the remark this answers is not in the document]
Tatskaari: Yeah but I'm changing things on both sides right now. Will lock it before we land.

third_party/go/BUILD
```diff
@@ -8,7 +8,6 @@ go_toolchain(
         "01cd67bbc12e659ff236ecebde1806f76452f7ca145c172d5ecdbf4f4803daae",  # freebsd_amd64
         "e85278e98f57cdb150fe8409e6e5df5343ecb13cebf03a5d5ff12bd55a80264f",  # linux_amd64
     ],
-    strip_srcs = CONFIG.BUILD_CONFIG != "dbg",
```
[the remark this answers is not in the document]
Tatskaari: No, it will never strip sources as the default is false. We need this now because we use this toolchain to bootstrap `please_go`, and `go build` doesn't work without the sources.

plugins/BUILD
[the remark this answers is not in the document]
Tatskaari: Updated :)

.plzconfig
Tatskaari: Oh, does that actually work? That's neat. The reason we introduced this was so that `./plz-out/please/please_go` was on the path which was required for `//tools/please_go`. I can leave it in if it's helping with development.

src/help/help_test.go
Tatskaari: What if it exits with a bad code? Then my test will fail. Wouldn't want that.

.plzconfig
tiagovtristao: The built please will now be available in `plz-out/bin/src` locally which changes things a bit. It's not too bad, a `.plzconfig.local` file can be created for the same effect. It was just nice that you didn't need the extra step.

tiagovtristao (approved):

peterebden (approved):

PR #2408, opened 2022-04-04
Sanitise environment variable values when executing remote actions

### LINKED ISSUE #2407: Remote execution fails for target with environment variable containing double-quotes
MWE:

```
genrule(
    name = "envtest",
    cmd = "true",
    env = {
        "DANGER_ZONE": """shell metacharacters:"[}:]{'""",
    },
)
```

This builds fine locally because the subshell's environment is created using `os/exec`:

```
$ plz build //:envtest --shell
Temp directories prepared, total time 50ms:
  //:envtest: plz-out/tmp/envtest._build
    Command: true

bash-5.0$ echo $DANGER_ZONE
shell metacharacters:"[}:]{'
```

However, building remotely fails because remote execution creates the environment using `export` commands before `cmd` is executed:

https://github.com/thought-machine/please/blob/dab003ffab79e340073526aeaab70bdc3fb0f48b/src/remote/action.go#L100

This will attempt to execute

```
export TMP_DIR="`pwd`" && export HOME= && export DANGER_ZONE="shell metacharacters:"[}:]{'"
```

which contains shell syntax errors:

```
$ plz build //:envtest --profile remote
Build stopped after 3.83s. 1 target failed:
    //:envtest
Remotely executed command exited with 1
    Execution failed: exit status 1
Failed action details: https://[...]
      Original action: https://[...]
Stderr:
/usr/bin/bash: -c: line 0: unexpected EOF while looking for matching `''
/usr/bin/bash: -c: line 1: syntax error: unexpected end of file
```

chrisnovakovic:
When executing remote actions to build targets that have a custom environment, Please sets up the environment in the remote shell by executing an `export` command for each environment variable before executing the build command. The environment variable values aren't sanitised, so shell constructs such as `$(...)` will be interpreted (which isn't the case when building locally, since the environment is set up by `os/exec` instead), and any value containing a double-quote
causes the action to fail altogether.

When constructing the `export` commands, sanitise environment variable values using [shellescape](https://github.com/alessio/shellescape). This ensures not only that double-quotes in the value don't cause failures when building the target remotely, but also that the environment when building the target remotely is identical to that when building it locally.

Closes #2407.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ require (
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/OneOfOne/cmap v0.0.0-20170825200327-ccaef7657ab8
 	github.com/ProtonMail/go-crypto v0.0.0-20210920135941-2c5829bbf927
+	github.com/alessio/shellescape v1.4.1
 	github.com/bazelbuild/buildtools v0.0.0-20210920153738-d6daef01a1a2
 	github.com/bazelbuild/remote-apis v0.0.0-20210718193713-0ecef08215cf
 	github.com/bazelbuild/remote-apis-sdks v0.0.0-20210909182119-af1232ee0d79
--- a/go.sum
+++ b/go.sum
@@ -48,6 +48,8 @@ github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751/go.mod h1:LOuy
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
 github.com/alecthomas/units v0.0.0-20190717042225-c3de453c63f4/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
 github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d/go.mod h1:rBZYJk541a8SKzHPHnH3zbiI+7dagKZ0cgpgrD7Fyho=
+github.com/alessio/shellescape v1.4.1 h1:V7yhSDDn8LP4lc4jS8pFkt0zCnzVJlG5JXy9BVKJUX0=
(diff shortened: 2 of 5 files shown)


Tatskaari (approved):

Tatskaari:
> Seems reasonable but I'm not mad about having to do this; why aren't these going into the environment variables of the action rather than generating a bunch of `export` statements in Bash?

Hmm good point.... why aren't they?

Tatskaari:
> Seems reasonable but I'm not mad about having to do this; why aren't these going into the environment variables of the action rather than generating a bunch of `export` statements in Bash?

Oh I remember! It was useful to be able to nest env vars inside each other: `SOME_PATH=$TMP_DIR/$PKG_DIR/foo`

peterebden:
Ergh right

I wish RBE would let you do something like that (even if only to identify the working directory). 
I'm not certain I want plz to get into the business of trying to parse shell expressions. I wonder if we should spot variables containing a dollar and `export` them this way, but chuck others into the environment field. That would be more elegant for _most_ build actions using this stuff.

chrisnovakovic:
> > Seems reasonable but I'm not mad about having to do this; why aren't these going into the environment variables of the action rather than generating a bunch of `export` statements in Bash?
> 
> Oh I remember! It was useful to be able to nest env vars inside each other: `SOME_PATH=$TMP_DIR/$PKG_DIR/foo`

In that case, this PR requires some more thought - I was under the impression that environment variable values weren't interpolated at all, but they are:

```
genrule(
    name = "envtest",
    cmd = "true",
    env = {
        "TEST": "$OS",
    },
)
```

```
$ plz build //:envtest --shell
Temp directories prepared, total time 60ms:
  //:envtest: plz-out/tmp/envtest._build
    Command: true

bash-5.0$ echo $TEST
linux
```

Only variable names are interpolated, though, which I'm guessing is a special case that's being handled in Please rather than in the shell it creates:

```
genrule(
    name = "envtest",
    cmd = "true",
    env = {
        "TEST": "$(echo $OS)",
    },
)
```

```
$ plz build //:envtest --shell
Temp directories prepared, total time 60ms:
  //:envtest: plz-out/tmp/envtest._build
    Command: true

bash-5.0$ echo $TEST
$(echo linux)
```

PR #2409, opened 2022-04-05
Fix config overrides for plugins

Tatskaari:


test/proto_plugin/test_repo/.plzconfig
```diff
@@ -1,9 +1,5 @@
-[Please]
-Version = 16.20.0-beta.14
-
 [Parse]
 BuildFileName = BUILD # We override this in the e2e test profile and the subrepo has BUILD files not BUILD_FILE
-PreloadSubincludes = ///go//build_defs:go
```
Tatskaari, line 6: Removing this gives us some coverage that the subincludes from the `///proto_go//build_defs:go` works

tiagovtristao (approved):

PR #2412, opened 2022-04-05
Stop trying to preload the subrepo's preloads

Tatskaari:
This is a data-race that came out on the go plugin PR (nice one @pebers) 

The RepoConfig field is populated when we build the subrepo target but we kick off the Parser preloading before that. This means that this might or might not be populated by the time we hit this code. 

We should probably kick off preloading the subincludes for the subrepo state when we have built its target and loaded its config. 

see: #2413


samwestmoreland:
So just trying to understand this. After this we won't be preloading the subrepo's preloads - do we pull them in some other way?

samwestmoreland (approved):

PR #2414, opened 2022-04-06
Handle booleans and ints in build config

Tatskaari:


tiagovtristao (approved):
I guess this makes sense

PR #2417, opened 2022-04-07
Don't debug non-test binaries in a sandbox

tiagovtristao:
This PR makes the following changes:

- Non-test binaries were debugged in a sandbox if the build configuration was set as such. This is no longer the case. The binary will always be debugged outside a sandbox.
- A new `$WD` environment variable is now made available that the debugger can use to launch the debugging session from a different location.
- Test binaries behave as before. They are executed in a sandbox if they were configured to be.

- The `$OUT`/`$TEST` env vars for `plz debug`/`plz test` are now absolute (required due to `$WD`) and this would be considered a **breaking change** for commands that used to be: `plz exec //:target -- ./\$OUT flag1 flag2`. Since v17 is around the corner it might make sense to wait for it before landing this.

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -509,7 +509,7 @@ var buildFunctions = map[string]func() int{
 		if !success {
 			return toExitCode(success, state)
 		}
-		return debug.Debug(state, opts.Debug.Args.Target, opts.Debug.Args.Args)
+		return debug.Debug(state, opts.Debug.Args.Target, opts.Debug.Args.Args, core.RepoRoot)
 	},
 	"exec": func() int {
 		success, state := runBuild([]core.BuildLabel{opts.Exec.Args.Target.BuildLabel}, true, false, false)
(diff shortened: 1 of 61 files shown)


src/please.go
```diff
@@ -509,7 +509,7 @@ var buildFunctions = map[string]func() int{
 		if !success {
 			return toExitCode(success, state)
 		}
-		return debug.Debug(state, opts.Debug.Args.Target, opts.Debug.Args.Args)
+		return debug.Debug(state, opts.Debug.Args.Target, opts.Debug.Args.Args, core.RepoRoot)
```
tiagovtristao, line 512: `core.RepoRoot` would be the default run location but this could be easily extended by making a `--wd` flag available if required.

tiagovtristao:
This wasn't meant to be closed, I guess github did it automatically when I purged my fork of the repo.

tiagovtristao:
I believe this is still relevant for v17 as an improvement to debugging.

PR #2419, opened 2022-04-07
Fix some races around plugins and subincludes

Tatskaari:


src/parse/asp/builtins.go
```diff
@@ -347,7 +347,11 @@ func subincludeTarget(s *scope, l core.BuildLabel) *core.BuildTarget {
 	// all available parser threads (easy to happen if they're all waiting on a single target which now can't start)
 	t := s.WaitForSubincludedTarget(l, pkgLabel)
 
-	pkg.RegisterSubinclude(l)
+	// TODO(jpoole): when pkg is nil, that means this subinclude was made by another subinclude. We're currently loosing
+	// this information here. We probably need a way to transitively record the subincludes.
+	if s.pkg != nil {
+		s.pkg.RegisterSubinclude(l)
+	}
```
Tatskaari, line 354: This was causing a race because `s.contextPackage()` can return a package that was written to by another go routine. It's also incorrect to register these on that package.

src/parse/asp/config.go
```diff
@@ -180,7 +180,11 @@ func pluginConfig(pluginState *core.BuildState, pkgState *core.BuildState) pyDic
 }
 
 func (i *interpreter) loadPluginConfig(pluginState *core.BuildState, pkgState *core.BuildState, c *pyConfig) {
-	pluginName := pluginState.Config.PluginDefinition.Name
+	if pluginState.RepoConfig == nil {
+		return
+	}
+
+	pluginName := pluginState.RepoConfig.PluginDefinition.Name
```
Tatskaari, line 187: This was just incorrect before. Doesn't fix any race condition.

src/parse/asp/interpreter.go
```diff
@@ -34,10 +34,15 @@ func newInterpreter(state *core.BuildState, p *Parser) *interpreter {
 		state:  state,
 		locals: map[string]pyObject{},
 	}
+	// If we're creating an interpreter for a subrepo, we should share the subinclude cache.
+	subincludes := &subincludeMap{m: map[string]subincludeResult{}}
+	if p.interpreter != nil {
+		subincludes = p.interpreter.subincludes
+	}
```
Tatskaari, line 41: This also doesn't fix any race conditions. It just means we don't duplicate work between subrepos.

src/parse/asp/parser.go
```diff
@@ -105,10 +105,11 @@ func (p *Parser) SubincludeTarget(state *core.BuildState, target *core.BuildTarg
 	}()
 	p.limiter.Acquire()
 	defer p.limiter.Release()
-	subincludePkgState := state
+	subincludePkgState := state.Root()
```
Tatskaari, line 108: This was causing a race because preloading from a subrepo was causing us to race on the RepoConfig field. We can guarantee that RepoConfig will be populated by now for the root repo and the subinclude target repo. The repo that this state object belongs to will most likely not be populated yet and will definitely race here.

PR #2420, opened 2022-04-07
Handle race between Build() and ShowProgress()

Tatskaari:


peterebden:
Looks good but the build failure needs a tweak.

peterebden (approved):

peterebden:
Think I'm seeing this atm too (guessing it's rareish and happens more with more parallelism?), let's pull the trigger

PR #2424, opened 2022-04-11
Bump 16.20.0 beta.15

tiagovtristao:


Tatskaari (approved):
we should probably do a full release soon. This seems pretty stable in core3 now.

tiagovtristao:
> we should probably do a full release soon. This seems pretty stable in core3 now.

I was going to this here but noticed there were some commits that fixed some plugin-related things, it's better to try them out first.

PR #2425, opened 2022-04-11
Fix scope issues regarding builtins

tiagovtristao:
This fixes the following issue that happens when `plz query changes --include_dependees=transitive --since origin/master > plz-out/changes` is run:

```
**.build_defs:482:40: error: Non-iterable type none; must be a list
def _dict_to_tfvar(inputdict:dict, key_formatter=lambda val: "{}".format(val)):
    return "{{{}}}".format(", \n".join(["{} = \"{}\"".format(key_formatter(k), v) for k, v in items(inputdict)]))
```

This seems to happen only at the "after" run build in `plz query changes` and looks like it might be related to previous state "pollution". This was introduced in https://github.com/thought-machine/please/commit/8c646de483af3e9240438393b15fff5df96c9d96 where builtin methods aren't re-registered with recent scopes which seems to be what's causing this issue. I've tried to wrap the whole contents of `registerBuiltins` in `sync.Once` but even more errors were showing up, leading to believe that the methods were being triggered with the wrong scope.


peterebden (approved):
Ah good point about the Once!

I like this tho, nice for them not to be globals.

samwestmoreland (approved):

Tatskaari (approved):
Ah nice one!

PR #2426, opened 2022-04-11
Load the subrepo config after the subrepo has been built

Tatskaari:
We also need to defer preloading subincludes until this has been loaded too.


samwestmoreland (approved):
Looks good to me!

tiagovtristao (approved):

PR #2429, opened 2022-04-12
Stop adding //_please:all in subrepos

Tatskaari:
Each subrepo was building it's own version of each rule.


tiagovtristao (approved):
Would this have any impact on plugins that use different plz versions?

PR #2432, opened 2022-04-12
Fix plugin_repo erroring out on top level BUILD

tiagovtristao:
This fixes https://github.com/thought-machine/please-issues/issues/67.

What was happening was that:
  `label == BuildLabel{PackageName:"build_defs", Name: "cc", Subrepo:"cc"}`
  `dependent == BuildLabel{PackageName:"", Name: ":all", Subrepo:""}`

and once `sl := label.SubrepoLabel(state, dependent.Subrepo)` is called:
  `sl == BuildLabel{PackageName:"", Name: "cc", Subrepo:""}`

therefore `localSubinclude := sl.PackageName == dependent.PackageName && forSubinclude` would always be true for the `preloadsubincludes = ///cc//build_defs:cc` on `.plzconfig`.

My understanding of the `localSubinclude` logic is that it should be true if the subrepo and package are the same between `label` and `dependent`.

### DIFF
--- a/src/parse/parse_step.go
+++ b/src/parse/parse_step.go
@@ -91,7 +91,7 @@ func checkSubrepo(tid int, state *core.BuildState, label, dependent core.BuildLa
 	sl := label.SubrepoLabel(state, dependent.Subrepo)
 
 	// Local subincludes are when we subinclude from a subrepo defined in the current package
-	localSubinclude := sl.PackageName == dependent.PackageName && forSubinclude
+	localSubinclude := label.Subrepo == dependent.Subrepo && label.PackageName == dependent.PackageName && forSubinclude
 
 	// If we're including from the same package, we don't want to parse the subrepo package
 	if !localSubinclude {


src/parse/parse_step.go
```diff
@@ -91,7 +91,7 @@ func checkSubrepo(tid int, state *core.BuildState, label, dependent core.BuildLa
 	sl := label.SubrepoLabel(state, dependent.Subrepo)
 
 	// Local subincludes are when we subinclude from a subrepo defined in the current package
-	localSubinclude := sl.PackageName == dependent.PackageName && forSubinclude
+	localSubinclude := label.Subrepo == dependent.Subrepo && label.PackageName == dependent.PackageName && forSubinclude
```
Tatskaari, line 94: Nice! I think this makes a lot more sense. I can see how this fixes the bug before and handles the following case well: 

```
filegroup(
    name = "my_def",
    ...
)

subinclude(":my_def")
```

The comment above implies that it's also meant to handle sub-including from subrepos. I was wondering what happens when you The other is when we have:
```
subrepo(
    name = "my_plugin",
    ...
)

subinclude("///my_plugin//build_defs")
```

I'm not sure we actually need to fix this (it might already work) but we should probably update the comment at least.

Tatskaari (approved):

src/parse/parse_step.go
tiagovtristao: So the `localSubinclude` logic lives within the `checkSubrepo` so it seems to be exclusive to subrepo logic. I tried the later example and it works fine.

I was testing this more and bumped into a race condition but it seems related to a race that we already know about. For instance, if you have:

`.plzconfig`
```ini
preloadsubincludes = ///my_plugin//build_defs
```

`BUILD`
```python
subrepo(
    name = "my_plugin",
    ...
)

subinclude("///my_plugin//build_defs")
```

This seems unrelated to my change, given that the subrepo (or the target that we are interested in) only becomes activated when the package is fully "resolved".

PR #2435, opened 2022-04-14
Expose `ExperimentalDir` to BUILD files

chrisnovakovic:
Expose the value of the `Parse.ExperimentalDir` configuration option to BUILD files as `CONFIG.EXPERIMENTAL_DIRS`, allowing build defs to distinguish between experimental and non-experimental code and therefore treat them differently.


peterebden:
> allowing build defs to distinguish between experimental and non-experimental code and therefore treat them differently.

I'm not convinced this is a thing that build defs should be doing (or plz should be encouraging). Things outside the experimental dir aren't really supposed to be aware of things in it.

chrisnovakovic:
> > allowing build defs to distinguish between experimental and non-experimental code and therefore treat them differently.
> 
> I'm not convinced this is a thing that build defs should be doing (or plz should be encouraging). Things outside the experimental dir aren't really supposed to be aware of things in it.

@Tatskaari said the same thing, and proposed #2436 after I explained to him why I thought this would be necessary (I'm trying to ensure that `remote_file` isn't used outside of `third_party` plus a few other directories including the experimental directories, and thought it'd make more sense to get that value from `.plzconfig` rather than assuming there'll always be only one and its name will always be `experimental`). His solution is a lot cleaner, and it has the added benefit that it can't be evaded simply by using `build_rule` directly.

We agreed there was no worthwhile use for this, I just forgot to close it 🙂

PR #2437, opened 2022-04-14
Move the remote_file() headers out of curl

Tatskaari:


rules/misc_rules.build_defs
```diff
@@ -451,41 +454,22 @@ def remote_file(name:str, url:str|list, hashes:list=None, out:str=None, binary:b
...
-
-    if curl:
-        for header, value in headers.items():
-            header_flag = f'{header_flag} --header "{header}: {value}"'
-        for header, value in secret_headers.items():
-            header_flag = f'{header_flag} --header "{header}: $(cat {value})"'
-
-        # Authentication info
-        user_info = f'--user {username}:$(cat {password_file})' if auth else ""
+    labels = labels + [f"remote_file:header:{h}:{v}" for h, v in headers.items()]
```
Tatskaari, line 460: I didn't want to add more args to `build_rule()` so I'm doing this via labels

samwestmoreland (approved):
Lgtm!

src/build/build_step.go
```diff
@@ -1131,6 +1136,65 @@ func fetchOneRemoteFile(state *core.BuildState, target *core.BuildTarget, url st
...
+		return str[:i], str[(i + 1):]
+	}
+
+	userName := ""
+	password := ""
+	for _, l := range target.Labels {
+		param, value := param(l)
+		switch param {
+		case "":
+			continue
```
tiagovtristao, line 1159: Does Go not allow an empty `case`?

src/build/build_step.go
```diff
@@ -1131,6 +1136,65 @@ func fetchOneRemoteFile(state *core.BuildState, target *core.BuildTarget, url st
...
+		case "header":
+			k, v := header(value)
+			log.Warningf(v)
```
tiagovtristao, line 1162: Should this be a `log.Debugf` instead?

src/build/build_step.go
```diff
@@ -1131,6 +1136,65 @@ func fetchOneRemoteFile(state *core.BuildState, target *core.BuildTarget, url st
...
+			v = os.Expand(v, env.ReplaceEnvironment)
+			log.Warningf("setting header: %v: %v", k, v)
```
tiagovtristao, line 1164: Same here

src/build/build_step.go
```diff
@@ -1131,6 +1136,65 @@ func fetchOneRemoteFile(state *core.BuildState, target *core.BuildTarget, url st
...
+			if err != nil {
+				return fmt.Errorf("failed to read password file: %v", err)
+			}
+			password = string(p)
+		default:
+			return fmt.Errorf("unknown remote file label: %v", l)
+		}
+	}
+
+	if userName != "" {
```
tiagovtristao, line 1187: Might be worth checking `password` here too?

rules/misc_rules.build_defs
tiagovtristao: The usage of `labels` is quite neat and extensible.

tiagovtristao (approved):
Nice, it's quite clean.

src/build/build_step.go
Tatskaari: Yeah, that's what the default case below is. This is matching the case where `param()` returns an empty string, which it will do when the label isn't a parameter. The default case handles when we do something like :`remote_file:nonsense_param:blah` which is different.

src/build/build_step.go
Tatskaari: I don't think it's necessarily wrong to have an empty password.

PR #2443, opened 2022-04-22
Increase bootstrap version

peterebden:
Fixes a spurious warning during bootstrap:
```
10:11:29.860 WARNING: Plugin "go" requires Please version 16.20.0-beta.14
10:11:29.907 WARNING: Plugin "go" requires Please version 16.20.0-beta.14
```
Historically that version never had any real significance; this makes it larger than any expected version so a master build will always be new enough for any plugin version.

### DIFF
--- a/test/remote/BUILD
+++ b/test/remote/BUILD
@@ -2,22 +2,22 @@
 remote_file(
     name = "elan",
     binary = True,
-    hashes = ["dd8f7d546d270a198b271841c0d687cd366c770589d34e2f654964d6447fdbd8"],
-    url = "https://github.com/thought-machine/please-servers/releases/download/v9.5.1/elan",
+    hashes = ["afabbbae6cb22045a7ee1d53edf0e298602ef6d41a3a2554e1e06555f9827eb4"],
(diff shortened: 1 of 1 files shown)


Tatskaari (approved):
Ah nice!

samwestmoreland (approved):

peterebden:
I'm not sure why I've seen so many test-rex failures on this PR; I can't think how they'd be related. Am bumping the Mettle server versions a bit in case that helps (there are some mildly nice fixes in there)

Tatskaari:
> I'm not sure why I've seen so many test-rex failures on this PR; I can't think how they'd be related. Am bumping the Mettle server versions a bit in case that helps (there are some mildly nice fixes in there)

I've never been able to figure out why the rex tests fail. We managed to get to the bottom of a few issues but it does seem to be happening quite frequently now.

peterebden:
Yeah this was just a surprising number in sequence on this apparently unrelated PR. My theory would be that something's getting killed (possibly OOMKilled?) given the way it seems to be running OK, then suddenly it's all "connection reset by peer" and gRPC errors.

PR #2446, opened 2022-04-27
Only log errors for the Please shim

tiagovtristao:
This will prevent warnings for unknown fields (as seen below) if the Please shim version differs from the Please binary installed.

```
17:14:13.022 WARNING: Error in config file: warning:
can't store data at section "featureflags", variable "ExcludeShellRules"
```

### DIFF
--- a/tools/please_shim/main.go
+++ b/tools/please_shim/main.go
@@ -28,7 +28,7 @@ var opts struct {
 	} `group:"Options controlling what to build & how to build it"`
 
 	OutputFlags struct {
-		Verbosity cli.Verbosity `short:"v" long:"verbosity" description:"Verbosity of output (error, warning, notice, info, debug)" default:"error"`
+		ShimDebugVerbosity bool `long:"shim_debug_verbosity" description:"Shim debug verbosity of output"`
 	} `group:"Options controlling output & logging"`
 
 	FeatureFlags struct {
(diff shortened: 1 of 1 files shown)


samwestmoreland (approved):
Lgtm!

tools/please_shim/main.go
```diff
@@ -27,10 +27,6 @@ var opts struct {
 		Profile  []core.ConfigProfile `long:"profile" env:"PLZ_CONFIG_PROFILE" env-delim:";" description:"Configuration profile to load; e.g. --profile=dev will load .plzconfig.dev if it exists."`
 	} `group:"Options controlling what to build & how to build it"`
 
-	OutputFlags struct {
-		Verbosity cli.Verbosity `short:"v" long:"verbosity" description:"Verbosity of output (error, warning, notice, info, debug)" default:"warning"`
```
peterebden, line 31: I wonder if you want to leave this in but change the name so it's still possible to run the shim in verbose mode if needed?

tiagovtristao: Agreed.

peterebden: Oh, I do think you might need to rename it though if you do that - I think we often invoke plz in CI like `plz -p -v 2 build //blahblah` which would override this

tiagovtristao: Ah, I see what you mean now. Will change.

tools/please_shim/main.go
```diff
@@ -71,6 +71,26 @@ func parseFlags() (*flags.Parser, error) {
 	return parser, nil
 }
 
+func setLogging() {
+	verbosity := cli.MinVerbosity
+	if opts.OutputFlags.ShimDebugVerbosity {
+		verbosity = cli.MaxVerbosity
+		os.Args = filterArgsFlag(os.Args, "--shim_debug_verbosity")
```
tiagovtristao, line 78: This is required otherwise the main binary will throw a fatal error at the unknown flag. Having this flag as boolean simplifies the filtering too.

PR #2449, opened 2022-04-29
Bring plz exec process to the foreground

tiagovtristao:
This brings the process group created by `plz exec` to the foreground by default, making it behave like `plz run`. This should be fine as far as Please is concern, since it is already listening to when the spawn process terminates.


tiagovtristao:
Might have to think how to get around the current test failures.

tiagovtristao:
> I've been seeing these errors pop up quite unexpectedly when trying to debug in intellij as well... not exactly sure what it means.

Is the Intellij plugin already using `plz debug`? If so, are you seeing these errors before this PR, or applying the changes of this PR?

PR #2450, opened 2022-05-03
rex: Fix logic for rerunning tests when --num_runs is set  

Tatskaari:


peterebden:
I guess this will end up making sense but could you clarify a bit what was wrong before / what's being fixed please?

Tatskaari:
> I guess this will end up making sense but could you clarify a bit what was wrong before / what's being fixed please?

When we `plz test -n 5 //some:test`, I kind of expect it to skip cache lookup, and actually run the tests just like when we provide `--rerun`. The logic before was a little flawed. If you don't pass `--rerun`, it would pick up the cache regardless. We also didn't check the number of runs when asking the remote APIs to skip cache lookup.

peterebden (approved):
Cool thanks, makes sense.

I don't think I ever looked at this very hard since there is always an issue where the remote server is allowed to merge concurrent runs of the same action (there's no way in REAPI to indicate not to do that). This should be better though!

PR #2461, opened 2022-05-11
unconditionally add remote_file label to remote_files

nmpowell:
So we can query for all remote_files like so: `plz query alltargets --hidden --include=remote_file //...`

### DIFF
--- a/rules/misc_rules.build_defs
+++ b/rules/misc_rules.build_defs
@@ -457,6 +457,7 @@ def remote_file(name:str, url:str|list, hashes:list=None, out:str=None, binary:b
     if bool(username) != bool(password_file):
         fail(f"Both or neither of username and password_file should be specified")
 
+    lables = labels + ["remote_file"]
     labels = labels + [f"remote_file:header:{h}:{v}" for h, v in headers.items()]
     labels = labels + [f"remote_file:secret_header:{h}:{v}" for h, v in secret_headers.items()]
 


nmpowell:
Other suggestions for how we could do this welcome!

rules/misc_rules.build_defs
```diff
@@ -457,6 +457,7 @@ def remote_file(name:str, url:str|list, hashes:list=None, out:str=None, binary:b
     if bool(username) != bool(password_file):
         fail(f"Both or neither of username and password_file should be specified")
 
+    lables = labels + ["remote_file"]
```
peterebden, line 460: lables -> labels

The perils of dynamic languages!

peterebden: Also WYH: could you change this to a single statement please? i.e.
```
labels += [
    ...
]
```
mildly nicer than repeating the initial addition each time!

peterebden:
Hmmmm also looks like we do have a `remote` label on them all here: https://github.com/thought-machine/please/blob/498599b3573e7c2422b3bde4be261aafbe2459f0/src/parse/asp/targets.go#L136

I'm not fussed about the specific name of the two, but would rather have just one. (Changing will need a little time though...)

nmpowell:
I think we can do what we want with `plz query print //third_party/python3/... --include remote`

PR #2463, opened 2022-05-13
Plugin halp messages

peterebden:
Adds a Help field on plugin config settings so they can self-describe what they do.

Also adds a mechanism to stop producing help for builtin sections when they are disabled by a feature flag; previously `plz halp go` always answered for the builtin section first, even when you're using the plugin.
That requires us to attempt to load the config before producing help, which I think we have to just suck up now plugins are becoming more widely used.

Finally, fiddled with the output a bit to make the non-help part a little more compact. Before:
```
go is a plugin defined in the .plzconfig file.

This plugin has the following options:
    (string) (optional) Default value: [cc]
    (string) (optional) Default value: []
   pleasegotool (string)Default value: [//tools:please_go]
    (string)Default value: [dlv]
    (bool)Default value: [false]
    (string) (optional) Default value: []
    (string) (optional) Default value: []
    (string)Default value: [go]
    (string) (optional) Default value: []
    (string)Default value: [ar]
    (bool)Default value: [false]
    (bool)Default value: [false]
```
After:
```
go is a plugin defined in the .plzconfig file.

This plugin has the following options:
   ar_tool (string) (default: ar) 
   c_flags (string) (optional) 
   cc_tool (string) (optional, default: cc) 
   cgo_enabled (string) (optional) 
   cpp_coverage (bool) (default: false) 
   default_static (bool) (default: false) 
   delve_tool (string) (default: dlv) 
   go_tool (string) (default: go) 
   import_path (string) (optional) 
   ld_flags (string) (optional) 
   pleasegotool (string) (default: //tools:please_go) 
   test_root_compat (bool) (default: false) 
```
(The actual plugins don't have any help text set yet)


Tatskaari (approved):

samwestmoreland (approved):
Nice, that's much cleaner. Thanks!

Tatskaari (approved):
LGTM

peterebden:
> Do a better job than Pebers

I like to think that my role is to provide inspiration which causes you to raise your performance ;-)

PR #2464, opened 2022-05-15
Experimental static variable analysis for asp

peterebden:
Basically it finds cases where variables are written to but never read from. I'm pretty happy with the extension to WalkAST, it feels pretty elegant to write the logic this way.

Has located one case in go_rules (although admittedly in `go_get` which is deprecated) and would have spotted the one in #2461 too (tried that locally)
Does have some false positives for our builtins where we do things like `cxx_binary = cc_binary`; maybe just needs some kind of whitelisting? If that is sorted it might be nice to use this as a premerge test (although also need to think how this would work for plugins)

### DIFF
--- a/src/parse/asp/main/main.go
+++ b/src/parse/asp/main/main.go
@@ -65,18 +65,14 @@ func parseFile(pkg *core.Package, p *asp.Parser, filename string) error {
 }
 
 type assignment struct {
+	Name string
 	Pos  asp.Position
 	Read bool // does it get read later on?
 }
 
-type astError struct {
-	Name string
-	Pos  asp.Position
-}
-
 // checkAST runs some static checks on a loaded AST.
 // Currently this checks for variables that are assigned to but not read.
-func checkAST(stmts []*asp.Statement, parentScopes ...map[string]assignment) (errs []astError) {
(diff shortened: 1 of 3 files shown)


src/parse/asp/main/main.go
```diff
@@ -31,6 +32,7 @@ var opts = struct {
 	NumThreads   int           `short:"n" long:"num_threads" default:"10" description:"Number of concurrent parse threads to run"`
 	ParseOnly    bool          `short:"p" long:"parse_only" description:"Only parse input files, do not interpret them."`
 	DumpAst      bool          `short:"d" long:"dump_ast" description:"Prints AST to stdout. Implies --parse_only."`
+	Check        bool          `short:"c" long:"check" description:"Runs some static checks on the parsed files. Implies --parse_only."`
```
Tatskaari, line 35: Can we just set this as part of `plz fmt`?

Tatskaari (approved):
Nice one! I'm sure this will turn up a decent amount of stuff in core3.

src/parse/asp/main/main.go
peterebden: Yeah that might be a good place for it to live. It's a _little_ weird since it's not actually a format thing though...

Issue #2462, opened 2022-05-12
Config and template cache URL(s) for remote_file rule to allow automatic fetching from a cache


nmpowell:
The `python_wheel` rule allows a template `CONFIG.PYTHON_WHEEL_NAME_SCHEME` together with a [`CONFIG.PYTHON_WHEEL_REPO`](https://github.com/thought-machine/please/blob/e030e9957b840c2cb9dee70fa5bdf2ccb683156e/rules/python_rules.build_defs#L515) to define URLs which can be checked before downloading wheels from external libraries (i.e. PyPI).

The `remote_file` rule [takes a list of urls](https://github.com/thought-machine/please/blob/e030e9957b840c2cb9dee70fa5bdf2ccb683156e/rules/misc_rules.build_defs#L378) to fetch from, which are checked in series until a fetch succeeds.

We would like to add similar config functionality as `python_wheel` to `remote_file` (which the wheel rule uses anyway under the hood).

We have devendorised most of our third_party dependencies. These now depend upon `remote_file` rules which download external resources as and when we `plz build`. The end goal is to `plz build` these targets and allow them to seamlessly fall back on a locally-managed storage bucket (e.g. S3 / cloud or local file cache) URL in case the external resource goes away / becomes unavailable. Just as happens for Python wheels with a custom storage bucket at present.

`.plzconfig` would provide something like:
```
[buildconfig]
cache-priority = low
cache-url = https://s3-zone.amazonaws.com/company-bucket

[remote_file]
name-scheme = {url_base}/{directory_name}/{file_name}
```

The priority above would indicate whether the cache took priority over the external resource.

This could be more generic and (perhaps eventually) could fully replace the `python_wheel` functionality; different rules would rely upon different name schemes and take different arguments to populate that string.

PR #2475, opened 2022-05-26
Remote file templated url

nmpowell:
See #2462. Allow `remote_file` rules to specify a template a Cache URL which can be prioritised over the `url` passed to the `remote_file` rule itself.

In a similar way to `python_wheel` rules specifying a custom wheel naming scheme, here custom cache name scheme(s) can be defined. This should allow escrow backups of remote files in e.g. a cloud storage bucket, or as local files with e.g. `cacheurl = file:///var/tmp/escrow/local_storage`.

Would use a `name_scheme` like this in `.plzconfig`:
```
[remote_file]
prioritisecache = true
cacheurl = https://some-cloud-provider.com/some-bucket-name
cachenamescheme = {url_base}/{cache_path}/{file_name}
```

Also add a `base64url` function.

By default `base64url` is used to generate the `cache_path` in the above name scheme using the `dirname` of a URL (i.e. omitting the file name). Thus for two remote files with the same dirname
`http://google.com/file.a`: `{cache_url}/aHR0cDovL2dvb2dsZS5jb20/file.a`
`http://google.com/file.b`: `{cache_url}/aHR0cDovL2dvb2dsZS5jb20/file.b`


docs/config.html
```diff
@@ -2251,6 +2251,57 @@ <h3 class="mt1 f6 lh-title" id="proto.jsdep">JsDep</h3>
...
 
+<section class="mt4">
+  <h2 id="remote_file" class="title-2">[RemoteFile]</h2>
+
+  <p>Properties that affect how the <code class="code">remote_file</code> build rule works.</p>
+
+  <ul class="bulleted-list">
+    <li>
+      <div>
+        <h3 class="mt1 f6 lh-title" id="remote_file.wheelrepo">WheelRepo</h3>
```
chrisnovakovic, line 2262: This should presumably be `CacheURL` 🙂

rules/builtins.build_defs
```diff
@@ -283,3 +283,7 @@ def is_semver(s:str) -> bool:
 
 def semver_check(version:str, constraint:str) -> bool:
     pass
+
+def base64_url(s:str) -> str:
```
chrisnovakovic, line 287: It'd be better to call this `base64url` (without the underscore), since that's the name of the function according to [RFC 4648](https://datatracker.ietf.org/doc/html/rfc4648#section-5), and the underscore implies that it only consumes URLs.

nmpowell:
I'm not sure whether I've done everything necessary to set up the `[remote_file]` stuff in `.plzconfig`.

peterebden:
`remote_file` supports multiple URLs passed to it, which are tried in sequence. That's the underlying mechanism used to back `python_wheel`. To me that seems sufficient; you simply put your cache URL in as the first entry and it'll be tried before the others, but it won't fail if it doesn't have the thing in question.

nmpowell:
> To me that seems sufficient; you simply put your cache URL in as the first entry and it'll be tried before the others

We want to be able to download from our bucket for _every_ remote file, for instance in case the external resource gets moved or deleted, we'll fall back to the cache. It would be arduous to define the cache url for every `remote_file` target.

nmpowell:
... Understand the concern though; I'll have a think about how to make this a bit more elegant ...

chrisnovakovic:
> `remote_file` supports multiple URLs passed to it, which are tried in sequence. That's the underlying mechanism used to back `python_wheel`. To me that seems sufficient; you simply put your cache URL in as the first entry and it'll be tried before the others, but it won't fail if it doesn't have the thing in question.

As Nick said, this is mostly about reducing the effort needed to effectively add an extra `src` for every `remote_file` - is there a better way to achieve that than what Nick is proposing here?

Tatskaari:
> You could define your own build_def that wraps remote file and adds the cache url first to the underlying remote file?

Yeah I think I prefer this idea. This seems a little too magical for a low level rule like remote file.

PR #2482, opened 2022-06-01
Fix building initial targets in a subrepo that race with preloaded subincludes

peterebden:
I have an internal example of this but difficult to produce a reduced test case.

Basically if you ask for `plz build ///cc/grpc/grpc//...`, it needs to build `//cc/grpc` upfront (in `plz.go`, because it has to download the subrepo to figure out what to do in it). That sets `forSubinclude` which bypasses the checks on the preloaded subincludes completing, and hence will fail if that build file expects one of them to be there.

It feels a bit fiddly having to break all these functions up but I think this is the right answer.


peterebden:
Yeah I think this is just an unintentional interaction of the two, normally you don't notice that `forSubinclude` getting set.

PR #2491, opened 2022-06-09
Trim path when cross compiling the go sdk

Tatskaari:


brianbianco:
When trying this PR I get the following error:

```
flag provided but not defined: -trimpath
```

Changing line 59 to instead be the following makes it complete successfully:

``` 
cmd += f' && (export GOOS={goos} && export GOARCH={goarch} && $OUT/bin/go install{tag_flag} --trimpath std)'
```

However when running strings on the resultant binary I still get the following entry

```
strings plz-out/gen/mybin | grep brian
/Users/brianbianco/code/myproject/plz-out/bin/tools/golang/1.17.6
```

All other pathing has been stripped out.

brianbianco:
I've up a PR with what I believe is the fix for this @ https://github.com/thought-machine/please/pull/2492

PR #2492, opened 2022-06-09
Trim path when cross compiling the go sdk

brianbianco:


brianbianco:
Should be a working version of PR #2491

Tatskaari:
Ah brilliant thanks :) I had made this change locally but didn't push :S

PR #2493, opened 2022-06-09
Add custom symlink wrappers for plz generate

### LINKED ISSUE #2488: Committing `plz generate`d sources to git
👋 , I'm trying to use `plz generate` to automatically generate protobufs into source that can then be committed and pushed to a git repository. The end-goal of this is to have Go proto types readily importable in a similar fashion to https://pkg.go.dev/google.golang.org/genproto.

To reproduce this, you can:

1. init a temporary git repo which uses plz locally.
    ```bash
    $ cd "$(mktemp -d)" && git init && plz init
    ```
2. make plz use hard links so the full source is copied into git.
    ```bash
    $ cat <<EOF >> .plzconfig
    
    [build]
    LinkGeneratedSources = hard
    UpdateGitignore = false
    EOF
    ```
3. add a `BUILD` file that generates something with the `codegen` label.
    ```bash
    $ cat <<EOF >> BUILD
    genrule(
        name = "generate",
        labels = ["codegen"],
        outs = ["generated.txt"],
        cmd = "echo foo > generated.txt"
    )
    EOF
    ```
4. run `plz generate` and observe the new `generated.txt` with `foo` inside.
    ```bash
    $ plz generate && cat generated.txt
    ```
5. change the genrule to output 'bar' instead of 'foo'
    ```bash
    $ sed -i 's/foo/bar/g' BUILD
    ```
6. run `plz generate` again and observe that the `generated.txt` hasn't changed. I was expecting it to now contain `bar`, so that I could commit the change to git. The only way I can get it to have `bar` is by removing the `generated.txt` file entirely.
    ```bash
    $ plz generate && cat generated.txt
    $ rm generated.txt && plz generate && cat generated.txt
    ```

Is this expected behaviour? I've tracked it down to this call of `fs.LinkIfNotExists`(https://github.com/thought-machine/please/blob/master/src/generate/generate.go#L58) which returns straight away if the path already exists. I guess we'd need to refactor that call slightly to something like:
```go
func Link(src, dest string, hard bool) {
	if PathExists(dest) && !hard {
		return
	}
	if PathExists(dest) && hard {
		// remove existing hard links as git won't follow them
		if err := os.Remove(dest); err != nil {
			log.Warning("Failed to remove link %s: %s", dest, err)
		}
	}
	Walk(src, func(name string, isDir bool) error {
		...
	})
}
```

Does that sort of make sense to you? Or i can see a solution where we wrap the `os.Symlink` and `os.Link` functions with that logic to avoid changing the signature, e.g. 
```go
func symlinkIfNeeded(src, dest string) error {
	if PathExists(dest) {
		return nil
	}
	
	return os.Symlink(src, dest)
}

func hardlink(src, dest string) error {
	if PathExists(dest) {
		// remove existing hard links as git won't follow them
		if err := os.Remove(dest); err != nil {
			return fmt.Errorf("could not remove link %s: %w", dest, err)
		}
	}
	
	return os.Link(src, dest)
}

func LinkGeneratedSources(state *core.BuildState, labels []core.BuildLabel) {
	linker := symlinkIfNeeded

	if state.Config.Build.LinkGeneratedSources == "hard" {
		linker = hardLink
	}
	
	...
	fs.Link(path.Join(srcDir, out), path.Join(destDir, out), linker)
}
```

VJftw:
Fixes #2488, this is an example implementation of a fix for it :)

I'm not sure how granular you'd like the checks to be, e.g. with this, a symbolic link pointing at the wrong file will still be skipped. 

Let me know if there's anything else you'd prefer :)

### DIFF
--- a/src/fs/copy.go
+++ b/src/fs/copy.go
@@ -1,6 +1,7 @@
 package fs
 
 import (
+	"fmt"
 	"os"
 	"path"
 )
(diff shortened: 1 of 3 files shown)


Tatskaari (approved):
Thanks VJ! I think this makes sense.

Issue #2495, opened 2022-06-11
`plz query deps --include ...` regression


VJftw:
:wave: again, since updating to Please 16.20.1+, I've been experiencing a regression in the `plz query deps --include ...` functionality.

With the given example Please workspace:
1.  Set-up Please in a temporary folder:
     ```bash
    $ cd "$(mktemp -d)" && git init && plz init
    # uncomment the please version configuration
    $ sed -i '/; \[please\]/s/; //' .plzconfig
    $ sed -i '/; version/s/; //' .plzconfig
    ```
2. Create a genrule that depends on another genrule that has a label `my_dep`:
    ```bash
    $ cat <<EOF >> BUILD
    genrule(
        name = "my_dep",
        labels = ["my_dep"],
        outs = ["my_dep.txt"],
        cmd = "echo foo > my_dep.txt",
    )
    
    genrule(
        name = "my_rule",
        labels = ["my_rule"],
        outs = ["my_rule.txt"],
        cmd = "cat \$(location :my_dep) > my_rule.txt",
        deps = [":my_dep"],
    )
    EOF
    ```

## With Please `16.20.0`

```bash
$ sed -i '/version =/s/= .*/= 16.20.0/' .plzconfig
$ plz query deps //:my_rule
//:my_rule
  //:my_dep
$ plz query deps --include my_dep //:my_rule
  //:my_dep
# note the above output
```

## With Please `16.20.1`+

However, with any more recent version, adding `--include my_dep` returns an empty list, e.g.:

```bash
$ sed -i '/version =/s/= .*/= 16.20.1/' .plzconfig
$ plz query deps //:my_rule
//:my_rule
  //:my_dep
$ plz query deps --include my_dep //:my_rule
# no output
```

Tatskaari:
Sorry about this! The behaviour before was pretty poorly designed and would end up printing things at odd indentation levels if we exclude them from the results. 

When I fixed the issue, I thought about this case but decided that if the target was excluded from the results, then we should stop traversing its dependencies. In hindsight, I think it makes sense to keep traversing, but just not increment the indentation level.

VJftw:
no problem! thank you for the quick fix :heart: i've been using it in the same way as: https://github.com/thought-machine/dracon/blob/master/scripts/development/kind/util.sh#L11-L12 to get us a list of docker images that a K8s config depends on 😄

PR #2496, opened 2022-06-13
Fix bug where we wouldn't traverse excluded deps

Tatskaari:
Fixes: #2495 

I wasn't sure how this should behave, but in hindsight I think this makes more sense. 

We used to stop traversing dependencies if that dependency was excluded from the results via a label filter. This instead just skips printing that dependency but still traverses it.


VJftw (approved):
Thank you! 🚀

PR #2503, opened 2022-06-21
Expose `go_binary` sources when included as a `go_test` dependency

chrisnovakovic:
Currently, the binary outputted by a `go_binary` is brought in as an input to a `go_test` when the `go_binary` is included as a dependency. This prevents the sources of a `go_binary` from being tested by a `go_test` without either mirroring the `srcs` and `deps` from the `go_binary` into the `go_test` (which is repetitive), or splitting the bits to be tested into a `go_library` that is then included as a dependency of the `go_binary` (which is cumbersome).

Use the `requires`/`provides` mechanism to output a `go_binary`'s sources instead of a binary when a `go_test` is being built, thus avoiding the need for either of the above workarounds.


peterebden (approved):
Thanks!

PR #2506, opened 2022-06-22
Fix hang in `plz query deps`

peterebden:
Think this happened in #2496 ; we still need to mark targets as done, even if we don't print them, otherwise we walk every possible path downwards through the dependency tree which can take eternity.

Bit tricky to give specific instructions to reproduce but it manifests as a hang when using `plz query deps` on a sufficiently complex tree of stuff.


Tatskaari (approved):
Ah right sorry... thanks for that.

PR #2507, opened 2022-06-24
Make tab completion work on test args

peterebden:
This means that if you want to test two targets simultaneously you can do `plz test //src/core:state_test //src/co <tab>` and get build label completion on the second one.


Tatskaari (approved):
Ah nice! Looks good to me.

PR #2508, opened 2022-06-28
Strip host arch from subrepo part of build label

samwestmoreland:
Had an issue when cross-compiling the python plugin:
```
Build stopped after 160ms. 2 targets failed:
    ///go_linux_amd64//build_defs:go                                                      
Subrepo go_linux_amd64 is not defined (referenced by ///linux_amd64//tools/please_pex:all)
```
The cross-compiled tools in the python plugin were depending on the `///go_{arch}//...` which is fine until the arch is the host arch, in which case we don't have a go subrepo by that name, and so we should just depend on the ///go//... subrepo.

### DIFF
--- a/src/core/build_label.go
+++ b/src/core/build_label.go
@@ -183,7 +183,7 @@ func ParseBuildLabelContext(target string, pkg *Package) BuildLabel {
 		if subrepo == "" && pkg.Subrepo != nil && (target[0] != '@' && !strings.HasPrefix(target, "///")) {
 			subrepo = pkg.Subrepo.Name
 		} else if arch := cli.HostArch(); strings.Contains(subrepo, "_"+arch.String()) {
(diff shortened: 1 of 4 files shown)


src/core/build_label.go
```diff
@@ -181,6 +182,10 @@ func ParseBuildLabelContext(target string, pkg *Package) BuildLabel {
 	if p, name, subrepo := parseBuildLabelParts(target, pkg.Name, pkg.SubrepoName); name != "" {
 		if subrepo == "" && pkg.Subrepo != nil && (target[0] != '@' && !strings.HasPrefix(target, "///")) {
 			subrepo = pkg.Subrepo.Name
+		} else if arch := cli.HostArch(); strings.Contains(subrepo, "_"+arch.String()) {
+			subrepo = strings.ReplaceAll(subrepo, "_"+arch.String(), "")
```
Tatskaari, line 186: nit: TrimSuffix would be more correct here.

PR #2511, opened 2022-07-12
Do not iterate over ls output in install script

egladman:
Iterating over `ls` output is fragile, and prone to breakage. It is best practice to use globbing.

Additional Reading
- https://www.shellcheck.net/wiki/SC2045


Tatskaari (approved):
Thank you for the PR! 

Looks much cleaner. Was using the old back-tick stuff which is deprecated too :disappointed:

PR #2513, opened 2022-07-22
Disallow recursive dirs

peterebden:
Seems that the SDK will incorrectly convert an empty directory to one containing a single directory named `.`. This causes us to infinitely recurse if we encounter such a thing.

This fixes that by simply ignoring it. It pretty much seems to work except for downloading the outputs of that action (which I currently can't find a nice solution for) - so this is strictly better but not a complete solution.


Garbett1 (approved):
LGTM

PR #2518, opened 2022-07-26
Use python plugin

samwestmoreland:


peterebden:
Do we actually need the Python plugin here? It seems like it's just fulfilling a bunch of tests which the plugin itself would also have.

Is it just duplicating the test coverage here to keep an eye on plz?

Tatskaari:
This has been done

PR #2521, opened 2022-08-01
Add plz exec parallel / sequential

peterebden:
After some discussion the other evening it turned out we had some uses for these.

There is still some work to be done; currently the new commands don't support the output file flags. I'd also like to give them the same output options as `plz run parallel` etc but want to leave some things for a future PR.

### DIFF
--- a/src/exec/exec.go
+++ b/src/exec/exec.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/sync/errgroup"
+
 	"github.com/thought-machine/please/src/cli/logging"
 	"github.com/thought-machine/please/src/core"
 	"github.com/thought-machine/please/src/process"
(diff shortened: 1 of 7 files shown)


Garbett1 (approved):
LGTM after the quick discussion.

src/exec/exec.go
```diff
@@ -18,25 +20,65 @@ var log = logging.Log
...
-func exec(state *core.BuildState, target *core.BuildTarget, runtimeDir string, env []string, overrideCmdArgs []string, foreground bool, sandbox process.SandboxConfig) error {
+func exec(state *core.BuildState, target *core.BuildTarget, runtimeDir string, env []string, overrideCmdArgs, additionalArgs []string, entrypoint string, foreground bool, sandbox process.SandboxConfig) error {
+	if err := exec2(state, target, runtimeDir, env, overrideCmdArgs, additionalArgs, entrypoint, foreground, sandbox); err != nil {
+		log.Error("Failed to execute %s: %s", target, err)
+		return err
+	}
+	return nil
+}
+
+func exec2(state *core.BuildState, target *core.BuildTarget, runtimeDir string, env []string, overrideCmdArgs, additionalArgs []string, entrypoint string, foreground bool, sandbox process.SandboxConfig) error {
```
tobyfielding1, line 81: could this have a better name?

PR #2523, opened 2022-08-03
Activate target in build_rule() rather than after parsing the whole package

### LINKED ISSUE #2516: `plz build` hangs when subincluding a build def from the package that defines and subincludes the build def
MWE:

```
# pkg/constant.build_defs:
CONSTANT = "constant"

# pkg/BUILD:
export_file(
    name = "constant",
    src = "constant.build_defs",
    visibility = ["PUBLIC"],
)

subinclude(":constant")

genrule(
    name = "output",
    outs = ["out"],
    cmd = "echo $CONSTANT > $OUTS",
    env = {
        "CONSTANT": CONSTANT,
    },
)

# pkg2/BUILD:
subinclude("//pkg:constant")

genrule(
    name = "output",
    outs = ["out"],
    cmd = "echo $CONSTANT > $OUTS",
    env = {
        "CONSTANT": CONSTANT,
    },
)
```

`//pkg:output` builds fine:

```
$ plz build //pkg:output
Build finished; total time 50ms, incrementality 100.0%. Outputs:
//pkg:output:
  plz-out/gen/pkg/out
$ cat $(plz query output //pkg:output)
constant
```

but building `//pkg2:output` hangs indefinitely. Looking at the output of `-v debug`, it hangs after the cycle detection has finished:

```
13:43:25.904    INFO: Total cache size: 11 GB
13:43:30.369   DEBUG: Running cycle detection...
13:43:30.369   DEBUG: Cycle detection complete, no cycles found
```

My best guess at what's happening here is that subincluding `//pkg:constant` from `//pkg2` causes Please to load `pkg/BUILD`, which then has no idea what to do when it encounters the same subinclude that it's already been asked to load. In retrospect this is a silly thing to do, but I was surprised that the latter caused Please to hang indefinitely and that the former worked at all - in fact it was the former working that gave me hope that the latter might also work...

Tatskaari:
Fixes #2516


src/core/state.go
```diff
@@ -917,6 +917,77 @@ func (state *BuildState) waitForTargetAndEnsureDownload(l, dependent BuildLabel,
...
+			if err := state.QueueTarget(l, dependent, forSubinclude || dependent.IsAllTargets()); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+
+// exportFile adds a single-file export target. This is primarily used for Bazel compat.
```
peterebden, line 972: Why do we need a special function here just for Bazel compat? Seems like the wrong place.

Tatskaari: Yeah, I agree but I don't have any better ideas at this time. Essentially we have a `load("//pkg:go.build_defs")` and load has to figure out if that's a real target or a file. Maybe we could queue the package up for parse, and if the target doesn't exist, we try adding this filegroup? Might make more sense in a followup PR though.

peterebden:
Long time since I reviewed this, but I don't think I had a lot of other complaints. Should we be going ahead if it fixes the issue? 
(I got asked about the same thing, or at least the same symptom, at work this week, so that's another case in the wild...)

PR #2525, opened 2022-08-06
`java_library`: output JNI headers generated by `javac`

chrisnovakovic:
The Java compiler outputs header files for Java sources that use the Java Native Interface (JNI), but there's currently no way for them to be consumed by the native-code components that depend on them, preventing them from being built.

Provide the JNI headers as outputs of `java_library` when `cc_hdrs` is required. This allows native-code JNI components written in C/C++ to be built by including the `java_library` target as a dependency.


rules/java_rules.build_defs
```diff
@@ -45,7 +48,7 @@ def java_library(name:str, srcs:list=None, src_dir:str=None, resources:list=[],
         else:
             javac_flags = CONFIG.JAVAC_TEST_FLAGS if test_only else CONFIG.JAVAC_FLAGS
 
-        javac_flags = '-g -encoding utf8 ' + javac_flags
+        javac_flags = '-h _jni -g -encoding utf8 ' + javac_flags
```
chrisnovakovic, line 51: There doesn't seem to be any significant overhead in invoking the compiler with `-h` even when the JNI headers aren't needed (at least not compared with the time it takes to generate the class files).

rules/java_rules.build_defs
```diff
@@ -84,6 +87,30 @@ def java_library(name:str, srcs:list=None, src_dir:str=None, resources:list=[],
...
+
+        jni_hdrs = build_rule(
+            name = name,
+            tag = "jni_hdrs",
+            srcs = {
+                "WORKER": srcs or [src_dir],
+            },
+            cmd = " && ".join([
+                javac_cmd,
+                "find _jni -type f -name '*.h' | sort",
```
chrisnovakovic, line 103: This should be safe because `javac` only ever outputs on stderr by default, so it shouldn't mess up the post-build output collection.

peterebden:
Thanks! Would you mind please opening this against https://github.com/please-build/java-rules instead? Those plugin repos are where future development of language-specific rules will happen.

I have a couple of comments too but let's pick the conversation up there :)

chrisnovakovic:
> Thanks! Would you mind please opening this against https://github.com/please-build/java-rules instead? Those plugin repos are where future development of language-specific rules will happen.
> 
> I have a couple of comments too but let's pick the conversation up there :)

Will do - I noticed a couple of CI failures in here that I didn't see over there, so I'll fix those, then force-push equivalent changes to https://github.com/please-build/java-rules/pull/13.

PR #2526, opened 2022-08-08
Update macos executors

peterebden:
The current version has gone away so it's broken on master now.

Also removes a bunch of Python tests which seem to be failing on the newer ones (I think because pip decides it needs to build grpc from scratch which takes too long). These have historically been troublesome on newer platforms (e.g. Ubuntu Jammy) and this kind of testing should now happen in the `python-rules` repo.


peterebden:
Cirrus failing because of Go 1.19 so will ignore for now (this change is also blocking that one...)

PR #2527, opened 2022-08-08
Queue data of targets before downloading outputs during `plz run`

peterebden:
Fixes a case where running a test with `plz run` can fail complaining about `could not find action digest for label`.
This is fairly obscure; the case we found was a test with data & a require/provide setup on that data (which is fulfilled differently for data than for a normal dependency).


Tatskaari:
I'm not sure I 100% follow why it was wrong before, but it makes sense to queue up the data ASAP rather than waiting.

Tatskaari (approved):

PR #2533, opened 2022-08-11
Add debug flag that implies plain output for breakpoint()

Tatskaari:
Currently we drop into the interactive debugger which fights with the interactive output and leaves the terminal in a seemingly irrecoverable state. I think gating breakpoints behind this flag is a nice feature anyway, but it can also imply plain output to avoid this issue.

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -213,6 +213,9 @@ type BuildState struct {
 	CurrentSubrepo string
 	// ParentState is the state of the repo containing this subrepo. Nil if this is the host repo.
 	ParentState *BuildState
+	// EnableBreakpoints enablese the breakpoint() build-in, and drops Please into an interactive debugger when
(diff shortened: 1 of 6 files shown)


src/core/state.go
```diff
@@ -213,6 +213,9 @@ type BuildState struct {
 	CurrentSubrepo string
 	// ParentState is the state of the repo containing this subrepo. Nil if this is the host repo.
 	ParentState *BuildState
+	// EnableBreakpoints enablese the breakpoint() build-in, and drops Please into an interactive debugger when
```
samwestmoreland, line 216: nit: enables* and builtin*

samwestmoreland (approved):
Lgtm

peterebden (approved):
LGTM but agree with Sam's comment

PR #2536, opened 2022-08-12
Output options for `plz exec parallel`

peterebden:
This commonises the output modes from `run` and applies them to `plz exec sequential` and `plz exec parallel`.


Tatskaari (approved):
Ah lovely! I'm sure this is going to make the testing team's life a lot easier :D

PR #2538, opened 2022-08-15
Implement updates from subprocesses using `plz exec parallel`

peterebden:
Produces periodic log messages showing the state of subprocesses, which can otherwise be a bit opaque for `quiet` / `group_immediate` output.

Bonus: `errgroup` can now rate-limit itself so get rid of our channel \o/


samwestmoreland (approved):
Nice!

PR #254, opened 2017-12-09
Add pyenv-specific addendum to getting started documentation

natpen:
See https://github.com/thought-machine/please/issues/253 for background context!

PR #2544, opened 2022-08-18
Fix get_pls.sh symlink creation

andremarianiello:
Fixes #254


tools/misc/get_plz.sh
```diff
@@ -38,7 +38,7 @@ echo "Downloading Please ${VERSION}..."
 curl -fsSL "${PLEASE_URL}" | tar -xzpf- --strip-components=1 -C "$DIR"
 # Link it all back up a dir
 for x in "${DIR}/"*; do
```
Tatskaari, line 40: Just trying to figure out why that PR broke this. Am I right in thinking that `ls $DIR/*` gives "please", whereas "$DIR/*" gives "$DIR/foo"?

Tatskaari (approved):

Tatskaari:
Thanks!

PR #2545, opened 2022-08-21
Allow limiting the number of parallel remote_file downloads

peterebden:
Have a case where I have quite a lot of `remote_file`s, but just because I have 24 cores available doesn't mean I necessarily want 24 parallel HTTP requests trying to happen.


Tatskaari (approved):
I think this makes sense. Did a tiny bit of googling to see what the common wisdom here is. Seems Chrome/Safari etc. etc. have all settled on 6 so i think something like 4 might make sense where we usually download larger files.

PR #2546, opened 2022-08-24
Rewrite `pleasew` in POSIX shell

scheatkode:
As promised, here's the working rewrite of the main `pleasew` script using a bit more defensive approach to string interpolation considering there are more footguns when using POSIX shell than say bash or zsh.

My understanding from skimming through the code is that [this build step](https://github.com/thought-machine/please/blob/ab7c8ce1e373ae94b41f47a74fc7b432c9b5a622/BUILD#L28-L33) pulls the main script before embedding it [here](https://github.com/thought-machine/please/blob/ab7c8ce1e373ae94b41f47a74fc7b432c9b5a622/src/plzinit/init.go#L117). Please do correct me if I'm wrong.

This is but an initial rewrite and there are more enhancements to be done. It would be my pleasure to take care of it in a following PR.

### DIFF
--- a/pleasew
+++ b/pleasew
@@ -24,14 +24,14 @@ else
 fi
 
 get_profile () {
-	 while [ "${#}" -gt 0 ]
-	 do
-		  case "${1}" in
-			   --profile=*) echo "${1#*=}"; return;;
-			   --profile) echo "${2}"; return;;
-			   *) shift;;
-		  esac
-	 done
+    while [ "${#}" -gt 0 ]
+    do
+        case "${1}" in
+            --profile=*) echo "${1#*=}"; return;;
+            --profile) echo "${2}"; return;;
+            *) shift;;
+        esac
+    done
 }
 
 # Check `PLZ_CONFIG_PROFILE` or fall back to arguments for a profile.
@@ -49,13 +49,13 @@ EOS
 )"
 
 read_config() {
-	 # Disable globbing to ensure word-splitting is safe.
-	 set -f
+    # Disable globbing to ensure word-splitting is safe.
+    set -f
 
-	 old_ifs="${IFS}"
-	 search_term="${1}"
+    old_ifs="${IFS}"
+    search_term="${1}"
 
-	 IFS='
+    IFS='
 '
 
     # This is intended, we *do* want word-splitting here.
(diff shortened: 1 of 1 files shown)


pleasew
```diff
@@ -1,99 +1,122 @@
...
+	${PROFILE:+.plzconfig.${PROFILE}}
+	.plzconfig_${OS}_${ARCH}
+	.plzconfig
+	${HOME}/.config/please/plzconfig
+	/etc/please/plzconfig
+EOS
+)"
+
+read_config() {
+    printf "%s" "${CONFIGS}" | xargs grep -i "${1}" 2>/dev/null | head -n 1
```
Tatskaari, line 45: This is going to grep over the file name, not the contents of the file unfortunately. We might just want a go old fashioned bash loop here.

pleasew
```diff
@@ -1,99 +1,122 @@
...
 fi
 
-# Check PLZ_CONFIG_PROFILE or fall back to arguments for a profile.
-PROFILE="${PLZ_CONFIG_PROFILE:-$(sed -E 's/.*--profile[= ]([^ ]+).*/\1/g' <<< "$*")}"
+# Check `PLZ_CONFIG_PROFILE` or fall back to arguments for a profile.
+PROFILE="${PLZ_CONFIG_PROFILE:-$(
+	sed -E 's/.*--profile[= ]([^ ]+).*/\1/g' <<-EOS
+		"${*}"
+	EOS
+)}"
```
Tatskaari, line 31: From my testing, this seems to add a trailing quote to the profile string. Not sure why though.

Tatskaari (commented):
Thanks a lot; much appreciated! I think I've found a couple issues though.

pleasew
scheatkode: Notice the use of `xargs` here. But I see your point and I may have a better way of doing this.

pleasew
scheatkode: Indeed, I just made a few tests and the use of `sed` here may not be the best choice. I realized it's not just a quoting problem since `sed` returns the string as-is when it doesn't match, for instance `./pleasew query alltargets` will give a profile of `"query alltargets"`.

scheatkode:
Thanks for the review, I wasn't sure about those either. Modifications coming up !

Tatskaari (approved):
Looks good to me :)

peterebden (approved):
Thanks!

PR #2547, opened 2022-08-30
Set sandbox true on text_file

samwestmoreland:
Think this might be a bug...


Tatskaari (approved):
Oh yeah I have no idea what this is about...

PR #2550, opened 2022-09-02
Reduce log level of a couple of warnings

peterebden:
These are a bit unnecessarily verbose; a 'large' set of stuff in experimental can result in tens of these, and you never really do anything about them anyway.

An alternative could have been to use a `sync.Once` to calm them down but I'm not sure it's really worth even emitting these once.


samwestmoreland (approved):
Agreed it was too verbose

PR #2553, opened 2022-09-14
Strip the subrepo when it's a plugin defined in the host

### LINKED ISSUE #2465: Can't resolve plugin name to subrepo when the plugin is the host repo
For example, we're trying to subinclude `///go//build_defs:go` here: https://github.com/please-build/go-rules/pull/27

This needs to work.

Tatskaari:
The test here contains a good example of this. We have a plugin "foo" and we're building some targets in it. We want to be able to refer to ourselves using the plugin name in the subrepo part of the build label, e.g. `///foo//test:test` should work just like `//test:test`. This doesn't currently work, because we can't resolve this to a subrepo at parse time. It's also harder to resolve this there because we can't modify the label at that point as it's used as a key to synchronise building and parsing dependencies. 

This problem is similar to how you should be able to refer to `//test:test` as `///linux_amd64//test:test`, on a `linux_amd64` host. To make this work, we strip out the subrepo part of a build label if it matches the host repo when parsing. This PR does something similar, but to avoid leaking a bunch of state from the current parse scope into the parse function, I've opted to move the function onto the `scope` struct. 

If we think this is a nicer approach, then I suggest we move all the `ParseBuildLabelContext` code into that scope leaving just the raw parsing in core. Parsing would then become a 2 step process: 1) in core, we parse out the various parts of the target, and 2) in the interpreter, we then contextualise that label to the current package being parsed. 

Fixes #2465


src/parse/asp/interpreter.go
```diff
@@ -3,6 +3,7 @@ package asp
 import (
 	"context"
 	"fmt"
+	"github.com/thought-machine/please/src/cli"
```
peterebden, line 6: appease the linter plz

peterebden (commented):
I think this makes sense to me. Mildly unsure though since the definition of the plugin name happens in the parent repo, so it seems weird for the Go plugin to try to refer to itself when technically it doesn't know what it'll be called.

Tatskaari:
> it seems weird for the Go plugin to try to refer to itself when technically it doesn't know what it'll be called.

The idea was that we ignore what the host repo named the subrepo, and just override it with the plugin ID. This is done when we check the subrepo of the label in `BuildLabel.SubrepoLabel()`. This means it does always know what it will be called because it defines its plugin ID.

src/core/build_label.go
```diff
@@ -137,67 +136,27 @@ func ParseBuildLabel(target, currentPath string) BuildLabel {
 	return label
 }
 
-func splitAnnotation(target string) (string, string) {
+// SplitLabelAnnotation splits the build label from the annotation
+func SplitLabelAnnotation(target string) (string, string) {
```
Tatskaari, line 140: Still not entirely happy with this. Ideally consumers of this package wouldn't have to worry about annotations. Everything depends on the concrete `BuildLabel` type so we can't add indirection without a huge refactor, and adding the annotation to that struct increases the memory footprint quite a bit because we construct potentially 100k or even millions of these during a build.

src/parse/asp/label_context_test.go
```diff
@@ -0,0 +1,115 @@
...
+	arch := cli.HostArch()
+
+	testCases := []struct {
+		testName           string
+		label              string
+		scope              *scope
+		subrepo, pkg, name string
+	}{
+		{
+			testName: "Test parse absolut label with subrepo using @",
```
samwestmoreland, line 34: nit: absolute

samwestmoreland (approved):
Lgtm!

PR #2554, opened 2022-09-15
fix(core/package): Fix dependency resolution before showing warning

scheatkode:
When two rules output to the same directory, the condition used to check for the parent of the rule only, this adds a check for the rule itself before showing the warning.

A better version of this would be to check whether there is a conflict of files between the two rules but I'm not sufficiently accustomed with the codebase yet to attempt that endeavor. For now, this PR opens an escape hatch for those who know what they're doing to at least silence the warning.

### DIFF
--- a/src/core/package.go
+++ b/src/core/package.go
@@ -213,7 +213,7 @@ func (pkg *Package) verifyOutputs() []string {
 	ret := []string{}
 	for filename, target := range pkg.Outputs {
 		for dir := path.Dir(filename); dir != "."; dir = path.Dir(dir) {
-			if target2, present := pkg.Outputs[dir]; present && target2 != target && !target.HasDependency(target2.Label.Parent()) {
+			if target2, present := pkg.Outputs[dir]; present && target2 != target && !(target.HasDependency(target2.Label.Parent()) || target.HasDependency(target2.Label)) {
 				ret = append(ret, fmt.Sprintf("Target %s outputs files into the directory %s, which is separately output by %s. This can cause errors based on build order - you should add a dependency.", target.Label, dir, target2.Label))
 			}
 		}


Tatskaari:
> A better version of this would be to check whether there is a conflict of files between the two rules

I think that's already checked here?
```
target2, present := pkg.Outputs[dir]
```

It's saying that if `target2` outputs a file into a parent directory of one of our outputs, we have a conflict. 

I think the real bug is that `target.HasDependency(target2.Label)` doesn't check transitive dependencies, but maybe that's too slow.

Tatskaari (approved):

scheatkode:
> I think that's already checked here?

I should have worded my statement better here since what I meant was recursively checking for the conflicting outputs, but you're right, maybe I was thinking of this the wrong way because ...

> We don't know what files it outputs until build time.

... and ...

> maybe that's too slow.

What threw me off to begin with was a simple case of reusing a rule's `outs` as another's `srcs` to build a single file in the same tree, both of which are depended on by another "main" rule. Like so:

```python
filegroup(
   # ...
   srcs = [ "dir/" ], # `dir/some_file_altered` doesn't exist yet.
   # ...
)

do_something(
   # ...
   deps = [ ":filegroup" ],
   srcs = [ "dir/some_file" ],
   outs = [ "dir/some_file_altered" ],
   # ...
)

main_rule(
   # ...
   deps = [
      ":filegroup",
      ":do_something",
   ],
   # ...
)
```

What would be a more idiomatic and pleasing way of doing so ?

---

Back to the issue at hand, I'm starting to think a better escape hatch would be to add a _special_ tag similar to `codegen`, maybe `noconflict`. For the time being, this could be a better way to tell Please that the user has manually checked for file conflicts instead of making the check more lax.

Just thinking out loud here, @Tatskaari what's your point of view on this ?

PR #2566, opened 2022-10-19
Use IsAbs for absolute checks which is windows aware

Tatskaari:


samwestmoreland (commented):
Nice, there are some instances of `strings.HasPrefix(foo, '/')` around as well. Some of those might need changing. Like I think the thing that checks for system-level config files does a `HasPrefix`.

samwestmoreland (approved):
Lgtm! 😄

PR #2568, opened 2022-10-24
Error on empty glob

samwestmoreland:
Reviving this! Made a new PR because the old one was a bit of a pain to rebase.

Have tested this in our internal repo, put it behind a feature flag, and added an opt-out.

### DIFF
--- a/src/parse/asp/builtins.go
+++ b/src/parse/asp/builtins.go
@@ -549,7 +549,7 @@ func glob(s *scope, args []pyObject) pyObject {
 
 	glob := s.globber.Glob(s.pkg.SourceRoot(), include, exclude, hidden, includeSymlinks)
 	if s.state.Config.FeatureFlags.ErrorOnEmptyGlob && !allowEmpty && len(glob) == 0 {
-		// Strip build file name from exclude list
+		// Strip build file name from exclude list for error message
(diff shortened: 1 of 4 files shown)


src/parse/asp/builtins.go
```diff
@@ -541,11 +541,18 @@ func glob(s *scope, args []pyObject) pyObject {
...
+	allowEmpty := args[4].IsTruthy()
 	exclude = append(exclude, s.state.Config.Parse.BuildFileName...)
 	if s.globber == nil {
 		s.globber = fs.NewGlobber(s.state.Config.Parse.BuildFileName)
 	}
-	return fromStringList(s.globber.Glob(s.pkg.SourceRoot(), include, exclude, hidden, includeSymlinks))
+
+	glob := s.globber.Glob(s.pkg.SourceRoot(), include, exclude, hidden, includeSymlinks)
+	if s.state.Config.FeatureFlags.ErrorOnEmptyGlob && !allowEmpty && len(glob) == 0 {
+		log.Fatalf("glob(%s) in %s returned no files", include, s.pkg.Filename)
```
goddenrich, line 552: the exclude might be the thing which causes the glob to be empty... do we want to log that to make that more obvious to the user?

goddenrich (approved):

PR #2571, opened 2022-10-28
improve the behaviour of update gitignore

goddenrich:
We now check if the file is ignored first using `git check-ignored` this matches based on globbing so allows us to add ignores to the root .gitignore file. If the generated file doesn't match already we generate a gitignore file in the targets pkg and add the files there rather than finding the closest gitignore file.

### DIFF
--- a/src/scm/scm.go
+++ b/src/scm/scm.go
@@ -25,8 +25,8 @@ type SCM interface {
 	ChangedFiles(fromCommit string, includeUntracked bool, relativeTo string) []string
 	// IgnoreFiles marks a file to be ignored by the SCM.
 	IgnoreFiles(gitignore string, files []string) error
-	// FindOrCreateIgnoreFile gets the ignore file name for the version control system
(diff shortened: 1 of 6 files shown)


src/scm/scm.go
```diff
@@ -25,8 +25,8 @@ type SCM interface {
 	ChangedFiles(fromCommit string, includeUntracked bool, relativeTo string) []string
 	// IgnoreFiles marks a file to be ignored by the SCM.
 	IgnoreFiles(gitignore string, files []string) error
-	// IgnoreFileName gets the ignore file name for the version control system
-	FindClosestIgnoreFile(path string) string
+	// FindOrCreateIgnoreFile gets the ignore file name for the version control system
```
samwestmoreland, line 28: This comment probably needs updating

samwestmoreland (approved):
Nice looks good! Just had a couple of minor comments

peterebden (approved):

Issue #2514, opened 2022-07-23
CPU info: Not yet implemented yet on Apple M1 Pro 


PeterBocan:
On my 14" Macbook Pro with Please version 16.21.3 I am getting this slightly annoying (makes `plz build --shell //target` unusable, due to constant printing into the console). 

```
22:10:27.461   ERROR: Error getting CPU info: not implemented yet
22:10:27.960   ERROR: Error getting CPU info: not implemented yet
22:10:28.461   ERROR: Error getting CPU info: not implemented yet
22:10:28.960   ERROR: Error getting CPU info: not implemented yet
22:10:29.460   ERROR: Error getting CPU info: not implemented yet
22:10:29.960   ERROR: Error getting CPU info: not implemented yet
22:10:30.461   ERROR: Error getting CPU info: not implemented yet
```

PeterBocan:
As I did a little investigation in #2515 the `cpu.Times(false)` calls `allCPUTimes()` which, depending on whether the CGO is enabled or not, has two different implementations:

https://github.com/shirou/gopsutil/blob/2f8da0a39487ceddf44cebe53a1b563b0b7173cc/cpu/cpu_darwin_nocgo.go
https://github.com/shirou/gopsutil/blob/2f8da0a39487ceddf44cebe53a1b563b0b7173cc/cpu/cpu_darwin_cgo.go

It seems like something went wrong with the cross-compilation and CGO not being enabled for Apple plaftorm.

alaroldai:
I see that #2515 has been merged, but I'm still seeing this on my M1 Mac Mini with Please 16.23.1:

```
13:56:37.567   ERROR: Error getting CPU info: not implemented yet
```

Are there any plans to address this? It also makes `plz watch` unusable.

namtzigla:
I'm not sure wh but all the binaries for darwin_amd64 downloaded from github main repo, including 16.23.1, print out this error `ERROR: Error getting CPU info: not implemented yet` but if I'm compiling please (v16.231.) on my local macos m1 computer the error disappear.

I'm using a macos 13.0 with xcode  14.1 (14B47b).

namtzigla:
new symptom, version 16.19.0 is the last known version where this does not happen.

peterebden:
I can see it's finding `cpu_darwin_nocgo.go`. I suspect this is because of cross-compilation (we don't have a `darwin_arm64` worker on CircleCI) although not certain of all the logic there - @Tatskaari might know more

[...]

PR #2575, opened 2022-11-07
Fix darwin_arm64 build

Tatskaari:
Fixes: #2514

Seems we broke this when we migrated to the Go plugin. We were passing the overrides to the built-in rules config rather than the plugin config.


samwestmoreland (approved):
Nice!

PR #2585, opened 2022-11-20
feat: Print out the Global Config object from within the BUILD file

PeterBocan:
This change allows printing out the builtin `CONFIG` file from the BUILD files.

### DIFF
--- a/src/parse/asp/objects.go
+++ b/src/parse/asp/objects.go
@@ -824,6 +824,11 @@ func (c *pyConfig) MarshalJSON() ([]byte, error) {
 	if c.overlay == nil {
 		return json.Marshal(c.base.dict)
 	}
+
+	return json.Marshal(c.Config())
+}
+
+func (c *pyConfig) Config() pyDict {
 	merged := make(pyDict, len(c.base.dict)+len(c.overlay))
 	for k, v := range c.base.dict {
 		merged[k] = v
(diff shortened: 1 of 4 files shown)


src/parse/asp/objects.go
```diff
@@ -822,7 +822,7 @@ type pyConfig struct {
 
 func (c *pyConfig) MarshalJSON() ([]byte, error) {
 	if c.overlay == nil {
-		return json.Marshal(c.overlay)
```
PeterBocan, line 825: this was possibly a bug? Looks a bit sus...

src/parse/asp/objects.go
```diff
@@ -835,7 +835,11 @@ func (c *pyConfig) MarshalJSON() ([]byte, error) {
 }
 
 func (c *pyConfig) String() string {
-	return "<global config object>"
+	marshalledConfig, err := c.MarshalJSON()
+	if err != nil {
+		panic(fmt.Sprintf("failed to marshal object: %s", err))
+	}
+	return fmt.Sprintf("GlobalConfig%s", string(marshalledConfig))
 }
```
Tatskaari, line 843: This is quite different to `pyDict` which uses this formatting:
```
func (d pyDict) String() string {
	var b strings.Builder
	b.WriteByte('{')
	started := false
	for _, k := range d.Keys() {
		if started {
			b.WriteString(", ")
		}
		started = true
		b.WriteByte('"')
		b.WriteString(k)
		b.WriteString(`": `)
		b.WriteString(d[k].String())
	}
	b.WriteByte('}')
	return b.String()
}
```

I feel like these two should behave the same.

PeterBocan: Fair point! Thanks, I have addressed that.

src/parse/asp/objects.go
Tatskaari: Yeah almost certainly. nice spot!

src/parse/asp/objects.go
```diff
@@ -822,20 +822,40 @@ type pyConfig struct {
...
+			b.WriteString(", ")
+		}
+		started = true
+		b.WriteByte('"')
+		b.WriteString(key)
+		b.WriteString(`": `)
+		b.WriteString(config[key].String())
+	}
+	b.WriteByte('}')
+	return b.String()
```
Tatskaari, line 858: Can't we just do `c.Config().String()`, using the `pyDict` implementation?

src/parse/asp/objects.go
```diff
@@ -822,20 +822,40 @@ type pyConfig struct {
...
 func (c *pyConfig) MarshalJSON() ([]byte, error) {
 	if c.overlay == nil {
-		return json.Marshal(c.overlay)
+		return json.Marshal(c.base.dict)
 	}
+
+	return json.Marshal(c.Config())
+}
+
+func (c *pyConfig) Config() pyDict {
```
Tatskaari, line 831: (nit: needs a doc comment)

but does this need to be public? Maybe make it private and re-name it `toPyDict()`?

PR #2588, opened 2022-11-21
Fix flake in plz-run test

Tatskaari:
The test was sensitive to the ordering that these targets get built in. We had a sleep to try and ensure the ordering, but this makes it insensitive to the order.

### DIFF
--- a/test/plz_run/BUILD
+++ b/test/plz_run/BUILD
@@ -29,10 +29,12 @@ please_repo_e2e_test(
 
 please_repo_e2e_test(
     name = "std_in_parallel_test",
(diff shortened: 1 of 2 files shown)


samwestmoreland (approved):
Why do we need to copy to wobble.txt? Why not just
```
expect_output_contains = {
    "wibble.txt": "wibble",
    "wibble.txt": "wobble",
},
```
The reason for the flake was the ordering, right?

samwestmoreland:
Ohhh sorry I understand your comment now. Fair enough, sorry!

PR #2592, opened 2022-11-23
Minor: Don't watch subincludes in subrepos

peterebden:
Saw some error messages from `plz watch` about files that didn't exist in the local repo. Looks like they're ultimately from subincludes in subrepos, which we shouldn't watch since they're generated.


Tatskaari (approved):
Ah nice one!

PR #2593, opened 2022-11-24
Add a warning for output dir conflicts

Tatskaari:
If we have `t1` that is depended on by `t2`, we:
1) Build `t1` and write its rule hash to the outputs
2) Build `t2`, writing its rule hash

When we next run a build, we:
1) Rebuild `t1` because an output has `t2`s rule hash
2) Write `t1`s hash to the output
3) Rebuild `t2` because it now has the wrong rule hash

This PR checks for this case, and logs a warning.


peterebden (approved):
Looks good but the linter is cross with you (and correct about it)

goddenrich (approved):

PR #2596, opened 2022-11-26
Replaces some reflective function calls with generics

peterebden:
Sadly can't remove reflection entirely since we still need to be able to walk the set of struct fields.

Moved the code for making multiple callbacks in one pass into the one (experimental) place that uses it.


Tatskaari (approved):
Nice!

PR #2597, opened 2022-11-26
Update blake3 implementation

peterebden:
Added a benchmark and did some comparisons; it's remarkably noisy but this implementation is consistently 25-50% faster than the previous one we had. Answers returned are the same across a number of random inputs.

We should probably consider pushing blake3 harder (maybe using internally more too), it is considerably faster than sha256.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,6 +47,9 @@ require (
 	google.golang.org/protobuf v1.27.1
 	gopkg.in/op/go-logging.v1 v1.0.0-20160211212156-b2cb9fa56473
 	lukechampine.com/blake3 v1.1.5
+)
+
+require (
 	cloud.google.com/go v0.65.0 // indirect
 	github.com/StackExchange/wmi v1.2.1 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
@@ -61,6 +64,7 @@ require (
 	github.com/juju/ansiterm v0.0.0-20210706145210-9283cdf370b5 // indirect
 	github.com/klauspost/compress v1.13.6 // indirect
 	github.com/klauspost/cpuid v1.3.1 // indirect
+	github.com/klauspost/cpuid/v2 v2.0.12 // indirect
 	github.com/kr/pretty v0.2.1 // indirect
 	github.com/lunixbochs/vtclean v1.0.0 // indirect
 	github.com/mattn/go-colorable v0.1.8 // indirect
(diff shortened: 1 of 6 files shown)


peterebden:
I don't really understand what's going on with the linter errors. The Go compiler seems fine, the file hasn't been changed recently (those bits since last year sometime), it seems to just be wrong about it.

peterebden:
> perhaps we should consider using it for incrementality too?

I'm not sure I understand - if `build.hashfunction = blake3`, we will use this for incrementality / all internal hashing? am I misunderstanding?

Tatskaari:
> > perhaps we should consider using it for incrementality too?
> 
> I'm not sure I understand - if `build.hashfunction = blake3`, we will use this for incrementality / all internal hashing? am I misunderstanding?

we use sha1 for building, e.g.:
```golang
// Calculate the hash of all sources of this rule
func sourceHash(state *core.BuildState, target *core.BuildTarget) ([]byte, error) {
	h := sha1.New()
	...
}
```

The hash function is used for remote execution and for checking the `hashes = []` field on `build_rule()`. We don't use it for incrementatlity though.  

For remote execution, the hash function must be `sha256` or `sha1`. We'd need to support blake3 in the API as REAPI only supports the following:
```
const (
	// It is an error for the server to return this value.
	DigestFunction_UNKNOWN DigestFunction_Value = 0
	// The SHA-256 digest function.
	DigestFunction_SHA256 DigestFunction_Value = 1
	// The SHA-1 digest function.
	DigestFunction_SHA1 DigestFunction_Value = 2
	// The MD5 digest function.
	DigestFunction_MD5 DigestFunction_Value = 3
	// The Microsoft "VSO-Hash" paged SHA256 digest function.
	// See https://github.com/microsoft/BuildXL/blob/master/Documentation/Specs/PagedHash.md .
	DigestFunction_VSO DigestFunction_Value = 4
	// The SHA-384 digest function.
	DigestFunction_SHA384 DigestFunction_Value = 5
	// The SHA-512 digest function.
	DigestFunction_SHA512 DigestFunction_Value = 6
	// Murmur3 128-bit digest function, x64 variant. Note that this is not a
	// cryptographic hash function and its collision properties are not strongly guaranteed.
	// See https://github.com/aappleby/smhasher/wiki/MurmurHash3 .
	DigestFunction_MURMUR3 DigestFunction_Value = 7
)
```

peterebden:
OK I see. We do still use the configured one for hashing files, going down a couple of lines more:
```
func sourceHash(state *core.BuildState, target *core.BuildTarget) ([]byte, error) {
	h := sha1.New()
	for source := range core.IterSources(state, state.Graph, target, false) {
		result, err := state.PathHasher.Hash(source.Src, false, true, false)
```
that `PathHasher` comes from
```
state.PathHasher = state.Hasher(config.Build.HashFunction)
```
so we combine the hashes of multiple files using sha1, but the files individually are hashed using the configured function. I suspect the benefit here is a lot less because it's not much data, but might still be worth changing  it.

I would also at some point like to revisit that and hash multiple files in parallel - probably the PathHasher should have an API for that.

PR #2598, opened 2022-11-28
Fix race condition in subrepo parse

Tatskaari:
We were seeing errors like this occasionally:
```
15:55:44.161   ERROR: ///cc/abseil-cpp/abseil-cpp//absl/strings:str_format failed:
///cc/abseil-cpp/abseil-cpp//absl/strings:str_format depends on ///cc/grpc/grpc//cc/abseil-cpp:all, but the directory plz-out/subrepos/cc/grpc/grpc/cc/abseil-cpp doesn't exist: cc/abseil-cpp
```

The issue was that another thread could parse the package between when we check if the subrepo already exists, and when we try and parse the subrepo package. 

1) We check for the subrepo already existing [here](https://github.com/thought-machine/please/blob/master/src/parse/parse_step.go#L89). 
2) We then attempt to parse the subrepo package [here](https://github.com/thought-machine/please/blob/master/src/parse/parse_step.go#L100). At this point, it might've been parsed by another thread. 
3) We find the package in the graph, so we return `false` to indicate that we didn't handle the package parse. The subrepo does exist at this point as the other thread has finished parsing. 
4) We then look for the subrepo relative to the current subrepo, which results in the above error because it doesn't exist there.

This PR checks for the subrepo if the package exists, which resolves this race condition.

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -790,7 +790,8 @@ func (state *BuildState) SyncParsePackage(label BuildLabel) *Package {
 	return state.Graph.PackageByLabel(label) // Important to check again; it's possible to race against this whole lot.
 }
 
-// WaitForPackage is similar to WaitForBuiltTarget however
+// WaitForPackage is similar to WaitForBuiltTarget however it waits for the package to be parsed, queuing it for parse
+// if necessary
 func (state *BuildState) WaitForPackage(l, dependent BuildLabel) *Package {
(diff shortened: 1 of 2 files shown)


peterebden (approved):
Nice find!

goddenrich (approved):
yea thats much clearer LGTM

PR #2605, opened 2022-12-02
Allow subrepos to output into a package root to avoid collisions

Tatskaari:
This seems to work nicely and allows us to depend on targets from multiple subrepos without worrying about them clobbering each other in the build directory!

### DIFF
--- a/src/plz/plz.go
+++ b/src/plz/plz.go
@@ -4,7 +4,6 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/peterebden/go-cli-init/v5/flags"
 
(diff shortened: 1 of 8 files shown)


src/plz/plz.go
```diff
@@ -146,6 +147,7 @@ func findOriginalTask(state *core.BuildState, target core.BuildLabel, addToList
 		prefix := ""
 		if target.Subrepo != "" {
 			state.WaitForInitialTargetAndEnsureDownload(target.SubrepoLabel(state, ""), target)
+			time.Sleep(time.Millisecond * 100)
```
samwestmoreland, line 150: Why do we need this? Is it that `WaitForInitialTargetAndEnsureDownload` doesn't block on the subrepo being initialised I guess?

samwestmoreland (approved):

src/plz/plz.go
Tatskaari: Ooops this shouldn't be here. This is because there's another race condition that I will fix in another PR.

PR #2606, opened 2022-12-05
Wildcard matching for includes and excludes

### LINKED ISSUE #2532: Allow wildcards for `--include`
As far as I can work out, at the moment it's not possible to do something like `plz query print //dir/... --include 'metadata*'`, but it might be nice to be able to do that.

This is potentially too narrow a use case and/or a misuse of `labels`...

samwestmoreland:
Works for patterns with an asterisk at the end. I did try with `filepath.Match` and with `path.Match` but they don't seem to like there being a colon in the label, which I think labels will quite often have.

Closes #2532.


Tatskaari (approved):
Oh nice one! This will be nice for some use cases.

PR #2608, opened 2022-12-06
Allow setting no hash verification as an env var

Tatskaari:
This will be useful during escrow. We want to set this off so scripts calling Please also disable hash verification.


goddenrich (approved):

goddenrich:
yep extra " at the end

samwestmoreland (approved):

PR #2611, opened 2022-12-07
Fix some race conditions with loading plugin config and migrate to the cc rules

Tatskaari:
This fixes some race conditions around architecture subrepos that were being hit because of the preloaded subinclude.


src/core/state.go
```diff
@@ -216,6 +216,52 @@ type BuildState struct {
...
+				return
+			}
+			if err = validateSubrepoNameAndPluginConfig(state.Config, state.RepoConfig, subrepo); err != nil {
+				return
+			}
+		}
+
+		go state.Parser.Init(state)
+	})
+	return
```
samwestmoreland, line 259: Do we need to return nil or error here depending on how `Do` went?

src/core/subrepo.go
```diff
@@ -25,20 +24,30 @@ type Subrepo struct {
 	Arch cli.Arch
 	// True if this subrepo was created for a different architecture
 	IsCrossCompile bool
-	// loadConfig is used to control when we load plugin configuration. We need access to the subrepo itself to do this
-	// so it happens at build time.
-	loadConfig            sync.Once
+	//
```
samwestmoreland, line 27: Missing comment here I think!

samwestmoreland (approved):
Nice! Think this is a lot neater. Just a couple of minor comments.

src/core/subrepo.go
Tatskaari: No comment

src/core/state.go
Tatskaari: Ah, this is a go thing. We name the return variable, so when we can set it in the closure. When we return like this, we return that variable from the function. It's a bit weird but useful in this sort of situation.

samwestmoreland: Oh I see, ok that makes sense then. Thanks!

PR #2613, opened 2022-12-09
[Fix] update codelabs web page asset sources

mauriceyap:
In the `claat` library, the path to the bucket for codelabs elements has been updated. This PR fixes our codelabs webpages, which are attempting to use CSS and JS assets from the old bucket (and therefore not working).

### DIFF
--- a/docs/build_defs/codelab_template.html
+++ b/docs/build_defs/codelab_template.html
@@ -7,7 +7,7 @@
     <title>{{.Meta.Title}}</title>
     <link rel="stylesheet" href="//fonts.googleapis.com/css?family=Source+Code+Pro:400|Roboto:400,300,400italic,500,700|Roboto+Mono">
     <link rel="stylesheet" href="//fonts.googleapis.com/icon?family=Material+Icons">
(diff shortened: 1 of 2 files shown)


tiagovtristao:
This upgrade has caused the builds to fail and I don't think its necessary. Looking at the codelabs page, the errors are:

```
GET https://storage.googleapis.com/codelab-elements/codelab-elements.css net::ERR_ABORTED 403
GET https://storage.googleapis.com/codelab-elements/codelab-elements.js net::ERR_ABORTED 403
GET https://storage.googleapis.com/codelab-elements/custom-elements.min.js net::ERR_ABORTED 403
GET https://storage.googleapis.com/codelab-elements/native-shim.js
```

I think the URLs in `https://github.com/thought-machine/please/blob/master/docs/build_defs/codelab_template.html` need to be fixed instead. It seems the `codelab-elements` part of the URL should be now `claat-public` instead, but it's worth double checking.

Tatskaari (approved):
Thanks! I was trying to figure out what went wrong here, but ran out of time.

PR #2617, opened 2022-12-21
Put plz query whatinputs errors on stderr & return exit code

peterebden:
If you do `plz query whatinputs nonsense` you currently get back a zero exit code and `Error: 'nonsense' is not a source to any current target` on stdout, which generally confuses a machine that's calling it (because it _looks_ like the call has been successful, and it'll try to interpret that string as build labels).

This makes it just error out instead. Maybe a nicer version would be to log errors and later exit with a code?


Tatskaari (approved):
LGTM but the tests need addressing

peterebden:
Yeah sorry, I ran the tests in `//src/query` but didn't check the e2e tests. Updated them.

PR #2621, opened 2022-12-23
Fix race condition when using a subrepo in original targets

Tatskaari:


peterebden (approved):
Seems reasonable, although am feeling a little rusty on some of this - but I'll trust you!

Tatskaari:
> Seems reasonable, although am feeling a little rusty on some of this - but I'll trust you!

You fool! We activate the target when parsing the BUILD file in `build_target()`. The call to `build_target()` has to appear before the call to `subrepo()` so these things race against each other. Typically this isn't a problem because we wait for the subrepo target but in this case, we're trying to walk the subrepo to find all the targets.

PR #2624, opened 2022-12-27
Add log file rotation

peterebden:
Have been wanting this for a while; sometimes you see something but you need the _previous_ build's logs to understand what's happened (e.g. if you're asking "why is this target building again", you ideally want all the details in the log of the previous time it built to compare the two, but that's just been stomped all over).

This doesn't do anything clever like gzipping old files and isn't configurable (log setup happens pretty much first thing so we don't have access to any config at this point). I imagine this will be sufficient but we can always do more later.

### DIFF
--- a/src/cli/logging.go
+++ b/src/cli/logging.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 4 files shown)


Tatskaari (commented):
Looks good to me but I wonder how it handles two concurrent Please processes both trying to rotate their log files at the same time? Have you thought about or tested this?

samwestmoreland (approved):

peterebden:
> Looks good to me but I wonder how it handles two concurrent Please processes both trying to rotate their log files at the same time? Have you thought about or tested this?

Good point. I have added a repo lock around the rotation; that works fine (as expected, we can rename the log file another process is writing into OK). The only real downside is that it tends to moan a little:
```
11:37:33.177 WARNING: Looks like process with PID 67274 has already acquired the lock for plz-out/.lock. Waiting for it to finish...
```
That isn't technically a big deal but it happens every time you pipe one plz into another which I think is likely going to be annoying (I'm sure I'll get questions internally even though it's only a warning and won't block anything)

peterebden:
OK, I've made the acquisition of the lock for this purpose quieter.

PR #2626, opened 2023-01-02
Get rid of FOUC from the website

tiagovtristao:
This PR sorts the https://en.wikipedia.org/wiki/Flash_of_unstyled_content issue when loading the website where the system's font is used until the expected font is finally downloaded. Once available it replaces the system's font but this causes a jarring user experience which isn't ideal.

I looked at several solutions and getting the base64 enconding of the fonts onto a new CSS file provides the best solution and user experience whether for fast or slow connections. I've defaulted to only providing the WOFF2 font format since it is already widely supported across the browser market. The new encoding itself adds a couple of ten of kB (80kB -> 110kb) on top of the original fonts but it's negligible for the user.


peterebden (approved):
Thanks Tiago!

PR #2637, opened 2023-01-12
New signer

Tatskaari:


peterebden (approved):
Seems legit, assuming this sigstore module is doing something sensible (but seems likely that it will be given that we're getting KMS to do the signing).

Not super enthusiastic about the dependencies for it but I doubt we'll get a better offer.

src/remote/action.go
```diff
@@ -474,7 +474,7 @@ func (c *Client) verifyActionResult(target *core.BuildTarget, command *pb.Comman
 
 // uploadLocalTarget uploads the outputs of a target that was built locally.
 func (c *Client) uploadLocalTarget(target *core.BuildTarget) error {
```
Tatskaari, line 476: I had to update this stuff because sigstore requires gRPC for the kms API, and that had a breaking change. Damn google and their disregard for Go packaging standards.

src/remote/remote.go
```diff
@@ -272,7 +273,7 @@ func (c *Client) initFetch() error {
 	if c.state.Config.Remote.Secure {
 		dialOpts = append(dialOpts, grpc.WithTransportCredentials(credentials.NewClientTLSFromCert(nil, "")))
 	} else {
-		dialOpts = append(dialOpts, grpc.WithInsecure())
+		dialOpts = append(dialOpts, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
Tatskaari, line 276: A quite migration for a deprecated function

src/remote/remote.go
```diff
@@ -700,11 +701,11 @@ func (c *Client) reallyExecute(tid int, target *core.BuildTarget, command *pb.Co
 		return nil, nil, c.wrapActionErr(fmt.Errorf("Failed to execute %s: %s", target, err), digest)
 	}
 	switch result := resp.Result.(type) {
-	case *longrunning.Operation_Error:
+	case *longrunningpb.Operation_Error:
```
Tatskaari, line 704: This is a deprecation migration too to shut the linter up

Tatskaari:
> Seems legit, assuming this sigstore module is doing something sensible (but seems likely that it will be given that we're getting KMS to do the signing).
> 
> Not super enthusiastic about the dependencies for it but I doubt we'll get a better offer.

Most of the new deps are only for signing via the KMS. This stuff isn't linked into the Please binary so I don't think it's too bad.

src/update/key.pub
```diff
@@ -0,0 +1,4 @@
+-----BEGIN PUBLIC KEY-----
```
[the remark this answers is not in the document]
andrewnicolalde: 👍🏻

src/remote/remote.go
andrewnicolalde: Why do we want to do this [without transport security?](https://pkg.go.dev/google.golang.org/grpc/credentials/insecure#NewCredentials)

src/update/verify.go
```diff
@@ -3,56 +3,60 @@ package update
...
 // verifySignature verifies an OpenPGP detached signature of a file.
 // It returns true if the signature is correct according to our key.
-func verifySignature(signed, signature io.Reader) bool {
-	entities, err := openpgp.ReadArmoredKeyRing(strings.NewReader(pubkey))
+func verifySignature(signed, sig io.Reader) bool {
+	return verifySignatureWithKey(signed, sig, key)
+}
+
+func verifySignatureWithKey(signed, sig io.Reader, key []byte) bool {
+	priv, err := cryptoutils.UnmarshalPEMToPublicKey(key)
```
andrewnicolalde, line 32: Why `priv`? Doesn't this return a public key?

src/update/key.pub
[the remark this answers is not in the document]
andrewnicolalde: Which elliptic algo are we using?

src/remote/remote.go
peterebden: There are some use cases where it appears unencrypted to the application - maybe the most legit is going via a service mesh (e.g. Istio) where the connection will be encrypted for us.
The default is that TLS is on though so a user has to explicitly opt out.

andrewnicolalde (commented):
I've left a few comments, mostly questions. I would like to get a second pair of eyes from someone else in Security before we merge this though, given how much it matters that we get this right.

src/update/key.pub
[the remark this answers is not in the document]
Tatskaari: Elliptic Curve P-256 key using sha256 to digest it

It's the default for gcp's kms so I assumed it's a good one

src/update/verify.go
Tatskaari: Ah, I think I just copied from an example.

andrewnicolalde: We should change this to pub then, since this does seem to return a public key: https://pkg.go.dev/github.com/sigstore/sigstore/pkg/cryptoutils#UnmarshalPEMToPublicKey

Tatskaari: Yup sorry, I hadn't pushed

src/update/verify_test.go
```diff
@@ -2,67 +2,78 @@ package update
...
+
+	bs, err := verifier.SignMessage(bytes.NewReader(msg))
+	if err != nil {
+		panic(err)
+	}
+
+	return bytes.NewReader(bs)
+}
+
 // readFile reads a file into an io.Reader
```
chrisnovakovic, line 48:
```suggestion
// readFile reads a file into a byte slice.
```

This needs updating now that the type signature has changed.

src/update/verify_test.go
```diff
@@ -2,67 +2,78 @@ package update
...
-	assert.False(t, verifySignature(signed, signature))
+	return b
 }
 
-func TestVerifyBadFile(t *testing.T) {
-	signed := readFile("src/update/test_data/bad.txt")
-	signature := readFile("src/update/test_data/test.txt.asc")
-	assert.False(t, verifySignature(signed, signature))
+func fileReader(filename string) io.Reader {
+	return bytes.NewReader(readFile(filename))
```
chrisnovakovic, line 62: Seems a bit circular to slurp an `io.Reader` into a byte slice and then return an `io.Reader` backed by that byte slice - could you just use `os.Open` instead? (It probably doesn't matter though, it doesn't look like we're reading large files anyway.)

tools/release_signer/signer/signer.go
```diff
@@ -32,6 +35,22 @@ func SignFile(filename, output, keyring, user, password string) error {
...
+	if err != nil {
+		return err
+	}
+
+	sig, err := signer.SignMessage(bytes.NewReader(message))
+	if err != nil {
+		return err
+	}
+
+	return os.WriteFile(output, sig, 0666)
```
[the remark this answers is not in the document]
Tatskaari: I suppose it could be read only (0444). I was basing this off the `os.Create()` method used by the pgp signer which sets it to 666

src/update/verify_test.go
Tatskaari: There's a circleci joke in there somewhere

chrisnovakovic (approved):

andrewnicolalde:
> We have cycled all access tokens so no further releases can be made using the compromised access token. 

To clarify, this means that no new objects can be added to the S3 bucket backing https://get.please.build/ from which user download Please. We've verified that no suspicious objects were uploaded to that bucket during the time the compromised access token was still valid.

However, since the signing key was exposed in the CircleCI breach, an attacker could still theoretically create a valid, signed release of Please, but they'd still need to MITM the Please client's network connection out to https://get.please.build/ to actually get clients to download the malicious release.

PR #2644, opened 2023-01-16
Log different build result when downloading data

Tatskaari:
This should help debug some scenarios where we get stuck in "preparing"


peterebden (approved):
The "log build result" version seems legit, although I assume it's visual only, right?

PR #2646, opened 2023-01-17
Add arm to release step

Tatskaari:


.circleci/config.yml
```diff
@@ -344,11 +344,6 @@ workflows:
       - build-darwin:
           requires:
             - build-alpine
-            - build-linux
-            - build-linux-alt
```
Tatskaari, line 348: Not sure why we were waiting for these but the darwin builds take a long time so we should start them earlier if possible. I figure we gate this behind the alpine build to avoid running the more expensive macos build when there are obvious build failures that can be caught in the alpine build.

peterebden: Yeah it was exactly that; don't start running the expensive darwin build (the workers are like 5x more or smthn) if we're just going to replicate a bunch of obvious test failures that we can find in the cheaper ones. Maybe the alpine one is enough.

peterebden (approved):

PR #2650, opened 2023-01-18
Change asp.Position to just be an int

peterebden:
This is pretty heavily inspired by / pinched from [go/token](https://pkg.go.dev/go/token#Pos)'s approach. Rather than storing a fully hunam-readable structured form, we just store byte offsets into the file. This is obviously quite a bit more compact (8 bytes vs. 40). At runtime we don't need info like line or column, only at the point of error, where we can reconstruct it from the file.

I see some reduction in live memory (~1305MB -> ~1290MB) after completing a build on our internal repo which is nice, although presumably only just so much difference because much of the AST is not kept live. TotalAlloc is approx 60MB less.

The only thing that isn't working is the globals stuff in the lsp, which I need to think a little more on (basically how we extract the file containing a builtin given its name). Otherwise error traces etc continue to work.


Tatskaari (approved):
Makes sense to me :)

PR #2653, opened 2023-01-21
Make --pre work with plz query print

peterebden:
Currently it hangs, because it doesn't think it needs to build things, but the pre-target addition code is waiting for it. This fixes by forcing the build as though it were needed for a subinclude.

e.g.
```
$ plz plz query print //test:_add_out_gen --pre //test:_add_out_gen
# //test:_add_out_gen:
build_rule(
    name = '_add_out_gen',
    cmd = 'echo hello > _add_out_gen.txt',
    outs = ['_add_out_gen.txt'],
    output_is_complete = True,
    build_timeout = 600,
    post_build = '<function <lambda>>',
)
```
observe that `outs` is set on the result (it is not without `--pre`)


peterebden:
Needs changes, we can't queue things up in this way.

PR #2654, opened 2023-01-21
Fix race condition in parser

peterebden:
Found this while fiddling with #2653; there's a concurrent map read/write on the top-level scope between the global subincludes (writing) and interpreting another file (reading). While the changes there definitely make this occur more often (for whatever reason) I don't think they are actually causing the problem.

This adds a mutex which fixes the race condition. Interestingly it then nearly always breaks with undefined names (`go_test` in this case but could be anything from a global subinclude) - so this is definitely not the real / full solution. I think it illustrates that something is wrong though.

What I'd ideally like to do is force `Init` to complete before `ParseFile` can begin (essentially removing `WaitForInit` and making that implicit. That's very hard though because the subincludes in `Init` may themselves queue targets which then call back to `ParseFile`...

### DIFF
--- a/src/parse/asp/parser.go
+++ b/src/parse/asp/parser.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/thought-machine/please/src/cli/logging"
 	"github.com/thought-machine/please/src/core"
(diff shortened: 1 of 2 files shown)


Tatskaari:
The idea here was that we can load the globals of the preloaded subincludes by doing the following:

1) We queue the original targets up as not for subinclude. These then just immediately block on waiting for init
2) We then queue up the preloaded targets for subinclude. These skip the wait for init check. We build them all but don't update the global state. 
3) Once all these targets have been built, we begin sub-including them into the root interpreter scope. The ideas is that nothing else should be parsing at this time. 

After step 2, we have completed all the parsing for the sub-include tasks, and the original tasks should all be blocked on waiting for init. This means that we have an opportunity to write into the parser's global scope. Is it possible this isn't true? Are there any other sources of parse tasks that mean we're parsing files during step 3?

peterebden:
Okay, maybe this is the original change then doing things that it doesn't expect. I'll revisit that.

In the meantime, I don't think this is correct anyway - I think it could just wind up with a deadlock (not that I've seen that failure mode so far).

PR #2656, opened 2023-01-23
Handle plugin and host in original task

Tatskaari:
We strip the plugin and host repo when parsing build files, but leave them in when finding the original targets. We also were overriding the subrepo on the command line when passing `--arch`. This PR fixes these issues.

### DIFF
--- a/src/plz/plz_test.go
+++ b/src/plz/plz_test.go
@@ -1,0 +1,71 @@
+package plz
+
+import (
+	"fmt"
+	"github.com/thought-machine/please/src/cli"
+	"testing"
+
+	"github.com/thought-machine/please/src/core"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestStripHost(t *testing.T) {
+	hostArch := cli.HostArch()
+	testData := []struct{
+		Name string
+		Plugin string
+		Label string
+		Expected string
+	} {
+		{
+			Name: "Leaves no subrepo",
+			Label: "//foo:bar",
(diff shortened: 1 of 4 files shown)


src/plz/plz_test.go
```diff
@@ -0,0 +1,70 @@
...
+			Label:    fmt.Sprintf("@%v//foo:bar", core.SubrepoArchName("go", cli.HostArch())),
+			Plugin:   "go",
+			Expected: "//foo:bar",
+		},
+		{
+			Name:     "Strips host but leaves different subrepo",
+			Label:    fmt.Sprintf("@%v//foo:bar", core.SubrepoArchName("python", cli.HostArch())),
+			Plugin:   "go",
+			Expected: "@python//foo:bar",
+		},
```
samwestmoreland, line 59: This might break if we define a subrepo in a plugin that's prefixed with the plugin name e.g. `go_subrepo` in the `go` plugin. I.e. `@go_subrepo//foo:bar` you'd expect to end up with an unchanged label I think? Don't we do something like that in the toolchain?

samwestmoreland (approved):

src/plz/plz_test.go
Tatskaari: Ah good spot! I don't think that works at the moment from either the command line or when defining dependencies in BUILD files. There hasn't been a need thus far, but maybe this is something we should think about. 

Right now, `@go_subrepo//:all` will try and parse the root package to find a subrepo called `go_subrepo`.

PR #2658, opened 2023-01-24
Preload subincludes into the package scope rather than the root scope

Tatskaari:
The previous idea was to try and treat preloaded subincludes like builtins but that broke too many of the assumptions. This approach treats them more like the `subinclude()` buitin, loading them in to the package scope.


src/core/state.go
```diff
@@ -1235,12 +1241,30 @@ func (state *BuildState) ForSubrepo(name string, bazelCompat bool) *BuildState {
...
 
+// GetPreloadedSubincludes gets the preloaded subincludes for this state, de-duplicating if there are duplicates
+func (state *BuildState) GetPreloadedSubincludes() []BuildLabel {
+	done := map[BuildLabel]struct{}{}
+	includes := make([]BuildLabel, 0, len(state.Config.Parse.PreloadSubincludes)+len(state.RepoConfig.Parse.PreloadSubincludes))
+
+	is := append(state.Config.Parse.PreloadSubincludes, state.RepoConfig.Parse.PreloadSubincludes...)
+
+	for _, i := range is {
+		_, ok := done[i]
```
peterebden, line 1257: nit: inline this into the if statement

peterebden (approved):

PR #2664, opened 2023-01-26
Delete java rules

samwestmoreland:


Tatskaari (commented):
Do we not also have Java tests that can be nuked?

PR #2667, opened 2023-01-27
Jarcat -> Arcat

samwestmoreland:
I've left the `Java.Jarcat` config in because we subinclude plugins (`cc` for example) that reference it and obviously blow up if it's not there. Not sure what's best here but I figure one way is to leave it in now, land this, migrate all the plugins to `Build.Arcat`, and then delete it?


docs/config.html
```diff
@@ -763,6 +763,36 @@ <h3 class="mt1 f6 lh-title" id="build.updategitignore">UpdateGitignore</h3>
...
+            rel="noopener"
+            >this</a
+          >
+          or
+          <a
+            class="copy-link"
+            href="http://d104xtrw2rzoau.cloudfront.net/wp-content/uploads/2013/06/15806-cat_stuck_jar_27_8_2012.jpg"
+            target="_blank"
+            rel="noopener"
+            >this</a
```
Tatskaari, line 788: This is a broken link :(

I'm not sure we should really be hot linking to other websites anyway. Maybe remove these links?

docs/config.html
```diff
@@ -763,6 +763,36 @@ <h3 class="mt1 f6 lh-title" id="build.updategitignore">UpdateGitignore</h3>
...
+          Defines the tool used to concatenate files which we use to build
+          the output of various build rules. Defaults to
+          <code class="code">arcat</code> (formerly <code class="code">jarcat</code>).<br />
+          Any resemblance to
+          <a
+            class="copy-link"
+            href="http://4.bp.blogspot.com/_fzolLOGJOn0/TRkl9BFE_4I/AAAAAAAAEGQ/vxs9pwhDCVA/s1600/Cat-Packed.in.Glass.jpg"
+            target="_blank"
+            rel="noopener"
+            >this</a
```
Tatskaari, line 780: this joke doesn't make any sense with the name change haha

Tatskaari (approved):
LGTM :) just need to remove those images from the docs.

PR #2668, opened 2023-01-27
Delete java rules

samwestmoreland:


Tatskaari (approved):
We will probably have to add an `exclude` tag for the config here too.

Issue #2257, opened 2022-01-11
Named outputs in `deps`/`exported_deps` produce syntax error


chrisnovakovic:
```
genrule(
    name = "one",
    srcs = ["one"],
    outs = {
        "txt": ["one.txt"],
        "dat": ["one.dat"],
    },
    cmd = "for i in $OUTS; do cp $SRCS $i; done",
)

genrule(
    name = "two",
    srcs = {
        "onetxt": [":one|txt"],
        "two": ["two"],
    },
    outs = ["two.txt"],
    cmd = "cat $SRCS_ONETXT $SRCS_TWO > $OUTS",
)
```

`plz build :two` succeeds, but adding `deps = [":one|txt"],` or `exported_deps = [":one|txt"],` outputs the following error:

```
Build stopped after 30ms. 1 target failed:
    //[...]:two
misc_rules.build_defs:114:12: error: Invalid build label: :one|txt
        fail('Can\'t specify both "out" and "outs".')
    return build_rule(
           ^
        name = name,

Traceback:
misc_rules.build_defs:114:12:       return build_rule(

           test/BUILD: 11: 1:   genrule(
```

peterebden:
This may actually be deliberate. I'm not sure it makes sense to say you depend on a rule but want specific outputs from it; that seems more appropriate for `srcs` which might want to get a particular set of output files.
Sort of feel like this is mixing up named outputs which are a file-y thing and deps which are a build label-y thing.

chrisnovakovic:
Makes sense - I can work around it by adding the named output as a `src` of a `filegroup` that also collects the outputs of the target that had `exported_deps`, then depending on the `filegroup` instead, which is consistent with what you said about named outputs being about files and `deps` being about build targets. Thanks!

PR #2670, opened 2023-01-28
Named Outputs can only be depended as a source

robxu9:
Per #2257, deps/exported_deps don't accept named outputs as inputs deliberately. Update the documentation to remove references to using named outputs in deps, and clarify that it can be depended on as sources only.


peterebden (approved):
Thanks!

PR #2690, opened 2023-02-08
removing persistent workers, erroring if discovered

goddenrich:
This removes the ability to use persistent workers. Workers are no longer started or stopped. The worker commands are still parsed but if they are encountered the build or test will error.


src/build/build_step.go
```diff
@@ -1215,57 +1211,16 @@ func (r *progressReader) Read(b []byte) (int, error) {
...
 
-// buildMaybeRemotely builds a target, either sending it to a remote worker if needed,
-// or locally if not.
-func buildMaybeRemotely(state *core.BuildState, target *core.BuildTarget, inputHash []byte) (*core.BuildMetadata, error) {
+// build builds a target locally, it errors if a remote worker is needed since this has beeen removed.
+func build(state *core.BuildState, target *core.BuildTarget, inputHash []byte) (*core.BuildMetadata, error) {
 	metadata := new(core.BuildMetadata)
 
-	workerCmd, workerArgs, localCmd, err := core.WorkerCommandAndArgs(state, target)
+	workerCmd, _, localCmd, err := core.WorkerCommandAndArgs(state, target)
```
samwestmoreland, line 1218: Should we delete the code in `WorkerCommandAndArgs` as well? Looks like it's to do with `$(worker)` which I guess we don't need anymore

src/test/test_step.go
```diff
@@ -546,24 +545,15 @@ func moveOutputFile(state *core.BuildState, hash []byte, from, to, dummy string)
 	return fs.RecordAttr(to, hash, xattrName, state.XattrsSupported)
 }
 
-// startTestWorkerIfNeeded starts a worker server if the test needs one.
-func startTestWorkerIfNeeded(tid int, state *core.BuildState, target *core.BuildTarget) error {
-	workerCmd, _, testCmd, err := core.TestWorkerCommand(state, target)
+// verifyWorkerNotNeeded returns an error if a persistent worker is needed.
+func verifyWorkerNotNeeded(tid int, state *core.BuildState, target *core.BuildTarget) error {
+	workerCmd, _, _, err := core.TestWorkerCommand(state, target)
```
samwestmoreland, line 550: This eventually calls the sequence replacement stuff for `$(worker)` as well so maybe we can get rid of this?

samwestmoreland (approved):
Lgtm! I made a comment about the `$(worker)` code but now I'm thinking maybe better to leave that in for now so it's clearer to the user what's changed? Rather than just deleting it completely. That's probably what you were thinking?!

src/build/build_step.go
goddenrich: Yea my aim is to keep this in for now this allows us to give people hints to what is wrong. Without this code in core we aren't able to detect it so easily

PR #2693, opened 2023-02-09
Load plugin config to top level scope of the subinclude too

Tatskaari:


Tatskaari:
This isn't as simple as it first seemed. The config is inherited from the root scope of the subincludes so we end up with plugin config in the scope of the build def function body that we shouldn't have.

PR #2695, opened 2023-02-10
Remove go plugin

Tatskaari:


test/entry_point/test_repo/.plzconfig
```diff
@@ -1,6 +1,11 @@
...
 BuildFileName = BUILD
 preloadsubincludes = ///shell//build_defs:shell
+preloadsubincludes = ///go//build_defs:go
 
 [Plugin "shell"]
 target = //plugins:shell
+
+
+[Plugin "go"]
+target = //plugins:shell
```
samwestmoreland, line 11: Should this be plugins:go?

plugins/BUILD
samwestmoreland, line 1: I guess these don't work with the latest versions of the plugins? Do you know what the issue is with the python one?

samwestmoreland (approved):
Nice!

PR #2713, opened 2023-03-04
Don't print alias info if a command is active

### LINKED ISSUE #2576: Don't display Available commands when viewing help for a specific sub command
If I do `plz run --help` I get the relevant help text and it also displays the "Available commands for this repository:" at the end. The latter is useful if you just do `plz --help`, but is not relevant to the help of a sub-command. It would be better to not display that text in this case.

peterebden:
Fixes #2576


peterebden:
I'm not sure what the linter is moaning about, maybe have to go down that rabbit hole again...

Tatskaari (approved):

PR #2716, opened 2023-03-07
Delete reference to non-existent struct field that was breaking `plz help`

samwestmoreland:
Think it would be nice to implement something here so that the function knows which plugin it was defined in, and then include that in the doc string instead?


peterebden (approved):

peterebden:
Thanks for that!

It would be nice to display the source plugin. Not sure how easy that is at the point we do help though...?

samwestmoreland:
We loop through the plugins and check whether the topic is a rule defined in that plugin when we're doing the help, so at that point we could potentially pass the plugin name into the template thing somehow...

PR #2719, opened 2023-03-08
Fix issues with `pleasew`

### LINKED ISSUE #2717: pleasew detects incompatible architecture name on (my) Intel mac
I've noticed that `pleasew` doesn't seem to correctly infer a please-compatible architecture name on my Intel Mac. The result is that `pleasew` fails to correctly download please.

The problem seems to lie in the way that `ARCH` is inferred on Darwin systems:

```
if [ "${OS}" = 'Darwin' ]; then
    # switch between mac amd64/arm64
    ARCH="$(uname -m)"
else
    # default to amd64 on other operating systems
    # because we only build intel binaries
    ARCH='amd64'
fi
```

On my Intel Mac (2019 MBP running Ventura), the output of `uname -m` is is `x86_64`, which results in `pleasew` using an invalid URL to try to fetch please and getting a 403:

```
...
+ printf '%bDownloading Please %s to %s...%b\n' '\x1B[32m' 16.27.2 .../.please/16.27.2 '\x1B[0m'
Downloading Please 16.27.2 to .../.please/16.27.2...
+ mkdir -p ../.please/16.27.2
+ curl -fsSL https://get.please.build/darwin_x86_64/16.27.2/please_16.27.2.tar.xz
+ tar -xJpf- --strip-components=1 -C .../.please/16.27.2
curl: (22) The requested URL returned error: 403
...
```

Note that this seems to relate to the newer POSIX-rewrite of `pleasew`; the older version only had support for `amd64` and so doesn't suffer the same issue.

This should be a fairly trivial thing to fix, and I'm happy to raise an MR, but not sure if anyone has particular feelings about how they prefer it fixed.

scheatkode:
- Remove redundant top-level set

 `set -o errexit` is equivalent to `set -e`.

- Better eye-candy

  - Add download progress bar.
  - Fix coloring on some terminals.
  - Support `$NOCOLOR`.

- Add .plzconfig indentation support

 `please` (the binary) supports indented config but `pleasew` does not.
 Now it does.

- Normalize `uname -m` output. Fixes #2717.

### DIFF
--- a/pleasew
+++ b/pleasew
@@ -3,12 +3,12 @@
 set -e
 set -u
 
-set -o errexit
+ESC="$(printf '\033')"
 
-RED='\x1B[31m'
-GREEN='\x1B[32m'
-YELLOW='\x1B[33m'
-RESET='\x1B[0m'
+RED="${ESC}[31m"
+GREEN="${ESC}[32m"
+YELLOW="${ESC}[33m"
+RESET="${ESC}[0m"
 
 DEFAULT_URL_BASE='https://get.please.build'
 
@@ -22,6 +22,11 @@ else
     # because we only build intel binaries
     ARCH='amd64'
 fi
+
+case "${ARCH}" in
+	aarch64_be|aarch64|armv8b|armv8l) ARCH='arm64' ;;
+	x86_64) ARCH='amd64' ;;
+esac
 
 get_profile () {
     while [ "${#}" -gt 0 ]
@@ -69,7 +74,7 @@ read_config() {
 }
 
 # We might already have it downloaded...
-LOCATION="$(read_config '^location' | cut -d '=' -f 2 | tr -d ' ')"
+LOCATION="$(read_config '^\s*location' | cut -d '=' -f 2 | tr -d ' ')"
 
 if [ "${LOCATION:+x}" != 'x' ]; then
     if [ "${HOME:+x}" != 'x' ]; then
@@ -92,7 +97,7 @@ if [ -f "${TARGET}" ]; then
     exec "${TARGET}" ${PLZ_ARGS:-} "${@}"
 fi
 
-URL_BASE="$(read_config '^downloadlocation' | cut -d '=' -f 2 | tr -d ' ')"
+URL_BASE="$(read_config '^\s*downloadlocation' | cut -d '=' -f 2 | tr -d ' ')"
 
 if [ "${URL_BASE:+x}" != 'x' ]; then
     URL_BASE="${DEFAULT_URL_BASE}"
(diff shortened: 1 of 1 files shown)


pleasew
```diff
@@ -133,7 +146,7 @@ fi
 
 printf >&2 '%bDownloading Please %s to %s...%b\n' "${GREEN}" "${VERSION}" "${DIR}" "${RESET}"
 mkdir -p "${DIR}"
-curl -fsSL "${PLEASE_URL}" | tar -xJpf- --strip-components=1 -C "${DIR}"
+curl -#fSL "${PLEASE_URL}" | tar -xJpf- --strip-components=1 -C "${DIR}"
```
peterebden, line 149: Do you know what curl will do here if the terminal isn't interactive? i.e. will this force the progress bar to show always or will it still select intelligently?

peterebden (commented):
Thanks for this!

pleasew
scheatkode: AFAIK, it forces the progress bar to show (tried with a non-interactive docker container). Let me know if that causes an issue and I should remove it.

Ultimately the script should use either `curl` or `wget` depending on the availability on the user's system. Should I add that in this PR ?

peterebden: Okay, maybe let's see how it goes. I'm mildly worried about it producing heaps of lines of output in a scenario like that but hopefully it's fine.

I like the idea of trying to detect `curl` or `wget`, don't mind if  you want to do that in this PR or another.

scheatkode: > I like the idea of trying to detect `curl` or `wget`, don't mind if you want to do that in this PR or another.

Done ! Other utilities such as `grep` and `sed` should be present in POSIX systems and don't require checks.

> Okay, maybe let's see how it goes. I'm mildly worried about it producing heaps of lines of output in a scenario like that but hopefully it's fine.

This is only to let the user know that *something* is happening (in my case, slow internet connection). As a compromise, would disabling the progress bars in CI be enough ? Otherwise let me know and I'll remove them entirely.

peterebden: What about just removing the `s`? That should leave it as a default, which is maybe less obvious than a progress bar, but I think it will more or less do "the right thing" in most cases. WDYT?

scheatkode: Yeah that works, it should reduce the amount of output for non-interactive terminals while visually signaling to the user that it's not stuck somewhere. I like it !

`wget` should have similar behavior by default according to its manpage where it changes the progress indicator for non-interactive terminals. Nothing to do handle on our end.

peterebden (approved):
Thanks again!

scheatkode:
Thank you for taking the time to review !

PR #2722, opened 2023-03-10
Exclude targets in plugin subrepos from `plz query changes`

chrisnovakovic:
Changing a `plugin_repo()` target causes all of the plugin subrepo's child targets to be outputted by `plz query changes`. This is problematic because it isn't necessarily desirable (or even possible, depending on the environment) to build every single target in the plugin subrepo.

Exclude all plugin subrepo child targets from the output of `plz query changes` by default, and add an `--include_plugins` option to revert to the previous behaviour if desired.

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -416,6 +416,7 @@ var opts struct {
 		Changes struct {
 			Since            string `short:"s" long:"since" default:"origin/master" description:"Revision to compare against"`
 			IncludeDependees string `long:"include_dependees" default:"none" choice:"none" choice:"direct" choice:"transitive" description:"Deprecated: use level 1 for direct and -1 for transitive. Include direct or transitive dependees of changed targets."`
+			IncludePlugins   bool   `long:"include_plugins" description:"Include changed targets that belong to plugin subrepos."`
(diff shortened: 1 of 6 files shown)


src/please.go
```diff
@@ -416,6 +416,7 @@ var opts struct {
 		Changes struct {
 			Since            string `short:"s" long:"since" default:"origin/master" description:"Revision to compare against"`
 			IncludeDependees string `long:"include_dependees" default:"none" choice:"none" choice:"direct" choice:"transitive" description:"Deprecated: use level 1 for direct and -1 for transitive. Include direct or transitive dependees of changed targets."`
+			IncludePlugins   bool   `long:"include_plugins" description:"Include changed targets that belong to plugin subrepos."`
```
peterebden, line 419: Maybe this should just be all subrepos? I feel like I would probably want the same behaviour most of the time from non-plugin repos as well (e.g. if you pulled in something like absl or googletest as a subrepo, I don't think I want to run all their tests on upgrade)

chrisnovakovic:
There's consensus that the right way to do this is for all subrepos, not just plugin subrepos, which I've implemented in #2723 - thanks everyone for the feedback.

PR #2725, opened 2023-03-13
[v16] Exclude targets in subrepos from `plz query changes`

chrisnovakovic:
Exclude all subrepo child targets from the output of `plz query changes` by default, and add an `--include_subrepos` option to revert to the previous behaviour if desired.


chrisnovakovic:
This is #2723 for the stable branch.

peterebden (approved):

PR #2731, opened 2023-03-15
Don't follow subincludes when querying changes

Tatskaari:
We changed this in an attempt to support subincludes that might affect a target only through pre/post-build actions, however it turns out this approach includes far too many targets and has a significant performance impact.


peterebden:
Agree with the revert, but looks like something needs a small fix.

peterebden (approved):

PR #2733, opened 2023-03-17
Remove please_pex from target list

peterebden:
It isn't there any more in v17. It isn't _used_ but can break builds (e.g. if you're looking for stuff changing with `plz query changes` or whatever).


Tatskaari (approved):

Tatskaari:
Sorry I did this in a different PR

PR #2737, opened 2023-03-21
Remove feature flags that were meant to be removed before

Tatskaari:
We've already removed these features. We just forgot to clean up the flags.

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -252,13 +252,6 @@ func (state *BuildState) Initialise(subrepo *Subrepo) (err error) {
 func (state *BuildState) ExcludedBuiltinRules() map[string]struct{} {
 	// TODO(jpoole): remove this function, including the changes to rules.AllAssets() in v17
 	ret := make(map[string]struct{})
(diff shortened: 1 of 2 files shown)


src/core/state.go
```diff
@@ -252,13 +252,6 @@ func (state *BuildState) Initialise(subrepo *Subrepo) (err error) {
 func (state *BuildState) ExcludedBuiltinRules() map[string]struct{} {
 	// TODO(jpoole): remove this function, including the changes to rules.AllAssets() in v17
 	ret := make(map[string]struct{})
-	if state.Config.FeatureFlags.ExcludeJavaRules {
-		ret["java_rules.build_defs"] = struct{}{}
-	}
-	if state.Config.FeatureFlags.ExcludeCCRules {
-		ret["c_rules.build_defs"] = struct{}{}
-		ret["cc_rules.build_defs"] = struct{}{}
-	}
```
Tatskaari, line 261: These files don't exist on disk

PR #2738, opened 2023-03-21
Remove iter sources ff

Tatskaari:


samwestmoreland (approved):
Nice! Need to delete the ff from `src/core/config.go` as well right?

PR #2739, opened 2023-03-21
Remove the proto rules

Tatskaari:


samwestmoreland (approved):

Tatskaari:
> Looks like docs wants an update to the python plugin

Yeah I've done that in the ff cleanup PR :)

PR #2745, opened 2023-03-23
Add `patches` parameter to subrepo rules

### LINKED ISSUE #2744: Patching code in subrepos derived from third-party repos is cumbersome
`github_repo` and friends make it easy to build code derived from third-party repos when combined with `build_file`. However, one difficulty I've faced is with applying patches (usually sourced from unmerged PRs, or merged PRs from a later commit) to the code before the targets in `build_file` do anything productive.

The way I typically apply patches now is to expose the patches as `export_file`s to the subrepo, apply them to individual groups of unpatched files and collect the patched files into named output groups with a `genrule`, then use the output groups as inputs to rules like `c_library`. This introduces a fair amount of repetitive boilerplate `genrule`s to build files, and means that I can't easily take advantage of globbing source files (e.g. I'd have to patch `kittens.c` in a `patch_kittens` `genrule`, then use `srcs = [":patch_kittens|srcs"] + glob(["*.c"], exclude = ["kittens.c"])` in `c_library` to avoid duplicate symbol errors at link-time). It can all get pretty complex if files need to be patched for multiple targets.

I feel it'd be more user-friendly to allow the necessary patches to be applied after the repo archive is downloaded and extracted, but before the build file is installed. The patched files would then already be visible to the targets in the build file, which would eliminate the need for the `genrule` hijinks.

chrisnovakovic:
Fixes #2744.

### DIFF
--- a/rules/subrepo_rules.build_defs
+++ b/rules/subrepo_rules.build_defs
@@ -24,7 +24,7 @@ def workspace(name:str):
     )
 
 
-def http_archive(name:str, urls:list, strip_prefix:str=None, labels:list=[],
+def http_archive(name:str, urls:list, strip_prefix:str=None, patches:list=[], labels:list=[],
                  hashes:str|list&sha256=None, config:str=None, bazel_compat:bool=False,
                  visibility:list=None):
     """Fetches a remote file over HTTP and expands its contents.
@@ -39,6 +39,8 @@ def http_archive(name:str, urls:list, strip_prefix:str=None, labels:list=[],
       urls: List of URLs to fetch from. These are assumed to be mirrors and will be
             tried in sequence.
       strip_prefix: Prefix to strip from the expanded archive.
+      patches: A list of patches to apply to the expanded archive (with `patch -p1`), in the order
+               in which they should be applied.
       labels: Labels to apply to this rule.
       hashes: List of hashes to verify the rule with.
       config: Configuration file to apply to this subrepo.
@@ -51,6 +53,7 @@ def http_archive(name:str, urls:list, strip_prefix:str=None, labels:list=[],
         name = name,
         urls = urls,
         strip_prefix = strip_prefix,
+        patches = patches,
         labels = labels,
         hashes = hashes,
         config = config,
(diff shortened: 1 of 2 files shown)


chrisnovakovic:
#2746 fixes the plugin docs lint CI failure.

samwestmoreland (approved):
Just that rules_test that needs fixing but otherwise seems sensible to me. Thanks!

sohaibiftikhar:
This does not really work I believe because the patch file is not staged into the dir from where the patch command is called. Is there a working test for this somewhere that I can use to verify?

chrisnovakovic:
The patch file being in a separate directory shouldn't be a problem because `patch(1)` is invoked with `-d "$OUT"`, which changes the directory to the subrepo output directory before applying the patch. The `-d` option is specified in POSIX, so it should be available in every `patch` implementation you're likely to come across.

I don't believe we have tests for this, but we've been relying on the new `patches` parameter for subrepos internally for several months now without any problems.

sohaibiftikhar:
So is it expected that the patch file is part of the subrepo? I have a subrepo I need to patch with my own patch file (exists in my repo not in suprepo). When I try that with something like:
```
github_repo(
    name = "some_repo",
    build_file = "some_repo.BUILD",
    repo = "someorg/somerepo",
    revision = "<commit>",
    patches = ["somerepo.patch"],
)
```
the file `somerepo.patch` is unfortunately not found. I also grepped the plz-out directory for its existence but did not find the patch file staged anywhere. The `somerepo.BUILD` on the other hand is staged and works. Let me know if this is not the intended usecase.

chrisnovakovic:
> So is it expected that the patch file is part of the subrepo?

No - the patch should be in the parent repo, alongside the definition of `github_repo`. Having said that, you're absolutely right that the patch file isn't visible to the rule that builds the subrepo, because it's never added as a source - there's a minor difference between the `new_http_archive` rule we use internally and the one that's built into Please that masked that, sorry. I'll raise a PR to fix this now.

chrisnovakovic:
Should be fixed by #2870. Thanks for reporting this!

PR #2750, opened 2023-03-24
CircleCI: correct `gsutil` option order in release script

chrisnovakovic:
`-h` is a global option to `gsutil` - its current position appears to be causing a `CommandException: Incorrect option(s) specified` error when a file with a specific Content-Type is uploaded.


Tatskaari (approved):
But why did this work before?

chrisnovakovic:
> But why did this work before?

Unsure, but there's nothing else in this script that could have caused https://app.circleci.com/pipelines/github/thought-machine/please/5593/workflows/5519d779-a82f-475c-a416-1774d43cac93/jobs/44126?invite=true#step-103-316, from what I can see...

PR #2753, opened 2023-03-24
Migrate from `go_module()` to `go_get()`

Tatskaari:
Use the new experimental go_get() rules that compile third party code as a subrepo. We will need a better way to update these targets as the new labels are quite unweildly.


src/parse/asp/BUILD
```diff
@@ -2,37 +2,58 @@ subinclude("//build_defs:benchmark")
...
+        "config_test.go",
+        "file_position_test.go",
+        "interpreter_test.go",
+        "label_context_test.go",
+        "lexer_test.go",
+        "logging_test.go",
+        "parser_test.go",
+        "targets_test.go",
+        "util_test.go",
+    ],
```
samwestmoreland, line 48: How come the globs are expanded? Is that expected?

samwestmoreland (approved):
Exciting!

peterebden (approved):
Seems good! I think (hope) this will be a lot less work to maintain `//third_party/go` although arguably it's a bit more to add deps for each library (but maybe they are more systematic so still okay)

Tatskaari:
> Seems good! I think (hope) this will be a lot less work to maintain `//third_party/go` although arguably it's a bit more to add deps for each library (but maybe they are more systematic so still okay)

There's an install` arg on the `go_get()` rules which makes them behave similarly to the `go_module()` rules, i.e. you can depend on `//third_party/go:testify`. I have already done the migration here though. We could start using them going forward if this is too much of a pain, but I'd like to use the Please repo as a test bed for writing a tool to automatically generate the library rules for us.

PR #2757, opened 2023-03-28
Fix perf test

peterebden:
#2755 removes the Java stuff, which breaks it.

This is just removing Java from the test. I looked at re-adding it but it's just building up more stuff specific to that test which I think is just gonna be brittle.


Tatskaari:
Ah man this keeps happening. We might want some kind of pre-merge testing on this...

PR #2761, opened 2023-04-01
Optimise generate of plugin config

peterebden:
Been looking at the [performance graphs](https://please.build/performance.html) which have gone in the bad direction during the v17 work. I can imagine it will be a _bit_ slower but this seems more than I'd expect, and defo wouldn't expect the memory usage to increase the way it has.

Turns out we're spending quite a bit of time & memory in v17 recreating the plugin config (it's redone each time at every `subinclude`, even the global ones). This caches the initial config for each plugin. It also represents the plugin configs as nested config objects rather than normal dicts, so we can take advantage of its base/overlay behaviour to override things.

Before:
```
INFO: Run 1 of 5
INFO: Complete in 7.94s, using 12859852 KB
INFO: Run 2 of 5
INFO: Complete in 7.97s, using 12466732 KB
INFO: Run 3 of 5
INFO: Complete in 7.89s, using 11885564 KB
INFO: Run 4 of 5
INFO: Complete in 7.66s, using 11138540 KB
INFO: Run 5 of 5
INFO: Complete in 7.86s, using 11697292 KB
INFO: Complete, median time: 7.89s, median mem: 11885564.00 KB
```
After:
```
INFO: Run 1 of 5
INFO: Complete in 6.97s, using 9329508 KB
INFO: Run 2 of 5
INFO: Complete in 6.86s, using 9855560 KB
INFO: Run 3 of 5
INFO: Complete in 6.96s, using 9520748 KB
INFO: Run 4 of 5
INFO: Complete in 7.07s, using 9695692 KB
INFO: Run 5 of 5
INFO: Complete in 7.07s, using 10228892 KB
INFO: Complete, median time: 6.97s, median mem: 9695692.00 KB
```
This will do better vs. our normal benchmark (the CPU is faster and has more cores than CircleCI's `xlarge` resource class) but we should see improvement there.

### DIFF
--- a/src/parse/asp/interpreter.go
+++ b/src/parse/asp/interpreter.go
@@ -237,9 +237,14 @@ func (i *interpreter) optimiseExpressions(stmts []*Statement) {
 			if expr.Val.Property == nil && len(expr.Val.Ident.Action) == 0 {
 				expr.Optimised = &OptimisedExpression{Local: expr.Val.Ident.Name}
 				return false
-			} else if expr.Val.Ident.Name == "CONFIG" && len(expr.Val.Ident.Action) == 1 && expr.Val.Ident.Action[0].Property != nil && len(expr.Val.Ident.Action[0].Property.Action) == 0 {
-				expr.Optimised = &OptimisedExpression{Config: expr.Val.Ident.Action[0].Property.Name}
-				expr.Val = nil
+			} else if expr.Val.Ident.Name == "CONFIG" && len(expr.Val.Ident.Action) == 1 && expr.Val.Ident.Action[0].Property != nil {
+				if prop := expr.Val.Ident.Action[0].Property; len(prop.Action) == 0 {
+					expr.Optimised = &OptimisedExpression{Config: prop.Name}
+					expr.Val = nil
+				} else if len(prop.Action) == 1 && prop.Action[0].Property != nil && len(prop.Action[0].Property.Action) == 0 {
+					expr.Optimised = &OptimisedExpression{Config: prop.Name, SubConfig: prop.Action[0].Property.Name}
+					expr.Val = nil
+				}
 				return false
 			}
 		}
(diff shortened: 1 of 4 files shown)


src/parse/asp/interpreter.go
```diff
@@ -19,35 +20,43 @@ type interpreter struct {
...
-	i := &interpreter{
-		scope:       s,
-		parser:      p,
-		subincludes: subincludes,
-		configs:     map[*core.BuildState]*pyConfig{},
-		limiter:     make(semaphore, state.Config.Parse.NumThreads),
+		i.subincludes = cmap.New[string, pyDict](cmap.SmallShardCount, cmap.XXHash)
+		i.pluginConfigs = cmap.New[pluginConfigKey, *pyConfigBase](cmap.SmallShardCount, func(key pluginConfigKey) uint64 {
+			return cmap.XXHash(key.Name) ^ uint64(uintptr(unsafe.Pointer(key.State)))
+		})
```
Tatskaari, line 59: Hmmm I feel like go must know how to hash these objects otherwise we wouldn't be able to use the config key in a map. Bit annoying that it doesn't expose that for us here.

src/parse/asp/interpreter.go
```diff
@@ -236,9 +245,14 @@ func (i *interpreter) optimiseExpressions(stmts []*Statement) {
...
-				expr.Optimised = &OptimisedExpression{Config: expr.Val.Ident.Action[0].Property.Name}
-				expr.Val = nil
+			} else if expr.Val.Ident.Name == "CONFIG" && len(expr.Val.Ident.Action) == 1 && expr.Val.Ident.Action[0].Property != nil {
+				if prop := expr.Val.Ident.Action[0].Property; len(prop.Action) == 0 {
+					expr.Optimised = &OptimisedExpression{Config: prop.Name}
+					expr.Val = nil
+				} else if len(prop.Action) == 1 && prop.Action[0].Property != nil && len(prop.Action[0].Property.Action) == 0 {
+					expr.Optimised = &OptimisedExpression{Config: prop.Name, SubConfig: prop.Action[0].Property.Name}
+					expr.Val = nil
+				}
```
Tatskaari, line 255: Nice! Does this make a big difference? I imagine we do this quite a lot of so it would help.

Tatskaari (approved):
Nice! Looks good to me.

src/parse/asp/interpreter.go
peterebden: It has some way of doing it - I thought the stdlib map hash was exposed somewhere but I can't find it right now.
All I really need is the pointer value anyway...

src/parse/asp/interpreter.go
peterebden: I think a bit of one. We're probably approaching a point where there are more three-level config lookups than two so seemed useful.

PR #2770, opened 2023-04-18
Store and retrieve dirs from remote execution

peterebden:
This should alleviate (if not entirely stop) the time spent retrieving directories from the remote in `addChildDirs`. Basically we stash them in the build metadata so we can reuse them on a cache hit; that trades some memory & CPU for fewer RPCs which is almost certainly a good thing.

Thought about controlling this via a feature flag but I don't think it's worth it. `encoding/gob` should do a fine job of compatibility (it will ignore the new field when reading from an old version, if it's empty in the new one it has no effect, and it does everything by field name so order doesn't matter) and ultimately it's just a set of content-addressed blobs that we don't have to download so they can't be 'incorrect'.

Also added some metrics & logging out the counter values on exit so I can see them.

### DIFF
--- a/src/core/build_target.go
+++ b/src/core/build_target.go
@@ -247,7 +247,8 @@ type BuildMetadata struct {
 	// Standard output & error
 	Stdout, Stderr []byte
 	// Serialised build action metadata.
-	RemoteAction []byte
+	RemoteAction  []byte
+	RemoteOutputs []byte
 	// Time this action was written. Used for remote execution to determine if
 	// the action is stale and needs re-checking or not.
 	Timestamp time.Time
--- a/src/metrics/prometheus.go
+++ b/src/metrics/prometheus.go
@@ -20,6 +20,16 @@ var registerer = prometheus.WrapRegistererWith(prometheus.Labels{
 
 // Push performs a single push of all registered metrics to the pushgateway (if configured).
 func Push(config *core.Configuration) {
+	if family, err := prometheus.DefaultGatherer.Gather(); err == nil {
+		for _, fam := range family {
+			for _, metric := range fam.Metric {
+				if metric.Counter != nil {
+					log.Debug("Metric recorded: %s: %0.0f", *fam.Name, *metric.Counter.Value)
+				}
+			}
+		}
+	}
+
 	if config.Metrics.PrometheusGatewayURL == "" {
 		return
 	}
--- a/src/remote/utils.go
+++ b/src/remote/utils.go
@@ -161,7 +161,7 @@ func (c *Client) getOutputsForOutDir(target *core.BuildTarget, outDir core.Outpu
 
 // readDirectory reads a Directory proto, possibly using a local cache, otherwise going to the remote
 func (c *Client) readDirectory(dg *pb.Digest) (*pb.Directory, error) {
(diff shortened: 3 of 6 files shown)


src/metrics/prometheus.go
```diff
@@ -20,6 +20,16 @@ var registerer = prometheus.WrapRegistererWith(prometheus.Labels{
 
 // Push performs a single push of all registered metrics to the pushgateway (if configured).
 func Push(config *core.Configuration) {
+	if family, err := prometheus.DefaultGatherer.Gather(); err == nil {
```
Garbett1, line 23: Intended to keep in?

src/remote/utils.go
```diff
@@ -181,14 +211,32 @@ func (c *Client) actionURL(digest *pb.Digest, prefix bool) string {
...
+		}
+		for _, dir := range tree.Children {
+			c.directories.Store(c.digestMessage(dir).Hash, dir)
+		}
+		directoriesStored.Add(float64(len(tree.Children)))
+		data, _ := proto.Marshal(&tree)
+		metadata.RemoteOutputs = data
+	}
+
+	if err := c.mdStore.storeMetadata(dg.Hash, metadata); err != nil {
```
Garbett1, line 239: Would be interesting to see how much bigger the cache gets if we _are_ storing all of this in the local MD file. I can image that for stuff like the AWS SDK might be a bit chunky.

Garbett1 (approved):
Small comments but not blocking, want to bump the VERSION file while we are at it?

src/metrics/prometheus.go
peterebden: Yep, I think it's nice to be able to read metrics from the end of a build log if you want. In this case, it's very hard to observe how many of these directories I'm creating/storing - it'd have to push it to Prometheus, then I'd have to go and query it back, just getting it out of the log file seems nice and easy.

We don't have very many metrics so it doesn't take much time or space in there.

src/remote/utils.go
peterebden:
```
$ du -sh ~/.cache/please/build-metadata-store/
47M	/home/pebers/.cache/please/build-metadata-store/
```

Seems fine. Even for something like the v1 AWS SDK I don't think the directory metadata is really all _that_ big.

peterebden:
> Small comments but not blocking, want to bump the VERSION file while we are at it?

I want to talk with Jon first (which means waiting for him to get back from wherever he's larking about), I'm seeing a bunch of errors in other repos which I think is #2765 but I'm not certain if that is expected or not at this point.

src/core/build_target.go
```diff
@@ -247,7 +247,8 @@ type BuildMetadata struct {
 	// Standard output & error
 	Stdout, Stderr []byte
 	// Serialised build action metadata.
-	RemoteAction []byte
+	RemoteAction  []byte
+	RemoteOutputs []byte
```
Tatskaari, line 251: I wonder if this stuff needs to be stored, or if it could just be read back from disk on demand?

Tatskaari (approved):
This looks good to me. Nice find there. I didn't think creating a directory proto would involve going to the CAS. The API is quite awkward like that...

PR #2776, opened 2023-04-21
sandbox: use bit mask to indicate sandbox features

chrisnovakovic:
Rather than passing individual booleans to `contain()` to control the sandboxing behaviour, define and use a bit mask scheme in which each feature is a separate bit flag. This allows additional sandboxing features to be added in future without blowing up `contain()`'s function signature.

### DIFF
--- a/tools/sandbox/main.c
+++ b/tools/sandbox/main.c
@@ -12,6 +12,8 @@
 #include "tools/sandbox/sandbox.h"
 
 int main(int argc, char* argv[]) {
+    int flags = 0;
+
     if (argc < 2) {
         fputs("please_sandbox implements sandboxing for Please.\n", stderr);
         fputs("It takes no flags, it simply executes the command given as arguments.\n", stderr);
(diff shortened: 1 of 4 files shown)


chrisnovakovic:
Tested by running `plz exec --share_network [...] /bin/bash` and `plz exec --share_mount [...] /bin/bash` and making sure the network/file system are correctly shared/isolated.

peterebden:
Are you anticipating there being more of these? I feel like two bools is clearer than one bitmask (I'm not sure how many there'd have to be before I'd swap but I think it's more than that). 

If this were a programmatic API we exposed then it'd be a lot more relevant for the signature to remain consistent.

chrisnovakovic:
> Are you anticipating there being more of these? I feel like two bools is clearer than one bitmask (I'm not sure how many there'd have to be before I'd swap but I think it's more than that).

We have some features in our private fork that it'd be nice to genericise and push upstream, if only so we're not maintaining two copies of the sandbox code any more (although with the acknowledgement that these features won't be useful/desirable to everyone, so it'd be good to be able to switch them on or off in `.plzconfig`). Potential future uses of this that I can envision are:

* enabling or disabling the `pid` namespace in the child process (internally we build a separate binary that enables only the `pid` namespace - I think it'd be cleaner to have this functionality in the main sandbox binary and control it using an environment variable, like we do with `net` and `mount`)
* enabling or disabling the `ipc` namespace in the child process (this might be a more robust way to stop shm being shared with the child process)
* choosing whether to send `SIGTERM` or `SIGKILL` to the child process when the sandbox exits (I can see arguments for wanting to use either - `SIGTERM` is cleaner, but it relies on the child process not ignoring it, which can't be guaranteed)

PR #2778, opened 2023-04-26
Print end-of-test messages directly to stdout

chrisnovakovic:
At the end of a test run, messages are concatenated with newlines into a single string that is then printed, which unnecessarily consumes time and memory if a large number of messages is to be printed. Rather than concatenating the list into a single string, iterate over the list and print the messages directly to stdout.


peterebden (approved):

peterebden:
Nice thanks!

PR #2782, opened 2023-04-28
Re-add glog patch

peterebden:
This got lost in the move to `go_get` and means we now get undesirable output when using remote execution.


third_party/go/BUILD
```diff
@@ -52,8 +52,8 @@ go_get(
 
 go_get(
     module = "github.com/matttproud/golang_protobuf_extensions",
-    version = "v1.0.1",
     requirements = ["github.com/golang/protobuf"],
+    version = "v1.0.1",
```
peterebden, line 56: These are just autoformatting.

samwestmoreland (approved):

Tatskaari (approved):

PR #2784, opened 2023-05-09
Allow provides to be a list

Tatskaari:
This changes 2 things:
1) it allows provides to be a list 
2) It recursively provides when building the sources env var. We were doing this when iterating through the sources to copy into the dir, but were only providing one level when building the sources list for the env vars.

### DIFF
--- a/src/parse/asp/targets.go
+++ b/src/parse/asp/targets.go
@@ -500,8 +500,15 @@ func addProvides(s *scope, name string, obj pyObject, t *core.BuildTarget) {
 		s.Assert(ok, "Argument %s must be a dict, not %s, %v", name, obj.Type(), obj)
 		for k, v := range d {
 			str, ok := v.(pyString)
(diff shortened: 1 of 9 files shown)


src/parse/asp/targets.go
```diff
@@ -500,8 +500,15 @@ func addProvides(s *scope, name string, obj pyObject, t *core.BuildTarget) {
...
 		for k, v := range d {
 			str, ok := v.(pyString)
-			s.Assert(ok, "%s values must be strings", name)
-			t.AddProvide(k, checkLabel(s, s.parseLabelInPackage(string(str), s.pkg)))
+			if ok {
+				t.AddProvide(k, []core.BuildLabel{checkLabel(s, s.parseLabelInPackage(string(str), s.pkg))})
+				continue
+			}
+			_, ok = v.(pyList)
+			s.Assert(ok, "%s values must be strings or a lists of strings", name)
```
samwestmoreland, line 508: list*

PR #2789, opened 2023-05-16
Fix issue with interleaving please builds

### LINKED ISSUE #2788: Race conditions with parallel targets
There's a race condition in Please where two instances might rebuild a target while the other Please requires it for a later action. This often manifests in subrepos such as this build failure: 

https://app.circleci.com/pipelines/github/thought-machine/please/5715/workflows/612bf9ec-e3f4-4040-a132-65e13c7cd7f6/jobs/45088/tests

We can see from the logs that the sequence of events is as such:

First, we can see that `query_alltargets_test` won the race to build the `#dl` rule for the subrepo:

```
query_alltargets_test.log 08:05:57.565   DEBUG: Attempting to acquire lock for plz-out/tmp/third_party/go/_github.com_google_go-cmp#dl._build.lock...
query_alltargets_test.log 08:05:57.565   DEBUG: Acquired lock for plz-out/tmp/third_party/go/_github.com_google_go-cmp#dl._build.lock
query_alltargets_test.log 08:05:57.565   DEBUG: Need to rebuild //third_party/go:_github.com_google_go-cmp#dl, metadata file is missing
query_alltargets_test.log 08:05:57.565    INFO: //third_party/go:_github.com_google_go-cmp#dl: Preparing...
```

We can then see that `query_revdeps_level1` waits for `query_alltargets_test`, as it should:
```
query_revdeps_level4.log 08:05:57.777   DEBUG: Attempting to acquire lock for plz-out/tmp/third_party/go/_github.com_google_go-cmp#dl._build.lock...
query_revdeps_level4.log 08:05:57.777   DEBUG: Looks like process with PID 153443 has already acquired the lock for plz-out/tmp/third_party/go/_github.com_google_go-cmp#dl._build.lock. Waiting for it to finish...
```

However `query_revdeps_level1` wins the race to build the repo rule:
```
query_revdeps_level4.log 08:06:01.270   DEBUG: Acquired lock for plz-out/tmp/third_party/go/_github.com_google_go-cmp#repo._build.lock
query_revdeps_level4.log 08:06:01.270   DEBUG: Need to rebuild //third_party/go:_github.com_google_go-cmp#repo, metadata file is missing
query_revdeps_level4.log 08:06:01.270    INFO: //third_party/go:_github.com_google_go-cmp#repo: Preparing...
query_revdeps_level4.log 08:06:01.270   DEBUG: Attempting to acquire lock for plz-out/tmp/third_party/go/_github.com_google_go-cmp#repo._build.lock...
```

Then `query_alltargets_test` waits patiently for this to finish: 
```
query_alltargets_test.log 08:06:01.454   DEBUG: Looks like process with PID 153351 has already acquired the lock for plz-out/tmp/third_party/go/_github.com_google_go-cmp#repo._build.lock. Waiting for it to finish...
```

The problem is that `query_alltargets_test` considered the `#dl` rule to have changed, which means that it will re-build the repo rule when it acquires the targets lock:

```
08:06:02.467   DEBUG: Acquired lock for plz-out/tmp/third_party/go/_github.com_google_go-cmp#repo._build.lock
08:06:02.567   DEBUG: Need to rebuild //third_party/go:_github.com_google_go-cmp#repo, //third_party/go:_github.com_google_go-cmp#dl has changed
```

This results in it clobbering the outputs and breaking the other Please targets that might need to use that subrepo. 

This is all because of an optimisation here:
```
	// Check the dependencies first, because they don't need any disk I/O.
	if target.NeedsTransitiveDependencies {
		if anyDependencyHasChanged(target) {
			return true // one of the transitive deps has changed, need to rebuild
		}
	} else {
		for _, dep := range target.Dependencies() {
			if dep.State() < core.Unchanged {
				log.Debug("Need to rebuild %s, %s has changed", target.Label, dep.Label)
				return true // dependency has just been rebuilt, do this too.
			}
		}
	}
```

The simplest solution would be to disable this, however it probably has a reasonable impact on performance.

Tatskaari:
This PR addresses issue #2788 by removing an optimization check to avoid hashing the outputs of our dependencies when we know they have changed because we built them earlier on. The removal of this check should not change the behavior since the sources hash would have been updated when any of the dependencies changed.

The problem arises when two Please instances race to build two targets that depend on each other, in this case #dl which is depended on by #repo. If the instances interleave, such that one instance builds #dl, the other builds #repo, then the first instance rebuilds #repo even though it was already built by the second instance. This occurs because the first instance marks #dl as changed and does not check if #repo has subsequently been built by another Please instance. This can lead to build failures due to the subrepo directory being deleted and recreated when we copy the build outputs to `plz-out/subrepos`, while another Please instance tries to read from it simultaneously.

By removing the check, the sources hash will be calculated normally. Since the second Please instance has already built the target, the sources will not have changed, and the build artifact will be reused as intended.

Although there may be a slight performance impact, it is expected to be minimal. If the dependencies have changed, the target will need to be rebuilt anyway, which will likely take significantly longer than the hashing process.


peterebden (approved):
I think this is correct? We basically just rely on hashing every input and that _should_ be correct (I worry that I'm missing some case here but it seems like it should be).

I don't think there is even a performance impact on the happy path, we actually do slightly less work than before. I guess on the unhappy path (rebuilding) it takes a bit longer to figure it out but I think that'll not be super noticeable if we're actually doing build things anyway.

PR #2793, opened 2023-05-23
Change plz exec args

peterebden:
This makes them behave similarly to `plz exec sequential`, `plz exec parallel` and `plz run` (and its subcommands too), i.e. it removes the command override business. I think that is no longer necessary since the debugging functionality it supported now has `plz debug` but I think this is quite a bit easier to understand (before it was tricky to understand how to pass flags into `plz exec`).


samwestmoreland (approved):
Looks like still some tests failing, but otherwise seems sensible to me!

PR #2795, opened 2023-05-24
Add strip option to text_file

izissise:
For multilines text_file it is usually convenient to strip all spaces on the leftside of the text so the text is correctly intended in the code.


Tatskaari (approved):
Nice! I think this makes sense. Thanks :)

Tatskaari (approved):
Nice! I think this makes sense. Thanks :)

Issue #2777, opened 2023-04-22
Investigate on where and how to add a minimal graphviz dot output


erik-adelbert:
This issue [originated on please gitter chan](https://matrix.to/#/!NGUhfgTASEHCAYHXIN:matrix.org/$QdtWnuvv95xEmvliMGGvQzopm_cBBBEU6NOX9jq3zs0?via=matrix.org&via=gitter.im), it invites a discussion about adding `graphviz dot` support to `plz` output:

1. Fabian Ponce was asking about a way to output the build graph in `graphviz dot` format
2. I replied by publishing our `awk` script that rewrites `plz query deps <target>` output in `dot`
3. @Tatskaari reacted positively to it and told us about a (nice) existing plan to do more than that
4. But in the meantime, his conclusion was:

> yeah, adding dot output should be quite simple if you want to have a go
> I think you might just want to switch on a flag here:
> https://github.com/thought-machine/please/blob/master/src/query/graph.go#L24
> 
> And spit out the structs using a dot file encoder (hopefully there's something in the wild we can use, though it's not a super complex formatting)

But imho, the proposed insertion point and technique doesn't fit so well:
- On one hand the existing `awk` script piggybacks `query deps` to rewrites the output, it is simple: the dependency graph is kept internally by please following a successful build, the translation is mechanical. This result is akin to [`Bazel`](https://blog.bazel.build/2015/06/17/visualize-your-build.html). It could be added to please [here](https://github.com/thought-machine/please/blob/master/src/query/deps.go),
- On the other hand, the graph output of `query graph` is flat and embeds extra data (build state and more), inserting a `dot` output here would probably require to rebuild the dependency graph and, imo, poses at the very least the open question on how to represent the data. Addressing these questions seems to exceed the purpose of an encoder.

The question is: **Should we add `graphviz dot` support now? If so, would it be acceptable to add it to `query deps`?**

PR #2801, opened 2023-05-26
Can generate dot graph on query deps command

izissise:
Add an option --dot on query deps which
generate a dot compatible graph.

Graph nodes will have different shapes depending
if it's a filegroup/remote_file/text_file/binary

One can get an image with
`plz query deps //src:please --dot | dot -Tpng > graph.png`

issue #2777

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -338,6 +338,7 @@ var opts struct {
 
 	Query struct {
 		Deps struct {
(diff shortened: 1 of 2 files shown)


Tatskaari (approved):
Nice one! Tried it out a little bit and it seems to work great. Thanks for the PR!

PR #2804, opened 2023-05-31
Fix the release file function in the release script

Tatskaari:
We we're uploading these files to the wrong place:
```
upload: ./VERSION to s3://please-releases//latest_version
```


samwestmoreland (approved):

peterebden (approved):
those sneaky slashes!

PR #2808, opened 2023-06-02
Retrieve named_outputs keys with get_named_outs

izissise:
When working with named outputs it is convenient to
be able to get the keys from the target after it was created.

With this addition, one can do the following:
```
target = genrule(
  name = "target",
  outs = {
    "a": ["a"],
    "b": ["b"]
  },
  ...
)

target = get_named_outs(target)

target2 = genrule(
   name = "target2",
   srcs = target.a,
   ...
)
```

This makes easy to re-export a target outs in a filegroup:
```
target = genrule(
  name = "target",
  outs = {
    "a": ["a"],
    "b": ["b"]
  },
  ...
)

fg = filegroup(
  name = "fg",
  srcs = get_named_outs(target) | {
    "c": [":another_rule"]
  }
)
```

### DIFF
--- a/rules/builtins.build_defs
+++ b/rules/builtins.build_defs
@@ -235,6 +235,8 @@ def add_out(target:str, name:str, out:str=''):
     pass
 def get_outs(target:str):
     pass
+def get_named_outs(target:str) -> dict:
+    pass
 def add_licence(target:str, licence:str):
     pass
 def get_licences(target:str):
--- a/test/get_outs/BUILD
+++ b/test/get_outs/BUILD
@@ -1,0 +1,90 @@
+nonamedouts = genrule(
+    name = "nonamedouts",
+    outs = ["x"],
+    cmd = """
+        echo 'x' > "$OUTS"
+    """,
+)
+
+gr_getouts = genrule(
+    name = "genrule_getouts",
+    outs = {
+        "wibble": ["wibble_file1"],
+        "wobble": ["wobble_file1"],
+    },
+    cmd = """
+        echo 'wibblewibblewibble' > "$OUTS_WIBBLE"
+        echo 'wobblewobblewobble' > "$OUTS_WOBBLE"
+    """,
+)
+
+fg_getouts = filegroup(
+    name = "filegroup_getouts",
+    srcs = {
+        "wibble": [text_file(name = 'wibble_file2', content = 'wibblewibblewibble')],
+        "wobble": [text_file(name = 'wobble_file2', content = 'wobblewobblewobble')],
+    },
+)
+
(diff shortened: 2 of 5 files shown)


rules/builtins.build_defs
```diff
@@ -238,7 +238,7 @@ def add_data(target:str, datum:str|list|dict):
     pass
 def add_out(target:str, name:str, out:str=''):
     pass
-def get_outs(target:str):
+def get_outs(target:str, keys:bool=False):
```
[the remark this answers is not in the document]
izissise: I agree in totality, I'll try to make the changes next week.

izissise:
It should be good, not sure why the CI is falling though

Tatskaari:
@izissise looks like you need to `go fmt` a few files. The other lint error is about keeping the docs site up to date with the latest plugin releases. You just need to update the plugin version for go in `//docs/BUILD` from 1.6.0 to 1.7.0 (I think)

Tatskaari (approved):

test/get_outs/BUILD
```diff
@@ -0,0 +1,90 @@
...
+assert_dict({}, get_named_outs(nonamedouts))
+
+assert_dict({
+    "wibble": ["//test/get_outs:genrule_getouts|wibble"],
+    "wobble": ["//test/get_outs:genrule_getouts|wobble"],
+}, get_named_outs(gr_getouts))
+
+assert_dict({
+    "wibble": ["//test/get_outs:filegroup_getouts|wibble"],
+    "wobble": ["//test/get_outs:filegroup_getouts|wobble"],
```
Tatskaari, line 42: Sorry! I was thinking more that the dict would return a mapping of the named output tag to the outs that constitute it i.e. 
```
{
    "wibble": ["wibble_file1"],
    "wobble": ["wobble_file1"]
}
```

Does that make sense? Apologies for the back and forth.

Tatskaari (changes requested):

test/get_outs/BUILD
izissise: No apologies necessary, It makes more sense with what get_outs is doing indeed.
I think it will make the result less useful to use with an another rule, but that's fine, one can still regenerate the `AnnotatedOutputLabel` using the keys.

Maybe something better for the usage I'm describing would be `get_target_outs` that would return the same thing as right now ?

What do you think?

izissise: I made the change you requested

Tatskaari (approved):

Tatskaari:
Nice one! Thanks for that :)

PR #2811, opened 2023-06-05
Quote source paths in plugin_repo

### LINKED ISSUE #2809: Please 17.0 fails to build projects with spaces in their paths.
Since 17.0, it appears that projects with spaces in their paths no longer build.
From a cursory look, it appears to be related to [these commands](https://github.com/thought-machine/please/blob/b4c5472eac3a7fcb76c9fbfeee941743acd69602/rules/subrepo_rules.build_defs#LL347C35-L347C35), which seem to be used as part of the plugin system that is now required for most rules.

peterebden:
I think this _should_ fix #2809 although it doesn't seem to pass bootstrap for me locally - but I can build it with a plz with this fix. Somewhat unsure exactly what is going on there, although I think this is clearly better.

### DIFF
--- a/rules/subrepo_rules.build_defs
+++ b/rules/subrepo_rules.build_defs
@@ -345,7 +345,7 @@ def plugin_repo(name:str, revision:str,  plugin:str="", owner:str="please-build"
         outs = [name],
         cmd = " && ".join([
           '$TOOL x "$SRCS" -o _tmp',
-          "mv $(dirname $(ls _tmp/*/.plzconfig)) $OUT",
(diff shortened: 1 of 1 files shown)


samwestmoreland (approved):

Tatskaari (approved):
Looks good but I'm curious what the error is locally?

peterebden:
> Looks good but I'm curious what the error is locally?

Same thing as before - as though the paths weren't quoted (IIRC it was "extra argument" type things)

PR #2813, opened 2023-06-07
Implement `any` and `all` built-ins natively

chrisnovakovic:
Switch from an ASP-based implementation of the `any` and `all` built-in functions to a native implementation. This significantly improves the worst-case performance of both functions.

### DIFF
--- a/rules/builtins.build_defs
+++ b/rules/builtins.build_defs
@@ -78,17 +78,12 @@ def isinstance(obj, types:function|list) -> bool:
 
 def range(start:int, stop:int=None, step:int=1) -> str:
     pass
-def any(seq:list) -> bool:
-    for x in seq:
-        if x:
-            return True
-    return False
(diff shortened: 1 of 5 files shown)


chrisnovakovic:
ngl, the impetus for doing this was curiosity after seeing the functions implemented in `rules/builtins.build_defs`, but I was surprised at how much faster the native implementations are so I thought it was worth opening the PR anyway. Poor man's benchmark using comically large lists, just to illustrate the point:

```
## BUILD

any([False for i in range(100000000)] + [True])
```

Before (ASP implementation):

```
$ ../plz-out/bin/src/please build
Build finished; total time 11.47s, incrementality 100.0%. Outputs:
```

After (native implementation):

```
$ ../plz-out/bin/src/please build
Build finished; total time 7.18s, incrementality 100.0%. Outputs:
```

PR #2816, opened 2023-06-10
Remove thread id

peterebden:
This removes the 'thread id' concept everywhere, which is only actually relevant in the output package. We now start a goroutine for every incoming task & just limit them to a certain number processing at a time. That _might_ alter latency a bit (because there are lots more goroutines, most of which are blocking and 'ready to go' as soon as the previous one finishes) but I doubt it's measurable.
The remote and local task queues are unified into one, as is the output which now scales up & down more like the remote one did, which is probably the most externally visible change here.

Most of the files touched are a mechanical removal of the `tid` argument. The interesting bits are in the `plz` and `output` packages - particularly the latter has to do more bookkeeping to assign tasks into slots (which is a little subtle to make it 'feel nice').

I think this is quite a bit nicer. I've wanted to get rid of this concept for a while now but hadn't really gotten motivated until our conversation the other day.

### DIFF
--- a/src/plz/plz.go
+++ b/src/plz/plz.go
@@ -86,7 +86,7 @@ func RunHost(targets []core.BuildLabel, state *core.BuildState) {
 	Run(targets, nil, state, state.Config, cli.HostArch())
 }
 
-func doTasks(tid int, state *core.BuildState, actions <-chan core.Task, remote bool) {
+func doTasks(state *core.BuildState, actions <-chan core.Task, remote bool) {
 	for task := range actions {
 		switch task.Type {
 		case core.TestTask:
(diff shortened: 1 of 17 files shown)


src/plz/plz.go
```diff
@@ -39,34 +39,44 @@ func Run(targets, preTargets []core.BuildLabel, state *core.BuildState, config *
...
+	go func() {
+		for task := range actions {
+			go func(task core.Task) {
+				if task.Remote {
+					remoteLimiter.Acquire()
+					defer remoteLimiter.Release()
+				} else {
+					localLimiter.Acquire()
+					defer localLimiter.Release()
+				}
```
Tatskaari, line 68: Really excited to see if we can defer this decision into build_step, and build locally, but spill over into remote execution if there's too much backpressure on the local limiter.

Tatskaari (approved):
Looks good to me :D

src/plz/plz.go
peterebden: I think it's possible - definitely more possible than before. The limiter would need to be quite a bit smarter though.

PR #2820, opened 2023-06-12
Fix language server crash when trying to format malformed content

jkimbo:
This PR attempts to fix a crash that happens when trying to format a BUILD file that is malformed. The lsp `Handle` function assumes that any error it encounters is already a `jsonrpc2.Error` but it can be a `ParseError` (and presumably other errors as well). In those cases this fix casts other error types to an internal jsonrpc2 error.

Note: I'm not sure how to test this change, it doesn't look like the `Handle` function is actually tested anywhere. The best I could think of doing it adding a test to show that the result of the formatting call is a `ParseError` type, not a `jsonrpc2.Error`.


Tatskaari (approved):
Nice one! Thanks :)

PR #2832, opened 2023-06-19
`github_repo`: strip `v` prefix if `revision` is any version number

### LINKED ISSUE #2829: `github_repo` fails when `revision` is a non-semver version number
MWE:

```
## java/snappy-java/BUILD

github_repo(
    name = "snappy-java",
    repo = "xerial/snappy-java",
    revision = "v1.1.10.1",
    hashes = ["b5dcf03f9618bd33fd9f424d9acf711a175aecd690cfca31440954723064d449"],
    build_file = "snappy-java.build",
)
```

```
$ plz build //java/snappy-java:snappy-java
Build stopped after 60ms. 1 target failed:
    //java/snappy-java:snappy-java
Error building target //java/snappy-java:snappy-java: exit status 1
mv: cannot move 'java/snappy-java/snappy-java.build' to '[...]/plz-out/tmp/java/snappy-java/snappy-java._build/snappy-java/BUILD': No such file or directory
```

This happens because the first command `github_repo` runs - namely, the GitHub archive zip extraction operation using arcat - fails because arcat is given the wrong prefix to strip by the rule, but exits successfully:

```
bash-5.1$ unzip -l $SRCS_REMOTE | head -n5
Archive:  java/snappy-java/snappy-java_v1.1.10.1.zip
737f3973ff392c351da2744820a23a0521c6cc77
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2023-06-14 22:51   snappy-java-1.1.10.1/
bash-5.1$ $TOOL x $SRCS_REMOTE -o "$OUT" -s snappy-java-v1.1.10.1
bash-5.1$ echo $?
0
```

GitHub doesn't publish the algorithm they use to decide when to strip the `v` prefix from a tag name during the archival process, so the rule currently assumes that any `revision` that looks like a semver will have any leading `v` stripped from the top-level directory name in the exported archive. In reality the logic appears to be more complicated than that. It's probably safe to assume that any dotted version number format will also cause the `v` prefix to be stripped on GitHub's side.

In addition, perhaps it's worth changing arcat so that it warns/errors if any value given by `-s` results in no files being extracted - the underlying cause would then become clearer here.

chrisnovakovic:
`github_repo` currently strips the `v` prefix from the file names in the zip file downloaded from GitHub if `revision` is a semver, since GitHub silently does the same if asked to archive the code for semver-like tags. It looks like this logic extends to more than just semvers, though - the same behaviour has been observed for strings that are simple dotted version numbers with any number of components (e.g. `v1.1.10.1`). Account for this in the prefix-stripping logic.

Fixes #2829.

### DIFF
--- a/rules/subrepo_rules.build_defs
+++ b/rules/subrepo_rules.build_defs
@@ -202,9 +202,19 @@ def github_repo(name:str, repo:str, revision:str, build_file:str=None, labels:li
     org, _, repo = repo.partition('/')
     assert repo, "Must pass a valid Github repo argument, e.g. thought-machine/please"
 
-    # If it looks like semver, we need to remove the v from it because github formats their zips differently for semver
-    # tags
-    if is_semver(revision):
+    # If revision looks like a version number, we need to strip any "v" prefix from it because
+    # GitHub also strips it from the top-level directory name for zips generated for tags that
+    # look like version numbers
+    def is_dotted_version_number(v):
+        v = v.removeprefix("v")
+        num = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9"]
+        for p in v.split("."):
+            for n in num:
+                p = p.replace(n, "")
+            if p != "":
+                return False
+        return True
+    if is_semver(revision) or is_dotted_version_number(revision):
         prefix = revision.removeprefix('v')
     else:
         prefix = revision


chrisnovakovic:
> This issue has been automatically marked as stale because it has not had any recent activity in the past 90 days. It will be closed if no further activity occurs. If you require additional support, please reply to this message. Thank you for your contributions.

Not stale, but I'm going to think of a more efficient way to implement this.

PR #2835, opened 2023-06-23
Adds some hostname + version metric

Garbett1:


src/metrics/prometheus.go
```diff
@@ -24,6 +26,23 @@ func Push(config *core.Configuration) {
 		return
 	}
 
+	if config.Metrics.PushHostinfo {
+		name, _ := os.Hostname()
+		gauge := prometheus.NewGauge(prometheus.GaugeOpts{
```
peterebden, line 31: why a gauge rather than a counter?

src/metrics/prometheus.go
```diff
@@ -24,6 +26,23 @@ func Push(config *core.Configuration) {
...
+			Namespace: "plz",
+			Subsystem: "metrics",
+			Name:      "hostinfo",
+			Help:      "Please host running info",
+			ConstLabels: prometheus.Labels{
+				"remote":   strconv.FormatBool(config.Remote.URL != ""),
```
peterebden, line 37: prefer `config.NumRemoteExecutors() > 0`, that handles both the URL and whether we're configured to execute anything

maybe we should add an explicit function for this too, there are several cases where things ask this question

src/core/config.go
```diff
@@ -708,6 +708,7 @@ type Configuration struct {
 	Metrics struct {
 		PrometheusGatewayURL string       `help:"The gateway URL to push prometheus updates to."`
 		Timeout              cli.Duration `help:"timeout for pushing to the gateway. Defaults to 2 seconds." `
+		PushHostinfo         bool         `help:"Whether to push host info"`
```
peterebden, line 711: nit: `PushHostInfo`

peterebden (commented):
thanks!

src/metrics/prometheus.go
Garbett1: Re-considered this, counter probably does make more sense from the perspective of gathering and summing them.

src/core/config.go
```diff
@@ -1080,6 +1081,13 @@ func (config *Configuration) NumRemoteExecutors() int {
 	return config.Remote.NumExecutors
 }
 
+func (config *Configuration) IsRemoteExecutution() bool {
```
peterebden, line 1084: Execututorisation -> Execution

Garbett1:
I'm landing despite the lint failing as I assume we might want to update the Go plugin soon.

PR #2838, opened 2023-06-24
Go toolchain update

peterebden:
Updates to latest Go, adds stdlib


peterebden:
OK so there's some strange issue with asm symabi generation where it's looking at `/home/runner/work/go-rules/go-rules` instead of `~/git/please`. This happens on my machine too, it isn't a CircleCI thing, but I can't find where it's getting that from

peterebden:
Also some awkward stuff with running Go binaries as subprocesses when they're built for coverage; they emit messages about "GOCOVERDIR not set" which seem to be hard to suppress.

PR #2841, opened 2023-06-28
Use the built in sandboxing by default

Tatskaari:
This is potentially a breaking change but I think we can safely consider this an implementation detail. The newer Go versions produce binaries that are far less expensive during init so anecdotally, the perf issues we were experiencing seem to be no more. I think this is a better default value as it means users no longer need to install `please_sandbox`.


peterebden:
> The newer Go versions produce binaries that are far less expensive during init

Do they? I wasn't aware anything had changed there.

This is probably fine though. I agree that we don't have to consider it a breaking change.

PR #2843, opened 2023-06-28
Preload the preloaded subincludes in files involved in normal (non-preloaded) subincludes 

Tatskaari:
This means that we can be more specific about where we don't preload subincludes. Before any package involved in any kind of subinclude would be skipped. Now it's only if that package is part of the preloading.


src/build/BUILD
```diff
@@ -36,3 +36,18 @@ go_test(
         "//src/fs",
     ],
 )
+
+go_test(
```
Tatskaari, line 40: oops

test/preloaded_subinc/test_repo/src/build_defs/BUILD_FILE
```diff
@@ -0,0 +1,8 @@
+# This should be available to us via preloads even though this file is involved in a subinclude
+blah = sh_test
```
Tatskaari, line 2: This used to fail but we're more specific about when we don't preload now

src/parse/parse_step.go
```diff
@@ -45,22 +45,22 @@ func parse(state *core.BuildState, label, dependent core.BuildLabel, forSubinclu
 
 	// Ensure that all the preloaded targets are built before we sync the package parse. If we don't do this, we might
 	// take the package lock for a package involved in a subinclude, and end up in a deadlock
-	if !forSubinclude {
+	if !mode.IsPreload() {
 		state.WaitForPreloadedSubincludeTargetsAndEnsureDownloaded()
 	}
```
Tatskaari, line 50: Being more specific here

src/core/state.go
```diff
@@ -631,7 +646,7 @@ func (state *BuildState) WaitForPreloadedSubincludeTargetsAndEnsureDownloaded()
 			// Queue them up asynchronously to feed the queues as quickly as possible
 			wg.Add(1)
 			go func(inc BuildLabel) {
-				state.WaitForTargetAndEnsureDownload(inc, OriginalTarget)
+				state.WaitForTargetAndEnsureDownload(inc, OriginalTarget, true)
 				wg.Done()
 			}(inc)
```
Tatskaari, line 651: this sets the preload flags

src/core/state.go
```diff
@@ -23,6 +23,21 @@ import (
 	"github.com/thought-machine/please/src/process"
 )
 
+type ParseMode uint8
+
+const (
+	ParseModeForSubinclude ParseMode = 1 << iota
```
peterebden, line 29: how about adding a `ParseModeNormal = 0` constant? We often call things with 0 which is a little unclear at the call site.

src/core/state.go
```diff
@@ -23,6 +23,21 @@ import (
...
+	ParseModeForPreload
```
peterebden, line 30: do you need `ParseModeForPreload ParseMode`? I feel I've been caught out by this before

peterebden (approved):

src/core/state.go
Tatskaari: It seems to compile fine so I assume not?

PR #2859, opened 2023-07-13
Support preloads in subincludes

Tatskaari:
This change preloads all subincludes of a preloaded target as well as the preloaded target.


test/preloaded_subinc/test_repo/src/build_defs/empty.build_defs
```diff
@@ -0,0 +1 @@
+foo()
```
zgeor, line 1: Should this be an empty file? If not then should there be an empty file test?

zgeor:
Can you add a brief description of the change?

test/preloaded_subinc/test_repo/src/build_defs/empty.build_defs
goddenrich: maybe the name should be clearer - there is a comment in the BUILD_FILE above. This is a build_def which doesn't define anything but does use a build def defined elsewhere which we expect to be able to be used because the def has been preloaded

src/parse/asp/interpreter.go
```diff
@@ -78,7 +78,7 @@ func (i *interpreter) getConfig(state *core.BuildState) *pyConfig {
 
 // LoadBuiltins loads a set of builtins from a file, optionally with its contents.
 func (i *interpreter) LoadBuiltins(filename string, contents []byte, statements []*Statement) error {
-	s := i.scope.NewScope(filename)
+	s := i.scope.NewScope(filename, 0)
```
goddenrich, line 81: I think using an enum or comment here to clarify what 0 means would be useful

goddenrich (approved):

src/core/state.go
```diff
@@ -648,6 +650,9 @@ func (state *BuildState) WaitForPreloadedSubincludeTargetsAndEnsureDownloaded()
 			wg.Add(1)
 			go func(inc BuildLabel) {
 				state.WaitForTargetAndEnsureDownload(inc, OriginalTarget, true)
+				if err := state.Parser.RegisterPreload(inc); err != nil {
+					panic(err)
```
peterebden, line 654: This seems a bit worrying, if the preload fails for any reason (including fairly likely things like errors in the file) this is gonna die with an enormous stacktrace.

Probably we should use an errgroup here to return the errors so they can be handled.

peterebden: Okay I've updated this to handle errors with an errgroup.

peterebden (approved):

PR #2861, opened 2023-07-14
Implement `min` and `max` built-in functions

chrisnovakovic:
```
min(seq[, key])

max(seq[, key])
```

These functions return the "least" and "greatest" items in a list respectively. The optional `key` parameter is a function that is applied to each item before they are compared.


Tatskaari (approved):

Tatskaari:
Thanks Chris!

PR #2870, opened 2023-07-24
`new_http_archive`: add `patches` files as sources to extraction target

chrisnovakovic:
Currently, any attempt to use the `patches` parameter on subrepo rules fails because the patch files referenced in the command aren't visible as sources to the target that extracts the subrepo archive. Add them so that the generated `patch` commands succeed.


samwestmoreland (approved):

sohaibiftikhar:
Thanks for the quick fix!

PR #2872, opened 2023-07-26
Add Filter and Map builtin functions

izissise:
```
filter(filter, seq)
```
returns a copy of the given list with the items filtered using function

```
map(mapper, seq)
```
returns a copy of the given list with the items transformed using function


Tatskaari (approved):
Thanks! I guess it's nice to have more builtins from Python. I suppose these might be much faster than list comprehensions.

PR #2873, opened 2023-07-26
Set CurrentSubrepo when it's a subrepo for arch

samwestmoreland:


goddenrich (approved):

samwestmoreland:
This isn't right. Better fix incoming.

PR #2874, opened 2023-07-26
`format`: split out test cases into separate files

chrisnovakovic:
This will make the purpose of each test clearer, make adding new test cases simpler, and make regressions easier to spot in failing test output.


Tatskaari (approved):
Nice one! Thanks for that Chris!

PR #2877, opened 2023-07-28
Wait for non pkg local targets in getX builtins

izissise:
Use WaitForBuiltTarget with `get*` (get_outs, get_licenses, get_labels, ...) allows querying non-local targets metadata.

It can be useful to be able to re-export filegroups inside another, but it is tedious if you need to repeat the keys. With these changes, one can write something like this:

```
filegroup(
    name = 'in_pkg_b',
    visibility = ['PUBLIC'],
    srcs = {
        'X': [...]
        'y': [...]
    }
)
---------------
filegroup(
    name = 'in_pkg_a',
    visibility = ['PUBLIC'],
    srcs = {
        'A': [...]
    } | {
        f'B_{k}': [ f'//b:in_pkg_b|{k}' ] for k in get_named_outs('//b:in_pkg_b').keys()
    }
)
```

Also it makes possible the following syntax:
```
def dictioniryze(target:str) -> dict:
    return { k: [ f'{target}|{k}' ] for k in get_named_outs(target).keys() }

a = dictioniryze('//a:in_pkg_a')

genrule(
    name = 'build',
    srcs = {
        'A': a.A,
        'B1': B.X,
        'B2': B.y,
    }
)

```

The Pr is separated into two commits:
- 61b839 - main changes, make add*, get* and set* use `getTarget()` to resolve targets, increasing consistency between these functions. `getTarget()` will `WaitForBuiltTarget` on targets not local to the current package.
- 6f084d - Implement a basic cycle detector with a e2e test for the case the user creates a cycle of dependencies, someone more experienced with please codebase could probably implement this better than me.

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -1463,7 +1463,6 @@ loop:
 				cyclestr += "[" + k.String() + "]{" + dstr + "}\n"
 			}
 			log.Fatalf("Dependencies cycle detected:\n%s", cyclestr)
-			panic("Fatal failed")
 		}
 	case ptask := <-state.waitTargetCycleDetectorRm:
 		if !ptask.Mode.IsForInformation() {
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -1463,6 +1463,7 @@ loop:
 				cyclestr += "[" + k.String() + "]{" + dstr + "}\n"
 			}
 			log.Fatalf("Dependencies cycle detected:\n%s", cyclestr)
+			panic("Fatal failed")
 		}
 	case ptask := <-state.waitTargetCycleDetectorRm:
 		if !ptask.Mode.IsForInformation() {
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -30,6 +30,7 @@ const (
 	ParseModeNormal ParseMode = 1 << iota
 	ParseModeForSubinclude
 	ParseModeForPreload
+	ParseModeForInformation
 	ParseModeForceBuild
 )
 
@@ -39,6 +40,10 @@ func (m ParseMode) IsPreload() bool {
 
 func (m ParseMode) IsForSubinclude() bool {
 	return m&ParseModeForSubinclude != 0
+}
+
+func (m ParseMode) IsForInformation() bool {
+	return m&ParseModeForInformation != 0
 }
 
 // startTime is as close as we can conveniently get to process start time.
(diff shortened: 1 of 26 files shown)


Tatskaari:
Just a heads up that this is quite involved and will require us to think through the various ways in which this changes how packages and targets can depend on each other. The team internally here is firefighting so Please work is on hold for a few more weeks, so we probably won't be able to review it until that's resolved. Will try and get back to this when the fires are all under control. :)

izissise:
Thank you for the heads up! I completely understand the situation. Good luck with resolving the issues, and I hope everything gets back on track soon! :)

peterebden:
I'm not convinced this is a good idea. This couples one target to another in a different package, I think it will ignore `visibility` and it seems like a fertile source of things deadlocking against one another.

izissise:
Do you want to wait for https://github.com/thought-machine/please/pull/2888 to be merged to further discuss this?

peterebden:
I'm not really expecting that to change a lot in this regard. I'm hoping it will make the queuing / building logic simpler and easier to understand, but it doesn't really affect my concern about deadlocking.

I also don't have a particular timeframe to finish it up; these days I'm not working in the infra department for my day job so my work on it is a bit more itinerant. Unclear if a bunch of plane travel in the next month or two will help or hinder it... :)

izissise:
I've implemented a runtime cycle detector that triggers a panic if `waitonbuildtarget()` could potentially cause a deadlock. However, there's a problem with the tests not running correctly on the Darwin platform, and regrettably, I don't have the capability to verify it on that platform.

Currently, when calling `get_outs('@//b:b')` from package a, the behavior is inconsistent and depends on the order in which packages were parsed. It works randomly. We should make a change to ensure that, at the very least, an error is returned.

izissise:
I'm closing this, 
I found a way to get target metadata with current please builtin
- download https://raw.githubusercontent.com/thought-machine/please/v{please_version}/rules/misc_rules.build_defs
- parse the ast and rewrite genrule function to add meta field like this
```
+__META_TEMPLATE = """
+def meta_{func}({func_params}, meta={{}}):
+    _dict_args = json({dict_args} | meta)
+    text_file(
+        name = f'{{name}}_+meta',
+        strip = True,
+        content = f'''
+            CONFIG.setdefault('{{META}}', {{{{}}}}) # should not be needed
+            null = None # cheat hack
+            true = True # cheat hack
+            false = False # cheat hack
+            CONFIG['{{META}}'] = CONFIG['{{META}}'] | {{{{ '{{name}}': {{_dict_args}} }}}}
+        ''',
+    )
+    return {func}({call_args})
+"""
```
- now you can use meta_genrule instead of genrule
- meta_genrule create a target with '_+meta'
- add meta_get
```
+def meta_get(pkg: str, target: str) -> dict:
+    subinclude(buildlabel(pkg=pkg, target=f'{target}_+meta', aslist=False))
+    return CONFIG[META].get(target)
```
- target_meta = meta_get('target')
- target_meta contains all parameters given to genrule
- now you can stop repeating key value all over the codebase, and put them only where they make sense

PR #2880, opened 2023-07-31
Add reduce builtin

izissise:
reduce(function, seq, initializer=None)

apply of function of two arguments cumulatively to the items of seq and return result


Tatskaari:
Thanks for that :D

PR #2884, opened 2023-08-01
Fix concurrent download woes

peterebden:
I'm seeing various failures from concurrent plz's which I'm pretty sure are due to one instance downloading a subrepo, going on to try to use it then another one re-downloading it during that process.
(I think there are other failure cases too that don't involve subrepos, but they're a bit less easily triggered and this is the most obvious one with v17 because basically any operation involves at least one subrepo now)

This basically hoists the locking a bit so it also covers checking for whether outputs exist. In the process I've simplified by having `Build` just call `Download` rather than reinventing its own take on things.

Before this I could fairly trivially trigger explosions by doing `plz clean` and scripting 10 plz processes to run at once trying to build basically anything, which then fight over the subrepos. After this the script doesn't fail any more.


Tatskaari (approved):
Oh nice! I think that makes sense.

PR #2888, opened 2023-08-05
Rework target queuing

peterebden:
This attempts to simplify the queuing logic by making the interfaces to it synchronous and leaning (even more) heavily on goroutines. There are a few top-level functions on State which other things use to do their things (`Parse`, `Build`, `Test` etc) and those internally do more things if they need. Ideally we end up in a state where we track completion of the build by calling those functions and waiting rather than counting 'pending actions' etc.

I find this quite a bit simpler but think it's time to get some feedback. It's (seemingly) working on the happy path but I still need to figure out what to do on the sad path - when a target fails to build, a bunch of other things are blocked on it and just hang around forever. (I tried to fix that at the end but it didn't really work and I don't think I like it - I might undo that although I've also rebased the commit into everything else...)

This gets rid of the 'force build' thing but I'm a little sad that I can't get rid of ParseMode. I think ForSubinclude is probably irrelevant but ForPreload is still needed because the parser has to know not to preload more stuff for those. I'd like to find another way but I can't see anything right now.

Would also like to redo the 'resolve dependencies' stuff on BuildTarget but that seems like a future PR which needs a bit more work. I feel like there's a bit more scope to simplify WaitForPackage but haven't got my head fully around that either.


peterebden:
Okay I've moved this on a bit more; it's working a bit better now but still quite a bit of wonkiness. It feels like it's going to be hard to do the initial piecemeal approach I was thinking of; getting everything to work ends up meaning that I need to move everything to be sync and started losing some of the old concepts.

Still a way away from this being fully ready but it still feels like the right direction to me.

PR #2895, opened 2023-08-18
Fix text_file strip

izissise:
Please doesn't really support tabs,
the strip command ended up removing t and \ from lines beginnning with them.
test_file_content.sh didn't correctly checked multilines.

Everything should be good with these changes


peterebden (approved):
thanks!

PR #2896, opened 2023-08-18
Add stacktrace builtin

izissise:
return a list of dict containing filename, package_name, subrepo and function leading to the current scope where stacktrace was called

This useful when using `breakpoint()` to have a stacktrace. 
Also it can be useful to tag current function name to a build_rule

### DIFF
--- a/docs/lexicon.html
+++ b/docs/lexicon.html
@@ -263,6 +263,16 @@
     <li>
       <span>
         <code class="code"
+          ><span class="fn-name">stacktrace</span
+          ><span class="fn-p">()</span></code
+        >
+        - return a list of dict containing filename, package_name, subrepo and function
+        leading to the current scope where stacktrace was called.
+      </span>
+    </li>
+    <li>
+      <span>
+        <code class="code"
           ><span class="fn-name">json</span><span class="fn-p">(</span
           ><span class="fn-arg">x</span><span class="fn-p">)</span></code
         >
--- a/rules/builtins.build_defs
+++ b/rules/builtins.build_defs
@@ -281,6 +281,10 @@ def breakpoint():
     """Breaks into an interactive debugging session."""
     pass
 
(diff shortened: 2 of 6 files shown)


peterebden:
I'm not convinced we need this. The build language is meant to be minimal and it isn't meant to need to do a lot of introspection - the syntax is Pythonic but it deliberately doesn't have a lot of Python's dynamism.

izissise:
While I understand the intention behind keeping the build language minimal, having stacktrace feature can improve debugging and development efficiency.
It can also be beneficial for newcomers as a way to understand how subinclude and subrepo work.

It is true that it introduce some dynamism, but I don't see any-other alternative to able to a genrule with the current function name.

Regards,
Hugues

izissise:
I still would like this feature, to auto generate rules name from the function name or the N-1 function name.
Is there an alternative way to do this ?

izissise:
Hello, It has been a while, is there still no chance for a similar feature to get merged?

toastwaffle:
@izissise could you explain exactly why you want to generate your target name based on the function name? That seems pretty unusual (especially doing it based on the parent function name). As @peterebden notes above, build definitions should be minimally dynamic.

As for showing a stacktrace when `breakpoint()` is used, that should be feasible to do (in the same way that `fail()` shows a stack trace) without exposing the stack to the build language.

izissise:
> build definitions should be minimally dynamic

I'm challenging this rule, this is not written in the documentation and if I'm going to write all the build definition in a dynamic language like python I'd rather it be dynamic.
An example where I would like this functionality: 
I want to calculate sha1 of file and put the hash in a filegroup with the source
```
def sha1(name, f):
  return genrule(
       name=name,
       tag='sha1',
       src=[f],
  )
```
If I have more hash function I could tag them automatically and even generate a name from the calling function + file.

PR #2901, opened 2023-09-03
Change to use third-party go test parser

peterebden:
I recall us looking at this sort of thing in the past but can't recall why we didn't just use it. I like the idea of us having a bunch less code here.

Have fiddled with a few cases and this _seems_ reasonable. It's definitely broadly working, unsure if maybe there is some edge case that is worse.
One minor change: on test failures we now emit just the failure and not stdout. I think this is better, Go tests tended to feel overly verbose before with the doubled output.

I've also rediscovered what I previously knew about `go test -json`: it is not inherently part of the test, it is a separate binary that still parses the (very similar) output. Seems not worth going through another subprocess given this can just parse it directly.

### DIFF
--- a/src/test/results_test.go
+++ b/src/test/results_test.go
@@ -78,7 +78,7 @@ func TestGoFailedTraceback(t *testing.T) {
 	require.NoError(t, err)
 
 	var failedTC = getFirstFailedTestCase(results)
-	assert.Equal(t, "\tresults_test.go:11: Unable to parse file: EOF", failedTC.Executions[0].Failure.Traceback)
+	assert.Equal(t, "results_test.go:11: Unable to parse file: EOF", failedTC.Executions[0].Failure.Message)
 }
 
 // Go 1.14 changes the ordering of failed messages in Go tests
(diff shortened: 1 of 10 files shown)


peterebden:
Hmmmm so this got tried in #982 and reverted again in #994. Reasons for reverting seem lost to history though.

src/test/test_data/go_test_failure.txt
```diff
@@ -1,9 +1,9 @@
 === RUN TestJSONExpectedFailure-6
---- FAIL: TestJSONExpectedFailure-6 (0.00s)
 	results_test.go:11: Unable to parse file: EOF
+--- FAIL: TestJSONExpectedFailure-6 (0.00s)
```
peterebden, line 3: I'm not fully certain about all the logic here but I don't observe this ordering happening to me when tests fail.

src/test/test_data/go_test_skip.txt
```diff
@@ -5,7 +5,7 @@
 === RUN TestLimitedPrintfAnsi
 --- PASS: TestLimitedPrintfAnsi (0.00s)
 === RUN TestLimitedPrintfAnsiNotCountedWhenReducing
---- SKIP: TestLimitedPrintfAnsiNotCountedWhenReducing (0.00s)
 	interactive_display_test.go:21: haven't written proper support for this yet
+--- SKIP: TestLimitedPrintfAnsiNotCountedWhenReducing (0.00s)
```
peterebden, line 9: Similarly here, the skip message appears between the RUN and SKIP lines

src/test/results_test.go
```diff
@@ -169,11 +169,6 @@ func TestGoIgnoreUnknownOutput(t *testing.T) {
 	assert.Equal(t, 0, results.Skips())
 }
 
-func TestGoFailIfUnknownTestPasses(t *testing.T) {
```
peterebden, line 172: I think go-junit-report handles the case we test here (of results happening out of order). I'm not sure to what degree we need to care.

samwestmoreland (approved):

peterebden:
Alright I'm going to land this and we'll see how we go. If we have to revert it again at least we'll try to remember why this time...

PR #2909, opened 2023-09-29
Check all plugin docs are up-to-date rather than exiting early

samwestmoreland:


docs/plugins/lint_plugin_docs.sh
```diff
@@ -3,28 +3,32 @@
...
-plugins=("python" "java" "go" "cc" "shell" "go-proto" "proto")
-
+PLUGINS=("python" "java" "go" "cc" "shell" "go-proto" "proto")
 URLPREFIX="https://github.com/please-build/"
 
-# loop through plugins and check version
-for PLUGIN in "${plugins[@]}"; do
-    LATEST=$(git ls-remote --tags ${URLPREFIX}${PLUGIN}-rules.git | sed 's/.*\///' | sed '/^v[0-9]\+\.[0-9]\+\.[0-9]\+$/!d' | tail -n 1)
+FAILED=0
+for PLUGIN in "${PLUGINS[@]}"; do
```
goddenrich, line 10: nit: as far as I'm aware convention is local variables generally is lower case. Given we are mixing and matching already it makes sense to standardise to upper

goddenrich (approved):

docs/plugins/lint_plugin_docs.sh
samwestmoreland: Oh ok I was going for upper case everywhere didn't know about the convention. I'll update

samwestmoreland: Yeah you're right - https://unix.stackexchange.com/questions/42847/are-there-naming-conventions-for-variables-in-shell-scripts#:~:text=Yes%2C%20there%20are%20full%20code%20style%20conventions%20for%20bash%2C%20including%20variable%20names.%20For%20example%2C%20here%27s%20Google%27s%20Shell%20Style%20Guide.

Variable Names: Lower-case, with underscores to separate words. Ex: my_variable_name
Constants and Environment Variable Names: All caps, separated with underscores, declared at the top of the file. Ex: MY_CONSTANT

PR #2913, opened 2023-10-01
Change usage of `which` to `command -v`

### LINKED ISSUE #2883: `which` is deprecated on some systems, better use command -v
Somewhere in the `./bootstrap.sh` process, the script or one of the sub-processes calls `which` to determine the location of some binaries.

The `which` command is not commonly installed on systems anymore. On Debian, it is deprecated and results in a warning:
```
/usr/bin/which: this version of `which' is deprecated; use `command -v' in scripts instead.
```

The warnings also shows potential solutions: use `command`; alternatives could be to use `type -P <comman-name>`

peterebden:
Resolves #2883


samwestmoreland (approved):

peterebden:
Oops, lost track of this one. Should be fine to go in though.

Issue #2906, opened 2023-09-26
Please LSP's autocompletion lost first character when inserting into neovim buffers


xulongwu4:
https://github.com/thought-machine/please/assets/16892446/4981e535-3bf4-4827-9119-dd53f6887e67

When using Please LSP with neovim (nvim-cmp or omni-complete), the selected completion item is not inserted correctly. As shown in the gif, I selected the item `check_config`, but got `heck_config` as the result. The first character `c` is lost.

izissise:
I have the same issue using Kate editor.
Looking a bit at the source code, it seems that the dependency https://github.com/tliron/glsp isn't maintained anymore.
One of the issue there mention a compatible library https://github.com/tliron/glsp
Not sure it would solve the issue though

PR #2917, opened 2023-10-09
Fix issue with completing idents in the language server

Tatskaari:
Not totally confident this is 100% correct but in my testing, this resolves the issue. 

Fixes: #2906

### DIFF
--- a/tools/build_langserver/lsp/completion.go
+++ b/tools/build_langserver/lsp/completion.go
@@ -111,7 +111,7 @@ func (h *Handler) completeIdent(doc *doc, s string, line, col int) (*lsp.Complet
 	list := &lsp.CompletionList{}
 	for name, f := range h.builtins {
 		if strings.HasPrefix(name, s) {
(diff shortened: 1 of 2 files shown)


tools/build_langserver/lsp/completion.go
```diff
@@ -111,7 +111,7 @@ func (h *Handler) completeIdent(doc *doc, s string, line, col int) (*lsp.Complet
 	list := &lsp.CompletionList{}
 	for name, f := range h.builtins {
 		if strings.HasPrefix(name, s) {
-			item := completionItem(name, s, line, col)
+			item := completionItem(name, "", line, col)
```
Tatskaari, line 114: If we're completing "check_" to "check_config", name would be "check_config", and s would have been "check_". We would then produce a replacement from "check_" to "config", rather than to the full "check_config" because we strip this prefix. I think this prefix is for when we have something like //foo: -> foo resulting in `//foo:foo`,

samwestmoreland (approved):

PR #2918, opened 2023-10-09
Update plugins

Tatskaari:


plugins/BUILD
```diff
@@ -1,7 +1,7 @@
 plugin_repo(
     name = "go",
     plugin = "go-rules",
-    revision = "v1.8.0",
+    revision = "v1.9.1",
```
peterebden, line 4: 1.9.2 is the latest now...

peterebden (approved):

PR #2921, opened 2023-10-11
format: use character semantics for length and index operations on strings

### LINKED ISSUE #2920: `len(str)` returns the number of bytes in the string, not the number of characters
MWE:

```
log.fatal(len("a"))  # [1]
log.fatal(len("世")) # [3]
```

This was a bit surprising to me, because ASP otherwise doesn't expose any byte-based semantics (it doesn't have the `bytes` data type of Starlark, for example), and also because in Python `len(str)` is the number of characters in the given string:

```
Python 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> print(len("a"))
1
>>> print(len("世"))
1
>>>
```

which makes sense, because in Python a string is conceptually a character sequence.

It'd be straightforward to change this behaviour, but are enough people relying on finding out the byte length of a given string containing high-end UTF-8 characters with `len()` that modifying it would represent a breaking change to Please?

chrisnovakovic:
Conceptually, strings (as in Python) are sequences of characters. Use character-based semantics when computing the length of a string or looking up what is in a string at a particular offset.

This more closely matches Python's behaviour for these operations:

```
Python 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> print(len("line"))
4
>>> print(len("línea"))
6
>>> print(len("линия"))
5
>>> print(len("خط"))
2
```

Fixes #2920.

### DIFF
--- a/src/parse/asp/builtins.go
+++ b/src/parse/asp/builtins.go
@@ -397,7 +397,7 @@ func objLen(obj pyObject) pyInt {
 	case pyFrozenDict:
 		return pyInt(len(t.pyDict))
 	case pyString:
-		return pyInt(len(t))
+		return pyInt(len([]rune(t)))
 	}
 	panic("object of type " + obj.Type() + " has no len()")
 }
--- a/src/parse/asp/interpreter_test.go
+++ b/src/parse/asp/interpreter_test.go
@@ -342,6 +342,14 @@ func TestInterpreterLen(t *testing.T) {
 	s, err := parseFile("src/parse/asp/test_data/interpreter/len.build")
 	assert.NoError(t, err)
 	assert.EqualValues(t, "sync", s.Lookup("y"))
+	assert.EqualValues(t, 4, s.Lookup("l1"))
+	assert.EqualValues(t, "l", s.Lookup("c1"))
+	assert.EqualValues(t, 6, s.Lookup("l2"))
+	assert.EqualValues(t, "n", s.Lookup("c2"))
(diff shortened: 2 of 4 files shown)


Tatskaari (approved):
I think this makes sense. A bit confused about this though:
```
>>> print(len("línea"))
6
```

Shouldn't that return 5 now?

chrisnovakovic:
> I think this makes sense. A bit confused about this though:
> 
> ```
> >>> print(len("línea"))
> 6
> ```
> 
> Shouldn't that return 5 now?

That string is actually six Unicode code points:

1. U+006C : LATIN SMALL LETTER L
2. U+0069 : LATIN SMALL LETTER I
3. U+0301 : COMBINING ACUTE ACCENT {stress mark; Greek oxia, tonos}
4. U+006E : LATIN SMALL LETTER N
5. U+0065 : LATIN SMALL LETTER E
6. U+0061 : LATIN SMALL LETTER A

Code point 3 modifies the appearance of code point 2. To confuse things further, normalising that string will result in code points 2 and 3 being replaced with a single code point, U+00ED (LATIN SMALL LETTER I WITH ACUTE). We could make `print(len("línea"))` return `5`, but we'd have to normalise every string in order to do so, which would be really slow for little benefit - we're probably better off defining "a character" as "a Unicode code point" like Python does.

PR #2925, opened 2023-10-13
Fix reduce builtin

erik-adelbert:
#2880 introduced a buggy reduce builtin which operates on the given list from right to left. So this:
```
reduce(lambda x, y: ''.join((x, y)), ["a", "b", "c", "d", "e"])
```
incorrectly returns "edcba" instead of "abcde".

The proposed patches hopefully solve this bug.

### DIFF
--- a/docs/lexicon.html
+++ b/docs/lexicon.html
@@ -186,11 +186,11 @@
       <span>
         <code class="code"
           ><span class="fn-name">reduce</span><span class="fn-p">(</span
-          ><span class="fn-arg">function</span>, <span class="fn-arg">seq</span
+          ><span class="fn-arg">reducer</span>, <span class="fn-arg">seq</span
           >, <span class="fn-arg">initializer=None</span
           ><span class="fn-p">)</span></code
         >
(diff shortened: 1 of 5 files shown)


docs/lexicon.html
```diff
@@ -186,11 +186,11 @@ <h2 class="title-2" id="python-builtins">
       <span>
         <code class="code"
           ><span class="fn-name">reduce</span><span class="fn-p">(</span
-          ><span class="fn-arg">function</span>, <span class="fn-arg">seq</span
+          ><span class="fn-arg">reducer</span>, <span class="fn-arg">seq</span
```
Tatskaari, line 189: This was copied from here:
https://docs.python.org/3/library/functools.html#functools.reduce

Looks like you've changed the arg name as well which makes it inconsistent with Python if I understand correctly. Should we switch this back?

Tatskaari (commented):
Thanks! This definitely wasn't intentional. Looks like the Python one is left associative which is inline with this change.

docs/lexicon.html
erik-adelbert: https://github.com/thought-machine/please/blob/8d9ffe8e07e26237f9738c566051d00f953109fd/docs/lexicon.html#L177-L181

PR #2926, opened 2023-10-15
Use errgroup limiter where appropriate

peterebden:
errgroup now has one (well it has for a while), use that where possible


samwestmoreland (approved):
Nice!

PR #2932, opened 2023-10-25
keep going flag

goddenrich:
In this PR we introduce a flag --keep_going which will allow us to continue as far as we can with tasks. This can be useful if we want to collect all build failures for reporting at once rather than just the first one encountered.

This doesn't really work with our current shell output. currently we stop early with our output if we have build failures. I took a quick look and it wasn't very easy to change the shell output to display test results after failing builds so I intend to resolve this in a separate PR.


test/keep_going/BUILD
```diff
@@ -0,0 +1,9 @@
+subinclude("//test/build_defs")
+
+please_repo_e2e_test(
+    name = "build_all",
+    expected_failure = True,
```
goddenrich, line 5: It would be good if we could test that we built fail, fail2, pass, dep_pass and fail_test rather than just we failed at some point

Tatskaari (approved):
LGTM! Might be worth having a pre-release for this change so we can test it in core3 before making it generally available.

PR #2942, opened 2023-11-02
Upgrade original targets

peterebden:
We've found cases where this can be notably slow when piping very large numbers of targets into plz.  The code there is pretty old - when it was written we didn't have repos with hundreds of thousands of targets or remote execution with hundreds of workers, so we never really noticed.

This upgrades the implementation to be more efficient rather than running over a huge slice again and again.

I've tested using the profiler locally and things look much more sensible after this (i.e. not all the time is going into `OriginalTarget` functions any more)


src/core/state_test.go
```diff
@@ -59,7 +59,8 @@ func TestExpandVisibleOriginalTargets(t *testing.T) {
 
 func TestExpandOriginalSubLabels(t *testing.T) {
 	state := NewDefaultBuildState()
-	state.AddOriginalTarget(BuildLabel{PackageName: "src/core", Name: "..."}, true)
```
peterebden, line 62: I don't think we ever actually needed to support this. The new implementation just panics but things like `plz test //...` are still fine.

ChangeLog
```diff
@@ -11,6 +11,7 @@ Version 17.4.0
     * Add all commit date formats to supported git_show() format verbs (#2930)
     * Fix reduce builtin (#2925)
     * Fix issue with completing idents in the language server (#2917)
+	* Improve performance when a large number of input targets are supplied (#2942)
```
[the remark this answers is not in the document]
peterebden: M-x untabify

src/core/state.go
```diff
@@ -423,14 +422,10 @@ func (state *BuildState) IsOriginalTarget(target *BuildTarget) bool {
...
-	defer state.progress.originalTargetMutex.Unlock()
-	for _, original := range state.progress.originalTargets {
-		if original == target.Label || (!exact && original.IsAllTargets() && original.PackageName == target.Label.PackageName && state.ShouldInclude(target)) {
-			return true
-		}
+	if exact {
+		return state.progress.originalTargets.MatchExact(target.Label)
 	}
-	return false
+	return state.progress.originalTargets.Match(target.Label) && state.ShouldInclude(target)
```
Tatskaari: This logic isn't quite right. I think what we want is:

```
func (state *BuildState) isOriginalTarget(target *BuildTarget, exactOnly bool) bool {
    partial, exact := state.progress.originalTargets.Match(target.Label)
    return exact || (!exactOnly && partial && state.ShouldInclude(target)) 
}
```

Tatskaari (changes requested):

Tatskaari (approved):
Wow okay slowpoke

PR #2947, opened 2023-11-07
Wildcard regex matching for includes

isobelormiston:
This adds regex-based wildcard matching to the 'include' flag, so that labels can be matched more flexibly. 

Unfortunately, this adds a little breaking change as matching on just `*` will no longer work as before. We could either add a hack to this version so that a single * at the end of the pattern matches like it did before, or we should really wait until the next major release :(

### DIFF
--- a/src/core/build_target.go
+++ b/src/core/build_target.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


peterebden (changes requested):
I'm afraid this is too slow. This is going to recompile a regex every time this is called, which I think will happen for every label on every target that's built, at least once a run.

isobelormiston:
Oh, that's a shame :(

peterebden:
It's possible we could look at precompiling the regex (I guess it's the thing someone passes as a flag so there won't be many of them). That would help although I'm not certain how much it's worth trying to work through the compatibility issues you note.

PR #2949, opened 2023-11-10
Implement a remote FS over the directory proto

Tatskaari:
This PR is being broken down into separate PRs

### DIFF
--- a/src/remote/fs/fs.go
+++ b/src/remote/fs/fs.go
@@ -4,6 +4,7 @@ package fs
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	iofs "io/fs"
 	"os"
@@ -22,24 +23,25 @@ type Client interface {
 
 // fs is an io/fs.FS implemented on top of a REAPI directory. This will download files on demand, as they are needed.
 type fs struct {
-	c   Client
-	dir *pb.Directory
+	c    Client
+	root *pb.Directory
 }
 
 // New creates a new filesystem on top of the given proto, using client to download files on demand.
 func New(c Client, root *pb.Directory) iofs.FS {
 	return &fs{
-		c:   c,
-		dir: root,
+		c:    c,
+		root: root,
 	}
 }
 
 // Open opens the file with the given name
 func (fs *fs) Open(name string) (iofs.File, error) {
-	if len(fs.dir.Symlinks) > 0 {
-		panic("not implemented yet")
-	}
-	for _, f := range fs.dir.Files {
+	return fs.open(".", name, fs.root)
+}
+
+func (fs *fs) open(path, name string, wd *pb.Directory) (iofs.File, error) {
+	for _, f := range wd.Files {
 		if f.Name == name {
 			d, err := digest.NewFromProto(f.Digest)
 			if err != nil {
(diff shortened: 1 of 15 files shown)


src/remote/fs/fs.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/digest"
+	pb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
+	"google.golang.org/protobuf/proto"
+)
+
+type Client interface {
+	ReadBlob(ctx context.Context, d digest.Digest) ([]byte, *client.MovedBytesMetadata, error)
+}
+
+// fs is an io/fs.FS implemented on top of a REAPI directory. This will download files on demand, as they are needed.
```
peterebden, line 24: super nit: `on demand, as they are needed` is redundant

src/remote/fs/fs.go
```diff
@@ -0,0 +1,210 @@
...
+func (fs *fs) open(path, name string, wd *pb.Directory) (iofs.File, error) {
+	name, rest, hasToBeDir := strings.Cut(name, string(filepath.Separator))
+	// Must be a dodgy symlink that goes past our root.
+	if name == ".." || name == "." {
+		return nil, os.ErrNotExist
+	}
+
+	for _, d := range wd.Directories {
+		if d.Name == name {
+			dg, err := digest.NewFromProto(d.Digest)
```
peterebden, line 52: FWIW you can probably get away with `NewFromProtoUnvalidated`, we shouldn't ever have invalid digest protos hanging around the place. Also fine as is tho

src/remote/fs/fs.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func (fs *fs) open(path, name string, wd *pb.Directory) (iofs.File, error) {
+	name, rest, hasToBeDir := strings.Cut(name, string(filepath.Separator))
+	// Must be a dodgy symlink that goes past our root.
+	if name == ".." || name == "." {
+		return nil, os.ErrNotExist
+	}
+
+	for _, d := range wd.Directories {
```
peterebden, line 50: You probably want to use `GetTree` instead of walking every directory one level at a time? It will be a bit slower up front but should overall be faster by using one RPC to fetch the whole tree instead of many.

src/remote/fs/fs.go
```diff
@@ -0,0 +1,210 @@
...
+					name:    f.Name,
+				},
+			}, nil
+		}
+	}
+
+	for _, l := range wd.Symlinks {
+		if l.Name == name {
+			if filepath.IsAbs(l.Target) {
+				// The doc comments suggest that sometimes this is supported. I'm not sure how this would be useful in
```
peterebden, line 110: I'm not sure that we ever really allow this. IIRC there are some flags about whether the server supports/allows it but it seems like a Bad Idea to me

peterebden (commented):
Looks good!

src/remote/fs/fs.go
Tatskaari: Cool cool, I will update the comment.

src/remote/fs/fs.go
Tatskaari: I think a lot of the time, I'm going to be downloading just one file in the whole directory tree (i.e. the BUILD file we're parsing). How does `GetTree` work? Doesn't that download the whole thing in one go?

Tatskaari: Oh, GetTree is done server side? That should be faster then!

Tatskaari:
This is being broken out into smaller PRs

Issue #2927, opened 2023-10-16
Add `--out` flag to `plz build`


Tatskaari:
It would be useful to have a `--out` flag e.g. `plz build --out ./dir //src:please` to copy `plz-out/bin/src/please` to `./dir/please`. 

When set, the [build](https://github.com/thought-machine/please/blob/master/src/please.go#L457) subcommand would loop through all the original targets i.e. `state.ExpandOriginalLabels()`. copying their outputs to the given directory:

```golang
for _, target := range state.ExpandOriginalLabels() {
	outs := state.Graph.TargetOrDie(target).FullOutputs()
        // copy the outs
}
```

peterebden:
I might argue for the name being `--out_dir` to be clearer about what it'll do. Like the idea otherwise though!

DefinitelyNotBen:
Hey, happy to work on this if you can assign it to me

Tatskaari:
> Hey, happy to work on this if you can assign it to me

I'm afraid I can't assign it to you if you're not part of the org, but please do have a crack!

samwestmoreland:
Closed by #2950

PR #2950, opened 2023-11-13
add: build out_dir flag for outputting to specified directory

DefinitelyNotBen:
Adds an output flag to build so you can place the outputs into a specified local directory. Mentioned in issue #2927


src/please.go
```diff
@@ -456,6 +457,19 @@ var opts struct {
...
+					for _, out := range outs {
+						err := fs.CopyFile(out, opts.Build.OutDir+"/"+out, 777)
+						if err != nil {
+							log.Fatalf("Failed to output build to provided directory: %s", err)
+						}
+					}
+				}
+			}
+		}
 		return toExitCode(success, state)
```
Tatskaari, line 473: Small one on code style, in Go people generally prefer to return early rather than nesting the main logic:
```
if !success {
}
// Logic here so it's not nested another level 
return 0
```

You _could_ arguably also do this for `if opts.Build.OutDir != "" {}` to reduce the nesting there too, but that's more personal preference.

src/please.go
```diff
@@ -456,6 +457,19 @@ var opts struct {
 var buildFunctions = map[string]func() int{
 	"build": func() int {
 		success, state := runBuild(opts.Build.Args.Targets, true, false, false)
+		if success {
+			if opts.Build.OutDir != "" {
+				for _, target := range state.ExpandOriginalLabels() {
+					outs := state.Graph.TargetOrDie(target).FullOutputs()
+					for _, out := range outs {
+						err := fs.CopyFile(out, opts.Build.OutDir+"/"+out, 777)
```
Tatskaari, line 465: Hmm, 777 isn't good. That would give everybody permissions to use this file, and make them all executable. We probably should stat the `out` to get it's existing permissions, using [os.Lstat()](https://pkg.go.dev/os#Lstat).

Tatskaari (commented):
Looks good! We'll have to fix the file permissions but other than that I think it can land. Let me know if you want to give adding an e2e test for this a crack.

src/please.go
```diff
@@ -457,20 +457,24 @@ var opts struct {
...
+		}
+		for _, target := range state.ExpandOriginalLabels() {
+			outs := state.Graph.TargetOrDie(target).FullOutputs()
+			for _, out := range outs {
+
+				fm, err := os.Lstat(out)
+				if err != nil {
+					log.Fatalf("Failed to get file mode on build output files: %s", err)
+				}
+				err = fs.CopyFile(out, opts.Build.OutDir+"/"+out, fm.Mode())
```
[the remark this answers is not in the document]
DefinitelyNotBen: Done!

Tatskaari (approved):
Nice on Ben or whoever you are

PR #2952, opened 2023-11-14
Make config loading use a virtual filesystem

Tatskaari:
This will be useful later on when we allow subrepos to have a filesystem based on the REAPI. We can plug in the remote FS in here so we can resolve config files without downloading the subrepo sources.

### DIFF
--- a/src/core/utils.go
+++ b/src/core/utils.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/sha1"
 	"fmt"
+	iofs "io/fs"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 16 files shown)


src/core/utils.go
```diff
@@ -36,6 +37,17 @@ func FindRepoRoot() bool {
...
+type osFS func(name string) (*os.File, error)
+
+func (r osFS) Open(name string) (iofs.File, error) {
+	return r(name)
+}
+
+// HostFS returns an io/fs.FS that behaves the same as the host OS i.e. the same way os.Open works.
+func HostFS() iofs.FS {
+	return osFS(os.Open)
+}
```
Tatskaari, line 49: Not sure why this doesn't exist in the go SDK tbh...

samwestmoreland (approved):

PR #2953, opened 2023-11-14
Make fs.Globber take in an iofs.FS

Tatskaari:


samwestmoreland (approved):
Noice, looks good to me

PR #2960, opened 2023-11-15
Use the subrepo's FS for reading config and globbing

Tatskaari:


src/core/state.go
```diff
@@ -255,7 +255,7 @@ func (state *BuildState) Initialise(subrepo *Subrepo) (err error) {
 		// handled for us already in plz.go
 		if state.CurrentSubrepo != "" {
 			state.RepoConfig = &Configuration{}
-			err = readConfigFilesInto(state.RepoConfig, append(subrepo.AdditionalConfigFiles, filepath.Join(subrepo.Root, ".plzconfig")))
+			err = readSubrepoConfig(state.RepoConfig, subrepo)
```
samwestmoreland, line 258: nit: Could we do an `if err` here?

samwestmoreland (approved):
Nice!

Tatskaari:
> Oh looks like globbing might have broken? Getting a nil pointer dereference somwhere

We were creating a dirfs from the string "" for architecture subrepos. This is explicitly not allowed by the documentation. I think we just need to handle this in the subrepo constructor.

PR #2965, opened 2023-11-16
Use the remote FS to handle subrepos

Tatskaari:
The final PR in this series: injects the remote FS when we're building a subrepo for remote execution. This results in far less data being downloaded by Please which can really help when using remote execution with 1000s of CI workers.

### DIFF
--- a/src/remote/action.go
+++ b/src/remote/action.go
@@ -262,11 +262,11 @@ func (c *Client) uploadInputDir(ch chan<- *uploadinfo.Entry, target *core.BuildT
 			for _, p := range i.Paths(c.state.Graph) {
 				subrepoPath, err := filepath.Rel(target.Subrepo.PackageRoot, p)
 				if err != nil {
-					return nil, fmt.Errorf("failed to ")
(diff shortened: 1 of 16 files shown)


src/remote/action.go
```diff
@@ -257,6 +258,30 @@ func (c *Client) uploadInputDir(ch chan<- *uploadinfo.Entry, target *core.BuildT
...
+					dir.Files = append(dir.Files, fileNode)
+				}
+				if dirNode != nil {
+					dir.Directories = append(dir.Directories, dirNode)
+				}
+				if symlinkNode != nil {
+					dir.Symlinks = append(dir.Symlinks, symlinkNode)
+				}
+			}
+			continue
```
samwestmoreland, line 283: Do we need this continue here?

samwestmoreland (approved):

src/remote/action.go
Tatskaari: Yup otherwise it will try and add this via `uploadInput()` on line 285.

PR #2970, opened 2023-11-20
Redact secret env vars from test step logging

rhysm:


Tatskaari (approved):
Nice! Thanks for the PR :)

PR #2980, opened 2023-11-28
Improve comments around checkSubrepo and remove superfluous check

Tatskaari:
We're encountering a lot of bugs in this area. I thought it would be good to document this prickly code to make sure I really get it, and make it easier for others to understand as well.

### DIFF
--- a/src/parse/parse_step.go
+++ b/src/parse/parse_step.go
@@ -63,7 +63,7 @@ func parse(state *core.BuildState, label, dependent core.BuildLabel, mode core.P
 	state.LogParseResult(label, core.PackageParsing, "Parsing...")
 
 	if subrepo != nil && subrepo.Target != nil {
-		// We have got the definition of the subrepo but it depends on something, make sure that has been built.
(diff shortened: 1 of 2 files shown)


src/parse/parse_step.go
```diff
@@ -83,39 +83,49 @@ func parse(state *core.BuildState, label, dependent core.BuildLabel, mode core.P
...
 	if label.Subrepo == "" {
 		return nil, nil
-	} else if subrepo := state.Graph.Subrepo(label.Subrepo); subrepo != nil {
-		return subrepo, nil
 	}
 
-	sl := label.SubrepoLabel(state)
-
-	// Local subincludes are when we subinclude from a subrepo defined in the current package
-	localSubinclude := label.Subrepo == dependent.Subrepo && label.PackageName == dependent.PackageName && mode.IsForSubinclude()
```
Tatskaari, line 103: This check is slightly different now. We no longer check the parse mode but if my reasoning is right, we shouldn't check the parse mode. If we were to continue under any circumstance, we would lock up because we're parsing the same package.

PR #2982, opened 2023-11-29
Add debug logs that mgiht help with lockups

Tatskaari:


src/core/state.go
```diff
@@ -1136,7 +1150,7 @@ func (state *BuildState) queueTargetAsync(target *BuildTarget, forceBuild, build
 		// Wait for these targets to actually build.
 		if building {
 			for _, t := range target.Dependencies() {
-				t.WaitForBuild()
+				t.WaitForBuild(target.Label)
```
samwestmoreland, line 1153: What's the reason for this change? (Just cuz it sticks out as not just a debugging change)

samwestmoreland (approved):

src/core/state.go
Tatskaari: I need the dependency information so I can add it to the log message

PR #2984, opened 2023-11-29
set remote url empty for all profiles except remote ones

goddenrich:
setting the plzconfig so we don't use remote execution to build this even if set on the system level. Also setting the value in all e2e tests


Tatskaari:
Why do we need this in all the profiles? Isn't just `.plzconfig` enough?

goddenrich:
> Why do we need this in all the profiles? Isn't just `.plzconfig` enough?

The tests use remote execution if not

PR #2985, opened 2023-11-29
rules: allow `genrule`'s `data` parameter to be a dict

chrisnovakovic:
This seems like an arbitrary restriction at the moment - `build_rule` allows `data` to be of type dict, and `gentest` is already passing through dicts too.


peterebden (approved):
Thanks!

PR #2989, opened 2023-12-01
Put test run number on BuildResult

peterebden:
Fixes running with `plz test -n X`; previously all the updates got indexed against a single output line because they're indexed only by the build label. This adds the run number so you get multiple lines again.


peterebden:
Hm does have an issue with remote execution which loses the run number & logs statuses without it... Not sure there is any good way of fixing that without plumbing this all through remote execution as well

peterebden:
Okay I've just bitten the bullet and done that. It's a bit fiddly since remote execution shares a lot of code which now all needs to be aware of `run` but it should be fine.

PR #2993, opened 2023-12-14
Add `add_entry_point` and `get_entry_points` built-in functions

chrisnovakovic:
`get_entry_points` enables the entry points for a build target to be looked up dynamically. `add_entry_point` enables entry points to be defined on build targets dynamically, e.g. in post-build functions.


peterebden (approved):
I think this makes sense. I don't completely love doing `get` dynamically like this but I guess it's equivalent to what we do with commands.

PR #2997, opened 2023-12-15
Add a check for duplicate versions in ChangeLog

samwestmoreland:


peterebden:
We could probably express this as an in-repo test rather than a separate Github Action?

PR #3006, opened 2023-12-27
Limit rate of open files

### LINKED ISSUE #2998: Get parser to provide line number and source file etc. with errors like "Non-iterable type %s; must be a list"
From a brief look at the code, I think this ought to be possible as we know what Expression we're parsing at the time of the error. It's just a question of whether we can get a FilePosition from the Expression's Pos.

peterebden:
This should fix 'too many open files' errors like https://app.circleci.com/pipelines/github/thought-machine/please/6351/workflows/b2452f66-0d80-4d03-9243-100d2ceab3c2/jobs/49808
(Surprised we haven't seen this internally tbh, it seems like it wouldn't be _that_ hard to reach)

The file-opening logic prior to #2961 was inside the Parser which has a limiter of how many files to actively parse at once. In that PR it was moved outside so it's possible for the number of open files to grow ahead of the rate they can be parsed at.
I don't want to introduce _another_ limiter so this pushes the opening back within the parser.

There are a couple of interesting edge cases where we're still reading off the local filesystem (e.g. subincludes); I've not attempted to change that here but we might want to think about it.

Seems to fix #2998 too.


Tatskaari (approved):
Ah nice one!

PR #3007, opened 2023-12-27
Improve error messages when we depend on a subrepo that doesn't exist

Tatskaari:
This PR make please handle a missing BUILD file in the same way as parsing a BUILD file that doesn't contain the subrepo definition, when parsing the subrepo package. This code is a little prickly so I'll do my best to explain. 

We used to get really confusing error messages when a target depends on a target in a non-existant subrepo. When we call `parseSubrepoPackage()`, we pass in the the the non-existant target as the dependant on the subrepo label i.e. `///third_party/go/golang.org_x_net//trace:trace` depends on it's subrepo label, `//third_party/go:golang.org_x_net`, so is the **dependant** when parsing `third_party/go` to find the subrepo. The dependant is used to give context when an error happens parsing and building that target. This results in a really confusing error message that has nothing to do with the original target that depended on the non-existant subrepo in the first place.

This happens a lot when a `go_repo()` depends on another `go_repo()` that hasn't been added to the BUILD file yet. In this case, we're depending on this from a subrepo e.g. `///third_party/go/google.golang.org_grpc//...`, so the subrepo we're looking for could also be defined in the current subrepo instead i.e. both `third_party/go` and `///third_party/go/google.golang.org_grpc//third_party/go`. We actually successfully parse `third_party/go/BUILD` but don't find the subrepo we're looking for, so we then try to parse this package in the current subrepo i.e. `///third_party/go/google.golang.org_grpc//third_party/go`, which doesn't exist at all.  Instead of successfully parsing the package like before, this results in a hard error, resulting in this nonsensical error message: 
```
Build stopped after 7.19s. 2 targets failed:
    ///third_party/go/golang.org_x_net//trace:trace
///third_party/go/golang.org_x_net//trace:trace depends on ///third_party/go/google.golang.org_grpc//third_party/go:all, but there's no BUILD or BUILD.plz file in third_party/go/
```
This indicates that the problem is with the target `///third_party/go/golang.org_x_net//trace:trace`, however this target doesn't even exist except in the deps list of `///third_party/go/google.golang.org_grpc//:grpc`. 

This PR suppresses the error from trying to parse a subrepo package that doesn't doesn't exist. This means that it behaves the same way as if the package did exist, but didn't define the subrepo we were looking for, resulting in the following error: 

```
Build stopped after 2.81s. 1 target failed:
    ///third_party/go/golang.org_x_net//trace:trace
Subrepo third_party/go/golang.org_x_net is not defined (referenced by ///third_party/go/google.golang.org_grpc//:grpc)
```

reporting correctly where teh subrepo was referenced from.

### DIFF
--- a/src/parse/parse_step.go
+++ b/src/parse/parse_step.go
@@ -6,6 +6,7 @@
 package parse
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	iofs "io/fs"
@@ -18,6 +19,8 @@ import (
 )
 
 var log = logging.Log
+
+var ErrMissingPackage = errors.New("missing package")
 
 // Parse parses the package corresponding to a single build label. The label can be :all to add all targets in a package.
 // It is not an error if the package has already been parsed.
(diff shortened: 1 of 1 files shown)


samwestmoreland (approved):
Nice!

PR #3011, opened 2024-01-03
Add a `--json` flag to `plz query outputs`

toastwaffle:
Produces a map from target name to an array of output files.

Tested using `plz-out/bin/src/please query outputs --json //src:please`

### DIFF
--- a/src/query/outputs.go
+++ b/src/query/outputs.go
@@ -35,11 +35,9 @@ func targetOutputsJSON(graph *core.BuildGraph, labels []core.BuildLabel) {
 			data[label.String()] = append(data[label.String()], filepath.Join(target.OutDir(), out))
 		}
 	}
-	bs, err := json.Marshal(data)
-	if err != nil {
-		log.Fatalf("failed to marshal JSON: %v", err)
-	}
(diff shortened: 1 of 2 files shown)


peterebden:
I don't have any particular problem with this although I do think it feels slightly wonky that only certain subcommands can output in JSON. Having said that, there's no obvious way of generalising it without writing this equivalent for each of them.

src/query/outputs.go
```diff
@@ -1,18 +1,45 @@
...
+
+func targetOutputsJSON(graph *core.BuildGraph, labels []core.BuildLabel) {
+	data := map[string][]string{}
+	for _, label := range labels {
+		target := graph.TargetOrDie(label)
+		for _, out := range target.Outputs() {
+			data[label.String()] = append(data[label.String()], filepath.Join(target.OutDir(), out))
+		}
+	}
+	bs, err := json.Marshal(data)
```
peterebden, line 38: I think this should probably use MarshalIndent so it's a bit more readable (or at least doesn't have million-character-long lines).

Or, better, an Encoder with an indent set so it can stream the writes to stdout.

peterebden (approved):
Thanks!

PR #3012, opened 2024-01-03
Simplify subincludes during reformat

peterebden:
This rewrites repeated calls to `subinclude` into one. See example in test_data.

Have run internally and am happy with the results (although other unrelated fmt issues have arisen...)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
    build-linux:
      working_directory: ~/please
      docker:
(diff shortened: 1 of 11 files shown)


peterebden:
I'll mention explicitly: This requires us to have Go 1.21+ to bootstrap Please now (because of the use of `slices`). I could use `x/exp/slices` but I think 1.21 has been around long enough that we can just make use of it.
(Of course building it using itself still works without any system version so that's fine)

Tatskaari (approved):

PR #3021, opened 2024-01-04
Add an environment variable containing the test run

peterebden:
I don't completely like that they can theoretically now take action based on this, but I think it's a fairly practical way of deduplicating them on the remote execution side.
Arguably there are more elegant ways of tackling that but this will work right now and won't require the backend to be able to separate multiple concurrent executions of the same action.

We could also only do this for remote execution, but I think it's a bad idea to start adding differences between the two (in some ways they already are but c'est la vie)

### DIFF
--- a/src/core/build_env.go
+++ b/src/core/build_env.go
@@ -162,7 +162,7 @@ func TestEnvironment(state *BuildState, target *BuildTarget, testDir string, run
 		// We shouldn't really have specific things like this here, but it really is just easier to set it.
 		"GTEST_OUTPUT=xml:"+resultsFile,
 		"PEX_NOCACHE=true",
-		"_TEST_RUN="+base64.RawStdEncoding.EncodeToString(hash[:12]),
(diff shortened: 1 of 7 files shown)


src/core/build_env.go
```diff
@@ -159,6 +159,7 @@ func TestEnvironment(state *BuildState, target *BuildTarget, testDir string) Bui
 		// We shouldn't really have specific things like this here, but it really is just easier to set it.
 		"GTEST_OUTPUT=xml:"+resultsFile,
 		"PEX_NOCACHE=true",
+		fmt.Sprintf("_TEST_RUN=%d", run),
```
Tatskaari, line 162: Could make this meaningless by hashing it and calling it something different.

Tatskaari (approved):
LGTM one small suggestion if you fancy it

src/core/build_env.go
peterebden: Good idea. Done.
Ideally I'd make it specific to the hash of the build action but I think this is sufficient.

PR #3029, opened 2024-01-09
Make the remotefs cache blobs

Tatskaari:


src/remote/remote.go
```diff
@@ -57,12 +58,13 @@ var remoteCacheReadDuration = metrics.NewHistogram(
...
 // It provides a higher-level interface over the specific RPCs available.
 type Client struct {
-	client      *client.Client
-	fetchClient fpb.FetchClient
-	initOnce    sync.Once
-	state       *core.BuildState
-	err         error // for initialisation
-	instance    string
+	client         *client.Client
+	remoteFSClient remotefs.Client
```
Tatskaari, line 62: Might've been nice to make these guys the same but i think the Client god object from their SDK is a bit too big to wrap up like this.

peterebden (approved):
This isn't threadsafe itself, but I guess that's fine because we control it at a higher level, right?

PR #3031, opened 2024-01-09
Remove `github.com/peterebden/buildtools` as a dependency

chrisnovakovic:
Looks like this isn't used any more (everything ought to be using `github.com/please-build/buildtools` by now).


peterebden (approved):
:sad-salute:

PR #3051, opened 2024-01-22
Remove mention of `--nocache` flag

jkimbo:
Remove the mention of the `--nocache` in the `plz clean` docs because the flag no longer exists.

The `--nocache` flag was removed in v16: https://github.com/thought-machine/please/blob/11052fd323e5a765fbda5e15e44c7d3fcab5f86d/ChangeLog#L776


peterebden (approved):
Thanks!

Issue #3052, opened 2024-01-22
UX optimization: reversed patch == hung fetch...


sean-:
This is embarrassing, and it took entirely too long to figure out, IMO, but I had a custom patch reversed (i.e., `diff -u new_file old_file`).  Behind the scenes `patch` was asking:

```
Reversed (or previously applied) patch detected!  Assume -R? [y]
```

Could the behavior of patch be tweaked to pass the `-f` or `-N` (or maybe `-t`, but I think this would be problematic to debug for different reasons)?  I'd rather have things fail fast than hang on user input and be hung/stalled.  Needless to say, I thought this was a VPN or WiFi issue while flying, not something stuck on user input because it was hung "Fetching".  From a UX perspective, having the output updated to be "Patching" vs "Fetching" would've also been a  sufficient clue.

Feel free to close, but I think this would be an easy UX win when someone does something dumb (i.e., flying and hacking while sleep-deprived).

Tatskaari:
I agree, we shouldn't be hanging on interactive input like this. Do you know what these flags do? I think we essentially want the default behavior (i.e. don't automatically reverse it for us) but when it detects this case, it should just error. From this description here, I assume `-t` would have applied this patch, a bit like if you passed `-R`:
```
  -t  --batch  Ask no questions; skip bad-Prereq patches; assume reversed.
  -f  --force  Like -t, but ignore bad-Prereq patches, and assume unreversed.
```

> From a UX perspective, having the output updated to be "Patching" vs "Fetching" would've also been a sufficient clue.

This is fairly easy to do but has a performance hit as patches are often used to patch large repos. We have to copy the files into a temp directory, and adding more intermediate targets means we have to repeat this step for each target.

sean-:
I'd use `-f`.  If the patch fails to apply, it should be an error (a reversed patch shouldn't magically apply because this is a form of non-determinism).  In my case, the `-f` would've caught this error and it would've been obvious there was a patch problem, not something blocked waiting for input on stdin.

Tatskaari:
Yeah, that sounds like the right thing to do. Which build defs are you experiencing this with? There are a number of places where we patch things in Please.

sean-:
`go_module()`:
```
go_module(
    name = "tengo",
    install = [
        "",
        "parser",
        "stdlib",
        "stdlib/json",
        "token",
    ],
    module = "github.com/d5/tengo/v2",
    patch = "d5-tengo-recover.patch",
    version = "v2.16.1",
[...]

PR #3053, opened 2024-01-23
Pass -f to patch to avoid hanging when patch is reveresed

Tatskaari:
See #3052


sean- (approved):

sean-:
Bump.  When is this going to get merged?  See https://github.com/thought-machine/please/issues/3052 for a new failure mode that is problematic.  I'm trying to convert an existing project to `puku` and am noticing new peculiarities that I'm having to deal with.

PR #3055, opened 2024-01-24
Fix calculation of cache size when using compression

rhysm:
When compression is on cached artifacts are stored as files. When walking the cache directory if we return SkipDir then it only counts the first artifact for a given target and ignores the rest.


peterebden:
Thanks! Sorry this slipped by us for a while there 😬

PR #2990, opened 2023-12-01
Make subinclude take annotated label and list

izissise:
This adds the possibility to subinclude a target named output 

Fix #2939

PR #3080, opened 2024-02-18
Allow type list for subinclude parameter

izissise:
Missing addition to allow to pass list in subinclude parameter

Relate to #2990


samwestmoreland (approved):
Nice spot, thanks!

PR #3099, opened 2024-03-07
upgrade v0 go third_party libraries

goddenrich:


third_party/go/BUILD
```diff
@@ -107,7 +107,7 @@ go_repo(
 
 go_repo(
     module = "github.com/ProtonMail/go-crypto",
-    version = "v0.0.0-20210920135941-2c5829bbf927",
```
peterebden, line 110: Let's be cautious on this one - I recall that it has screwed us before by removing support for some crypto nuance

goddenrich: ok i'll revert it and we can upgrade that one independently

Tatskaari (approved):

PR #3102, opened 2024-03-08
Expand documentation on the .plzconfig.local precedence.

matglas:
Small improvement to the documentation on using a config profile. Something that was not clear was that the `.plzconfig.local` always has precedence over anything set by the arguments. This should become more clear with this small change.

### DIFF
--- a/docs/config.html
+++ b/docs/config.html
@@ -28,13 +28,19 @@
     >
   </li>
   <li>
+    <span>
+      <code class="code">.plzconfig.*</code> read from <code class="code">--profile</code>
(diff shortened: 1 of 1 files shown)


Tatskaari (approved):
Nice! Thanks for that :)

Tatskaari:
>  And let me know if there is anything I can do better if needed for other PRs in the future. 

Looks good from my side! If you're going to contribute a feature, it can be useful to raise an issue first to get feedback on your approach to avoid wasted effort, but otherwise we're pretty easy going about this stuff.

matglas:
Thank you!

PR #3103, opened 2024-03-12
dont parse subrepo package again if the subrepo label subrepo is the same as the dependent subrepo

goddenrich:
Previously we were maybeParseSubrepoPackage for the host repo and if we don't find the subrepo we would check in the dependent label's subrepo. This is unnecessary if the dependent label's subrepo is the same as the host repo.

There is also a bug which this fixes where we indefinitely wait on parsing a subrepo if the subrepo doesn't exist. This change should just be an optimisation but it fixes the bug too.

### DIFF
--- a/docs/BUILD
+++ b/docs/BUILD
@@ -59,7 +59,7 @@ genrule(
 plugins = {
     "python": "v1.5.0",
     "java": "v0.4.0",
-    "go": "v1.16.8",
+    "go": "v1.16.9",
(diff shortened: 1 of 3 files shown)


peterebden (approved):
LGTM. 
I don't have a lot of insight into why it fixes the hang - possibly it makes it parse subrepos less often so that case doesn't come up. @Tatskaari is more familiar with that bit than I am these days though.

PR #3104, opened 2024-03-12
Add gRPC metrics

peterebden:
I'm interested in comparing what we do with remote execution given certain changes. This should be pretty minimal overhead (it's only on if we use rex and we already import all the prometheus & protobuf gubbins anyway)

### DIFF
--- a/src/metrics/prometheus.go
+++ b/src/metrics/prometheus.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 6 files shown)


src/metrics/prometheus.go
```diff
@@ -23,13 +24,12 @@ var registerer = prometheus.WrapRegistererWith(prometheus.Labels{
...
+		var buf strings.Builder
 		for _, fam := range family {
-			for _, metric := range fam.Metric {
-				if metric.Counter != nil {
-					log.Debug("Metric recorded: %s: %0.0f", *fam.Name, *metric.Counter.Value)
-				} else if metric.Gauge != nil {
-					log.Debug("Metric recorded: %s: %0.0f", *fam.Name, *metric.Gauge.Value)
-				}
+			buf.Reset()
+			if _, err := expfmt.MetricFamilyToText(&buf, fam); err == nil {
```
peterebden, line 30: I got some way through implementing label printing myself then realised I was just reimplementing expfmt. Bonus: this now prints all metric types for 'free'

PR #3107, opened 2024-03-14
Remove AddChildDirs calls

peterebden:
I've been noodling around the remote execution bit and am not certain why we need this. I think maybe it's intended for uploading filegroup inputs but so far in (admittedly somewhat empirical) testing it doesn't seem to suffer without.

If it works, this would be a nice boost to the happy path where things don't need to rebuild.


peterebden:
I think I've found an edge case on this; looks like it's related to `output_dirs = ["blah/**"]` :(

PR #3110, opened 2024-03-16
Add buildlabel function

izissise:
This function facilitates the creation
and modification of labels in bulk or for obtaining the full label.

It also enables referencing other packages and subrepos relatively from the current one using '../' resolution.

Examples:
- `buildlabel(target=genrule(name="xxx", ...))` -> `["///subrepo//pkg:xxx"]`
- `buildlabel([":a", ":b"], out="file")` -> `["///subrepo//pkg:a|file", "///subrepo//pkg:b|out"]`
- `buildlabel(["subdir", "../uppkg"], target="build")` -> `["///subrepo//pkg/subdir:build", "///subrepo//uppkg:build"]`
- `vis("*")` -> `["PUBLIC"]`
- `vis("//anotherpkg/...")` -> `["/////anotherpkg/..."]`

More examples can be found in the test file.

### DIFF
--- a/rules/builtins.build_defs
+++ b/rules/builtins.build_defs
@@ -297,3 +297,142 @@ def is_semver(s:str) -> bool:
 
 def semver_check(version:str, constraint:str) -> bool:
     pass
+
+
+# Returns build label
+# https://github.com/thought-machine/please/blob/v17.3.1/src/core/build_label.go#L159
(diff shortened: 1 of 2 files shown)


izissise:
What would be really nice is the addition of a python type `label` that would be returned by `build_rule` this would clearly separate normal strings and label ones.

PR #3111, opened 2024-03-18
Only download child dirs for filegroups where necessary

peterebden:
More selective version of #3107 . It doesn't seem as clean but it deals with some edge cases where that did seem to be necessary.


Tatskaari (approved):
Not sure i really understand but looks reasonable. Might be worth doing a sizeable build on core3?

peterebden:
I've been running this on my machine for some time and haven't seen any issues (I'm pretty sure we won't since it falls back to the old behaviour)

Tatskaari:
Let's get it merged and give it a try then

PR #3115, opened 2024-03-29
[XS] Fix syntax in basics doc

juh2600:
`java_binary`'s `main` arg should  be `main_class` per https://please.build/plugins.html#java; just `main` throws an error about an unknown argument.

### DIFF
--- a/docs/basics.html
+++ b/docs/basics.html
@@ -134,7 +134,7 @@
 
     java_binary(
         name = 'my_binary',
-        main = 'Main.java',
+        main_class = 'Main.java',
         deps = [':my_library'],
     )
 
--- a/docs/basics.html
+++ b/docs/basics.html
@@ -134,7 +134,7 @@
 
     java_binary(
         name = 'my_binary',
-        main_class = 'Main.java',
+        main_class = 'com.example.Main',
         deps = [':my_library'],
     )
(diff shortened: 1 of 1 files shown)


juh2600:
I'm not sure if this is the only change necessary:

```
[I] jtreed ~/s/p/java> cat staque/BUILD
java_binary(
  name = "staque",
  main_class = "Main.java",
)
[I] jtreed ~/s/p/java> plz run //staque:staque
Error: Could not find or load main class Main.java
Caused by: java.lang.ClassNotFoundException: Main.java
```

Changing "Main.java" to just "Main" in BUILD results in a corresponding change in this error message, so I'm definitely doing something else wrong, but I suspect we should be supplying the class name (`"Main"`) rather than the file name (`"Main.java"`). This is also supported by the plugin doc linked above. I just can't confirm that this is right since I haven't got my project working yet (unrelated).

juh2600:
Got the project working and confirmed that `Main.java` should just be `Main`, or perhaps some more illustrative path like `com.example.Main`. I'll mark this PR as a draft and attempt to add that change---still a noob with GitHub's PR system.

Tatskaari (approved):
Thanks! Appreciate the fix :)

PR #3125, opened 2024-04-11
plz gc

peterebden:
With changes from #3124, this identifies some (apparent) garbage that we can remove.

Will see what the test run does but at least `plz build //src:please` seems to work.


goddenrich:
the third_party/go targets removed will just be added back in the next time someone does a puku sync unless they are also removed from the go.mod

peterebden:
Sadly I can't get the go.mod to do anything sensible with `go mod tidy`. I'm not convinced that Go/puku are doing the optimal thing here though; there seems to be a bunch of stuff here that isn't required that they're populating anyway.

PR #3126, opened 2024-04-11
Cache subinclude ASTs per path only

peterebden:
Currently we re-parse all globally preloaded subincludes for every subrepo (this caught me by surprise a bit).
It's not trivial to avoid interpreting them for each one (although I do want us to do better there) but this is reasonably low-hanging fruit: we only need to parse them once each since the AST representation doesn't change.

I'm seeing nearly a 20% CPU drop for parsing an internal repo. Sadly, I don't think this is well exercised by the postmerge performance tests. (Maybe we should try to improve that...)

### DIFF
--- a/src/parse/asp/interpreter.go
+++ b/src/parse/asp/interpreter.go
@@ -19,6 +19,7 @@ type interpreter struct {
 	scope       *scope
 	parser      *Parser
 	subincludes *cmap.Map[string, pyDict]
+	asts        *cmap.Map[string, []*Statement]
 
 	configs      map[*core.BuildState]*pyConfig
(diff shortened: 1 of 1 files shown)


src/parse/asp/interpreter.go
```diff
@@ -38,15 +39,19 @@ func newInterpreter(state *core.BuildState, p *Parser) *interpreter {
 	}
 	// If we're creating an interpreter for a subrepo, we should share the subinclude cache.
```
peterebden, line 40: This isn't actually useful any more, although I haven't experimented with whether it's better or worse if we share / don't share them.
(It obviously is very useful for the asts though)

PR #3138, opened 2024-04-24
Allow 'default' entry in provide map for when there's no match

Tatskaari:
With #3140 this should allow us to default to no dependency when the require can't be satisfied, which would be nice in a number of circumstances.


src/core/build_target.go
```diff
@@ -557,14 +557,15 @@ func (target *BuildTarget) resolveDependencies(graph *BuildGraph, callback func(
...
+func (target *BuildTarget) resolveOneDependency(graph *BuildGraph, dep *depInfo, ffDefaultProvide bool) error {
 	t := graph.WaitForTarget(*dep.declared)
 	if t == nil {
 		return fmt.Errorf("Couldn't find dependency %s", dep.declared)
 	}
 	dep.declared = &t.Label // saves memory by not storing the label twice once resolved
 
-	labels := t.provideFor(target)
+	// TODO(jpoole): shouldn't this use the ProvideFor wrapper that handles resolving to self if there's no match?
+	labels := t.provideFor(target, ffDefaultProvide)
```
Tatskaari, line 568: This probably requires a bit of thought

peterebden (approved):
Makes sense. The plumbing for the feature flag is slightly awkward but I think it's the right thing to do

PR #3139, opened 2024-04-25
Use local paths only for determining changes

peterebden:
I think this is a better version of #3137 

 - If sources are local to the package, we don't need to compare their package part. 
 - If they're off the PATH, we don't need to worry about the PATH changing (that's handled elsewhere) so their name is enough
 - Build labels are excluded from this check in `sourceHash` anyway.

Hence this should be fine, and removes the possibility of them panicking.

### DIFF
--- a/src/query/changes.go
+++ b/src/query/changes.go
@@ -10,8 +10,6 @@ import (
 	"github.com/thought-machine/please/src/core"
 )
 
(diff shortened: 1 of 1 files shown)


Tatskaari (approved):
I think this makes sense :)

PR #3146, opened 2024-05-02
Rework str.format() implementation

peterebden:
It was pretty brute-force before. This reimplements it in a more sensible way that doesn't involve repeatedly replacing (potentially large) strings.

It should be pretty much the same as before except for a few really degenerate edge cases that I don't think we specifically supported - e.g. people building up JSON strings via things like
```
"""
[{
    "key": "{value}",
}]
""".format(value=x)
```
(note that this doesn't work in Python proper either, it chucks a KeyError)

Benchmark results:
```
BenchmarkStrFormat
BenchmarkStrFormat-12            1000000              1497 ns/op             208 B/op          5 allocs/op
BenchmarkStrFormatBig
BenchmarkStrFormatBig-12           32077             33884 ns/op           36587 B/op         11 allocs/op
PASS
```
vs.
```
BenchmarkStrFormat
BenchmarkStrFormat-12       	 2716239	       626.2 ns/op	      64 B/op	       2 allocs/op
BenchmarkStrFormatBig
BenchmarkStrFormatBig-12    	  135216	     10041 ns/op	    7568 B/op	       3 allocs/op
PASS
```
which looks considerably better.

### DIFF
--- a/src/parse/asp/builtins.go
+++ b/src/parse/asp/builtins.go
@@ -609,6 +609,11 @@ func strFormat(s *scope, args []pyObject) pyObject {
 			// Don't interpolate ${X} (but ${{X}} -> ${X})
 			if self[start+1] == '{' {
 				buf.WriteString(self[start+1 : end])
+			} else if start == end-1 {
+				// ${} interpolates as $ + positional arg
+				s.Assert(arg < len(args), "format string specifies at least %d positional arguments, but only %d were supplied", arg, len(args)-1)
+				buf.WriteString(args[arg].String())
+				arg++
 			} else {
 				buf.WriteString(self[start : end+1])
 			}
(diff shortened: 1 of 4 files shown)


goddenrich (approved):

src/parse/asp/builtins.go
```diff
@@ -587,13 +587,49 @@ func strRFind(s *scope, args []pyObject) pyObject {
...
+			// We may want to error here in some future revision
+			buf.WriteString(self)
+			break
+		}
+		buf.WriteString(self[:start])
+		end = start + end
+		if start > 0 && self[start-1] == '$' {
+			// Don't interpolate ${X} (but ${{X}} -> ${X})
+			if self[start+1] == '{' {
+				buf.WriteString(self[start+1 : end])
```
goddenrich, line 611: this could mean
```
${ {X}} -> $ {X}
```
which isn't ideal but unlikely so in the interest of not complicating things more I think its ok especially because in the previous implementation
```
${ {X}} -> ${ {X}
```
which is invalid too

peterebden: Yeah there's a bunch of stuff in there that's a bit gross. This is erring as close as possible to before, but I think we should ban quite a few of these constructions.

PR #3150, opened 2024-05-13
Fix hang on preloading subincludes

peterebden:
Found a  case where it hangs forever.
 - I have subincludes `//a` and `//b`, both of which are preloaded.
 -  `//a` contains `subinclude("//b")`. 
 - `//b` contains a syntax error
We never exit from `RegisterPreloads`; the errgroup has the error from `//b` but is waiting for `//a` to complete, which it never does.

This fixes it by dying, although it's a bit brutal. It would of course be better to have `//a` receive the parse error too...

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -18,7 +18,6 @@ import (
 
 	"github.com/cespare/xxhash/v2"
 	"github.com/zeebo/blake3"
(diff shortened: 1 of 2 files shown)


Tatskaari (approved):
Probably makes sense. can't do much if preloading failes.

PR #3153, opened 2024-05-13
Change to math/rand/v2

peterebden:
We don't use it for anything important but might as well move on to the newer version.
Also bump golangci so it's a version that understands 1.22 stdlib packages.


Tatskaari (approved):

peterebden:
oh we still have builders on 1.21, I'll come back to this later.

PR #3162, opened 2024-05-14
Try disabling caching on golangci-lint action

peterebden:
We seem to have https://github.com/golangci/golangci-lint-action/issues/244 (like everyone else), maybe this will make it better


peterebden:
Cool, this doesn't seem notably slower, but it doesn't have 15,000 lines of nonsense in the output any more.

PR #3167, opened 2024-05-16
Fix list concat overwriting

### LINKED ISSUE #3166: List concatenation bug
MRE:
```
fruit = [
    "apple",
    "banana",
]

fruit_and_veg = fruit + [
    "edamame",
    "fennel",
]

fruit_veg_and_canned_food = fruit_and_veg + [
    "tuna",
    "baked beans",
]

fruit_veg_canned_food_and_sweets = fruit_veg_and_canned_food + [
    "haribo",
]

fruit_veg_canned_food_and_drinks = fruit_veg_and_canned_food + [
    "matcha latte",
]

fail(', '.join(fruit_veg_canned_food_and_sweets))
```

```
apple, banana, edamame, fennel, tuna, baked beans, matcha latte
```

During the assignment/concatenation of "matcha latte" to the last list, the `fruit_veg_canned_food_and_sweets` list gets modified. There would appear to be something going on here with a shared underlying slice but I can't seem to get to the bottom of it. Also I can't seem to reproduce it with fewer concatenations than the above, weirdly, if anyone was wondering why there are so many.

peterebden:
Fixes #3166 

Looks like this is due to the slices overwriting since on append they both think they have extra capacity, but are in fact sharing it. A pure Go example: https://go.dev/play/p/pLl6MlMr3V4

This fixes this particular case; I don't think there are others at present, we seem to construct lists with capacity equal to their final size in other instances.

### DIFF
--- a/src/parse/asp/objects.go
+++ b/src/parse/asp/objects.go
@@ -3,6 +3,7 @@ package asp
 import (
 	"encoding/json"
 	"fmt"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -363,7 +364,7 @@ func (l pyList) Operator(operator Operator, operand pyObject) pyObject {
 			}
 			panic("Cannot add list and " + operand.Type())
 		}
-		return append(l, l2...)
+		return slices.Clip(append(l, l2...))
 	case In, NotIn:
 		for _, item := range l {
 			if item == operand {
@@ -429,7 +430,7 @@ func (l pyList) Freeze() pyObject {
 
 // Repeat returns a copy of this list, repeated n times
 func (l pyList) Repeat(n pyInt) pyList {
-	var ret pyList
+	ret := make(pyList, 0, int(n)*len(l))
 	for i := 0; i < int(n); i++ {
 		ret = append(ret, l...)
 	}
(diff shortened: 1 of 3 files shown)


goddenrich (approved):

samwestmoreland (approved):
Nice! Thanks for the fix 😄

src/parse/asp/objects.go
```diff
@@ -363,7 +364,7 @@ func (l pyList) Operator(operator Operator, operand pyObject) pyObject {
 			}
 			panic("Cannot add list and " + operand.Type())
 		}
-		return append(l, l2...)
+		return slices.Clip(append(l, l2...))
```
HartBlanc, line 367: Do we need to apply this change to the `pyFrozenList` case a few lines above too?

HartBlanc, line 367: Do we need to apply this change to the pyFrozenList case a few lines above too?

HartBlanc (commented):
Do we need to apply the slices.Clip to the pyFrozenList case a few lines above too?

src/parse/asp/objects.go
peterebden: can't hurt, thanks!

PR #3170, opened 2024-05-16
Added tab escape sequence.

### LINKED ISSUE #3163: Support for tab escape charcter missing.
It is possible to create a string with a newline using `\n` in please strings, but not for a tab character with `\t`.
It is however still possible to use a literal tab character in string declarations to get the same effect, but saying that that is unreadable is putting it mildly.

For instance `stringWithTabs.lstrip(" 	")` is the current workaround to the missing tab escape character, which would much more legibly read `stringWithTabs.lstrip(" \t")` with the escape character.

cemeceme:
Fixes #3163.


samwestmoreland (approved):
Makes sense to me, thanks!

PR #3173, opened 2024-05-17
Remove mention of pleasings from plz init message

### LINKED ISSUE #3155: plz init still suggests getting auxiliary rules from the Pleasings repo
It should suggest looking in the please-rules repo for plugins.

Just putting this here so we don't forget.

samwestmoreland:
Closes #3155


peterebden (approved):
Nice!

Issue #3168, opened 2024-05-16
Race condition when depending on rules created in post_build.


cemeceme:
Here is an example of a BUILD file with a post_build function adding dependencies:
```
def post_func(ruleName, cmdOutput):
    fileRule = text_file(name="postFile", content = "Some content", out="postFile.txt")
    add_dep(ruleName, fileRule, True)

genrule(name="postRule", cmd="", post_build=post_func)

genrule(name="postRuleConsumer", cmd="", deps=[":postRule"])
```

Running this with `plz build //:postRuleConsumer` will result in an output of `cannot calculate hash for plz-out/gen/postFile.txt: file does not exist`.

However, if you somehow force `post_func` to run without dependents (in this case `postRuleConsumer`), for example by calling `//:postRule` on its own, then `postFile.txt` will get created.
Once the file is created, subsequent runs will start to succeed and have the file available until the next `plz clean` call.

What seems to be happening, is that the commands of rules in post_builds aren't getting executed. I have confirmed this by changing the `text_file` rule into `genrule(name="postFile", cmd = "sleep 10; touch postFile.txt", out="postFile.txt")`. When `//:postRule` is run on its own, it will take 10 seconds to run, however when run with `//:postRuleConsumer` the command will immediately finish, either with the same error, or by succeeding if `//:postRule` was run on its own beforehand.

cemeceme:
I have been looking into the code base to track down this issue, and found that it was actually a race condition.
It turns out that if there is another rule that can block the dependent rule until the post build completes, then everything works as expected. Otherwise, the same issue as mentioned occurs.

Here is an example build script that will succeed:
```
def post_func(ruleName, cmdOutput):
    fileRule = genrule(name="postFile", cmd = "sleep 10; touch postFile.txt", out="postFile.txt")
    add_dep(ruleName, fileRule, True)

genrule(name="postRule", cmd="", post_build=post_func)

genrule(name="waitRule", cmd="sleep 11; touch waitFile.txt", out="waitFile.txt")

genrule(name="postRuleConsumer", cmd="", deps=[":postRule", ":waitRule"])
```

Note the `sleep 10` in the file rule and `sleep 11` in the wait rule. As long as the "postFile" rule in this example completes before the "waitRule" everything works.

PR #3174, opened 2024-05-22
Added dependency check after post_build step.

cemeceme:
Fixes the race condition for dependencies created within post_build steps #3168.

The issue is that a target will mark itself complete, even though the post_build step could have added additional dependencies to it. This meant that any dependents would start running, without waiting for the post_build dependencies to complete, resulting in a race.

The added code will now re-check all the dependencies of a target, after its post_build step has completed. This way, additional dependencies will properly get waited on, avoiding the race.

I would like to note, however, that the addition is essentially a copy of some code in [queueTargetAsync](https://github.com/thought-machine/please/blob/03ae086cd5c75bb1f42f7e823b9d61b81809a163/src/core/state.go#L1176), where the initial dependency checks occur. In this case, this is somewhat duplicating the dependency checks and also re-checking already resolved dependencies, even if a post_build step doesn't change the dependencies.

### DIFF
--- a/src/build/build_step.go
+++ b/src/build/build_step.go
@@ -78,7 +78,6 @@ func Build(state *core.BuildState, target *core.BuildTarget, remote bool) {
 	} else {
 		successfulLocalTargetBuildDuration.Observe(float64(time.Since(start).Milliseconds()))
 	}
-
 	// Mark the target as having finished building.
 	target.FinishBuild()
 	if target.IsTest() && state.NeedTests && state.IsOriginalTarget(target) {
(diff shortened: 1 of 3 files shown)


cemeceme:
Fixes https://github.com/thought-machine/please/issues/3168.
Just wanted to make sure that the issue was actually linked.

goddenrich:
Thanks for this. Im taking a look at it now and while it initially seems reasonable there are a few things I want to check before approving.

cemeceme:
Any updates on this? I would love to see the issue this fixes get resolved.

PR #3177, opened 2024-06-13
log aquiring target lock

goddenrich:
Previously we would log we are downloading when we are actually waiting to acquire a target lock. This could cause people to wait indefinitely thinking downloads are happening when another process has the target lock.


Tatskaari (approved):
Spelling mistake aside  ;P

PR #3188, opened 2024-07-02
Add new field allowing tests to opt out of creating coverage files

peterebden:
This has historically been fairly okay because we treat the coverage files as optional. However it's kind of bad with remote execution, which has no concept of optional outputs, so we have to ask for it and the remote will then not be able to cache the action result.

I thought about just having tests do `touch $COVERAGE_FILE` but then we print a non-fatal but annoying error. This seems cleaner, although it is yet another target field...


rules/misc_rules.build_defs
```diff
@@ -184,6 +184,10 @@ def gentest(name:str, test_cmd:str|list|dict, labels:list&features&tags=None, cm
       flaky (bool | int): If true the test will be marked as flaky and automatically retried.
       no_test_output (bool): If true the test is not expected to write any output results, it will
                              pass if the command exits with 0 and fail otherwise.
+      no_test_coverage (bool): If true, the test is not expected to produce any coverage output under
+                               `plz cover`. By default this is false, but coverage is optional, so t
```
toastwaffle, line 188: Orphaned `t` at end of line

rules/misc_rules.build_defs
```diff
@@ -184,6 +184,10 @@ def gentest(name:str, test_cmd:str|list|dict, labels:list&features&tags=None, cm
...
+                               Please will not fail if not output; this is mostly relevant for remote
```
toastwaffle, line 189: "Please will not fail if no coverage report is output"

src/test/test_step.go
```diff
@@ -504,7 +504,10 @@ func parseTestOutput(stdout string, stderr string, runError error, duration time
 }
 
 // Parses the coverage output for a single target.
-func parseCoverageFile(target *core.BuildTarget, coverageFile string, run int) *core.TestCoverage {
+func parseCoverageFile(state *core.BuildState, target *core.BuildTarget, coverageFile string, run int) *core.TestCoverage {
+	if !target.NeedCoverage(state) {
+		return core.NewTestCoverage()
```
toastwaffle, line 509: Why not return nil?

peterebden: I don't think the caller is expecting to get nil; this matches what happens in the error case below.

toastwaffle: I'd expect anything that's returning a pointer and an error to always return nil if it returns an error, but yes, having looked at `parseTestCoverageFile` that's not the case here

toastwaffle (approved):

Issue #3192, opened 2024-07-05
Add search functionality to docs website


mauriceyap:
The Please website does not currently provide a way to search for content across the site. [Using Google search is an option to achieve this](https://www.google.com/search?q=foo+site%3Aplease.build), but it is inelegant and relies on its cache.

The idea is to implement content search on the client side, in keeping with the architecture of the existing simple static site.

peterebden:
I'm not clear why we want to build this ourselves - it feels a lot lower effort to leverage Google, I wouldn't describe that as inelegant.
Maybe it's nice to have something on the site to search directly, but it's easy enough for that to just link back to Google with a `site:please.build` search

PR #3195, opened 2024-07-06
Split out lexicon, plugins and config page content

mauriceyap:
In the documentation site, extract out the pre-templater content of the pages, `lexicon.html`, `plugins.html` and `config.html` into separate build targets, so that they can be used to build out a search index. In the search index, we do not want the content which the templater adds from `template.html`.

I have tested this by ensuring that these three generated templated pages  before, and after these changes are the same:
```bash
$ cp $(plz build //docs:lexicon_html) tmp/lexicon_old.html
$ cp $(plz build //docs:config_html) tmp/config_old.html
$ cp $(plz build //docs:plugins_html) tmp/plugins_old.html

# Make the changes

$ cp $(plz build //docs:lexicon_html) tmp/lexicon_new.html
$ cp $(plz build //docs:config_html) tmp/config_new.html
$ cp $(plz build //docs:plugins_html) tmp/plugins_new.html

$ diff tmp/lexicon_old.html tmp/lexicon_new.html  # no output
$ diff tmp/config_old.html tmp/config_new.html # no output
$ diff tmp/plugins_old.html tmp/plugins_new.html # no output
```

### DIFF
--- a/docs/BUILD
+++ b/docs/BUILD
@@ -35,23 +35,32 @@ genrule(
 )
 
 genrule(
-    name = "lexicon_html",
-    srcs = deps + [
+    name = "lexicon_content",
+    srcs = [
         "lexicon.html",
         "lexicon_entry.html",
-        "template.html",
         ":rules",
+    ],
+    outs = ["lexicon_content.html"],
+    cmd = [
+        '"$TOOL" -i docs/lexicon.html -i docs/lexicon_entry.html docs/rules.json > "$OUT"',
+    ],
+    tools = ["//docs/tools/lexicon_templater"],
+)
+
+genrule(
+    name = "lexicon_html",
+    srcs = deps + [
+        ":lexicon_content",
+        "template.html",
     ],
     outs = ["lexicon.html"],
     cmd = [
-        '"$TOOLS_LEX" -i docs/lexicon.html -i docs/lexicon_entry.html docs/rules.json > "$OUT"',
-        '"$TOOLS_TMPL" --template docs/template.html --in lexicon.html > tmp.html',
+        'mv docs/lexicon_content.html "$OUT"',
+        '"$TOOL" --template docs/template.html --in lexicon.html > tmp.html',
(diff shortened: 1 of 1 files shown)


docs/BUILD
```diff
@@ -35,23 +35,32 @@ genrule(
...
+    name = "lexicon_html",
+    srcs = deps + [
+        ":lexicon_content",
+        "template.html",
+    ],
     outs = ["lexicon.html"],
     cmd = [
-        '"$TOOLS_LEX" -i docs/lexicon.html -i docs/lexicon_entry.html docs/rules.json > "$OUT"',
-        '"$TOOLS_TMPL" --template docs/template.html --in lexicon.html > tmp.html',
+        'mv docs/lexicon_content.html "$OUT"',
```
toastwaffle, line 59: Why?

And more accurately, why not:

```
genrule(
    name = "lexicon_html",
    srcs = deps + [
        ":lexicon_content",
        "template.html",
    ],
    outs = ["lexicon.html"],
    cmd = [
        '"$TOOL" --template docs/template.html --in docs/lexicon_content.html > "$OUT"',
    ],
    tools = ["//docs/tools/templater"],
    visibility = ["//docs/test/..."],
)
```

similarly below.

I'm not a fan of how the command relies on filenames rather than using `$SRCS`, but I'm guessing the template implicitly depends on the files in `deps`, which makes it hard to use the dict-style `srcs`

docs/BUILD
```diff
@@ -80,43 +89,62 @@ filegroup(
 )
 
 genrule(
-    name = "plugins_html",
+    name = "plugins_content",
     srcs = deps + [
```
toastwaffle, line 93: You removed deps in the other two _content targets?

toastwaffle (changes requested):

peterebden (changes requested):
As I've mentioned on #3192 , I'm unclear that building a search index ourselves is the right approach to take. We should agree on the approach to take before making more code changes.

PR #3200, opened 2024-07-22
Add sharing options to debug as well

Garbett1:
Sometimes I need to share file system for a `plz debug` run, so this exposes the options similar to Exec.

### DIFF
--- a/src/please.go
+++ b/src/please.go
@@ -171,7 +171,11 @@ var opts struct {
 	} `command:"cover" description:"Builds and tests one or more targets, and calculates coverage."`
 
 	Debug struct {
-		Port int               `short:"p" long:"port" description:"Debugging server port"`
+		Port  int `short:"p" long:"port" description:"Debugging server port"`
+		Share struct {
+			Network bool `long:"share_network" description:"Share network namespace"`
+			Mount   bool `long:"share_mount" description:"Share mount namespace"`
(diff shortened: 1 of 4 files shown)


peterebden (commented):
Makes sense!

Could you add a note in the changelog please? given we're on a beta it has the notes for the next release ready to go

peterebden (approved):

src/please.go
```diff
@@ -171,7 +171,11 @@ var opts struct {
 	} `command:"cover" description:"Builds and tests one or more targets, and calculates coverage."`
 
 	Debug struct {
-		Port int               `short:"p" long:"port" description:"Debugging server port"`
+		Port  int `short:"p" long:"port" description:"Debugging server port"`
+		Share struct {
+			Network bool `long:"share_network" description:"Share network namespace"`
+			Mount   bool `long:"share_mount" description:"Share mount namespace"`
+		} `group:"Options to override mount and network namespacing on linux, if configured"`
```
peterebden, line 178: FYI that you could do
```
Mount   bool `long:"mount" description:"Share mount namespace"`
} `group:"...." namespace:"mount"`
```
doesn't matter much here but it's less repetitive when there are lots of them

Garbett1: Fair! I copy-pasted the `exec` options in this case, but maybe could do some tidying.

PR #3202, opened 2024-07-29
Allow file-based completion for various arguments

peterebden:
Several of them weren't offering completions, when they should do it based on paths on the file system. Have swapped them to Filepaths to get this behaviour, adding the ability to it to read from stdin when given `-` as an input.

This also means that some _other_ flags will now interpret the `-` as stdin as well where previously they didn't. I think this makes sense to be more consistent although maybe not all of them are very useful.

Outcome: `plz query whatinputs src/ple<tab>` -> `plz query whatinputs src/please.go`, whereas previously it sat there unhelpfully.

### DIFF
--- a/src/parse/asp/config.go
+++ b/src/parse/asp/config.go
@@ -136,9 +136,6 @@ func getConfigKey(aspKey, configKey string) string {
 
 // pluginConfig loads the plugin's config into a pyDict. It will load con
 func pluginConfig(pluginState *core.BuildState, pkgState *core.BuildState) pyDict {
-	log.Debug("pluginState %s", pluginState.CurrentSubrepo)
-	log.Debug("pkgState %s", pkgState.CurrentSubrepo)
-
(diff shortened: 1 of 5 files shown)


src/parse/asp/config.go
```diff
@@ -136,6 +136,9 @@ func getConfigKey(aspKey, configKey string) string {
 
 // pluginConfig loads the plugin's config into a pyDict. It will load con
 func pluginConfig(pluginState *core.BuildState, pkgState *core.BuildState) pyDict {
+	log.Debug("pluginState %s", pluginState.CurrentSubrepo)
```
toastwaffle, line 139: Just to check, did you intend to leave these debug logs here?

toastwaffle (approved):
I'm not entirely sure how these changes make the tab completion work, but I trust that you have tested it :)

src/parse/asp/config.go
peterebden: good spot thanks, no I definitely did not! forgot I had some local changes floating around before :)

PR #3204, opened 2024-07-30
Query deps rewrite

### LINKED ISSUE #3203: plz query deps shouldn't return the queried target
In-repo example:
```
plz query deps --level=1 //tools/performance:parse_perf_test 
//tools/performance:parse_perf_test
```
I would _expect_ to get `//third_party/python:absl` and `//third_party/python:colorlog` which are the two it specifies.

This is clearly something to do with hidden targets...

peterebden:
This is a bit of a rewrite to try to have it be more sensible about what it prints, and also separate the recursion logic from the printing logic (it was split between the normal and dot versions before)

Fixes #3203

### DIFF
--- a/src/query/deps.go
+++ b/src/query/deps.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/thought-machine/please/src/core"
 )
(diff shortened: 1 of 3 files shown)


src/query/deps.go
```diff
@@ -3,88 +3,75 @@ package query
...
-			fmt.Fprintf(out, "   node [shape=%s] \"%s\";\n", shape, target)
+	for _, l := range target.DeclaredDependencies() {
+		dep := state.Graph.TargetOrDie(l)
+		if !state.ShouldInclude(dep) || done[dep] {
+			continue // target is filtered out
 		}
-		if parent != nil {
-			fmt.Fprintf(out, "   \"%s\" -> \"%s\";\n", parent, target)
+		done[dep] = true
+		if hidden || !dep.HasParent() {
```
goddenrich, line 40: I can't understand why whether the dep has no parent is the condition for whether it should be printed... I know this hasn't changed but a comment might help.

goddenrich (approved):

goddenrich (approved):

Tatskaari (approved):
Much nicer! Thanks :D

PR #3207, opened 2024-08-04
Add Google programmable site search

### LINKED ISSUE #3192: Add search functionality to docs website

peterebden:
I've twiddled this a bit to (mostly) match our colour scheme.
The search box background isn't directly customisable so I've left it as is; it's possible to hack with CSS but needs a bit of care to not look naff.

Resolves #3192 

Screenshot from 2024-08-04 17-01-23

### DIFF
--- a/docs/template.html
+++ b/docs/template.html
@@ -52,6 +52,7 @@
     <link rel="stylesheet" href="/styles.css" />
     <script src="/action.js"></script>
     <script src="/tabs.js"></script>
+    <!-- The ID here is the Google Programmable Search Engine ID. See https://developers.google.com/custom-search/docs/overview for more information. -->
     <script async src="https://cse.google.com/cse.js?cx=a49b06597fd4f470d"></script>
   </head>
   <body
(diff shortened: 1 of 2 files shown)


docs/template.html
```diff
@@ -52,6 +52,7 @@
     <link rel="stylesheet" href="/styles.css" />
     <script src="/action.js"></script>
     <script src="/tabs.js"></script>
+    <script async src="https://cse.google.com/cse.js?cx=a49b06597fd4f470d"></script>
```
toastwaffle, line 55: Could you add a comment to explain where you got this from?

toastwaffle (approved):

docs/template.html
peterebden: Sure thing, done.

Issue #3185, opened 2024-06-30
plz init go should set up go_toolchain / go_stdlib


peterebden:
It essentially doesn't work without them, you get
```
Error building target //data/archidekt:archidekt: exit status 1
find: ‘/usr/local/go1.22.4/pkg/linux_amd64’: No such file or directory
compile: -importcfg: open importconfig: no such file or directory
```
which isn't particularly intuitive in terms of leading you towards the solution.

PR #3212, opened 2024-08-08
Make init plugin go add the go stdlib and toolchain targets

Tatskaari:
Fixes: #3185

### DIFF
--- a/third_party/binary/BUILD
+++ b/third_party/binary/BUILD
@@ -1,4 +1,4 @@
-GO_CI_LINT_VERSION = "1.54.2"
+GO_CI_LINT_VERSION = "1.58.2"
 
 remote_file(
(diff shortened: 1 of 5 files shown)


third_party/binary/BUILD
```diff
@@ -1,16 +1,15 @@
-GO_CI_LINT_VERSION = "1.54.2"
+GO_CI_LINT_VERSION = "1.58.2"
```
peterebden, line 1: Oh I'd forgotten about this. I dunno how much we use it (the github action doesn't)

peterebden (approved):

third_party/binary/BUILD
Tatskaari: I use it to debug lint issues.

PR #3219, opened 2024-08-12
Return `None` for empty optional plugin config fields

chrisnovakovic:
Specifying an empty value for an optional plugin configuration field causes the field to be set to the type's default value (`""` for strings, `0` for integers, `False` for booleans) from the perspective of ASP. This makes it impossible to tell whether an optional field was actually assigned a value inside a `BUILD` file.

For optional fields, return `None` if no value is assigned.


peterebden (approved):
This seems legit to me. I guess it could technically be breaking if someone was previously explicitly checking for False, but it'd pretty much have to be `if CONFIG.THING is False:` which seems like it wouldn't be common

PR #3223, opened 2024-08-14
Don't compare unexported fields when comparing subrepos

### LINKED ISSUE #3222: "Found multiple definitions for subrepo <arch>" error when cross-compiling
Spotted today in the context of the [go-rules release workflow](https://github.com/please-build/go-rules/actions/runs/10388260693/job/28783305957), but has been seen before too:

```
$ ./pleasew build --profile release -o plugin.go.pleasegotool:$(pwd)/plz-out/bin/tools/please_go/please_go -p -v notice //package:please_go_release_files //package:driver_release_files
20:00:09.384 NOTICE: Build running for 0s, 0 / 1 tasks done, 0 workers busy
20:00:09.407 CRITICAL: Found multiple definitions for subrepo 'darwin_arm64' (&{Name:darwin_arm64 Root: PackageRoot: Target:<nil> State:0xc00021a3c0 Arch:{OS:darwin Arch:arm64} IsCrossCompile:true AdditionalConfigFiles:[] fs:{} fsSync:{done:{_:{} v:1} m:{state:0 sema:0}}} s &{Name:darwin_arm64 Root: PackageRoot: Target:<nil> State:0xc00021a3c0 Arch:{OS:darwin Arch:arm64} IsCrossCompile:true AdditionalConfigFiles:[] fs:<nil> fsSync:{done:{_:{} v:0} m:{state:0 sema:0}}})
Error: Process completed with exit code 1.
```

The architecture subrepo causing the problem isn't consistent - I've seen it fail with `darwin_arm64` and `linux_amd64` on different runs of the workflow. From the function that produces the error message, it seems that two subrepos can be equal if the structs representing them are deeply equal:

https://github.com/thought-machine/please/blob/382cfd6f8ab1bb08d70ef150f963d1a00c848847/src/core/graph.go#L105-L114

Perhaps this condition is slightly too strict - the unexported `fs` and `fsSync` fields were added relatively recently, in #3038, and from the error message above it appears that the subrepos are identical in other respects, so perhaps they should be excluded from the equality comparison.

chrisnovakovic:
Two subrepos can be equal even if their unexported `fs` and `fsSync` fields differ. Exclude them when checking the equality of two subrepos.

Fixes #3222.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/dustin/go-humanize v1.0.1
 	github.com/fsnotify/fsnotify v1.7.0
 	github.com/golang/protobuf v1.5.4
+	github.com/google/go-cmp v0.6.0
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/google/uuid v1.6.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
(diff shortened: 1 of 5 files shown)


Tatskaari:
Nice! I think this might have been the issue I was hitting when trying to cross compile recently.

PR #3232, opened 2024-08-20
Stop using cmp

peterebden:
`cmp.Equal` deferences pointers & recursively walks all their members. Subrepos have a pointer to the state, which includes the entire build graph, which isn't feasible to compare this way.

This just implements a very simple top-level comparison on most fields which I think should be adequate.


src/core/subrepo.go
```diff
@@ -73,7 +70,8 @@ func (s *Subrepo) IsRemoteSubrepo() bool {
 
 // Equal returns true if this subrepo is equivalent to another, or false if it is not.
 func (s *Subrepo) Equal(other *Subrepo) bool {
-	return cmp.Equal(s, other, cmpopts.IgnoreFields(Subrepo{}, "fs", "fsSync"))
+	return s.Name == other.Name && s.Root == other.Root && s.PackageRoot == other.PackageRoot &&
+		s.Target == other.Target && s.State == other.State && s.Arch == other.Arch && s.IsCrossCompile == other.IsCrossCompile
```
toastwaffle, line 74: Just to double check, pointer equality makes sense for `Target` and `State`?

toastwaffle (approved):

src/core/subrepo.go
peterebden: It does in the trivial definition of "if they are equal they are the same thing" :) 

In each case there should be only one of each instance of the thing (i.e. there are not two targets with the same label) so it should be ok.

goddenrich: is there a way we can ensure we don't miss a new field when the struct gets updated?

chrisnovakovic (approved):

PR #3237, opened 2024-08-28
Deal with require/provide in query deps

peterebden:
Seeing some cases where this dies with "target not found in build graph"
I don't have a nice minimal case but it does fix that on our internal repo.

I don't know if there's a neater way to make `query` stuff be less explicitly aware of this. Maybe an idea for a future refactor.

### DIFF
--- a/src/query/deps.go
+++ b/src/query/deps.go
@@ -37,19 +37,21 @@ func deps(out io.Writer, state *core.BuildState, target *core.BuildTarget, done 
 			continue // target is filtered out
 		}
 		done[dep] = true
-		if hidden || !dep.HasParent() {
-			// dep is to be printed; either we're printing hidden deps or it has no parent (i.e. is not hidden)
-			if formatdot {
-				printTargetDot(out, dep, target)
+		for _, l := range dep.ProvideFor(target) {
+			if dep := state.Graph.TargetOrDie(l); hidden || !dep.HasParent() {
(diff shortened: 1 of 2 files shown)


src/query/deps.go
```diff
@@ -37,19 +37,21 @@ func deps(out io.Writer, state *core.BuildState, target *core.BuildTarget, done
 			continue // target is filtered out
 		}
 		done[dep] = true
-		if hidden || !dep.HasParent() {
-			// dep is to be printed; either we're printing hidden deps or it has no parent (i.e. is not hidden)
-			if formatdot {
-				printTargetDot(out, dep, target)
+		for _, l := range dep.ProvideFor(target) {
+			if dep := state.Graph.TargetOrDie(l); hidden || !dep.HasParent() {
```
SpangleLabs, line 41: So this just turns the target into a label, and the label back into a target?

SpangleLabs: I see now, I needed to read [the docs on require/provide](https://please.build/require_provide.html), I was reading provide in a generic sense, rather than please-specific sense.
Makes sense!

SpangleLabs (approved):

src/query/deps.go
peterebden: Yeah, it resolves require/provide for that target then keeps going with that

PR #3242, opened 2024-08-28
Skip special handling of experimental directories when inside a subrepo

toastwaffle:


SpangleLabs:
I partially wondered if it's more a case of overriding the plzconfig for the subrepos, in case a different one wants specifying, but .. I can't imagine a proper use-case for an experimental directory in a subrepo, so this is probably perfectly good!

SpangleLabs (approved):

peterebden (approved):

PR #3245, opened 2024-09-01
Refresh Docker images

peterebden:
I haven't yet moved anything to 1.23 because of the telemetry stuff there, but this updates other things in preparation (so hopefully we only have that to worry about when we get there and not other things)

### DIFF
--- a/tools/images/ubuntu/Dockerfile
+++ b/tools/images/ubuntu/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:jammy
+FROM ubuntu:noble
 MAINTAINER peter.ebden@gmail.com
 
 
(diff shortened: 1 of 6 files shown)


tools/images/ubuntu/Dockerfile
```diff
@@ -11,12 +11,12 @@ RUN truncate -s0 /tmp/preseed.cfg; \
     echo "tzdata tzdata/Zones/Europe select London" >> /tmp/preseed.cfg; \
     apt-get update && \
     apt-get install -y python3 python3-dev python3-pip time \
-    curl unzip git locales pkg-config zlib1g-dev psmisc awscli \
```
peterebden, line 14: awscli doesn't appear to be available in noble - it's probably a snap or smthn. I don't think we need it any more anyway

toastwaffle (approved):

PR #3249, opened 2024-09-06
Update toolchain to Go 1.23.1

peterebden:
Forgot in #3248 that this would be necessary too.

### DIFF
--- a/plugins/BUILD
+++ b/plugins/BUILD
@@ -1,7 +1,7 @@
 plugin_repo(
     name = "go",
     plugin = "go-rules",
-    revision = "v1.23.1",
+    revision = "v1.21.3",
 )
 
 plugin_repo(
--- a/plugins/BUILD
+++ b/plugins/BUILD
@@ -1,8 +1,7 @@
 plugin_repo(
     name = "go",
     plugin = "go-rules",
-    revision = "02f5910e44bb8274b5bdb80ec80188a9d180f59f",
-    owner = "peterebden",
+    revision = "v1.21.4",
 )
 
 plugin_repo(
--- a/plugins/BUILD
+++ b/plugins/BUILD
@@ -1,7 +1,7 @@
 plugin_repo(
     name = "go",
     plugin = "go-rules",
-    revision = "v1.20.1",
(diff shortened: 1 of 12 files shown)


toastwaffle (approved):

peterebden:
Looks like something's up with the coverage code on the alt build. Think that will need a proper fix.

src/test/coverage_test.go
```diff
@@ -12,10 +12,11 @@ import (
 var target = &core.BuildTarget{Label: core.BuildLabel{PackageName: "src/test", Name: "coverage_test"}}
 
 const (
-	pythonCoverageFile    = "src/test/test_data/python-coverage.xml"
-	goCoverageFile        = "src/test/test_data/go_coverage.txt"
-	goCoverageFile2       = "src/test/test_data/go_coverage_2.txt"
-	goCoverageFile3       = "src/test/test_data/go_coverage_3.txt"
+	pythonCoverageFile = "src/test/test_data/python-coverage.xml"
+	// TODO(peterebden): Remove the 'x' on these once we update go-rules again
```
peterebden, line 16: I think there's an issue here where the coverage variables we create have the prefix `goCover` and so do these; they later get somehow detected as coverage vars. I guess something in 1.23 must have changed to find them.

The easiest / more correct fix is probably to make go-rules use a less usual coverage variable name. I'm changing these for now so I don't have to block on another round of updates there.

plugins/BUILD
```diff
@@ -1,7 +1,8 @@
 plugin_repo(
     name = "go",
     plugin = "go-rules",
-    revision = "v1.20.1",
+    revision = "02f5910e44bb8274b5bdb80ec80188a9d180f59f",
```
chrisnovakovic, line 4: Need to revert this to the official go-rules before merging.

goddenrich (approved):

plugins/BUILD
peterebden: Yup thanks, have done - wanted to get a build on here against it to check that really did fix it.

chrisnovakovic (approved):

PR #3250, opened 2024-09-06
Avoid storing package cache in Alpine build image

chrisnovakovic:
`apk add --no-cache` updates the package cache but doesn't store it on the file system after the transaction completes, making the preceding `apk update` unnecessary.


peterebden (approved):
TIL that the flag does an update as well!

PR #3252, opened 2024-09-09
Add Go build tags to stdlib

peterebden:
This produces static Go binaries by default (on Linux, anyway). There shouldn't be much real difference since the release binaries are built on Alpine anyway and are already static, so they don't support the weird dynamic DNS stuff that Go doesn't.

Remove a now redundant static stamp test.

I was hoping to get rid of the Alpine build stage completely, but I think we still need it to produce a static sandbox binary, so it would probably be predicated on moving to the Go sandbox (wherever that is at).


src/BUILD.plz
```diff
@@ -48,3 +48,12 @@ filegroup(
     binary = True,
     visibility = ["//:install"],
 )
+
+# On Linux, we should produce statically linked binaries. Emit a test to verify this is the case.
+if is_platform(os = "linux"):
+    gentest(
+        name = "static_test",
+        test_cmd = "ldd $DATA && exit 1 || exit 0",
```
chrisnovakovic, line 56: Clever! Perhaps we should be more explicit though (`file $DATA | grep 'statically linked'`)?

chrisnovakovic (approved):

src/BUILD.plz
peterebden: I tried that first out :) It failed on most of our Docker images that don't have `file` in them, this seems like a slightly more common denominator.

PR #3254, opened 2024-09-10
Remove alpine

peterebden:
I don't think this is necessary any more. We're building static Go binaries anyway as of #3252 and I seem to just be able to build static versions of the sandbox anyway. It feels simpler not to have to manage another OS configuration.

I assume that I will find out from circleci if it has any issues.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,10 +58,6 @@ jobs:
        - run:
            name: Bootstrap & Build
            command: ./bootstrap.sh --test_results_file plz-out/results/please/test_results.xml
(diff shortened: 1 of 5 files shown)


toastwaffle (approved):

peterebden:
Hmmmmm the sandbox binary is like three times the size with glibc :-o I'm not sure I love that

peterebden:
I'm gonna put this on ice (for now?) given this size issue. It's sort of nice but not _that_ big a deal, and I don't like that kind of regression without enough benefit.

PR #3255, opened 2024-09-10
Change iterable interface to be more flexible

peterebden:
Rather than using length + index for everything, this leverages 1.23 iterators. It doesn't really change much fundamental yet, but does move towards being able to iterate things like dicts without extra list allocations.

Benchmarks locally actually seem to do slightly better. I'm not really sure why - I don't especially expect this to improve things dramatically as is - but the main thing is it's not getting worse.

### DIFF
--- a/src/parse/asp/objects.go
+++ b/src/parse/asp/objects.go
@@ -1079,7 +1079,7 @@ func (r *pyRange) Item(index int) pyObject {
 func (r *pyRange) Iter() iter.Seq[pyObject] {
 	return func(yield func(pyObject) bool) {
 		for i := r.Start; i < r.Stop; i += r.Step {
-			if !yield(pyInt(i)) {
+			if !yield(i) {
 				break
 			}
 		}
--- a/src/parse/asp/objects.go
+++ b/src/parse/asp/objects.go
@@ -3,6 +3,7 @@ package asp
 import (
 	"encoding/json"
 	"fmt"
+	"iter"
 	"slices"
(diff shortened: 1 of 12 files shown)


third_party/binary/BUILD
```diff
@@ -1,16 +1,10 @@
-GO_CI_LINT_VERSION = "1.60.3"
+GO_CI_LINT_VERSION = "1.61.0"
 
 remote_file(
     name = "golangci-lint",
     binary = True,
     exported_files = ["golangci-lint-%s-${OS}-${ARCH}/golangci-lint" % GO_CI_LINT_VERSION],
     extract = True,
-    hashes = [
```
peterebden, line 8: I'm not sure we really need the hashes on this, we don't actually build it into anything

src/parse/asp/objects.go
```diff
@@ -37,7 +38,11 @@ type freezable interface {
 // Not all pyObjects implement this.
 type iterable interface {
 	pyObject
-	// This isn't super generic but it works fine for all cases we have right now.
+	Iter() iter.Seq[pyObject]
+}
+
+// An indexable represents an object that knows its length and can be indexed into.
+type indexable interface {
```
chrisnovakovic, line 45: Potentially missing `pyObject` here?

peterebden: It doesn't really matter as such - most of the others don't have it. I'll make them consistent.

PR #3256, opened 2024-09-11
plz query changes --inexact is more selective about parsing

peterebden:
I'm not sure why we didn't do this already - we already have all the tech for `query whatinputs`. Maybe that got added later.

Anyway, `--inexact` and friends only care about a set of files that have changed, and hence only need to parse the relevant packages for them, not the whole graph. That is pretty easy to do - most of this is a bit of cleanup, & sharing some other code to make paths relative so now `plz query changes /home/peter/git/please/src/please.go` works the same as `plz query changes src/please.go`, before it just returned nothing.

### DIFF
--- a/src/core/build_target.go
+++ b/src/core/build_target.go
@@ -961,7 +961,7 @@ func (target *BuildTarget) CheckTargetOwnsBuildOutputs(state *BuildState) error 
 			continue
 		}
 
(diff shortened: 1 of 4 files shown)


peterebden:
Aargh no of course, this only works with the default `level` but not with `--level=-1` or similar.

PR #3259, opened 2024-09-24
Minor clean up to len()

peterebden:
Use an interface rather than a type switch, continuing on the theme of objects implementing additional interfaces for extra functionality


chrisnovakovic (approved):
Much nicer!

PR #3264, opened 2024-10-03
Update to Go 1.23.2

peterebden:
I haven't bothered updating all the Docker images, this update doesn't seem especially critical (it's not security)


chrisnovakovic:
The failing plugin docs lint test is fixed in #3268.

chrisnovakovic (approved):

PR #3270, opened 2024-10-09
Expand home path in `fs.PathExists`

cezarelnazli:
A build rule like this

```python
genrule(
    name = name,
    secrets = ["~/test.yaml"],
)
```

fails building with this error

```
failed to calculate hash: Attempting to record rule hash: cannot calculate hash for ~/test.yaml: file does not exist
```

The hash functions use `os.Lstat` to check for file existence, and `lstat(2)` does not seem to expand the home path itself.

So this PR calls the `ExpandHomePath` before `os.Lstat` in the functions which check file presence. I've tested it by bootstrapping `please` on this branch and building the rule that initially failed.

Not sure if this is the best approach, if there are better suggestions I'm happy to look into it.

### DIFF
--- a/src/build/incrementality.go
+++ b/src/build/incrementality.go
@@ -483,7 +483,7 @@ func secretHash(state *core.BuildState, target *core.BuildTarget) ([]byte, error
 	}
 	h := sha1.New()
 	for _, secret := range target.Secrets {
-		ph, err := state.PathHasher.Hash(secret, false, false, false)
+		ph, err := state.PathHasher.Hash(fs.ExpandHomePath(secret), false, false, false)
 		if err != nil && os.IsNotExist(err) {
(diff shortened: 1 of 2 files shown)


toastwaffle:
Hey, thanks for sending us a PR!

Looking at how [`ExpandHomePath` is used elsewhere](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/thought-machine/please%24+ExpandHomePath&patternType=keyword&sm=0), I think this is the wrong approach, and we should be making sure we expand the path before we pass it into these functions (particularly because I don't think we want to allow srcs of build targets to be paths in a home directory).

Are you alright to chase that down?

cezarelnazli:
Sounds good, I'll look into it, thanks!

toastwaffle (approved):

PR #3292, opened 2024-11-08
Add some goroutine labels & dump stacks if we think we've hung

peterebden:
This is a little rough and ready but attempts to track goroutine labels correctly (bit tricky since we don't routinely propagate contexts but I think I got it)


src/core/state.go
```diff
@@ -630,6 +632,10 @@ func (state *BuildState) forwardResults() {
 		}
 	}()
 	activeTargets := map[*BuildTarget]struct{}{}
+	go func() {
+		time.Sleep(4 * time.Second)
+		dumpGoroutineInfo()
+	}()
```
Tatskaari: Oops

PR #3298, opened 2024-11-19
Auto-update plugin docs versions

samwestmoreland:
This replaces the docs plugin version lint action with a cron job that runs every morning at 9 that _should_ create a pull request to update the versions if needed. Not really sure how to test it!


.github/workflows/plugindocs-update.yml
```diff
@@ -0,0 +1,63 @@
+name: Update Plugin Docs
+on:
+  schedule:
+    - cron: '0 9 * * *'
+    workflow_dispatch:
```
toastwaffle, line 5: What does this line do?

samwestmoreland: That apparently allows manual triggering of the action

samwestmoreland:
> Have you been able to test this at all?

I haven't... yolo? Will revert if it's not working!

toastwaffle:
> Have you been able to test this at all?

Oh, I just saw your "Not really sure how to test it!" comment. In theory you could save (most of) the shell commands as a script and run it locally, but practically I don't think this can do much harm so we can just test "in prod" as it were

samwestmoreland:
Ah good point! The file update part at least seems to do the job:
```
$ git diff
diff --git a/docs/BUILD b/docs/BUILD
index f8505d5dd..105daf7dd 100644
--- a/docs/BUILD
+++ b/docs/BUILD
@@ -57,9 +57,9 @@ genrule(

 # Plugin versions to pull the docs from
 plugins = {
-    "python": "v1.7.0",
-    "java": "v0.4.0",
-    "go": "v1.17.3",
+    "python": "v1.12.1",
+    "java": "v0.4.2",
+    "go": "v1.21.5",
     "cc": "v0.4.0",
     "shell": "v0.2.0",
     "go-proto": "v0.3.0",
```

PR #3305, opened 2024-11-28
Fix potential subinclude lockup

peterebden:
Think we have an issue in here where `isLocal == true` and `s.pkg == nil`; we never activate the target and it never gets built. This happens in the attached test repo when going through `//a` because of the double-subinclude; it works okay if the subinclude from`//c` gets there first (because then we are specifically parsing for that target which gets handled elsewhere).

I think this is a reasonably tactical fix; we don't really _need_ the package in this case, it's only used for:
 - `:all` targets which you can't subinclude anyway
 - Some edge case of Bazel compat which I can't believe anyone is using any more and we should just bin. Anyway I don't think it would end up in this situation since it wouldn't use `subinclude()`
 - Suggesting target names if they mistype, which is only a nice-to-have UI thing


src/core/state.go
```diff
@@ -1029,6 +1031,9 @@ func (state *BuildState) ActivateTarget(pkg *Package, label, dependent BuildLabe
 	if state.ParsePackageOnly && !mode.IsForSubinclude() {
 		return nil // Some kinds of query don't need a full recursive parse.
 	} else if label.IsAllTargets() {
+		if pkg == nil {
```
Tatskaari, line 1034: How do we end up here?

Tatskaari (approved):
I think this makes sense. I wonder why pkg is nil in activate target though. Should it be set based on the context?

PR #3306, opened 2024-11-28
Fix limiter deadlock with queuing data for tests

Tatskaari:
We would hold the action limiter for too long. The build action is holding the lock while waiting for built targets, but if those targets can't acquire the limitter then we deadlock.

### DIFF
--- a/src/core/build_target.go
+++ b/src/core/build_target.go
@@ -712,8 +712,8 @@ func (target *BuildTarget) FinishBuild() {
 }
 
 // WaitForBuild blocks until this target has finished building.
-func (target *BuildTarget) WaitForBuild() {
-	<-target.finishedBuilding
+func (target *BuildTarget) WaitForBuild(dependant BuildLabel) {
+	waitOnChan(target.finishedBuilding, "Still waiting on (target %v).WaitForBuild(dependant %v)", target.Label, dependant)
 }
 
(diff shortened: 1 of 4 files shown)


src/core/build_target.go
```diff
@@ -712,8 +712,8 @@ func (target *BuildTarget) FinishBuild() {
 }
 
 // WaitForBuild blocks until this target has finished building.
-func (target *BuildTarget) WaitForBuild() {
-	<-target.finishedBuilding
+func (target *BuildTarget) WaitForBuild(dependant BuildLabel) {
```
Tatskaari, line 715: Just some improvements to debugging. This would have helped get to the root cause faster.

src/plz/plz.go
```diff
@@ -45,6 +45,32 @@ func Run(targets, preTargets []core.BuildLabel, state *core.BuildState, config *
...
+		} else {
+			localLimiter.Release()
+		}
+
+		if task.Type != core.BuildTask {
+			state.TaskDone()
+			return
+		}
+
+		if state.NeedTests && task.Target.IsTest() && state.IsOriginalTarget(task.Target) {
```
Tatskaari, line 60: This is the thing that needed to be moved to after we release the limiter

peterebden (approved):
nice find!

PR #3308, opened 2024-11-29
Add a yaml() builtin like the json one

Tatskaari:
Converts a asp object to a yaml string.

### DIFF
--- a/rules/builtins.build_defs
+++ b/rules/builtins.build_defs
@@ -291,6 +291,11 @@ def json(value, pretty=False) -> str:
     pass
 
 
+def yaml(value) -> str:
+    """Returns a YAML-formatted representation of a plz value.
+    """
+    pass
+
 def breakpoint():
     """Breaks into an interactive debugging session."""
     pass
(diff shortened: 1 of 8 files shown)


zgeor (approved):
Thanks! Exactly what I needed

rules/builtins.build_defs
```diff
@@ -291,6 +291,11 @@ def json(value, pretty=False) -> str:
     pass
 
 
+def yaml(value) -> str:
+    """Returns a YAML-formatted representation of a plz value.
+    """
+    pass
```
peterebden, line 297: nit: `pass` is unnecessary

src/parse/asp/objects.go
```diff
@@ -107,6 +107,13 @@ func (b pyBool) MarshalJSON() ([]byte, error) {
 	return []byte("false"), nil
 }
 
+func (b pyBool) MarshalYAML() (interface{}, error) {
```
peterebden, line 110: should this be part of the pyObject interface? I think MarshalJSON is

peterebden (approved):

peterebden (changes requested):
Thinking further about this, I'm not sure this is really necessary. As someone pointed out internally, JSON is a subset of YAML, so emitting it via the `json()` builtin would also be valid YAML for anything that wanted to consume that.

PR #3310, opened 2024-12-11
Add local IP and default gateway to network for sandbox tool

peterebden:
These were internal TM changes that I'm trying to move back now. Basically adds a 10.x IP address and routing to it for k8s envtest which can't run on localhost.

### DIFF
--- a/tools/sandbox/sandbox.c
+++ b/tools/sandbox/sandbox.c
@@ -12,6 +12,7 @@
 #include <sched.h>
 #include <signal.h>
 #include <net/if.h>
+#include <net/route.h>
 #include <sys/ioctl.h>
 #include <sys/mman.h>
 #include <sys/mount.h>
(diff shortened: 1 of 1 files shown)


toastwaffle (approved):
I'm not sure I can be trusted to review C; I don't see anything that seems obviously wrong, but you may want to get somebody else to have a look at this.

That said, I'm assuming you've run it locally and this looks simple/linear enough that it's probably correct, so lgtm or whatever?

peterebden:
> That said, I'm assuming you've run it locally and this looks simple/linear enough that it's probably correct, so lgtm or whatever?

Yup. This is also copied pretty much verbatim from our internal version of this (which ultimately I'd like to de-fork if possible) so I'm pretty confident it works already. (Sadly some of the other changes are less obvious & simple, although also are working for us)

PR #3316, opened 2024-12-13
Fix BuildTarget.ProvideFor not checking named data when deciding whether to skip resolution.

toastwaffle:


src/core/build_target.go
```diff
@@ -1205,6 +1205,22 @@ func (target *BuildTarget) ProvideFor(other *BuildTarget) []BuildLabel {
 	return []BuildLabel{target.Label}
 }
 
+func (target *BuildTarget) isDataFor(other *BuildTarget) bool {
+	for _, data := range other.Data {
```
Tatskaari, line 1209: I think you can just do `for _, data := range other.AllData()...`

Tatskaari (approved):
Approved with a suggestion

PR #3317, opened 2024-12-14
Using remote builds with buildbarn

gfleury:
Using remote builds with buildbarn (bb-deployment with docker-compose)

Changes were made accordingly to existing project already using please, so they might be biased (wrong).
So, these changes are more like questions. 

* Ignore symlinks while checking outputs. This one seems a legit fix.

* Use sri.checksums only if just one hash is available. bb-remote-asset currently only takes one hash and I could not find anything defining if multiple hashes are good or not. 

* Entrypoints checks on outputs considering OutputPaths. Currently we have multiple targets with entrypoints, and some have `outs` and other `output_dirs`. The ones with `output_dirs` seems to forced to have these dirs in the entrypoints path.

### DIFF
--- a/src/remote/action.go
+++ b/src/remote/action.go
@@ -470,7 +470,19 @@ func (c *Client) verifyActionResult(target *core.BuildTarget, command *pb.Comman
 			}
 
 			for ep, out := range target.EntryPoints {
-				if _, ok := flatOuts[out]; !ok {
+				ok := false
+				// Single output file
+				if _, ok = flatOuts[out]; ok {
+					continue
+				}
+				// Multiple output files or directories
+				for _, outputPath := range command.OutputPaths {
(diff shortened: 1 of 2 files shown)


peterebden:
Thanks for this! Nice to know the remote execution subsystem is getting exercised by others as well :D

I think the other two parts make a lot of sense but the subresource integrity change is a bit of an issue. I'm pretty sure we do take advantage of it as is; I also think that it's pretty misleading if one hash applies SRI but two hashes suddenly don't.
I do think the format should be fine though, per [MDN](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity):

> Note: An integrity value may contain multiple hashes separated by whitespace. A resource will be loaded if it matches one of those hashes.

I assume we could get in touch with Ed about having bb-remote-asset support multiple values?

gfleury:
Noice, I will take a look on these!

PR #3333, opened 2025-01-20
Record transitive subincludes for plz exports

Tatskaari:
Working through a few issues to try and uncover all the blockers with exporting targets. We're currently missing transitive subincludes. This PR records these so we can export them as well. 

This might be useful elsewhere for things like `plz query deps` but I haven't hooked that in yet. So far, I've used this in `plz export` to make sure these are picked up.

### DIFF
--- a/src/core/graph.go
+++ b/src/core/graph.go
@@ -24,7 +24,7 @@ type BuildGraph struct {
 	subrepos *cmap.Map[string, *Subrepo]
 	// Subincludes that are subincluded by other subincludes
 	subincludeSubincludes map[BuildLabel]labelSet
-	subincMux             sync.RWMutex
+	subincMux             sync.Mutex
 }
 
 // AddTarget adds a new target to the graph.
@@ -173,8 +173,8 @@ func (graph *BuildGraph) DependentTargets(from, to BuildLabel) []BuildLabel {
 
 // TransitiveSubincludes returns all the subincludes made by a given subinclude
 func (graph *BuildGraph) TransitiveSubincludes(l BuildLabel) []BuildLabel {
-	graph.subincMux.RLock()
-	defer graph.subincMux.RUnlock()
+	graph.subincMux.Lock()
+	defer graph.subincMux.Unlock()
 
 	incs := make(labelSet)
 	graph.findTransitiveSubincludes(l, incs)
--- a/src/core/graph.go
+++ b/src/core/graph.go
@@ -5,13 +5,24 @@
 package core
 
 import (
+	"maps"
+	"slices"
 	"sort"
 	"sync"
 
 	"github.com/thought-machine/please/src/cmap"
 )
 
-type labelSet = map[BuildLabel]struct{}
+type labelSet map[BuildLabel]struct{}
+
+func (ls labelSet) add(l BuildLabel) {
+	ls[l] = struct{}{}
+}
+
+func (ls labelSet) contains(l BuildLabel) bool {
+	_, ok := ls[l]
+	return ok
+}
 
 // A BuildGraph contains all the loaded targets and packages and maintains their
 // relationships, especially reverse dependencies which are calculated here.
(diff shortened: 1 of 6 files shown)


src/core/graph.go
```diff
@@ -19,6 +22,9 @@ type BuildGraph struct {
 	packages *cmap.Map[packageKey, *Package]
 	// Registered subrepos, as a map of their name to their root.
 	subrepos *cmap.Map[string, *Subrepo]
+	// Subincludes that are subincluded by other subincludes
+	subincludeSubincludes map[BuildLabel]labelSet
```
peterebden, line 26: is there a reason for going with map+mutex rather than cmap as the others? I guess it doesn't need all the notify-on-write stuff, it just feels like an odd one out

src/core/graph.go
```diff
@@ -19,6 +22,9 @@ type BuildGraph struct {
...
+	subincMux             sync.RWMutex
```
peterebden, line 27: I doubt you get any benefit from RWMutex. I don't think it ever reads in parallel so a normal mutex is probably slightly faster + simpler

test/export/BUILD
```diff
@@ -0,0 +1,6 @@
+subinclude("//test/build_defs")
+
+plz_e2e_test(
+    name = "export_src_please_test",
+    cmd = "plz export --output plz-out/plzexport //src:please && plz --repo_root=$(plz query reporoot)/plz-out/plzexport build //src:please",
```
peterebden, line 5: might want to keep an eye on the performance of this guy. rebuilding the entire thing in a test could be expensive - unless it can utilise already cached artifacts?

src/core/graph.go
Tatskaari: Because the object in the map here is not atomic. We need to guard against mutating not only the map, but the label set too.

src/core/graph.go
Tatskaari: Oh is it faster? I guess the rw mutex is actually two mutexes?  Yeah, I can change this to a normal mutex.

test/export/BUILD
Tatskaari: Yeah, the directory cache should mean these are all cached on the machine but I wonder if that's actually true. Might explain why the darwin build is timing out.

Tatskaari: Just reduced this to do //src/core only which has good coverage (a few subrepo and most of the same subincludes). From scratch the test takes about 20s on my machine. I think it's able to re-use stuff from the directory cache between the main build and the test.

src/core/graph.go
toastwaffle: Please add a comment to record that

src/core/graph.go
```diff
@@ -163,3 +170,41 @@ func (graph *BuildGraph) DependentTargets(from, to BuildLabel) []BuildLabel {
...
+	}
+}
+
+func (graph *BuildGraph) RegisterTransitiveSubinclude(from, to BuildLabel) {
+	graph.subincMux.Lock()
+	defer graph.subincMux.Unlock()
+
+	incs, ok := graph.subincludeSubincludes[from]
+	if !ok {
+		incs = map[BuildLabel]struct{}{}
```
toastwaffle, line 206:
```suggestion
		incs = labelSet{}
```

src/core/package.go
```diff
@@ -101,6 +101,26 @@ func (pkg *Package) RegisterSubinclude(label BuildLabel) {
...
+	includes := make(labelSet, len(pkg.Subincludes))
+
+	for _, s := range pkg.Subincludes {
+		includes[s] = struct{}{}
+		for _, inc := range graph.TransitiveSubincludes(s) {
+			includes[inc] = struct{}{}
+		}
+	}
+
+	ret := make(BuildLabels, 0, len(includes))
```
toastwaffle, line 115: `return slices.Sorted(maps.Keys(includes))`

?

test/export/BUILD
```diff
@@ -0,0 +1,6 @@
+subinclude("//test/build_defs")
+
+plz_e2e_test(
+    name = "export_src_please_test",
+    cmd = "plz export --output plz-out/plzexport //src/core && plz --repo_root=$(plz query reporoot)/plz-out/plzexport build //src/core",
```
toastwaffle, line 5: Can we verify anything about the output here?

Tatskaari: We're validating that after the export we're able to build it by the fact that we get an exit code 0. I think that's what I care about.

toastwaffle: oh, d'oh, I clearly did not read this well enough. Very elegant!

toastwaffle (approved):

PR #3334, opened 2025-01-20
plz export: transitive subrepos

Tatskaari:
A subrepo might depend on other subrepos or even first party targets. This change walks the dependency tree to find these and export them.

### DIFF
--- a/src/export/export.go
+++ b/src/export/export.go
@@ -33,7 +33,7 @@ func ToDir(state *core.BuildState, dir string, targets []core.BuildLabel) {
 			continue // This isn't a real package to be copied
 		}
 		if pkg.Subrepo != nil {
-			continue // Don't copy subrepo BUILD files... they don't exist
+			continue // Don't copy subrepo BUILD files... they don't exist in our source tree
 		}
 		dest := filepath.Join(dir, pkg.Filename)
 		if err := fs.RecursiveCopy(pkg.Filename, dest, 0); err != nil {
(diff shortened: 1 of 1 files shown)


src/export/export.go
```diff
@@ -32,6 +32,9 @@ func ToDir(state *core.BuildState, dir string, targets []core.BuildLabel) {
 		if pkg.Name == parse.InternalPackageName {
 			continue // This isn't a real package to be copied
 		}
+		if pkg.Subrepo != nil {
+			continue // Don't copy subrepo BUILD files... they don't exist
```
toastwaffle, line 36: I think the behaviour is correct here, but the comment is misleading - we don't want to copy anything from the subrepo, we just want to have the same subrepo configuration, right? But subrepos do actually have BUILD files (real, in the case of please plugins, and generated in the case of go_repo)

src/export/export.go
```diff
@@ -75,6 +71,22 @@ func export(graph *core.BuildGraph, dir string, target *core.BuildTarget, done m
...
+}
+
+// export implements the logic of ToDir, but prevents repeating targets.
+func export(graph *core.BuildGraph, dir string, target *core.BuildTarget, done map[*core.BuildTarget]bool) {
+	// We want to export the package that made this subrepo available, but we still need to walk the target deps
+	// as it may depend on other subrepos or first party targets
+	if target.Subrepo != nil {
+		export(graph, dir, target.Subrepo.Target, done)
+	}
+	if done[target] {
```
toastwaffle, line 83: Why is this not first?

src/export/export.go
Tatskaari: Yeah, fair enough. They don't exist in our source tree though so we don't want to export them. Updated :)

src/export/export.go
```diff
@@ -75,6 +71,22 @@ func export(graph *core.BuildGraph, dir string, target *core.BuildTarget, done m
...
+func export(graph *core.BuildGraph, dir string, target *core.BuildTarget, done map[*core.BuildTarget]bool) {
+	if done[target] {
+		return
+	}
+	// We want to export the package that made this subrepo available, but we still need to walk the target deps
+	// as it may depend on other subrepos or first party targets
+	if target.Subrepo != nil {
+		export(graph, dir, target.Subrepo.Target, done)
+	}
+	if target.Subrepo == nil {
```
toastwaffle, line 86:
```suggestion
	else {
```

(am I missing something, or is the GitHub UI for suggested edits terrible?)

toastwaffle (approved):

PR #3336, opened 2025-01-20
Preserve symlinks when copying files

Tatskaari:
This makes sure we preserve the symlinks when copying files. Resolving them can cause issues with broken symlinks in target output.


src/fs/copy.go
```diff
@@ -64,27 +64,26 @@ func RecursiveCopyOrLinkFile(from string, to string, mode os.FileMode, link, fal
...
 	if err != nil {
 		return err
 	}
-	if !filepath.IsAbs(resolvedPath) {
-		resolvedPath = filepath.Join(filepath.Dir(name), resolvedPath)
-	}
-	if err := RecursiveCopyOrLinkFile(resolvedPath, dest, mode, link, fallback); err != nil {
-		return err
+
+	if resolvedPath == "../mocha/bin/_mocha" {
```
toastwaffle, line 82: Is this supposed to be here?

src/fs/copy.go
```diff
@@ -64,27 +64,26 @@ func RecursiveCopyOrLinkFile(from string, to string, mode os.FileMode, link, fal
...
 			}
 			return CopyOrLinkFile(name, dest, fileMode.ModeType(), mode, link, fallback)
 		})
 	}
 	return CopyOrLinkFile(from, to, info.Mode(), mode, link, fallback)
 }
 
-// recursiveCopyOrLinkSymlink will resolve the symlink, and recursively copy the content to the destination
-func recursiveCopyOrLinkSymlink(name, dest string, mode os.FileMode, link, fallback bool) error {
+// copySymlink will resolve the symlink, and recursively copy the content to the destination
```
toastwaffle, line 75:
```suggestion
// copySymlink will resolve the symlink and create an equivalent symlink at dest. Assumes the symlink is relative, not absolute.
```

toastwaffle (approved):

src/fs/copy.go
Tatskaari: Yes this is absolutely critical to the operation of the buld system ;P

PR #3345, opened 2025-01-29
Ensure the dir we're exporting to exists

Tatskaari:
Otherwise we get errors trying to clean out the old `.plzconfig` file before copying ours in.

### DIFF
--- a/src/export/export.go
+++ b/src/export/export.go
@@ -37,6 +37,10 @@ func ToDir(state *core.BuildState, dir string, noTrim bool, targets []core.Build
 		exportedTargets:  map[core.BuildLabel]bool{},
 	}
 
+	if err := os.MkdirAll(dir, fs.DirPermissions); err != nil {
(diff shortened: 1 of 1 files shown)


toastwaffle (approved):

src/export/export.go
```diff
@@ -37,6 +37,10 @@ func ToDir(state *core.BuildState, dir string, noTrim bool, targets []core.Build
 		exportedTargets:  map[core.BuildLabel]bool{},
 	}
 
+	if err := os.MkdirAll(dir, fs.DirPermissions); err != nil {
+		log.Fatalf("failed to create export dir: %v", dir)
```
chrisnovakovic, line 41: Worth outputting the error too?

chrisnovakovic, line 41:
```suggestion
		log.Fatalf("failed to create export dir: %s", dir)
```

PR #3351, opened 2025-03-11
Add metrics for alias and command invocations

toastwaffle:
TIL that the mechanism by which Please invokes aliases is kinda sketchy...

This required some slightly awkward changes to break an input cycle between `core` and `metrics`:

- Move `core.PleaseVersion` to `version.PleaseVersion`
- Define the metrics config inside the `metrics` package
- Pass whether remote execution was used as a bool to `metrics.Push`

Tested this by adding the prom aggregation gateway to the config and running under mitmproxy to see that the metrics were generated as expected


src/core/config.go
```diff
@@ -1027,6 +1025,13 @@ func (config *Configuration) Completions(prefix string) []flags.Completion {
 	return ret
 }
 
+var aliasInvocationCount = metrics.NewCounterVec(
+	"alias",
+	"invocation",
+	"How many times each alias is used",
+	[]string{"alias"},
+)
```
SpangleLabs, line 1033: Ordinarily I would recommend that something initialises all valid metric labels here to zero, otherwise the first hit isn't counted. But because this is pushing to an aggregation gateway, I suppose that is not required!

src/metrics/prometheus.go
```diff
@@ -11,18 +11,25 @@ import (
...
 
 var log = logging.Log
 
 var registerer = prometheus.WrapRegistererWith(prometheus.Labels{
-	"version": core.PleaseVersion,
+	"version": version.PleaseVersion,
 }, prometheus.DefaultRegisterer)
 
+type Config struct {
+	PrometheusGatewayURL string       `help:"The gateway URL to push prometheus updates to."`
```
SpangleLabs, line 26:
```suggestion
	PrometheusGatewayURL string       `help:"The aggregation gateway URL to push prometheus updates to."`
```
This should specifically be an aggregation gateway, sending them to a standard push gateway would be worthless. So may as well clarify that slightly here.

SpangleLabs (approved):
Awesome! Thank you!

src/core/config.go
toastwaffle: indeed

PR #3358, opened 2025-03-25
Add a `ci` label to all Please metrics

toastwaffle:


SpangleLabs (approved):
Not sure this needs to be a minor, could just be a patch really. But looks good!

PR #3360, opened 2025-04-15
Cirrus CI: bootstrap using Go 1.24

chrisnovakovic:
`go121` has gone away from ports; use `go124` instead.


chrisnovakovic:
> I'll close this, but we might want to consider downloading and installing the official FreeBSD Go distribution from https://go.dev/dl on workers that need it.

#3361 for this.

PR #3365, opened 2025-05-14
continue on empty globs when keepgoing=true; add config option

grantzvolsky:
- Make empty globs non-critical when KeepGoing is enabled
- Add a please.keepgoing config option to complement the CLI flag

This allows us to execute nested `plz run` commands on partial trees exported via `plz export`.

A possible alternative implementation is to reintroduce the `ErrorOnEmptyGlob` flag. #2730

### DIFF
--- a/src/parse/asp/builtins.go
+++ b/src/parse/asp/builtins.go
@@ -678,9 +678,12 @@ func glob(s *scope, args []pyObject) pyObject {
 	if !allowEmpty && len(glob) == 0 {
 		// Strip build file name from exclude list for error message
 		exclude = exclude[:len(exclude)-len(s.state.Config.Parse.BuildFileName)]
-		log.Fatalf("glob(include=%s, exclude=%s) in %s returned no files. If this is intended, set allow_empty=True on the glob.", include, exclude, s.pkg.Filename)
+		if s.state.KeepGoing {
+			log.Warningf("glob(include=%s, exclude=%s) in %s returned no files; continuing due to enabled KeepGoing option.", include, exclude, s.pkg.Filename)
+		} else {
+			log.Fatalf("glob(include=%s, exclude=%s) in %s returned no files. If this is intended, set allow_empty=True on the glob.", include, exclude, s.pkg.Filename)
+		}
 	}
-
 	return fromStringList(glob)
 }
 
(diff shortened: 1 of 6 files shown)


src/parse/asp/builtins.go
```diff
@@ -678,9 +678,12 @@ func glob(s *scope, args []pyObject) pyObject {
 	if !allowEmpty && len(glob) == 0 {
 		// Strip build file name from exclude list for error message
 		exclude = exclude[:len(exclude)-len(s.state.Config.Parse.BuildFileName)]
-		log.Fatalf("glob(include=%s, exclude=%s) in %s returned no files. If this is intended, set allow_empty=True on the glob.", include, exclude, s.pkg.Filename)
+		if s.state.KeepGoing {
+			log.Warningf("glob(include=%s, exclude=%s) in %s returned no files; continuing due to enabled KeepGoing option.", include, exclude, s.pkg.Filename)
+		} else {
+			log.Fatalf("glob(include=%s, exclude=%s) in %s returned no files. If this is intended, set allow_empty=True on the glob.", include, exclude, s.pkg.Filename)
+		}
```
goddenrich, line 685: Im not sure but will this not cause a deadlock scenario if there are dependents of the rule that contains the problematic glob?

src/please.go
```diff
@@ -89,7 +89,7 @@ var opts struct {
 		KeepWorkdirs       bool    `long:"keep_workdirs" description:"Don't clean directories in plz-out/tmp after successfully building targets."`
 		HTTPProxy          cli.URL `long:"http_proxy" env:"HTTP_PROXY" description:"HTTP proxy to use for downloads"`
 		Debug              bool    `long:"debug" description:"When enabled, Please will enter into an interactive debugger when breakpoint() is called during parsing."`
-		KeepGoing          bool    `long:"keep_going" description:"Continue as much as possible after an error. While the target that failed and those that depend on it cannot be build, other prerequisites of these targets can be."`
+		KeepGoing          bool    `long:"keep_going" description:"Continue as much as possible after an error. Builds prerequisites of failed targets and disables safety checks."`
```
goddenrich, line 92: Im not sure I understand this description. It doesn't only build prerequisites of failed builds. Also not sure what safety checks it disables.

src/parse/asp/builtins.go
grantzvolsky: The code path introduced in this change is equivalent to explicitly passing the `allow_empty = True` argument to the glob. Now new deadlocks introduced.

grantzvolsky:
Closing this as no longer needed, may reopen if it becomes relevant again. For future possible implementations, the preference is not to combine the allow empty globs option with the keep going option.

PR #3369, opened 2025-05-29
fix(query.deps): memoize provided labels rather than declared targets

grantzvolsky:
The `plz query deps` graph traversal previously memoized targets directly from DeclaredDependencies. This was problematic because a single declared dependency can provide different actual build labels (and therefore different actual dependencies) depending on the context of the target that requires it (via `dep.ProvideFor(target)`).

If a declared dependency target `D` was processed for a pareent `P1`, `D` would be marked as done. If another parent `P2` also declared `D` as a dependency, and `D.ProvideFor(P2)` would have yielded additional build labels, these provisions for `P2` would be skipped because `D` was already memoized.


Tatskaari (approved):
Ah yeah nice catch. I think this makes sense.

PR #3370, opened 2025-06-09
Add `//` operator

chrisnovakovic:
This is equivalent to Python's `//` operator: the mathematical division of the first operand by the second with the `floor` function applied to the result.


mhbardsley (approved):
:smiley:

PR #3393, opened 2025-07-28
Add support for suppressing SRCS env variables (in the case of rules with large numbers of sources)

toastwaffle:


chrisnovakovic:
> I think it's reasonable for a target to be able to declare explicitly that it does not need the environment variables that would be generated (hence the move away from making please suppressing the variables when there were just too many sources). Is your concern that having this option would open the gate to other build defs abusing it?

Not abusing it, per se (it's opt-in, after all), but I can see other situations where downstream users might run into the same (or a similar) problem but in a way that can't be worked around with this approach. The long/short named sources example above is one of them; another is one in which two excessively long lists of sources are presented to the build environment and the command is still somehow expected to process them separately. We'd then have to think carefully about how to solve _that_ problem in a way that's backwards-compatible with what's being proposed here.

> I think the problem we're trying to fix is an exceptional case for which I expect this escape hatch to be used very rarely, and I don't think it justifies the additional effort to add support for putting the build environment into files (and the potential sharp edges that could come with doing that), especially when we have no other use case for it and a working solution for Go.

I think this is actually the crux of my concern: it feels like a highly specific solution developed in response to a highly specific problem. I appreciate the rarity of the problem - after all, we haven't encountered any other reports of this in nearly a decade of public Please development - but I feel that any solution implemented within the main Please code base ought to be generic enough to solve similar problems that might crop up in future, however unlikely we think they might be right now.

toastwaffle:
Okay, colour me convinced :) do you have any suggestions for where to put those files in the sandbox? I was thinking something like `<sandbox_dir>/__plz_build_environment/SRCS` (i.e. filenames matching what the env var would otherwise be called). And do you think it should be `SRCS_GO_FILE`, or `SRCSFILE_GO`, or `SRCS_FILE_GO`?

chrisnovakovic:
> Okay, colour me convinced :) do you have any suggestions for where to put those files in the sandbox? I was thinking something like `<sandbox_dir>/__plz_build_environment/SRCS` (i.e. filenames matching what the env var would otherwise be called). And do you think it should be `SRCS_GO_FILE`, or `SRCSFILE_GO`, or `SRCS_FILE_GO`?

IMO we should keep the directory structure as "Please-like" as possible - there should be files named `src` and `srcs` (analogous to `$SRC` and `$SRCS` respectively) as well as `srcs/xyz` (analogous to `$SRCS_XYZ` for named sources), all in a top-level directory in the sandbox. Unsure what it should be called, but it should be short enough that it isn't a pain to keep writing or reading it in the command while being generic enough that we can put other stuff in there later if we feel we need to. I just did a quick search of the plugins and couldn't find any with commands that created a top-level directory named `_plz`, so perhaps that?

If we make this explicitly an opt-in feature, I don't think we need the environment variables at all (although perhaps it'd be neater to refer to `$SRCSFILE_XYZ` than `$TMP_DIR/_plz/srcs/xyz`)...

PR #3398, opened 2025-08-12
Updated docstring for gen rule to amend incorrect info - outs cannot be a dict

avdes96:


chrisnovakovic (changes requested):
This documentation is accurate - `outs` of type `dict` is how [named outputs](https://please.build/build_rules.html#named-outs) are implemented.

PR #3399, opened 2025-08-20
Add number of tasks left in plain display update

orfeasa:


orfeasa:
Before:

image

After:
image

PR #3424, opened 2025-10-08
Use arcat v1.2.0

chrisnovakovic:
This version of arcat fixes the long-running problem with long file names in GNU-style ar archives, along with more robust ar handling generally.


toastwaffle (approved):

chrisnovakovic:
Looks like this needs to be done concurrently with the cc-rules v0.5.1 bump to keep BSD ar happy...

PR #3439, opened 2025-10-26
Keep os env vars for custom cache cmd

adox:
Host env vars are not available during cahce retrieval/store custom command ie. one can't make env vars based auth for aws.

Put some dummy custom cache cmd to .plzconfig:
```
[Cache]
RetrieveCommand="echo key $AWS_ACCESS_KEY_ID >&2 && aws s3 cp s3://YOUR-OWN-CACHE-BUCKET/please/$CACHE_KEY -"
```
Set proper env var in your os: 
```
export AWS_ACCESS_KEY_ID=asdfghjkl
```

Sample output:

Before: 07:17:45.405 WARNING: Custom command output:key
After: 07:34:22.196 WARNING: Custom command output:key asdfghjkl

Run `plz build` with verbose output `plz build -v 3` to see custom cache command warnings.

### DIFF
--- a/src/cache/cmd_cache.go
+++ b/src/cache/cmd_cache.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/hex"
 	"io"
+	"os"
 	"os/exec"
 	"path/filepath"
 
@@ -53,7 +54,7 @@ func (cache *cmdCache) Retrieve(target *core.BuildTarget, key []byte, _ []string
 	log.Debug("Retrieve %s: %s from custom cache...", target.Label, strKey)
 
 	cmd := exec.Command("sh", "-c", cache.retrieveCommand)
-	cmd.Env = append(cmd.Env, "CACHE_KEY="+strKey)
+	cmd.Env = append(os.Environ(), "CACHE_KEY="+strKey)
 
 	var cmdOutputBuffer bytes.Buffer
 	cmd.Stderr = &cmdOutputBuffer


src/cache/cmd_cache.go
```diff
@@ -53,7 +54,7 @@ func (cache *cmdCache) Retrieve(target *core.BuildTarget, key []byte, _ []string
 	log.Debug("Retrieve %s: %s from custom cache...", target.Label, strKey)
 
 	cmd := exec.Command("sh", "-c", cache.retrieveCommand)
-	cmd.Env = append(cmd.Env, "CACHE_KEY="+strKey)
+	cmd.Env = append(os.Environ(), "CACHE_KEY="+strKey)
```
toastwaffle: Doesn't this need to be done for `Store()` too?

toastwaffle (commented):
I'm a little wary of this, given we try to avoid the ambient environment leaking into the build process at all unless explicitly configured to do so. It also feels like a potential attack vector for anybody running Please in a not fully trusted repository, where a malicious cache command could slurp up the contents of your environment.

@chrisnovakovic what do you think?

One potential thing we could do is add a config option for an explicit list of environment variables that do get passed through. That would slightly limit (but not fully solve) the problems above.

In either case, it would be good to add an end-to-end test to prove this works somehow.

chrisnovakovic:
As @toastwaffle says, this violates Please's principle of reproducibility - the idea is that the build environment is minimal (and is something that the caching mechanism relies on: the hash that accompanies the output of each build target, which determines whether a target needs to be rebuilt, is derived in part from the environment that existed at build time).

If you want the values of certain variables defined in the parent environment to be "passed through" to the build sandbox, you can set

```
[Build]
PassEnv = ENV_VAR_NAME
PassEnv = ANOTHER_ENV_VAR_NAME
```

in `.plzconfig`, or

```
[Build]
PassUnsafeEnv = ENV_VAR_NAME
PassUnsafeEnv = ANOTHER_ENV_VAR_NAME
```

to additionally exclude the variables' names/values from the computation of the build target hash (although be aware that the former will lead to decreased incrementality and more frequent rebuilding of targets, and the latter will reduce reproducability to varying degrees depending on which environment variables you specify).

PR #3453, opened 2025-11-13
Download run-time dependencies when running a target remotely

chrisnovakovic:
The code path taken when `plz run`ning a binary target is slightly different for remote execution vs local execution - ensure the target's transitive run-time dependencies are all downloaded by the client before it runs the target.

### DIFF
--- a/src/remote/remote.go
+++ b/src/remote/remote.go
@@ -351,8 +351,11 @@ func (c *Client) Build(target *core.BuildTarget) (*core.BuildMetadata, error) {
 		if err := c.Download(target); err != nil {
 			return metadata, err
 		}
-		// TODO(peterebden): Should this not just be part of Download()?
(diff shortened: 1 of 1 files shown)


src/remote/remote.go
```diff
@@ -351,10 +351,13 @@ func (c *Client) Build(target *core.BuildTarget) (*core.BuildMetadata, error) {
 		if err := c.Download(target); err != nil {
 			return metadata, err
 		}
-		// TODO(peterebden): Should this not just be part of Download()?
+		// TODO(peterebden): Should these not just be part of Download()?
```
chrisnovakovic: My best answer is "maybe, but probably not" - Chesterton's Fence and all that.

toastwaffle (approved):

PR #3472, opened 2025-12-16
fix: set only one hash per version to "arcat" binary for compatibility with Buildbarn and SRI standard 

njskalski:
Buildbarn assumes that only one SHA per binary is provided, or the SHAs are of different algorithms. 

To enable Buildbarn as a backend to Please, this change is necessary.

With this change, current Buildbarn works as a backend.


src/parse/internal_package.go
```diff
@@ -27,15 +27,33 @@ func GetInternalPackage(config *core.Configuration) (string, error) {
 		url = fmt.Sprintf("%s/%s_%s/%s/please_tools_%s.tar.xz", config.Please.DownloadLocation, runtime.GOOS, runtime.GOARCH, version.PleaseVersion, version.PleaseVersion)
 	}
 
+	arcatHash := ""
```
toastwaffle: I assume the linter wants this to be `var arcatHash string`?

PR #3481, opened 2026-01-28
test showing bug with whatinputs

goddenrich:
This minimal example demonstrates a bug with plz query what inputs. I belive there is a bug somewhere in the parsing of [packages](https://github.com/thought-machine/please/blob/master/src/parse/asp/parser.go#L76) which causes the parents of targets to drop the underscore if there is one.

```
rgodden@rgodden-tm05522:~/repos/please$ plz test //test/whatinputs:hidden_targets 
Fail: //test/whatinputs:hidden_targets   0 passed   0 skipped   0 failed   1 errored Took 50ms
Error: TestFailed in hidden_targets
Test failed
exit status 1
//test/whatinputs:hidden_targets 1 test run in 53ms; 0 passed, 1 errored
    hidden_targets  ERROR
1 test target and 1 test run; 0 passed, 1 errored.
Total time: 370ms real, 50ms compute.
Messages:
12:59:57.797   ERROR: //test/whatinputs:hidden_targets failed: Test failed
```


goddenrich:
ah found the issue:
https://github.com/thought-machine/please/blob/master/src/core/build_label.go#L407
just trims underscores and removes # and anything after it

Issue #3471, opened 2025-12-09
`please format` does not canonicalise local build labels


HartBlanc:
The `please format` command helpfully canonicalises and deduplicates build labels. For example the following BUILD file:
```
go_library(
    name = "name",
    srcs = [
        "src.go",
    ],
    deps = [
        "//path/to/my/package:package",
        "//path/to/my/package:package",
        "//path/to/my/package",
        "//path/to/my/package",
    ],
)
```

is formatted as:
```
go_library(
    name = "name",
    srcs = [
        "src.go",
    ],
    deps = ["//path/to/my/package"],
)
```

However, the same is not true for local build labels.

For example the following BUILD file:
```
# path/to/my/package/BUILD
go_library(
    name = "name",
    srcs = [
        "src.go",
    ],
    deps = [
        "//path/to/my/package:package",
        "//path/to/my/package:package",
        "//path/to/my/package",
        "//path/to/my/package",
        ":package",
        ":package",
    ],
)
```

is formatted as:
```
go_library(
    name = "name",
    srcs = [
        "src.go",
    ],
    deps = [
        ":package",
        "//path/to/my/package",
    ],
)
```

I think it would be preferable if it was formatted as:
```
go_library(
    name = "name",
    srcs = [
        "src.go",
    ],
    deps = [
        ":package",
    ],
)
```

This would ensure that the target is not duplicated and would ensure that there is a single canonical label.
If there is consensus that this is preferable, I'd be happy to raise a PR - thanks.

hartblanc:
I've had another go at this in #3492

PR #3492, opened 2026-02-27
Add ShortenAbsoluteLabelsToRelative configuration option

hartblanc:
This change ensures that absolute labels to package-local targets (e.g. `//a/b/c:d`) are transformed into local labels (e.g. `:d`) by the please format command when the new `Format.ShortenAbsoluteLabelsToRelative` config option is enabled.

The aim of this change is to address https://github.com/thought-machine/please/issues/3471.

This change supercedes https://github.com/please-build/buildtools/pull/10.

### DIFF
--- a/docs/config.html
+++ b/docs/config.html
@@ -1212,6 +1212,24 @@
 </section>
 
 <section class="mt4">
+  <h2 id="format" class="title-2">[Format]</h2>
+
+  <ul class="bulleted-list">
+    <li>
+      <div>
+        <h3 class="mt1 f6 lh-title" id="format.shortenabsolutelabelstorelative">
+          ShortenAbsoluteLabelsToRelative <span class="normal">(bool)</span>
+        </h3>
+
+        <p>
+          Shortens absolute labels to relative when running <code class="code">plz format</code> if the label is in the same package as the associated target.
+        </p>
+      </div>
+    </li>
+  </ul>
+</section>
+
+<section class="mt4">
   <h2 id="test" class="title-2">[Test]</h2>
 
   <ul class="bulleted-list">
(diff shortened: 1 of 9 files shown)


toastwaffle (approved):

toastwaffle (commented):
Please could you fix the lint (otherwise this is good to go)

peterebden:
> After pondering about [please-build/puku#147](https://github.com/please-build/puku/issues/147), I'm actually wondering if we want to make `plz format` or `buildtools/build` configurable (as this change does).
> 
> The approach of gofmt for example, is to not allow any configuration and '[Gofmt's style is no one's favorite, yet gofmt is everyone's favorite.](https://go-proverbs.github.io/)'. This works very nicely for tools which modify go source code, as they are able to always produce well-formatted output by simply calling gofmt (as a library).
> 
> For this reason, whilst it seems a little crude at first, I'm wondering if [please-build/buildtools#10](https://github.com/please-build/buildtools/pull/10) might actually be preferable to this approach.

I tend to agree with this. It might be initially disruptive having this reformat but it seems nicer in the long run to have one canonical format than many with lots of flags.

FWIW I don't think we have to change the library in the PR you mentioned; the approach here of calling `FormatWithRewriter` seems good, but maybe we just set that field to `true`.

hartblanc:
> FWIW I don't think we have to change the library in the PR you mentioned; the approach here of calling FormatWithRewriter seems good, but maybe we just set that field to true.

My thinking was that it would be preferable to have other tools that format build rules via `buildtools/build` (e.g. puku) behave the same, which would be achieved by initial approach. If we use the `FormatWithRewriter` approach we would need to do that everywhere if we wanted to be consistent, which feels easy to forget to do.

PR #3506, opened 2026-03-24
Implement e2e tests for export functionality

DuBento:
This changes introduce small, targeted export test cases. The goal is to use these test for validating upcoming changes to the export logic.
- Introduced a new `please_export_e2e_test` build def that is heavily inspired by `please_repo_e2e_test` but branches away to avoid injecting any base configuration files. 
- Favoured readability by using golden-master tests: for each case we define a `test_repo` and an `expected_repo` with the entire files for testing and for validating the export. 
- Non-breaking changes to the generic `please_repo_e2e_test` were added to accommodate additional logic.


test/build_defs/test.build_defs
```diff
@@ -1,27 +1,33 @@
...
 def please_repo_e2e_test(
         name: str,
-        plz_command: str,
+        plz_command: str|list,
         repo: str,
         data: dict={},
         deps: list=[],
+        defer_cmd=[],
         tools: dict={},
+        tool_content_checker = ["//test/build_defs:content_checker"],
```
toastwaffle:
```suggestion
        content_checker_tool = ["//test/build_defs:content_checker"],
```

nit

toastwaffle (changes requested):

test/export/BUILD
```diff
@@ -1,6 +1,186 @@
 subinclude("//test/build_defs")
 
+sh_binary(
+    name = "build_statement_checker",
+    main = "build_statement_checker.sh",
+)
+
+# Runs an export e2e test. It purposely tests for different aspect (e.g. BUILD files, sources)
+# within the same test to avoid the overhead of exporting multiple times.
+def please_export_e2e_test(
```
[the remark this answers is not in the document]
DuBento: Added to `test/export`. Happy to change if needed

test/export/BUILD
```diff
@@ -1,6 +1,186 @@
 subinclude("//test/build_defs")
 
+sh_binary(
+    name = "build_statement_checker",
+    main = "build_statement_checker.sh",
+)
+
+# Runs an export e2e test. It purposely tests for different aspect (e.g. BUILD files, sources)
+# within the same test to avoid the overhead of exporting multiple times.
```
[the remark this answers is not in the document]
DuBento: Kept the building step but it's easy to remove if we wish.

test/export/please_export_e2e_test.build_defs
```diff
@@ -0,0 +1,66 @@
...
+        config_override += "-o plugin.go.gotool:$TOOLS_GO"
+        tools["GO"] = [CONFIG.GO.GO_TOOL]
+
+    test_cmd = [
+        # Export
+        f'plz {config_override} --repo_root="$DATA_TEST_REPO" export --output "{EXPORT_DIR}" {targets}',
+        # Golden-Master validation with expected repo. Done before any building to avoid plz-out
+        f'diff -r "{exported_repo}" "$DATA_EXPECTED_REPO"',
+        # Tests building the exported target which in turn ensures the sources are included
+        f'plz {config_override} --repo_root="{exported_repo}" build {targets}',
```
toastwaffle: Does it make sense for this to just be `plz build //...`? Each repo should be tiny, so building everything shouldn't be a problem, and it asserts our requirement that the entire export should be buildable?

test/export/BUILD
```diff
@@ -1,5 +1,124 @@
...
+filegroup(
+    name = "export_e2e_test_build_def",
+    srcs = ["please_export_e2e_test.build_defs"],
+)
+
+subinclude(":export_e2e_test_build_def")
+
+# Export a target generated by a native genrule target and trim unused
+# build statements.
+please_export_e2e_test(
```
toastwaffle: Please could you move each of these `please_export_e2e_test` targets to a BUILD file alongside the test/expected repos?

test/export/test_custom_def/expected_repo/build_defs/BUILD_FILE
```diff
@@ -0,0 +1,5 @@
+filegroup(
```
toastwaffle, line 1: Either this should contain an entry for dummy.build_defs, or dummy.build_defs shouldn't exist

test/export/BUILD
```diff
@@ -1,5 +1,124 @@
...
+    name = "export_custom_target_in_file",
+    dir = "test_custom_in_file_def",
+    export_targets = ["//:simple_custom_target"],
+)
+
+# Export a target generated by a custom build def.
+please_export_e2e_test(
+    name = "export_standard_children_custom_target",
+    cmd_on_export = [
+        # Child of build def
```
toastwaffle: Let's avoid referring to children, as that's not well-defined terminology. `adjacent_targets_in_build_def` is probably better naming for this case, and the fact that one is a "child" of the other doesn't make a difference.

What you're essentially capturing here is that a target A created in a build def alongside the target B that we're trying to export will result in both A and B being buildable

test/export/BUILD
```diff
@@ -1,5 +1,124 @@
...
+# Export a custom target from a repo that preloads the build def used.
+please_export_e2e_test(
+    name = "export_preload_use_preloaded_def",
+    dir = "test_use_preloaded",
+    export_targets = ["//:use_preloaded_def"],
+)
+
+# # Test go binary target.
+please_export_e2e_test(
+    name = "export_go_binary",
```
toastwaffle: Other than the target being exported moving to a subdirectory, I don't see any behavioural difference between this and `export_go_target_with_go_dep`

test/export/test_native_target_with_go_dep/source_repo/main.go
```diff
@@ -0,0 +1,11 @@
+package main
```
toastwaffle, line 1: There's no build target for this file

test/export/BUILD
```diff
@@ -1,5 +1,124 @@
...
+# Test go target with go third_party dependency.
+please_export_e2e_test(
+    name = "export_go_target_with_go_dep",
+    dir = "test_go_bin_with_go_dep",
+    export_targets = ["//:bin_go_dep"],
+    include_go = True,
+)
+
+# Test outputs export.
+please_repo_e2e_test(
```
toastwaffle: Why do you not use `please_export_e2e_test` here? I think both would be clearer as golden-master tests

test/export/BUILD
```diff
@@ -1,5 +1,124 @@
...
+    name = "export_multiple_targets",
+    dir = "test_multiple_targets",
+    export_targets = [
+        "//:target1",
+        "//:target2",
+    ],
+)
+
+# Export a target from a repo that preloads a build def.
+# This test purposely doesn't use the custom def but checks that the source files are still included.
```
toastwaffle:
```suggestion
# The exported target doesn't use the preloaded build def, but the preloaded build def is still present in the export.
```

test/export/please_export_e2e_test.build_defs
```diff
@@ -0,0 +1,66 @@
...
+    Args:
+        name (str): Name of the test.
+        dir (str): A relative path to the directory containing `test_repo` and `expected_repo`.
+        export_targets (list): Targets to export as part of the test.
+        cmd_on_export (list): Optional. Additional commands to run on the exported directory.
+        include_go (bool): Optional. A flag to include Go dependencies in the e2e test. Use when
+            testing repos that include Go targets.
+    """
+    EXPORT_DIR = "plz-out/plzexport"
+    test_repo = f"{dir}/test_repo"
```
toastwaffle: nit: can we call this `source_repo` rather than `test_repo` throughout please

test/export/please_export_e2e_test.build_defs
```diff
@@ -0,0 +1,66 @@
...
+    exported_repo = f"$DATA_TEST_REPO/{EXPORT_DIR}"
+    data = {}
+    tools = {}
+
+    targets = " ".join(export_targets)
+
+    config_override = ""
+    if include_go:
+        config_override += "-o plugin.go.gotool:$TOOLS_GO"
+        tools["GO"] = [CONFIG.GO.GO_TOOL]
```
toastwaffle: is this necessary at all?

test/export/please_export_e2e_test.build_defs
```diff
@@ -0,0 +1,66 @@
...
+    expected_repo = f"{dir}/expected_repo"
+    exported_repo = f"$DATA_TEST_REPO/{EXPORT_DIR}"
+    data = {}
+    tools = {}
+
+    targets = " ".join(export_targets)
+
+    config_override = ""
+    if include_go:
+        config_override += "-o plugin.go.gotool:$TOOLS_GO"
```
toastwaffle: can't we just put this in the `.plzconfig` in the source repo?

toastwaffle (changes requested):
Let's ensure that each test repo has at least one target that does not get exported (getting it as close as possible to the edge of being exported). You've got quite a few cases here where the expected repo is identical to the test repo, which doesn't prove anything for the "export as little as possible" goal of exporting. Bonus points for making the e2e enforce that the test repo and expected repo are different.

Also, what's the behaviour around comments in BUILD files - is it worth a test case which exposes that?

test/export/test_custom_def/expected_repo/build_defs/BUILD_FILE
DuBento: It was originally done to validate that an unbound source would also be trimmed, but I've added the filegroup as it makes more sense to test that it is fully trimmed.

test/export/BUILD
DuBento: Agreed, child gets overloaded quickly. Thanks.

test/export/BUILD
DuBento: ~The source_repo for that test includes a `go_repo` for a package that is not used and expected to be trimmed. It might not be worth having a separate test for that. I can always add an unused dependency to `export_go_target_with_go_dep`~
Yep, will make one test out of the 2.

test/export/BUILD
DuBento: Updated.
Indeed it would be good to follow the same format but the "output export" doesn't produce a repo, the resulting directory simply includes the output files of those targets (i.e. no .plzconfig or BUILD files). Because of this I'm not reusing the build_def but I've updated the test do validate against the expected files.

test/export/please_export_e2e_test.build_defs
DuBento: If not included the tests fail with "go not found in the path" error.
My understanding with this is that we are injecting from the current go config (plz config of the caller) into the test repo, forwarding the go binary from `//third_party/go:toolchain`.
I based it on this test: https://sourcegraph.com/r/github.com/thought-machine/please/-/blob/test/proto_plugin/BUILD?L10-15

DuBento: Another possible alternative could be to specify a go_toolchain in the source_repo to make it self-contained and closer to a real plz repo. You were probably hinting at that from the beginning. I'll check if this is possible.

DuBento: Updated.
It works but let me know if this is what you were thinking, I'm slightly concern about tests starting to fail because of changes to go_toolchain (or versions) but I don't see that happening often.

test/export/please_export_e2e_test.build_defs
DuBento: Updated. Refer to comment above.

test/export/please_export_e2e_test.build_defs
toastwaffle: You give me too much credit saying I was hinting at that! I just saw something that looked odd! 

I think having the go toolchain defined inside the test repo is the right move though; hopefully it won't be a maintenance burden.

toastwaffle (approved):

PR #3510, opened 2026-03-26
Improve UX when PATH-resolved plz binary differs from updated version

sean-:
## Summary

- Suppress the "Upgrading Please from version X to Y" message when the target version already exists in `~/.please/`. Previously this printed on every  invocation when the PATH binary (e.g. from Homebrew) was older than the pinned version, even though no download was needed.
- After `plz update`, warn if the binary resolved by PATH is not the one that was just updated. This helps users discover that e.g. `/opt/homebrew/bin/plz` is shadowing `~/.please/plz` and they need to fix their PATH.

## Reproducer

```bash
# With /opt/homebrew/bin/plz (v17.0.0) first in PATH
# and ~/.please/17.29.1/ already installed:

mkdir -p /tmp/plz-repro
cat > /tmp/plz-repro/.plzconfig << 'EOF'
[please]
version = 17.29.1
EOF
cat > /tmp/plz-repro/BUILD << 'EOF'
genrule(
    name = "hello",
    cmd = "echo hello > $OUT",
    outs = ["hello.txt"],
)
EOF

cd /tmp/plz-repro
plz --version          # Please version 17.0.0
plz build //:hello     # "Upgrading Please from version 17.0.0 to 17.29.1" (every time)
plz --version          # Still "Please version 17.0.0"

### DIFF
--- a/src/update/BUILD
+++ b/src/update/BUILD
@@ -48,6 +48,7 @@ go_test(
         "///third_party/go/github.com_sigstore_sigstore//pkg/cryptoutils",
         "///third_party/go/github.com_sigstore_sigstore//pkg/signature",
         "///third_party/go/github.com_stretchr_testify//assert",
+        "///third_party/go/github.com_stretchr_testify//require",
         "///third_party/go/gopkg.in_op_go-logging.v1//:go-logging.v1",
         "//src/cli",
         "//src/core",
--- a/src/update/update.go
+++ b/src/update/update.go
@@ -15,6 +15,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strconv"
@@ -90,6 +91,9 @@ func CheckAndUpdate(config *core.Configuration, updatesEnabled, updateCommand, f
 
 	// Clean out any old ones
 	clean(config, updateCommand)
+
+	// Warn if the binary in PATH isn't the one we just updated.
+	warnIfNotInPath(config)
 
 	// Now run the new one.
 	core.ReturnToInitialWorkingDir()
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by toastwaffle** [comment]
I'm not sure I follow what the problem is here are you trying to force Please to be used at whatever version Homebrew installs?

We don't expect `~/.please` to be in the path at all; whatever you have installed as `plz` on your `$PATH` (be it `pleasew`, `please_shim`, or an actual full Please binary) takes care of ensuring that the version you want (according to the repository config) exists in `~/.please`, and then `exec`s it

For all we care, the installed `plz` resolved on your `$PATH` can be Please v1.0.0, but if your `.plzconfig` specifies `version = 17.29.1` that's the version which will be used to execute your build (and if you don't specify that, the default I think is to use the latest available version).

**Comment by sean-** [comment]
The bootstrap/exec model makes sense and the correct version does end up running the build. The issue is purely a UX one. Here's a minimal repro:

```
# Setup: /opt/homebrew/bin/plz is v17.0.0 and appears first in PATH
# ~/.please/17.29.1/ already exists from a prior update

mkdir -p /tmp/plz-repro
cat > /tmp/plz-repro/.plzconfig << 'EOF'
[please]
version = 17.29.1
EOF
cat > /tmp/plz-repro/BUILD << 'EOF'
genrule(
    name = "hello",
    cmd = "echo hello > $OUT",
    outs = ["hello.txt"],
)
EOF

cd /tmp/plz-repro
plz --version          # Please version 17.0.0
plz build //:hello     # "Upgrading Please from version 17.0.0 to 17.29.1"
plz build //:hello     # Same message again
plz --version          # Still "Please version 17.0.0"
```

The current results are:
```
$ cd /tmp/plz-repro
$ plz --version
Please version 17.0.0
$ plz build //:hello
Upgrading Please from version 17.0.0 to 17.29.1
Build finished; total time 10ms, incrementality 100.0%. Outputs:
//:hello:
  plz-out/gen/hello.txt
$ plz build //:hello
Upgrading Please from version 17.0.0 to 17.29.1
Build finished; total time 10ms, incrementality 100.0%. Outputs:
//:hello:
  plz-out/gen/hello.txt
$ plz --version
Please version 17.0.0
```

The build works correctly every time — 17.29.1 runs it. But:

1. `plz --version` always reports `17.0.0` because that's the homebrew binary
2. `Upgrading Please from version 17.0.0 to 17.29.1` prints on every single invocation, even though `17.29.1` is already installed in `~/.please/`
3. `plz update` says `Up to date (version 17.29.1)` but `plz --version` immediately after says `17.0.0`

None of this is broken, but it's surprising. A user who sees `Upgrading` on every build naturally thinks something is wrong, and the `--version` / `update` mismatch is confusing.

A small warning after `plz update` pointing out that the `PATH`-resolved binary differs from the updated one would help users understand what's going on and how to fix it.

And actually, let me also fix that second part, too.  Now, when the target version already exists in `~/.please/`, we skip the `Upgrading Please from version X to Y` message. Currently it prints on every non- `plz update` invocation because the bootstrap binary's version doesn't match the config, even though the target is already downloaded and just needs to be exec'd. This makes the common case silent.

toastwaffle:
The best solution to fixing the UX here is to make Homebrew install `please_shim` rather than Please itself, such that you no longer have a version of Please on your path that doesn't get updated. The script at https://get.please.build/ installs `pleasew` as `plz`, although it should probably be using `please_shim`.

As it is, your change is going to add confusing/misleading messages for anybody using `please_shim` (e.g. Thought Machine). In addition, the fix your message suggests to add entries to `PATH` will never work, because the binaries in `~/.please` are all called `please`, not `plz`.

The message shown when "updating" even when the version already exists in `~/.please` is actually quite useful, as it highlights when you have different repos or branches with different Please versions (although the wording could be changed to "Switching Please version to x.y.z" to distinguish from actually downloading a new version). If you really want to suppress that message, please add a config field to do so.

I do wonder if the `please` binary should actually have updating disabled by default, although the question is then what it should do if its version does not match a version specified in `.plzconfig`. That would solve an annoyance for local testing, where you always have to use `--noupdate` to prevent your locally-built please from replacing itself with a version that doesn't have the changes you're testing.

sean-:
Good catch on the binary name: the binaries in `~/.please/` are `please` not `plz`, so the `PATH` suggestion was wrong. I've fixed this by creating a `plz` -> `please` relative symlink in `~/.please/` if it doesn't exist, so the `PATH` advice now actually works (which is how my environment invariably always gets set up, and has been running for years).

On the "Upgrading" message: the change only suppresses it when the target version already exists in `~/.please/` (i.e., nothing is being downloaded and nothing is upgrading). A "Switching to version X.Y.Z" wording for that case would be a reasonable follow-up and I could add that to this PR if you'd like.

On making Homebrew install `please_shim`: I think that's the wrong direction. A package-manager-installed binary that silently replaces itself based on a remote config file bypasses the verification and reproducibility that package managers exist to provide. That's a real supply-chain concern. It works well as an opt-in for orgs that want project-pinned versions, but users who installed via a package manager reasonably expect their package manager to be managing the binary. `plz update --version=X` already covers explicit version management. The right default for package-manager installs is "use what you installed" with `.plzconfig` overrides available but not silently self-modifying.  

There's good precedent for version managers (e.g. `tfenv`, `rbenv`, `goenv`) that let users explicitly switch versions, but none of them do so implicitly based on a project config file without the user's knowledge.  That's the distinction here.

PR #3512, opened 2026-03-26
Fix race in CloseResults that drops test failure events

sean-:
CloseResults could close the external results channel while forwardResults still had pending items to forward from the internal channel. This caused a `send on closed channel` panic that was silently recovered, dropping the result. On FreeBSD CI this manifested as `shell_output_test` failing because the test failure event was lost, leaving `failedTargets` empty and suppressing the detailed error output.

Fix by draining the internal channel before closing, and setting the external channel to nil after close so late arrivals are skipped rather than panicking.

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"bytes"
+	"context"
 	"crypto/sha1"
 	"crypto/sha256"
 	"fmt"
@@ -287,7 +288,6 @@ type stateProgress struct {
 	numParses  atomic.Int64
 	mutex      sync.Mutex
 	closeOnce  sync.Once
-	resultOnce sync.Once
 	// Used to track subinclude() calls that block until targets are built. Keyed by their label.
 	pendingTargets *cmap.Map[BuildLabel, chan struct{}]
 	// Used to track general package parsing requests. Keyed by a packageKey struct.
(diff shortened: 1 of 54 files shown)


sean-:
The `build-linux-alt` failure is likely due to https://github.com/thought-machine/please/pull/3510 not being merged yet (which has the fix for the rate limit).

src/core/state.go
```diff
@@ -424,12 +425,27 @@ func (state *BuildState) Stop() {
...
+		// Set to nil so forwardResults won't attempt to send on the closed
+		// channel if it reads a late result from internalResults.
+		state.progress.results = nil
+	}
+}
+
+// drainResults waits until all pending results in the internal channel have
+// been forwarded to the external results channel.
+func (state *BuildState) drainResults() {
+	for len(state.progress.internalResults) > 0 {
```
toastwaffle: I am concerned that this could lead to a deadlocked process - would it make sense to add a watchdog timer here to make it abandon this attempt to clean up?

sean-: Yeah, the spin loop was racy in a second, subtler way too: even after `len(internalResults)` hit 0, `forwardResults` could have read a result but not yet forwarded it under the mutex, so `CloseResults` could still `nil` out the channel and drop it.

Rather than adding a watchdog timer (which would mask the problem rather than fix it), I've reworked this to use context-based shutdown:

- `stateProgress` now has a `workerCtx` (cancelled in `Stop()`) and a `dispatcherCtx` (cancelled in `CloseResults()`)
- `forwardResults` selects on `ctx.Done()` at every blocking point.  When cancelled, it drains remaining items via a non-blocking select loop, then closes the results channel itself
- `CloseResults` just cancels the dispatcher context and waits on a `WaitGroup` for `forwardResults` to finish

This eliminates the spin loop, the recover/panic pattern, and the race window entirely. The shutdown ordering is: workers finish -> `wg.Wait()` in `plz.go` -> `CloseResults()` cancels dispatcher -> dispatcher drains and closes.

One thing I'd like to follow up on separately: `NewBuildState()` currently creates its own `context.Background()` internally. Ideally it would accept a `context.Context` parameter so callers can propagate cancellation from signal handlers, but that's a larger change I'd rather do in a follow-up.

PR #3519, opened 2026-04-14
fix go coverage when parsing blocks out of order

goddenrich:
previously we were assuming the last block contained the greatest end line. This is not always the case if we parse blocks out of order. Doing so causes plz cover to panic.


SpangleLabs (approved):

SpangleLabs:
This seems to fix some issues we were seeing on the internal repo, and looks good, thank you!

PR #3520, opened 2026-04-15
feat: minimal sandbox support in remote execution

njskalski:
Hi,

I am using a custom sandbox and remote execution. This is a minimal patch that enables just that.

I believe most people have a separate profile for remote execution, where they can disable sandbox to keep the old semantics if desired.

### DIFF
--- a/src/remote/action.go
+++ b/src/remote/action.go
@@ -116,7 +116,7 @@ func (c *Client) buildCommand(target *core.BuildTarget, inputRoot *pb.Directory,
 			Arguments: []string{
 				"fetch", strings.Join(target.AllURLs(state), " "), "verify", strings.Join(target.Hashes, " "),
 			},
-			EnvironmentVariables: c.buildEnv(target, map[string]string{}, false),
+			EnvironmentVariables: c.buildEnv(target, map[string]string{}, process.NoSandbox),
 			OutputPaths:          outs,
 		}, nil
 	}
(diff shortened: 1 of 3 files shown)


src/remote/action.go
```diff
@@ -572,10 +572,31 @@ func reallyTranslateOS(os string) string {
...
+
 // buildEnv translates the set of environment variables for this target to a proto.
 func (c *Client) buildEnv(target *core.BuildTarget, env core.BuildEnv, sandbox bool) []*pb.Command_EnvironmentVariable {
 	if sandbox {
 		env["SANDBOX"] = "true"
+		if c.state.Config.Sandbox.Tool != "" {
+			// Mirror what local execution sets so the sandbox tool sees the same interface.
+			// SHARE_NETWORK/SHARE_MOUNT=0 means "don't share" i.e. sandbox that namespace.
+			env["SHARE_NETWORK"] = "0"
+			env["SHARE_MOUNT"] = "0"
```
sean-: This seems brittle.  Shouldn't these be attributes of the named sandbox somehow?

njskalski: You are right. I passed them down and used here. Thank you for reviewing!

njskalski:
@sean- @toastwaffle Hi. I have addressed all comments you made. The feature is gated behind non-default setting. Can we merge it?

peterebden (commented):
Pretty late on this one but it looks interesting...

I think the expectation here is that the remote worker will have the sandbox tool available? That is probably not unreasonable.

FWIW the approach we tool is that the remote workers always sandbox actions according to their own policies (and they know about the `SANDBOX` env var for some things), but that does require them to cooperate more.

PR #3521, opened 2026-04-15
Update docker images used for CICD, to support golang 1.26

SpangleLabs:
This is extracted from #3498, hopefully should work independently of those changes, so we can land it before those


SpangleLabs:
Perhaps this cannot be extracted from #3498, dang

SpangleLabs:
Abandoning this, #3498 will do it

PR #3523, opened 2026-04-15
Add `git_tags` built-in function

chrisnovakovic:
`git_tags()` returns a list consisting of the tags that point at the repo's current commit.

### DIFF
--- a/rules/builtins.build_defs
+++ b/rules/builtins.build_defs
@@ -244,6 +244,8 @@ def git_show(fmt:str) -> str:
     fail('Disabled in config')
 def git_state(clean_label:str="clean", dirty_label:str="dirty") -> str:
     fail('Disabled in config')
+def git_tags() -> list:
+    fail("Disabled in config")
 
 def debug(args):
     pass
--- a/src/parse/asp/exec.go
+++ b/src/parse/asp/exec.go
@@ -43,16 +43,23 @@ func init() {
 	execPromises = make(map[execKey]*execPromise, initCacheSize)
 }
 
-// doExec executes a command and returns what it outputted on stdout. cmdIn is the command to
-// execute followed by the command's arguments, either as a list or a space-delimited string. If
-// cacheOutput is true, the command's output is memoised to prevent side effects and improve the
+// doExec executes a command and returns what it outputted on stdout.
+//
+// cmdIn is the command to execute followed by the command's arguments, either as a list or a
+// space-delimited string.
+//
+// If cacheOutput is true, the command's output is memoised to prevent side effects and improve the
 // performance of duplicate calls to the same command with the same arguments (e.g. `git rev-parse
 // --short HEAD`); to gain any benefit from this, the output from executed commands must be
-// reproducible. If storeNegative is true, the command's output will be memoised (and doExec will
-// return success = true) even if the command exits with a non-zero exit code; this is useful when
-// a command is expected to fail. If outputAsList is true, pyObj will be a pyList consisting of one
-// pyStr per line of the command's output (i.e. split on newlines); if outputAsList is false, pyObj
-// will be the command's entire output in a single pyStr.
+// reproducible.
+//
+// If storeNegative is true, the command's output will be memoised (and doExec will return with
+// success = true) even if the command exits with a non-zero exit code; this is useful when
+// a command is expected to fail.
+//
+// If outputAsList is true, pyObj will be a pyList consisting of one pyStr per line of the command's
+// output (i.e. split on newlines); if outputAsList is false, pyObj will be the command's entire
+// output in a single pyStr.
 //
 // NOTE: Commands that rely on the current working directory must not have their output cached.
 func doExec(s *scope, cmdIn pyObject, cacheOutput, storeNegative, outputAsList bool) (pyObj pyObject, success bool, err error) {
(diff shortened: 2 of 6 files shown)


rules/builtins.build_defs
```diff
@@ -244,6 +244,8 @@ def git_show(fmt:str) -> str:
     fail('Disabled in config')
 def git_state(clean_label:str="clean", dirty_label:str="dirty") -> str:
     fail('Disabled in config')
+def git_tags() -> list:
+    fail("Disabled in config")
```
toastwaffle, line 248: I appreciate you're following the existing pattern, but do you think we should be explicit here about it being `parse.GitFunctions` in the config?

src/parse/asp/exec.go
```diff
@@ -43,17 +43,19 @@ func init() {
...
+// reproducible. If storeNegative is true, the command's output will be memoised (and doExec will
+// return success = true) even if the command exits with a non-zero exit code; this is useful when
+// a command is expected to fail. If outputAsList is true, pyObj will be a pyList consisting of one
+// pyStr per line of the command's output (i.e. split on newlines); if outputAsList is false, pyObj
+// will be the command's entire output in a single pyStr.
 //
-// NOTE: Commands that rely on the current working directory must not be cached.
-func doExec(s *scope, cmdIn pyObject, cacheOutput bool, storeNegative bool) (pyObj pyObject, success bool, err error) {
+// NOTE: Commands that rely on the current working directory must not have their output cached.
+func doExec(s *scope, cmdIn pyObject, cacheOutput, storeNegative, outputAsList bool) (pyObj pyObject, success bool, err error) {
```
toastwaffle, line 65: I'm not suggesting you fix this now, but why on earth does this take a pyObject? As far as I can tell, it is only used with hard-coded lists of commands...

src/parse/asp/exec.go
```diff
@@ -189,7 +207,7 @@ func execGitBranch(s *scope, args []pyObject) pyObject {
 
 	cacheOutput := true
 	storeNegative := true
-	gitSymRefResult, success, err := doExec(s, pyList(cmdIn), cacheOutput, storeNegative)
+	gitSymRefResult, success, err := doExec(s, pyList(cmdIn), cacheOutput, storeNegative, false)
```
toastwaffle:
```suggestion
	outputAsList := false
	gitSymRefResult, success, err := doExec(s, pyList(cmdIn), cacheOutput, storeNegative, outputAsList)
```

ditto below

src/parse/asp/exec.go
```diff
@@ -330,6 +348,34 @@ func execGitState(s *scope, args []pyObject) pyObject {
...
+		pyString("tag"),
+		pyString("-l"),
+		pyString("--sort=refname"),
+		pyString("--points-at"),
+		pyString("HEAD"),
+	}
+
+	// No special error handling required here - `git tag` exits with exit code 0 even if no tags
+	// point at HEAD.
+	pyResult, success, err := doExec(s, pyList(cmdIn), true, false, true)
```
toastwaffle, line 363:
```suggestion
	cacheOutput := true
	storeNegative := false
	outputAsList := true
	pyResult, success, err := doExec(s, pyList(cmdIn), cacheOutput, storeNegative, outputAsList)
```

src/parse/asp/exec.go
```diff
@@ -330,6 +348,34 @@ func execGitState(s *scope, args []pyObject) pyObject {
...
+	if !success {
+		return s.Error("git_tags() failed: %v", err)
+	}
+
+	// If no tags point at HEAD, return an empty list, rather than the list consisting of a single
+	// empty string element that doExec returned to us.
+	if list, _ := pyResult.(pyList); list.Len() == 1 && list.Item(0) == pyString("") {
```
toastwaffle:
```suggestion
	if list, ok := pyResult.(pyList); ok && list.Len() == 1 && list.Item(0) == pyString("") {
```

Is this too defensive?

src/parse/asp/exec.go
```diff
@@ -43,17 +43,19 @@ func init() {
...
+// doExec executes a command and returns what it outputted on stdout. cmdIn is the command to
+// execute followed by the command's arguments, either as a list or a space-delimited string. If
+// cacheOutput is true, the command's output is memoised to prevent side effects and improve the
+// performance of duplicate calls to the same command with the same arguments (e.g. `git rev-parse
+// --short HEAD`); to gain any benefit from this, the output from executed commands must be
+// reproducible. If storeNegative is true, the command's output will be memoised (and doExec will
+// return success = true) even if the command exits with a non-zero exit code; this is useful when
+// a command is expected to fail. If outputAsList is true, pyObj will be a pyList consisting of one
+// pyStr per line of the command's output (i.e. split on newlines); if outputAsList is false, pyObj
+// will be the command's entire output in a single pyStr.
```
toastwaffle: This could do with some line breaks/bullet points

toastwaffle (approved):

src/parse/asp/exec.go
chrisnovakovic: My guess is that this function was originally written with the intention of it being exposed directly to ASP in future, but that (thankfully) never happened.

rules/builtins.build_defs
chrisnovakovic: Yes - I'll do that for all of them in a follow-up PR.

src/parse/asp/exec.go
chrisnovakovic: Perhaps (the return value of `doExec` is guaranteed to be a `pyList`), but it can't hurt to be sure.

rules/builtins.build_defs
chrisnovakovic: #3525

PR #3527, opened 2026-04-20
Upgrade google.golang.org/grpc to v1.80.0

chrisnovakovic:
This fixes CVE-2026-33186.

google.golang.org/grpc's `Dial` method has been deprecated since v1.63.0; use its replacement, `NewClient`, instead. There are subtle differences between them, but for our purposes they shouldn't matter.


toastwaffle (approved):

peterebden:
AFAICT this was done in #3546 - shall we close this?

PR #3529, opened 2026-04-26
Output default values for plugin config fields from plz query config

marcuscaisey:
## Problem
`plz query config` correctly outputs the values of a plugin config field if it has been explicitly set in `.plzconfig`. However, it errors if not. 

For example, in this repository: 
* `plz query config plugin.go.gotool` outputs `//third_party/go:toolchain|go` which has been explicitly set in [.plzconfig](https://github.com/thought-machine/please/blob/857c8aea4138d8a828442b04eb26334c547831a7/.plzconfig?plain=1#L11)
* `plz query config plugin.go.delvetool` errors with `Failed to get plugin.go.delvetool: Settable field not defined: plugin.go.delvetool`

This is annoying for external consumers of these values as they have to hardcode the default values in their logic.

## Solution
Update `config.Config` so that when we fail to look up a value from the `"plugin"` section of the configuration, we attempt to read the default values from the plugin's `.plzconfig`. 

Default values which look like build labels are normalised. For example: `plz query config plugin.go.pleasegotool` outputs `///go//tools:please_go` instead of `//tools:please_go` which was what was actually specified by the go plugin. This allows consumers of the values to not care were they came from (if they want to build the target for example).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 __pycache__
 
 .idea
+.vscode
 *.iml
 /venv
 /out
--- a/src/query/BUILD
+++ b/src/query/BUILD
@@ -14,6 +14,7 @@ go_library(
         "//src/core",
         "//src/fs",
         "//src/parse",
+        "//src/plz",
     ],
 )
 
--- a/src/query/config.go
+++ b/src/query/config.go
@@ -9,6 +9,7 @@ import (
 	"github.com/please-build/gcfg"
 
 	"github.com/thought-machine/please/src/core"
+	"github.com/thought-machine/please/src/plz"
 )
 
 // Config prints configuration settings in human-readable format.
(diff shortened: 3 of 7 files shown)


marcuscaisey:
Thanks for having a look at that build failure pebers, I missed that 🤦‍♂️

I don't really understand the cause (maybe the `build.xattrs = false` in the `linux-build-alt` `.plzconfig`?) but I can believe that your change will avoid it and it makes sense anyway since the test doesn't actually need the thing the label points to to exist.

Now the failure we're hitting in the `build-darwin` job is the race detector sounding the alarm (different from #3556). Seems unrelated to my change though since I see the same failure on my arm mac on master (also building plz wouldnt go through the code path i changed). I ran the pipeline myself and all of the jobs passed.

peterebden:
Okay took a while but I think that failure is fixed now. Have rebased this and hopefully this time it should work 🤞

PR #3543, opened 2026-07-08
Update hash to remove any built targets unless we run in update mode

tmjlewis:
Currently run plz hash will build it's targets in the local cache in /plz-out., Even if the target built doesn't match it's hash in the build rule. 

It is still tagged with a valid `user.plz_build`, so the next time we run `plz build` we don't rebuild that target  thinking it's valid based on this rule hash attached to it.

To get around this `plz hash` now removes any targets built from the cache if it wasn't ran in update mode. 

If it's update mode there is no issue because the hash of the target will be used to update the rule, so there is no issue with them falling out of sync.


.gitignore
```diff
@@ -20,3 +20,5 @@ __pycache__
 
 # The parse perf test repo
 /tree
+
+.vscode
```
toastwaffle: plz add trailing newline

src/please.go
```diff
@@ -508,6 +508,8 @@ var buildFunctions = map[string]func() int{
 			}
 			if opts.Hash.Update {
 				hashes.RewriteHashes(state, state.ExpandOriginalLabels())
+			} else {
+				clean.Targets(state, state.ExpandOriginalLabels())
```
toastwaffle, line 516: Add a comment to explain why (because our jira tickets aren't public)

toastwaffle (changes requested):

src/please.go
```diff
@@ -508,6 +508,11 @@ var buildFunctions = map[string]func() int{
 			}
 			if opts.Hash.Update {
 				hashes.RewriteHashes(state, state.ExpandOriginalLabels())
+			} else {
+				// plz hash was building targets to the cache and tagging it with a valid
+				// user.plz_build xattr even if the computed hash didn't match the hash in the rule.
+				// Make sure we clean up any built targets if we aren't updating the hash aswell
```
toastwaffle: Comments in the past tense are generally not a good idea

```suggestion
				// The code above will build the target(s) and unconditionally write to
				// the cache. If we're not updating the hash, we need to clear the target(s)
				// from the cache again to ensure that future builds depending on the
				// target(s) are not able to proceed despite a potentially mismatched hash.
```

toastwaffle (changes requested):

.gitignore
tmjlewis: There is a new line now, but it doesn't show in the diff. You can see the new line if you edit the file.

toastwaffle (approved):

PR #3546, opened 2026-07-13
Upgrade all the Go toolchain, packages and Please plugin

toastwaffle:


src/remote/remote.go
SpangleLabs: Do you wanna fix this lint error? Or leave it for later?

SpangleLabs (approved):

PR #3547, opened 2026-07-17
Allow __TEST_ARGS__ placeholder in test commands

chris-franklin:
Test args are currently appended to the end of the test command. This may result in an invalid command though, as is the case with go_test, which pipes the output into `tee`, so while you want to pass the arguments to your test binary, they actually get passed to `tee`.

This introduces a `__TEST_ARGS__` placeholder which can be put into the test command, and will be used as the location to put the test args, instead of just appending them.

Tested on a local repository that uses please and a fork of go-rules that made use of the placeholder. The args could
be passed to test via `-- --test-arg`, though it did interfere with test name pattern matching which necessitated appending a `.` on the end. That's a separate issue though and doesn't invalidate this change.

### DIFF
--- a/src/test/test_step.go
+++ b/src/test/test_step.go
@@ -343,7 +343,14 @@ func pluralise(word string, quantity int) string {
 func testCommandAndEnv(state *core.BuildState, target *core.BuildTarget, run int) (string, core.BuildEnv, error) {
 	replacedCmd, err := core.ReplaceTestSequences(state, target, target.GetTestCommand(state))
 	env := core.TestEnvironment(state, target, filepath.Join(core.RepoRoot, target.TestDir(run)), run)
-	if len(state.TestArgs) > 0 {
+	if strings.Contains(replacedCmd, "__TEST_ARGS__") {
+		args := ""
+		if len(state.TestArgs) > 0 {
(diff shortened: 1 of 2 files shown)


src/test/test_step.go
```diff
@@ -343,7 +343,14 @@ func pluralise(word string, quantity int) string {
...
 	replacedCmd, err := core.ReplaceTestSequences(state, target, target.GetTestCommand(state))
 	env := core.TestEnvironment(state, target, filepath.Join(core.RepoRoot, target.TestDir(run)), run)
-	if len(state.TestArgs) > 0 {
+	if strings.Contains(replacedCmd, "__TEST_ARGS__") {
+		args := ""
+		if len(state.TestArgs) > 0 {
+			args = strings.Join(state.TestArgs, " ")
+		}
+		newCmd := strings.ReplaceAll(replacedCmd, "__TEST_ARGS__", args)
+		replacedCmd = newCmd
```
toastwaffle:
```suggestion
		replacedCmd = strings.ReplaceAll(replacedCmd, "__TEST_ARGS__", args)
```

toastwaffle (approved):

PR #3552, opened 2026-07-22
Add support for test arguments to remote test commands

chris-franklin:
This adds support for `BuildState.TestArgs` to remote builds. This includes support for `TestFields.ArgsPlaceholder`.
It does this by introducing a central function for doing it, which wraps the call to `ReplaceTestSequences`.

It also updates `shell_output.go` so that the correct command is shown when using `--shell`.

The remote test build that was previously failing due to the presence of the placeholder passed when using this change, and using `--shell` also showed the correct command.

### DIFF
--- a/src/core/command_replacements.go
+++ b/src/core/command_replacements.go
@@ -59,7 +59,7 @@ import (
 	"runtime/debug"
 	"strings"
 
-	"github.com/peterebden/go-deferred-regex"
+	deferredregex "github.com/peterebden/go-deferred-regex"
 
 	"github.com/thought-machine/please/src/fs"
 )
@@ -106,6 +106,8 @@ func TestCommand(state *BuildState, target *BuildTarget) (string, error) {
 				args = strings.Join(state.TestArgs, " ")
 			}
 			cmd = strings.ReplaceAll(cmd, placeholder, args)
+		} else {
+			return "", fmt.Errorf("command %q does not contain expected arguments placeholder %q", cmd, target.Test.ArgsPlaceholder)
 		}
 	} else if len(state.TestArgs) > 0 {
 		cmd += " " + strings.Join(state.TestArgs, " ")
--- a/src/core/command_replacements.go
+++ b/src/core/command_replacements.go
@@ -100,15 +100,14 @@ func TestCommand(state *BuildState, target *BuildTarget) (string, error) {
 	}
 	if target.Test != nil && target.Test.ArgsPlaceholder != "" {
 		placeholder := target.Test.ArgsPlaceholder
-		if strings.Contains(cmd, placeholder) {
-			args := ""
-			if len(state.TestArgs) > 0 {
-				args = strings.Join(state.TestArgs, " ")
-			}
-			cmd = strings.ReplaceAll(cmd, placeholder, args)
-		} else {
+		if !strings.Contains(cmd, placeholder) {
 			return "", fmt.Errorf("command %q does not contain expected arguments placeholder %q", cmd, target.Test.ArgsPlaceholder)
 		}
+		args := ""
+		if len(state.TestArgs) > 0 {
+			args = strings.Join(state.TestArgs, " ")
+		}
+		cmd = strings.ReplaceAll(cmd, placeholder, args)
 	} else if len(state.TestArgs) > 0 {
 		cmd += " " + strings.Join(state.TestArgs, " ")
 	}
--- a/src/core/command_replacements.go
+++ b/src/core/command_replacements.go
@@ -92,6 +92,27 @@ func ReplaceTestSequences(state *BuildState, target *BuildTarget, command string
 	return replaceSequencesInternal(state, target, command, true)
 }
 
+// TestCommand returns the command to run for a test target, with all sequences and test arguments processed.
+func TestCommand(state *BuildState, target *BuildTarget) (string, error) {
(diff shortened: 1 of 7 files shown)


src/core/command_replacements_test.go
```diff
@@ -297,3 +297,89 @@ func (h *testHasher) OutputHash(target *BuildTarget) ([]byte, error) {
 
 func (h *testHasher) SetHash(target *BuildTarget, hash []byte) {
 }
+
+func TestTestCommand(t *testing.T) {
+	state := NewDefaultBuildState()
+
+	t.Run("Case 1: No TestArgs", func(t *testing.T) {
```
toastwaffle: Is there convention elsewhere to include the "Case N"? Feels odd to me

src/output/shell_output.go
```diff
@@ -421,20 +421,22 @@ func printTempDirs(state *core.BuildState, duration time.Duration, shell, shellR
...
 	for _, label := range state.ExpandVisibleOriginalTargets() {
 		target := state.Graph.TargetOrDie(label)
-		cmd := target.GetCommand(state)
+		var cmd string
 		dir := target.TmpDir()
 		env := core.StampedBuildEnvironment(state, target, nil, filepath.Join(core.RepoRoot, target.TmpDir()), target.Stamp)
 		shouldSandbox := target.Sandbox
 		if state.NeedTests {
-			cmd = target.GetTestCommand(state)
+			cmd, _ = core.TestCommand(state, target)
```
toastwaffle: please at least log this error

ditto below

src/remote/remote_test.go
```diff
@@ -353,3 +353,26 @@ func (c *Client) Store(target *core.BuildTarget) error {
...
+		ArgsPlaceholder: "__TEST_ARGS__",
+	}
+	target.IsBinary = true
+
+	cmd, err := c.buildTestCommand(state, target, 1)
+	assert.NoError(t, err)
+
+	assert.True(t,
+		strings.HasSuffix(cmd.Arguments[len(cmd.Arguments)-1], "$TEST --foo --bar 2>&1"),
+		"incorrect suffix on %q", cmd.Arguments[len(cmd.Arguments)-1],
```
toastwaffle: please make the error include the suffix we expected to see

src/core/command_replacements_test.go
```diff
@@ -297,3 +297,89 @@ func (h *testHasher) OutputHash(target *BuildTarget) ([]byte, error) {
...
+			Command:         "python -m unittest __TEST_ARGS__ 2>&1",
+			ArgsPlaceholder: "__TEST_ARGS__",
+		}
+		state.TestArgs = []string{"foo", "bar"}
+		cmd, err := TestCommand(state, target)
+		assert.NoError(t, err)
+		assert.Equal(t, "python -m unittest foo bar 2>&1", cmd)
+	})
+
+	t.Run("Case 4: Placeholder is specified but not present in the command", func(t *testing.T) {
```
toastwaffle: I think this case should actually be an error, but we'll see what @chrisnovakovic thinks

toastwaffle (changes requested):

src/output/shell_output.go
```diff
@@ -421,20 +421,26 @@ func printTempDirs(state *core.BuildState, duration time.Duration, shell, shellR
...
 			shouldSandbox = target.Test.Sandbox
 			if len(state.TestArgs) > 0 {
 				env["TESTS"] = strings.Join(state.TestArgs, " ")
 			}
+		} else {
+			cmd = target.GetCommand(state)
+			cmd, err = core.ReplaceSequences(state, target, cmd)
+		}
+		if err != nil {
+			log.Errorf("Error pre-processing command: %s", err.Error())
```
chris-franklin, line 442: There didn't seem like an obviously-correct way to output the error

src/output/shell_output.go
chris-franklin: I went for logging here, but in the other places I relay it upwards, and it does seem like that will eventually be logged.

src/core/command_replacements.go
```diff
@@ -92,6 +92,29 @@ func ReplaceTestSequences(state *BuildState, target *BuildTarget, command string
...
+		placeholder := target.Test.ArgsPlaceholder
+		if strings.Contains(cmd, placeholder) {
+			args := ""
+			if len(state.TestArgs) > 0 {
+				args = strings.Join(state.TestArgs, " ")
+			}
+			cmd = strings.ReplaceAll(cmd, placeholder, args)
+		} else {
+			return "", fmt.Errorf("command %q does not contain expected arguments placeholder %q", cmd, target.Test.ArgsPlaceholder)
+		}
```
toastwaffle:
```suggestion
		if !strings.Contains(cmd, placeholder) {
			return "", fmt.Errorf("command %q does not contain expected arguments placeholder %q", cmd, target.Test.ArgsPlaceholder)
		}
		args := ""
		if len(state.TestArgs) > 0 {
			args = strings.Join(state.TestArgs, " ")
		}
		cmd = strings.ReplaceAll(cmd, placeholder, args)
```

Guard clause mildly preferable?

src/output/shell_output.go
toastwaffle: Yeah, this seems reasonable to me

toastwaffle (approved):

Issue #3556, opened 2026-07-24
CircleCI build on Darwin (arm64) frequently segfaults and sometimes deadlocks


toastwaffle:
There seems to be a data race which only triggers on darwin_arm64. This is particularly bad, because sometimes Please doesn't detect it and instead gets deadlocked, and then spends 3 hours burning CircleCI credits to do nothing

```
WARNING: DATA RACE
Write at 0x00c000229b30 by goroutine 55:
  runtime.recvDirect()
      /Users/distiller/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.1.darwin-arm64/src/runtime/chan.go:405 +0x7c
  github.com/thought-machine/please/src/output.(*buildingTargets).handleOutput.(*BuildState).Stop.func1()
      /Users/distiller/project/src/core/state.go:419 +0x40
  sync.(*Once).doSlow()
      /Users/distiller/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.1.darwin-arm64/src/sync/once.go:78 +0x94
  sync.(*Once).Do()
      /Users/distiller/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.1.darwin-arm64/src/sync/once.go:69 +0x40
  github.com/thought-machine/please/src/core.(*BuildState).Stop()
      /Users/distiller/project/src/core/state.go:418 +0x2e4
  github.com/thought-machine/please/src/output.(*buildingTargets).handleOutput()
      /Users/distiller/project/src/output/targets.go:107 +0x288
  github.com/thought-machine/please/src/output.(*buildingTargets).ProcessResult.deferwrap1()
      /Users/distiller/project/src/output/targets.go:84 +0x34
  runtime.deferreturn()
      /Users/distiller/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.1.darwin-arm64/src/runtime/panic.go:668 +0x5c
  github.com/thought-machine/please/src/output.MonitorState()
      /Users/distiller/project/src/output/shell_output.go:57 +0x7c0
  main.runPlease.func1()
      /Users/distiller/project/src/please.go:1230 +0x94
Previous read at 0x00c000229b30 by goroutine 17861:
  runtime.chansend1()
      /Users/distiller/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.1.darwin-arm64/src/runtime/chan.go:162 +0x2c
[...]

PR #3557, opened 2026-07-24
Make the build-darwin CircleCI job time out after 15 minutes

toastwaffle:
Relates to #3556 

Unfortunately darwin doesn't have the coreutils `timeout` command which would make this much nicer. I considered whether we should have this timeout on every task, and then decided against it.

When it does succeed, the `build-darwin` job takes ~7 minutes, so this is a relatively generous timeout.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -222,7 +222,16 @@ jobs:
            command: ./.circleci/setup_osx.sh
        - run:
            name: Bootstrap & Build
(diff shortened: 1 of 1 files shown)


chrisnovakovic (approved):

chrisnovakovic:
https://app.circleci.com/pipelines/github/thought-machine/please/7787/workflows/e37fe7f1-b79d-4378-9cfd-5c32f7fbb9d3/jobs/62868

Well, at least we know it's working... 🙂

PR #3564, opened 2026-07-31
Add Github Action to also sync third_party/go/BUILD with changes from dependabot

peterebden:
Hopefully this will stop us from having to run after dependabot doing this ourselves.

Of course this is impossible to test without landing it.

### DIFF
--- a/.github/workflows/dependabot-sync.yml
+++ b/.github/workflows/dependabot-sync.yml
@@ -1,0 +1,28 @@
+name: Dependabot Build Sync
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  sync-build-files:
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+
+    permissions:
+      contents: write
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-sync.yml
```diff
@@ -0,0 +1,28 @@
...
+  sync-build-files:
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
chrisnovakovic:
```suggestion
        uses: actions/checkout@v7
```

Any reason not to go for `v7`? I've been seeing Node.js deprecation warnings from some of our actions, and IIRC `actions/checkout@v4` is one of them...

chrisnovakovic (approved):

.github/workflows/dependabot-sync.yml
peterebden: Nope, I just copied this.

PR #3566, opened 2026-08-03
tools/sandbox: update with new features, variants, and tests

### LINKED ISSUE #3541: Linux Sandbox: Support multi UID/GID mappings via `newuidmap` and `newgidmap`
> Is your feature request related to a problem? Please describe.

Currently, Please's Linux sandbox uses user namespaces to isolate build actions, mapping the single UID and GID of the user running plz inside the sandbox.

While this single-ID mapping works fine for standard compilation steps, it limits local developer environments when trying to run more complex tasks inside the sandbox. If a build rule or test needs to execute tools that expect multiple UIDs/GIDs (e.g., building rootless container images, using fakeroot, running package managers like apt/dnf), the process fails with EPERM (Operation not permitted). This happens because any attempt to chown files or setuid to an unmapped ID is blocked by the kernel.

> Describe the solution you'd like

Please's Linux sandbox should optionally support multi-UID and multi-GID mappings by leveraging the standard newuidmap and newgidmap utilities.

> Describe alternatives you've considered

* Disabling the sandbox: allows these tools to run, but sacrifices hermeticity.

gzvolsky-tm:
This PR brings several enhancements, reliability improvements, and tests to the sandbox:

- Add support for `SANDBOX_UID_MAP` and `SANDBOX_GID_MAP` environment variables. This allows mapping multiple UID/GID ranges into the user namespace utilizing `newuidmap` and `newgidmap`. Both must be set together.
- Add `MOUNT_PROC=0` to skip remounting `/proc`. This provides a workaround for environments (like newer systemd or certain container runtimes) that prevent remounting a partially masked `/proc`.
- Introduce `SANDBOX_FILE_MOUNTS` to allow bind-mounting a comma-separated list of specific files into the sandbox.
- The extra loopback IP address (`10.1.1.1`, added in #3310) can now be overridden via `SANDBOX_LOCAL_IP`, or disabled by setting it to the empty string.
- Mount a tmpfs over `/dev/shm` so nothing can be inadvertently shared through it.
- Introduce a pipe (`sync_fd`) to synchronize the parent and child processes. This prevents the child from hanging indefinitely if the parent dies before the sandbox setup is complete.
- Add `SIGTERM` forwarding from the parent to the sandboxed child process.
- Add a README documenting the sandbox, its configurable environment variables, and how to use them with Please.
- Add a `sandbox_test.go` test suite to test the network/mount isolation, `/proc` remounting, ID mapping, and process lifecycle across all combinations of the environment knobs.
- Install uidmap (shadow-uidmap on alpine) in the CI images and delegate root:0:65536 in /etc/subuid and /etc/subgid, since the sandbox execs newuidmap/newgidmap even for root. The sandbox tests must stay `manual`/`localonly` until the images are updated.

## Behaviour changes

- `/proc` is now mounted read-only. This may affect tests that write to it, e.g. to `/proc/self/oom_score_adj`.
- The `TMP_DIR` bind mount is now recursive (`MS_REC`), so submounts of `TMP_DIR` are visible inside the sandbox.
- When `TMP_DIR` is `/tmp` or lives under it, the sandbox no longer returns early: it still bind-mounts it to `/tmp/plz_sandbox`, remounts the root read-only, and changes into it.
- The root read-only remount now preserves the existing `nosuid`/`nodev`/`noexec`/`noatime` mount flags instead of silently dropping them.

## Breaking change

- Removing `nonet_sandbox` is a breaking change for downstream repositories that referenced it, and for remote execution workers that invoked a variant binary and relied on its defaults. The migration path is to use a wrapper that defines the environment (see tools/sandbox/README.md).

Resolves #3541


peterebden:
Maybe let's think about what the structure of these binaries should look like. I'm not sure that it's very optimal to ship multiple different ones with Please for the different combinations of things that it could sandbox - maybe that should be some flag / env var thing instead on a single binary.

tools/sandbox/test/BUILD
```diff
@@ -0,0 +1,15 @@
+go_test(
```
toastwaffle, line 1: Why put the tests in a subdirectory?

tools/sandbox/sandbox.c
```diff
@@ -20,8 +22,12 @@
...
 #include <sys/wait.h>
-#include <linux/rtnetlink.h>
+#include <sys/statvfs.h>
+#include <limits.h>
 #include <arpa/inet.h>
+#include <linux/rtnetlink.h>
+
+static int cloned_pid;
 
 int perror_sock(char *errmsg, const int sock) {
```
toastwaffle, line 32:
```suggestion
inline int perror_sock(char *errmsg, const int sock) {
```

tools/sandbox/README.md
```diff
@@ -0,0 +1,86 @@
...
+  these arguments to newuidmap/newgidmap to configure uid/gid mappings. The format is 1..n
+  space-delimited triples of [id lowerid count], see `man newuidmap`;
+
+Mount and network namespaces can be disabled setting the `SHARE_MOUNT` and `SHARE_NETWORK`
+environment variables to `1`. Remounting of `/proc` can be disabled by setting `MOUNT_PROC=0`,
+e.g. for systems that don't allow mounting a full `/proc` from a new user namespace (see the
+mount namespace notes below).
+
+When the network namespace is used, the loopback interface is brought up with an additional IP
+address. `SANDBOX_LOCAL_IP` defines, defined empty string disables the extra address.
```
toastwaffle:
```suggestion
When the network namespace is used, the loopback interface is brought up with an additional IP
address, which defaults to 10.1.1.1. The `SANDBOX_LOCAL_IP` environment variable can be used
to change this IP, and setting it to the empty string disables the extra address entirely.
```

tools/sandbox/test/BUILD
```diff
@@ -0,0 +1,15 @@
...
+    name = "sandbox_test",
+    srcs = ["sandbox_test.go"],
+    data = {
+        "please_sandbox": "//tools/sandbox:please_sandbox",
+    },
+    labels = [
+        "localonly",
+        "manual",
+    ],
```
toastwaffle, line 11: Presumably these tests aren't currently being run by CircleCI? Can we remove these labels? Do we need to make changes to the CircleCI config?

toastwaffle (dismissed):

tools/sandbox/sandbox.c
grantzvolsky: The `inline` was removed here: https://github.com/thought-machine/please/pull/3310/changes/5b8d7108cd4d90b93df6c9f48818ad7b49f92b9c . I assume it was removed because the keyword has different meanings C/C++ and based on whether it's just `inline`, `inline static`, or `inline extern`, and removing it made the code more readable than changing it to `inline static`.

tools/sandbox/test/BUILD
grantzvolsky: We got this lint error when the C and Go files were colocated:

```
  level=error msg="[linters_context] typechecking error: C source files not allowed when not using cgo or SWIG: main.c nonet_main.c nonetproc_main.c noproc_main.c sandbox.c"
```

tools/sandbox/test/BUILD
grantzvolsky: If I removed manual right now, it would be executed in this step: https://app.circleci.com/pipelines/github/thought-machine/please/7814/workflows/d70458df-9432-4518-bcf0-06c5bde6eec3/jobs/63145/parallel-runs/0/steps/0-105 .

I would like to remove it and see whether it passes in CI once we update the CI image references to a version that includes the dependencies added in this PR. It does pass when run locally in the images defined here.

toastwaffle: Let's split out the image changes to a separate PR and get that landed first?

tools/sandbox/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+In addition to the above, when enabling the mount namespace, the sandbox will remount /proc, so any
+process that use it will have access to accurate information of the PID namespace (otherwise they'd
+still have access to /proc from the parent namespace). There's however a specific edge case in the
+Linux kernel, that prevents /proc from being remounted in a mount namespace, when the parent /proc
+is not fully visible, [see this commit](https://github.com/torvalds/linux/commit/1b852bceb0d1).
+
+This is an issue with most container runtimes (and therefore Kubernetes), as by default, they will
+hide some part of /proc in a container to reduce attack surface, eg
+[see this docker PR](https://github.com/docker/cli/pull/1808).
```
peterebden, line 87: I'm a little unclear what the conclusion of this section is? I remember the issue (remember bashing my head against it for a while) but is this something the sandbox works around, is it just a limitation of cases where it doesn't work, or do we intend to try to fix it in the future?

peterebden (commented):
This is good by me, although I like Sam's idea of splitting out the image changes so the tests will work.

tools/sandbox/README.md
grantzvolsky: The conclusion is that users of the sandbox have two options: Either unmask proc (k8s pod `procMount: Unmasked` securityContext option, which is less of an issue when combined with `hostUsers: false`), or disable the `/proc` isolation, in which case the sandboxed process will be able to see (but not send signals to) processes outside the sandbox. I'll update the README to mention this.

tools/sandbox/test/BUILD
grantzvolsky: Happy to split the image changes if you'd prefer, but while testing them I found that they're no longer a pre-requisite: the sandbox tests can't run with the docker executor that is used by the current jobs (mount syscall is blocked and CircleCI doesn't expose configuration to allow it). To work around this, I updated this PR with new jobs that run the tests using [machine executors](https://circleci.com/docs/guides/execution-managed/using-linuxvm/) and added them to the release requirements. See [example run from the latest commit](https://app.circleci.com/pipelines/github/thought-machine/please/7832/workflows/edf82a75-c459-4ec8-b504-cc279c386459/jobs/63324). The jobs currently apply the image changes at container startup. Once the images are updated, [these steps](https://github.com/thought-machine/please/pull/3566/changes#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R362-R366) can be removed (tracked in #3571), but the rest of this setup will stay.

peterebden (approved):
This LGTM now although I'm not sure that we test sandboxing extensively on CircleCI. Might need some careful testing internally to be sure it still matches expected behaviours.

PR #3569, opened 2026-08-09
Convert build parallelism to a more synchronous model

peterebden:
This will (hopefully) make the parallelism easier to understand and avoids some of the weird edge cases we had before. The fan-out logic is now all in `src/plz` and has been cut out of `src/core`. The model there is quite a lot more synchronous - essentially it calls a function like Build or Test which builds all a target's dependencies, then that target, blocking until it's done.

We can drop a few of the concepts we had before (like "parse mode", "resolving dependencies" and things like that) and we can lose the maps of "pending things" on the build state. It's simpler around things like subincludes because when we get there we just ask to build the thing, we don't have to ask it to parse then figure out at the end of that whether we should trigger things to build or not.

I ran the in-repo perf tests and they are actually faster than before. That's nice to see although it was not expected - it appears to be that we have fewer concurrently live goroutines because all the dependency resolution stuff is gone. I think it's a bit of a parse-only effect though.


src/cmap/cmap.go
```diff
@@ -138,32 +126,19 @@ func (s *shard[K, V]) Set(key K, val V, overwrite bool) bool {
...
-		s.m[key] = awaitableValue[V]{Val: v}
-		close(existing.Wait)
-		existing.Wait = nil
-		return v, true
-	}
-	v := f()
-	s.m[key] = awaitableValue[V]{Val: v}
-	return v, true
+// get returns the value for a key, or its zero value if it isn't present.
+// Unlike Get it never inserts anything, so it's safe for callers that only want to read.
```
toastwaffle: This comment doesn't match the function

src/cmap/cmap.go
```diff
@@ -138,32 +126,19 @@ func (s *shard[K, V]) Set(key K, val V, overwrite bool) bool {
...
+func (s *shard[K, V]) Get(key K) V {
+	s.l.RLock()
+	defer s.l.RUnlock()
+	return s.m[key].Val
 }
 
 // Get returns the value for a key or, if not present, a channel that it can be waited
 // on for.
 // Exactly one of the target or channel will be returned.
 // The third value is true if it is the first call that is waiting on this value.
```
toastwaffle: This comment doesn't match the function

src/cmap/cmap_test.go
```diff
@@ -26,47 +26,38 @@ func TestMap(t *testing.T) {
...
+	assert.Equal(t, 0, m.Get(5))
+	// A failed lookup must not leave an entry behind; anything that later tries to set this key
+	// would find something already waiting on it and never get to do the work.
+	assert.False(t, m.Contains(5))
+}
+
 func TestReAdd(t *testing.T) {
 	m := New[int, int](DefaultShardCount, hashInts)
 	assert.True(t, m.Add(5, 7))
 	assert.False(t, m.Add(5, 7))
```
toastwaffle:
```suggestion
	assert.False(t, m.Add(5, 8))
```

to assert that the second m.Add doesn't mutate the map?

src/core/build_target.go
```diff
@@ -558,293 +538,268 @@ func (target *BuildTarget) AllURLs(state *BuildState) []string {
...
-// TODO(peterebden, tatskaari): See if we can get rid of this.
-func (target *BuildTarget) ResolveDependencies(graph *BuildGraph) error {
-	return target.resolveDependencies(graph, func(*BuildTarget) error { return nil })
-}
-
-// DeclaredDependencies returns all the targets this target declared any kind of dependency on (including sources and tools).
-func (target *BuildTarget) DeclaredDependencies() []BuildLabel {
+// Dependencies returns the resolved dependencies of this target, applying any require/provide
+// relationships to map each declared dependency to the target(s) that actually satisfy it.
+// It requires the graph to look targets up, since a BuildTarget no longer caches these itself.
```
toastwaffle: I'm not a fan of comments which refer to historic states of the world - imo they should only refer to the current state, or any future intentions (in the form of TODOs). If you really want to reference previous behaviour, please also reference this pull request in the comment as well.

src/core/build_target.go
```diff
@@ -558,293 +538,268 @@ func (target *BuildTarget) AllURLs(state *BuildState) []string {
...
+	return ret, unresolved
 }
 
-// ExternalDependencies returns the non-internal dependencies of this target (i.e. not "_target#tag" ones).
-func (target *BuildTarget) ExternalDependencies() []*BuildTarget {
+// ExternalDependencies returns the resolved dependencies of this target, with any internal
+// dependencies (i.e. "_target#tag" ones sharing this target's parent) flattened out to the
+// external targets they in turn depend on. Require/provide relationships are applied as in Dependencies,
+// as is the second return of any dependencies that aren't in the graph.
+func (target *BuildTarget) ExternalDependencies(graph *BuildGraph) ([]*BuildTarget, []BuildLabel) {
```
toastwaffle, line 608: What purpose does `ExternalDependencies` serve? The rough conclusion from the work on export was that the `_target#tag` convention is insufficiently well used for please to rely on it in any meaningful way.

Looks like the only place this is used is for [coverage](https://sourcegraph.iap.tmachine.io/r/github.com/thought-machine/please/-/blob/src/test/coverage.go?L77). I wonder if we could get rid of it, although that is outside of the scope of this PR

peterebden:
Note to self: I think there might be something slightly off about debug tools somewhere (have observed this internally but I have little understanding yet of what the test in question is doing)

peterebden:
> Note to self: I think there might be something slightly off about debug tools somewhere (have observed this internally but I have little understanding yet of what the test in question is doing)

Nope not caused by this. I assumed a failure that obvious would be, but it seems broken on master. Fixed in #3593

src/core/build_target.go
```diff
@@ -1041,8 +1016,9 @@ func (target *BuildTarget) CanSee(state *BuildState, dep *BuildTarget) bool {
 // Returns an error if not, or nil if all's well.
 func (target *BuildTarget) CheckDependencyVisibility(state *BuildState) error {
 	for _, d := range target.dependencies {
-		dep := state.Graph.TargetOrDie(*d.declared)
-		if !target.CanSee(state, dep) {
+		if dep := state.Graph.Target(d.Label); dep == nil {
```
toastwaffle, line 1020: Why have we moved off TargetOrDie here?

src/core/build_target_test.go
```diff
@@ -1035,11 +1040,28 @@ func makeTarget1(label, visibility string, deps ...*BuildTarget) *BuildTarget {
 	}
 	for _, dep := range deps {
 		target.AddDependency(dep.Label)
-		target.resolveDependency(dep.Label, dep)
 	}
 	return target
 }
 
+// resolved unwraps a call to Dependencies / ExternalDependencies, requiring that everything resolved.
+func resolved(deps []*BuildTarget, unresolved []BuildLabel) []*BuildTarget {
```
toastwaffle: Can we make this take a `t *testing.T`, call `t.Helper()`, and then `t.Fatalf` instead of panicking?

src/core/build_target_test.go
```diff
@@ -1035,11 +1040,28 @@ func makeTarget1(label, visibility string, deps ...*BuildTarget) *BuildTarget {
...
+	if len(unresolved) > 0 {
+		panic(fmt.Sprintf("dependencies not in graph: %s", unresolved))
+	}
+	return deps
+}
+
+// graphWith returns a graph populated with the given targets, for tests that need dependency
+// resolution (which now happens against the graph rather than being cached on the target).
```
toastwaffle: ditto previous comment about comments referencing history

src/core/command_replacements_test.go
```diff
@@ -270,16 +270,13 @@ func makeTarget2(name string, command string, dep *BuildTarget) *BuildTarget {
 	target := NewBuildTarget(ParseBuildLabel(name, ""))
 	target.Command = command
 	target.AddOutput(target.Label.Name + ".py")
+	// Dependency resolution now happens against the graph at replacement time rather than being
```
toastwaffle: And again

src/core/graph.go
```diff
@@ -153,20 +139,21 @@ func (graph *BuildGraph) AllTargets() BuildTargets {
 	return targets
 }
 
-// PackageMap returns a copy of the graph's internal map of name to package.
+// PackageMap returns a map of name to package.
+// TODO(peterebden): Change this to an iterator.
```
toastwaffle, line 132: File an issue?

src/core/state.go
```diff
@@ -21,41 +21,21 @@ import (
...
-func (m ParseMode) IsForSubinclude() bool {
-	return m&ParseModeForSubinclude != 0
-}
-
 // startTime is as close as we can conveniently get to process start time.
 var startTime = time.Now()
 
-// cycleCheckDuration is the length of time we allow inactivity for before we trigger cycle detection.
-const cycleCheckDuration = 5 * time.Second
+// resultsChanSize is the buffer size of the channel we report build results on.
```
toastwaffle, line 33: Why do we use a buffered channel?

src/core/state.go
```diff
@@ -304,12 +321,8 @@ type stateProgress struct {
...
 	// True if >= 1 target has failed test cases
 	testFailed atomic.Bool
-	// Stream of results from the build
-	results chan *BuildResult
-	// Internal result stream, used to intermediate them for the cycle checker.
-	internalResults chan *BuildResult
-	// The cycle checker itself.
-	cycleDetector cycleDetector
+	// Streams of results from the build
+	results []chan *BuildResult
```
toastwaffle: Why is this a list of channels?!

src/core/utils.go
```diff
@@ -171,8 +172,20 @@ func IterInputs(state *BuildState, graph *BuildGraph, target *BuildTarget, inclu
 
 			done[dependency.Label] = true
 			if target == dependency || (target.NeedsTransitiveDependencies && !dependency.OutputIsComplete) {
-				for _, dep := range dependency.BuildDependencies() {
-					for dep2 := range recursivelyProvideFor(graph, target, dependency, dep.Label) {
+				// TODO(peterebden): We are maintaining a particular ordering here to avoid causing rebuilds.
```
toastwaffle, line 175: File an issue?

src/core/state.go
```diff
@@ -237,15 +213,70 @@ type BuildState struct {
...
+type preloadChain struct {
+	run    *PreloadRun
+	parent *preloadChain
+}
+
+type preloadChainKey struct{}
 
-	// preloadDownloadOnce is used
-	preloadDownloadOnce *sync.Once
+// WithPreloading returns a context noting that we are resolving the given repo's preloads. Anything
```
toastwaffle: There is no repo given here as the comment suggests?

src/core/utils.go
toastwaffle: It's unclear what the "this" in "push this up" means

src/parse/asp/interpreter.go
```diff
@@ -402,20 +413,22 @@ func (s *scope) subincludePackage() *core.Package {
 	return nil
 }
 
-// NewScope creates a new child scope of this one.
-func (s *scope) NewScope(filename string, mode core.ParseMode) *scope {
-	return s.newScope(s.pkg, mode, filename, 0)
+// NewScope creates a new child scope of this one, continuing on the same context.
+// Use newScope directly if the new scope belongs to a different chain of work to this one; the context
```
toastwaffle: Can we define "different chain of work" more explicitly? This feels like it's going to be an important and potentially confusing distinction

src/parse/asp/parser.go
```diff
@@ -23,6 +24,12 @@ type semaphore chan struct{}
 func (s semaphore) Acquire() { s <- struct{}{} }
 func (s semaphore) Release() { <-s }
 
+// Callbacks is the interface we require from something that we can call back to for builds / parses.
+type Callbacks interface {
+	Parse(context.Context, core.BuildLabel, core.BuildLabel) (*core.Package, error)
+	BuildAndDownload(context.Context, core.BuildLabel, core.BuildLabel) (*core.BuildTarget, error)
```
toastwaffle: Please put argument names on this interface

src/parse/init.go
```diff
@@ -16,36 +17,29 @@ import (
...
 
-// InitParser initialises the parser engine. This is guaranteed to be called exactly once before any calls to Parse().
-func InitParser(state *core.BuildState) *core.BuildState {
-	if state.Parser == nil {
-		p := &aspParser{parser: newAspParser(state)}
-		state.Parser = p
+// InitParser initialises the parser engine.
+func InitParser(state *core.BuildState, callbacks asp.Callbacks) *asp.Parser {
+	// There is some awkward coupling here for the benefit of the language server, which wants to get its
+	// hands on the parser, but it cannot create a fully functional one any more.
```
toastwaffle: any more?

src/test/coverage.go
```diff
@@ -59,7 +59,9 @@ func collectCoverageFiles(state *core.BuildState, includeAllFiles bool) map[stri
 	doneTargets := map[*core.BuildTarget]bool{}
 	coverageFiles := map[string]bool{}
 	for _, label := range state.ExpandAllOriginalLabels() {
-		collectAllFiles(state, state.Graph.TargetOrDie(label), coverageFiles, includeAllFiles, true, doneTargets)
+		if target := state.Graph.Target(label); target != nil { // It won't be if it failed to parse
```
toastwaffle: It won't be what?

src/please.go
```diff
@@ -1222,15 +1222,17 @@ func runPlease(state *core.BuildState, targets []core.BuildLabel) {
 	pretty := prettyOutput(opts.OutputFlags.InteractiveOutput, opts.OutputFlags.PlainOutput || opts.BehaviorFlags.Debug, opts.OutputFlags.Verbosity) && state.NeedBuild && !streamTests
 	state.Cache = cache.NewCache(state)
 
-	// Run the display
-	state.Results() // important this is called now, don't ask...
```
toastwaffle, line 1257: I am delighted to see the end of this comment

test/subrepo/parse_deadlock/test_repo/defs/BUILD_FILE
```diff
@@ -0,0 +1,17 @@
...
+
+build_rule(
+    name = "sub",
+    srcs = {
+        "build": ["files/sub_build"],
+        "defs": ["files/defs_build_defs"],
+    },
+    outs = ["sub"],
+    cmd = f'mkdir -p "$OUT" && cp "$SRCS_BUILD" "$OUT/{build_file_name}" && cp "$SRCS_DEFS" "$OUT/defs.build_defs"',
+    _subrepo = True,
```
toastwaffle: Why is this marked as subrepo when you also have the subrepo rule?

src/plz/plz.go
```diff
@@ -25,166 +34,623 @@ var log = logging.Log
...
+
+type runner struct {
+	tasks         *errgroup.Group
+	state         *core.BuildState
+	parser        *asp.Parser
+	arch          cli.Arch
+	progress      *Progress
+	buildOnce     *cmap.ErrMap[core.BuildLabel, *core.BuildTarget]
+	parseOnce     *cmap.Map[core.BuildLabel, struct{}]
+	preloadOnce   *cmap.ErrMap[string, struct{}]
```
toastwaffle: Please document what the key is here (I'm surprised it's not a BuildLabel)?

src/plz/plz.go
```diff
@@ -25,166 +34,623 @@ var log = logging.Log
...
-	anyRemote := config.NumRemoteExecutors() > 0
+	return g.Wait()
+}
 
-	completeAction := func(remote bool, task core.Task) {
-		if remote {
-			remoteLimiter.Release()
-		} else {
-			localLimiter.Release()
+// RunHostAsync is a convenience function that uses the host architecture, the given state's
```
toastwaffle: Comment does not match function name

src/plz/plz.go
```diff
@@ -25,166 +34,623 @@ var log = logging.Log
...
+			}
+		}
+	}
+	// Nothing defines it, so the only remaining possibility is an architecture subrepo.
+	if arch, ok := couldBeArch(label.Subrepo); ok {
+		r.state.Graph.MaybeAddSubrepo(core.SubrepoForArch(r.state, arch))
+		return nil
+	} else if err != nil {
+		return err
+	}
```
toastwaffle, line 287: Worth a comment clarifying that is to return the `parse.ErrMissingBuildFile` that we ignored above?

src/plz/plz.go
```diff
@@ -25,166 +34,623 @@ var log = logging.Log
...
+	if r.state.KeepGoing {
+		return &errgroup.Group{}, ctx
+	}
+	return errgroup.WithContext(ctx)
+}
+
+// inSamePackage returns true if the two labels are in the same package (and hence, if one of them is
+// currently being parsed, both are).
+func inSamePackage(label, dependent core.BuildLabel) bool {
+	return !dependent.IsOriginalTarget() && label.Subrepo == dependent.Subrepo && label.PackageName == dependent.PackageName
```
toastwaffle, line 306: It's pre-existing code you've moved, but why do we check dependent.IsOriginalTarget() here?

src/plz/plz.go
```diff
@@ -25,166 +34,623 @@ var log = logging.Log
...
+			if !yield(r.parseTarget(ctx, p, dependent.Label)) {
+				break
+			}
+		}
+	}
+}
 
-		if state.NeedTests && task.Target.IsTest() && state.IsOriginalTarget(task.Target) {
-			state.QueueTestTarget(task.Target)
+// buildDep builds a single dependency of a target (which might of course turn into multiple when resolved)
```
toastwaffle: nit: I don't agree with your use of "of course" - it's not obvious to me why the dependency might turn into multiple

src/plz/plz.go
```diff
@@ -25,166 +34,623 @@ var log = logging.Log
...
+		}
+		if err := g.Wait(); err != nil {
+			return err
+		}
+	}
+
+	if !target.ModifiedByCallback {
+		return nil
+	}
+	// It could have modified itself with its own post-build function, so we have to check runtime dpendencies again.
```
toastwaffle: typo: dpendencies -> dependencies

src/plz/plz.go
```diff
@@ -25,166 +34,623 @@ var log = logging.Log
...
+	}
+	return g.Wait()
 }
 
-// RunHost is a convenience function that uses the host architecture, the given state's
-// configuration and no pre targets. It is otherwise identical to Run.
-func RunHost(targets []core.BuildLabel, state *core.BuildState) {
-	Run(targets, nil, state, state.Config, cli.HostArch())
+// buildJustOne calls the build for a single target.
+func (r *runner) buildJustOne(ctx context.Context, target *core.BuildTarget) error {
```
toastwaffle: I don't think the distinction between `buildOne` and `buildJustOne` is clear enough. I think `buildJustOne` could be `runBuildAction`?

src/plz/plz.go
```diff
@@ -25,166 +34,623 @@ var log = logging.Log
...
+// Build is the main entrypoint to build a label
+func (r *runner) Build(ctx context.Context, label, dependent core.BuildLabel) (*core.BuildTarget, error) {
+	if label.IsAllTargets() {
+		return r.buildOnce.GetOrSetCtx(ctx, label, func() (*core.BuildTarget, error) {
+			return nil, r.buildAll(ctx, label, dependent)
+		})
+	}
+	// N.B. We must parse the target _before_ claiming its entry in buildOnce; parsing its package can
+	//      re-enter here for the same label (e.g. a BUILD file that subincludes a target it defines
+	//      earlier in the same file) and we'd then deadlock waiting on ourselves.
```
toastwaffle, line 548: Why does this comment not apply in the buildAll case?

src/plz/plz.go
```diff
@@ -25,166 +34,623 @@ var log = logging.Log
...
+		r.progress.numTotal.Add(int64(r.state.NumTestRuns))
+	}
+	if _, err := r.Build(ctx, target.Label, dependent); err != nil {
+		return err
+	}
+	if !target.IsTest() {
+		return nil
+	}
+	// Now we're ready to test this target.
+	// TODO(peter): Is it okay for none of these to return errors? I _think_ so and we will capture it later?
```
toastwaffle: "I _think_ so" does not fill me with confidence... Also not clear what "these" is

How do we test that this is in fact okay?

src/plz/plz.go
```diff
@@ -25,166 +34,623 @@ var log = logging.Log
...
-	if state.NeedDebugDeps {
-		if len(targets) != 1 {
-			log.Fatalf("expected exactly 1 target in debug mode; got %d", len(targets))
+	pkg, err := r.Parse(ctx, label, dependent)
+	if err != nil {
+		return err
+	}
+	g, ctx := r.group(ctx)
+	for _, target := range pkg.AllTargets() {
+		if r.state.ShouldInclude(target) && (target.IsTest() || r.state.NeedCoverage) && !target.AddedPostBuild {
```
toastwaffle, line 626: I was under the impression that `plz test //foo/bar:all` would build everything in that package even if it's not a test, but this condition seems to imply otherwise. Have I been using it wrong all this time?

src/plz/plz.go
```diff
@@ -25,166 +34,623 @@ var log = logging.Log
...
-func findOriginalTaskSet(state *core.BuildState, targets []core.BuildLabel, addToList bool, arch cli.Arch) {
+// limiter returns either a local or remote limiter that ensures we don't build too many things at once.
+func (r *runner) limiter(remote bool) limiter {
+	if remote {
+		return r.remoteLimiter
+	}
+	return r.localLimiter
+}
+
+func (r *runner) FindOriginalTaskSet(ctx context.Context, targets []core.BuildLabel, needTest, needBuild bool) {
```
toastwaffle: naming nit: I'd expect something called `Find` to return a list of things; should this be `QueueOriginalTasks` or similar? Ditto `findOriginalTask` below

src/plz/plz.go
```diff
@@ -207,35 +673,70 @@ func stripHostRepoName(config *core.Configuration, label core.BuildLabel) core.B
...
+			return err
+		} else if err := r.state.EnsureDownloaded(target); err != nil {
+			return err
 		}
-		for filename := range FindAllBuildFiles(state.Config, dir, "") {
-			dirname, _ := filepath.Split(filename)
-			l := core.NewBuildLabel(strings.TrimLeft(strings.TrimPrefix(strings.TrimRight(dirname, "/"), prefix), "/"), "all")
-			l.Subrepo = target.Subrepo
-			state.AddOriginalTarget(l, addToList)
+		// Targets now get activated during parsing, so can be built before we finish parsing their package.
```
toastwaffle:
```suggestion
		// Targets get activated during parsing, so can be built before we finish parsing their package.
```

src/plz/plz.go
```diff
@@ -207,35 +673,70 @@ func stripHostRepoName(config *core.Configuration, label core.BuildLabel) core.B
...
+	if r.state.IsExcluded(target) {
+		return
+	}
+	r.state.AddOriginalTarget(target)
+	r.tasks.Go(func() error {
+		if needTest {
+			return r.Test(ctx, target, core.OriginalTarget)
+		} else if needBuild {
+			_, err := r.Build(ctx, target, core.OriginalTarget)
+			// TODO(peter): Ensure this gets downloaded if needed
```
toastwaffle: This looks like something that needs to be solved in this PR?

toastwaffle (changes requested):

src/core/build_target.go
peterebden: Possibly yes. My goal here was just to update the previous function to what it needs to be now (i.e. it needs the graph to resolve targets and it has to deal with the possibility that one of them isn't in there, although conceptually that should never happen in the one place it's called).

I think this could probably move to coverage if it really cares about this thing; maybe it's just `Dependencies` with a bit of a filter. I'd like to keep that separate though since it represents a potential functional change. I'll leave a TODO for it.

src/core/build_target.go
peterebden: It seems unnecessary to have TargetOrDie since this function returns an error. It's easier to debug what's going wrong if it bubbles the error up rather than just killing the process.

src/core/build_target_test.go
peterebden: Okay I actually ended up getting rid of this - it's not quite as neat as it sounds because of how callers supply the results of a function call (which I find a bit surprising in Go) and it seems easier to just have explicit assertions and bin this helper.

src/core/state.go
peterebden: So there's a bit of laxity between the thing submitting the result (which is a possibly ongoing build action which is trying to fire-and-forget this) and the thing receiving it and ultimately printing it out. Basically smooths out a bit if the receiver is blocked on I/O or whatever.

This is roughly what was going on before - there was a similar thing between the two channels in here mediated on an internal goroutine.

src/core/utils.go
peterebden: "this" is the next line, i.e. the sorting operation

It's a little bit awkward to create a ticket for because it isn't obvious what should happen in the current state. It's a lot easier to write the ticket describing this PR, but this isn't merged yet.

src/parse/init.go
peterebden: Debatable to what degree it could //really// before, but it //seemed// like it could because it only needed to supply a `BuildState` - there's an implicit expectation that something's running that state but that isn't clear type-wise.

After the change to add `callbacks` more explicitly, it's obvious what you have to give it, but the only place that can do that in a real sense is `//src/plz`.

src/plz/plz.go
peterebden: Because of the way that thing is defined, this would otherwise return true for any target at the root of the main repo

src/plz/plz.go
peterebden: Yea, good idea. I started to run out of creativity for names for these - at first there was just `Build`, `Test` and `Parse` but over time I ended up needing a bunch of others as I needed slightly different entry points to these.

src/plz/plz.go
peterebden: because buildAll calls back into here anyway

although I think ideally maybe it can shortcut that a bit

src/plz/plz.go
peterebden: It wasn't supposed to...

I am now confident that we don't need to here. Test errors have a more nuanced way of reporting their outcomes back to the user (in theory we could build some error hierarchy to cover it but we have not done that)

src/plz/plz.go
peterebden: I don't believe it does? https://github.com/thought-machine/please/blob/8cddc2516b384c512e7573017312b078a54c7b13/src/core/state.go#L1079 is doing the equivalent (bits of the check exist in different places now but it is still filtering out non-test targets).

Practical test:
```
$ ~/.please/17.31.2/please test //src/parse/asp/main:all
0 test targets and 0 tests run; 0 passed.
Total time: 20ms real, 0s compute.
$ ls plz-out/bin/src/parse/asp/main/asp
ls: cannot access 'plz-out/bin/src/parse/asp/main/asp': No such file or directory
```
vs.
```
$ ~/.please/17.31.2/please build //src/parse/asp/main:all
Build finished; total time 440ms, incrementality 99.0%. Outputs:
//src/parse/asp/main:asp:
  plz-out/bin/src/parse/asp/main/asp
$ ls plz-out/bin/src/parse/asp/main/asp
plz-out/bin/src/parse/asp/main/asp
```

src/test/coverage.go
peterebden: An excellent question.

I think this might have been facilitating an investigation of something else that was going wrong.

src/please.go
peterebden: It was all worth it then? ;-)

src/plz/plz.go
peterebden: It turns out that it doesn't need to be done here. It sort of seemed like it did but it's handled in `build_step.go`.

toastwaffle (approved):

toastwaffle (approved):

src/cmap/cerrmap.go
```diff
@@ -57,10 +58,15 @@ func (m *ErrMap[K, V]) Get(key K) (V, error) {
 // GetOrSet returns the value if set, or an error if one has been set.
 // If nothing has been set for the key, it runs the given function to generate the value and then sets it.
 func (m *ErrMap[K, V]) GetOrSet(key K, f func() (V, error)) (V, error) {
-	v, wait, first := m.m.getOrWait(key)
+	v, wait, first := m.m.GetOrWait(key)
 	if v.Err != nil {
 		return v.Val, v.Err
 	} else if first {
+		defer func() {
```
toastwaffle, line 65: Please document the fact that this will recover if `f` panics

src/parse/asp/builtins.go
```diff
@@ -313,15 +313,14 @@ func (s *scope) WaitForSubincludedTarget(ctx context.Context, l, dependent core.
 	return s.interpreter.callbacks.BuildAndDownload(ctx, l, dependent)
 }
 
-// WaitForPackage drops the interpreter lock and waits for the given package to be parsed. Like
+// WaitForSubrepo drops the interpreter lock and waits for the given package to be parsed. Like
```
toastwaffle, line 316: "given package" -> "given subrepo"

"to be parsed" -> "to be defined"?

src/plz/plz.go
```diff
@@ -120,6 +120,7 @@ func RunHost(targets []core.BuildLabel, state *core.BuildState) {
 
 type runner struct {
 	tasks         *errgroup.Group
+	ctx           context.Context //nolint:containedctx
```
toastwaffle, line 123: Please justify why it's reasonable to store the context?

src/plz/plz.go
```diff
@@ -132,21 +133,33 @@ type runner struct {
 	anyRemote     bool
 }
 
+// EnsureSubrepo makes sure a subrepo is available, blocking until it is, or returns an error if it could not be found.
+func (r *runner) EnsureSubrepo(ctx context.Context, subrepo string, defining, dependent core.BuildLabel) error {
+	s, wait := r.state.Graph.SubrepoOrWait(subrepo)
+	if s != nil {
+		return nil
+	}
+	if _, err := r.parse(ctx, defining, dependent, true, wait); err != nil && !errors.Is(err, parse.ErrMissingBuildFile) {
```
toastwaffle, line 142: If we ignore `ErrMissingBuildFile`, doesn't that imply that this can return `nil` and the subrepo still is not available?

PR #3570, opened 2026-08-10
Fix description of `:all` in `plz build` docs

SpangleLabs:
Just stumbled across this, not sure how it's ended up as a new list item there.
I neatened up some of the HTML nearby too

### DIFF
--- a/docs/commands.html
+++ b/docs/commands.html
@@ -416,19 +416,16 @@
       >
     </li>
     <li>
-      <span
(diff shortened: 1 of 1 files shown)


peterebden (approved):
Nice spot. Zealous but incompetent autoformatter maybe?

SpangleLabs:
Yeah, seems likely to be something like that!

PR #3573, opened 2026-08-15
Add file locking to dir cache

peterebden:
This should make it safer to run concurrent processes without them corrupting the shared cache.

I'm going to have a look at sharing it between the e2e tests, which are very slow on weaker machines (like CI) because they heavily rebuild the Go stdlib. That's nontrivial though so I will leave for a separate PR.

### DIFF
--- a/src/core/lock_test.go
+++ b/src/core/lock_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestOpenLockFile(t *testing.T) {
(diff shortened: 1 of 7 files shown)


src/core/lock_test.go
```diff
@@ -162,9 +164,10 @@ func TestAcquireExclusiveFileLockTwice(t *testing.T) {
...
 
 func TestReleaseFileLock(t *testing.T) {
-	file := AcquireExclusiveFileLock("path/to/file")
+	file, err := AcquireExclusiveFileLock("path/to/file")
+	require.NoError(t, err)
 
 	ReleaseFileLock(file)
-	err := file.Close()
+	err = file.Close()
 	assert.Error(t, err, "file already closed")
```
toastwaffle:
```suggestion
	assert.ErrorIs(t, err, os.ErrClosed)
```

toastwaffle (approved):

PR #3574, opened 2026-08-15
Share cache between e2e tests

peterebden:
Stacked on #3573 

Now the cache is threadsafe, we can share it between e2e tests. It's not super obvious but these have slowed down a _lot_ over time - as they moved to `plz_repo_e2e_test` which is better separated but rebuilds more, plugins added more downloads and (most significantly) we now build the Go stdlib in tree rather than assuming it's on the system. This isn't so bad on a powerful machine but it is bad on a weaker one (i.e. all the CI machines).

I've also rationalised some of the plugin / Go versions in there. There is possibly some future work to keep all these better aligned with the parent repo - honestly we usually just forget (hence 10 version old plugins and Go 1.24).

I've added the cache to CircleCI as well; otherwise it will always be cold, and we can't deduplicate the work between everything in that case, so we'd get no benefit without.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,12 +9,17 @@ jobs:
        PLZ_ARGS: "--profile ci --profile alpine --exclude no-musl"
      steps:
        - checkout
+       - run:
+           name: E2E cache key
+           command: ./tools/misc/e2e_cache_key.sh > /tmp/e2e-cache-key
        - restore_cache:
            key: go-mod-alpine-v9-{{ checksum "go.mod" }}
        - restore_cache:
            key: go-alpine-main-v9-{{ checksum "third_party/go/BUILD" }}
        - restore_cache:
            key: python-alpine-main-v3-{{ checksum "third_party/python/BUILD" }}
+       - restore_cache:
+           key: e2e-alpine-v1-{{ checksum "/tmp/e2e-cache-key" }}
        - run:
            name: Bootstrap & Build
            command: ./bootstrap.sh --test_results_file plz-out/results/please/test_results.xml
@@ -39,6 +44,9 @@ jobs:
        - save_cache:
            key: python-alpine-main-v3-{{ checksum "third_party/python/BUILD" }}
            paths: [ ".plz-cache/third_party/python" ]
+       - save_cache:
+           key: e2e-alpine-v1-{{ checksum "/tmp/e2e-cache-key" }}
+           paths: [ "plz-out/e2e_cache" ]
 
    build-linux:
      working_directory: ~/please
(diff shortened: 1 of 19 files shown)


.circleci/config.yml
```diff
@@ -211,12 +235,19 @@ jobs:
...
+           name: E2E cache key
+           command: ./tools/misc/e2e_cache_key.sh > /tmp/e2e-cache-key
        - restore_cache:
            key: go-mod-darwin-arm64-v8-{{ checksum "go.mod" }}
        - restore_cache:
            key: go-darwin-arm64-go123-v1-{{ checksum "third_party/go/BUILD" }}
        - restore_cache:
            key: python-darwin-arm64-v3-{{ checksum "third_party/python/BUILD" }}
+       # Keyed separately from the Linux jobs: the cache holds darwin_arm64 artifacts, so sharing a
+       # key with them would just mean a cache full of things this job can't use.
```
toastwaffle: bit odd that this comment is here specifically, given every single task uses different keys for every cache

test/build_defs/test.build_defs
```diff
@@ -1,3 +1,16 @@
+# Every e2e test runs in its own copy of a repo with its own $HOME, so by default they each get an
+# empty dir cache and repeat a lot of expensive work (fetching plugins, building the Go stdlib etc).
+# The dir cache is content-addressed and safe to use concurrently, so we point them all at a single
+# cache instead. $TMP_DIR is <repo root>/plz-out/tmp/..., so trimming from the last /plz-out/ gets us
+# back to the root of the repo we're being run from.
+#
+# Cleaning is off: the suite needs about 1.5G and 50k files of this, and the cleaner walks the whole
```
toastwaffle: any concerns about an ever-growing cache consuming CircleCI credits?

tools/misc/e2e_cache_key.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+# This globs rather than listing files so that a newly added test repo is picked up automatically.
+
+set -euo pipefail
+
+cd "$(dirname "$0")/../.."
+
+# cksum prints the checksum, the size and the name, so a file being renamed changes the output too.
+# Sorted so the ordering doesn't depend on how the filesystem hands them back.
+find test \( -path '*/plugins/BUILD_FILE' -o -path '*/third_party/go/BUILD_FILE' \) -type f \
```
toastwaffle, line 19: do we need to enforce the convention of using BUILD_FILE in e2e tests?

toastwaffle (approved):

test/build_defs/test.build_defs
peterebden: Not really, I'm pretty sure CircleCI reclaims caches as needed (in the same way as our go cache can continually generate new artifacts as we modify `go.mod` etc)

tools/misc/e2e_cache_key.sh
peterebden: I think we mostly do follow it, I think that is good enough.

The thing I might like to do more on is making sure they all have the same versions of things, but I'm leaving that for another day.

PR #3576, opened 2026-08-16
Fix lost-wakeup race on packageWaits in WaitForPackage

jackmarsh:
I've been building [rust-rules](https://github.com/becomeliminal/rust-rules), a Rust plugin for Please. While testing it on a repo with ~40 packages, each subincluding the plugin's build_defs, cold builds would intermittently hang at `N tasks left, 0 workers busy, parsing M BUILD files`.

The goroutine dump showed a single parser stuck in `subincludeTarget → WaitForPackage → waitOnChan`. The cause is a race in `WaitForPackage`. Two parsers racing for the same unparsed package can both see `packageWaits.Get(key) == nil` and both call `Set`. The second `Set` overwrites the first thread's channel, and `LogParseResult` closes only the channel currently in the map, so the first thread waits forever.

The fix makes the insertion atomic with `AddOrGet`, matching `pendingPackages` and `pendingTargets` in the same file. One caller inserts the canonical channel and queues the parse; everyone else waits on that channel and is woken by the existing close.

Added a regression test that races 32 callers against the package parse completing. It fails on master within a few iterations and passes with the fix, including under `-race`.

Reproduced on v17.27.0 and v17.31.2. Stock plz wedged 11 of 30 cold cycles on my machine; patched ran 30 clean. `plz test //src/core/...` passes.

### DIFF
--- a/src/core/state.go
+++ b/src/core/state.go
@@ -884,14 +884,16 @@ func (state *BuildState) WaitForPackage(l, dependent BuildLabel, mode ParseMode)
 	}
 
 	// If something has already queued the package to be parsed, wait for them
-	if ch := state.progress.packageWaits.Get(key); ch != nil {
+	// (atomically: a racing Get-then-Set here can orphan the first caller's channel)
+	if ch, inserted := state.progress.packageWaits.AddOrGet(key, func() chan struct{} {
+		return make(chan struct{})
+	}); !inserted {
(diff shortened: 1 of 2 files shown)


peterebden (approved):
Ah yes, that sounds right; I think I found a similar thing in #3569 . Ended up changing `Get` completely, but this seems fine here.

PR #3578, opened 2026-08-17
Consider .rs files for coverage by default

jackmarsh:
Please parses Rust coverage output correctly (the Rust plugin emits go-cover format), but `Cover.FileExtension` has no `.rs`, so `collectCoverageFiles` filters every Rust file out. The result is a coverage run that reports:

```
Total coverage:  No data
```

with every file at 0%, which reads as "coverage is broken" rather than "add a line of config". Every other language Please covers has its extension in this list already.

Adding it means Rust repos get working coverage without each of them discovering the same workaround. Three separate repos hit this before the cause was identified.

`plz test //src/core:core_test` is green (279 tests).


peterebden:
Makes sense. You can of course set it in config, but I think maybe you then have to list everything out again, and there's no real cost to having this as a default.

Looks like you have #3577 on this branch as well?

peterebden (approved):

PR #3579, opened 2026-08-19
Update to go 1.27.0

peterebden:
I think this works. Some unrelated test issues locally, will see what CI says.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
    build-alpine:
      working_directory: ~/please
      docker:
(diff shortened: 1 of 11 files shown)


peterebden:
OK - CI says many unhappy things, I guess :joy: Apparently we need the Alpine image (not pushed yet), golangci-lint and I guess I pushed one architecture of the linux image but not that one. Solvable in a day or two.

peterebden:
Right this should be good to go now

chrisnovakovic (approved):

PR #3583, opened 2026-08-26
Plugin updates to fix linux_arm64 build

peterebden:
Notably this moves the Python plugin to v2. I don't think it's very different as far as this repo is concerned but one of the options goes away.

Also a minor test fix that goes a little wonky if it gets cached.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -15,8 +15,8 @@ check_path_for_excludes() {
   EXCLUDES=""
 
   if ! hash python3 2>/dev/null ; then
-      warn "python3 not found, excluding python3 tests"
-      EXCLUDES="${EXCLUDES} --exclude=py3 --exclude python3"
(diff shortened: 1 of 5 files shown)


chrisnovakovic (approved):

test.sh
```diff
@@ -15,8 +15,8 @@ check_path_for_excludes() {
   EXCLUDES=""
 
   if ! hash python3 2>/dev/null ; then
-      warn "python3 not found, excluding python3 tests"
-      EXCLUDES="${EXCLUDES} --exclude=py3 --exclude python3"
+      warn "python3 not found, excluding python tests"
+      EXCLUDES="${EXCLUDES} --exclude=py"
```
peterebden, line 19: The plugin seems to only label things with `py` now, so that's what we should use

PR #3586, opened 2026-08-29
Update some Python libraries

peterebden:
Fixing a warning about a missing licence and bringing some things more up to date. Tried a script or two locally and they still work.

I do wonder a little if we should drop all Python here and just do it all in Go, but it is sort of nice to illustrate that this is a multi-language build system.

### DIFF
--- a/third_party/python/BUILD
+++ b/third_party/python/BUILD
@@ -14,8 +14,8 @@ package(
 python_wheel(
     name = "six",
     outs = ["six.py"],
-    hashes = ["8abb2f1d86890a2dfb989f9a77cfcfd3e47c2a354b01111771326f8aa26e0254"],
-    version = "1.16.0",
+    hashes = ["4721f391ed90541fddacab5acf947aa0d3dc7d27b2e1e8eda2be8970586c3274"],
+    version = "1.17.0",
 )
 
 python_wheel(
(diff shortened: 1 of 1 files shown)


SpangleLabs:
Should we be marking the licences for the rest of these?

SpangleLabs (approved):

peterebden:
> Should we be marking the licences for the rest of these?

We don't generally in this repo. I only added it for that one because `pip_library` is emitting a warning about it.

peterebden:
> > Should we be marking the licences for the rest of these?
> 
> We don't generally in this repo. I only added it for that one because `pip_library` is emitting a warning about it.

Hmmmm I've changed it to `python_wheel` (which seems better) but it still emits the warning. So I guess it's not `pip_library` but some property of the wheel metadata.

toastwaffle (approved):

peterebden:
> > > Should we be marking the licences for the rest of these?
> > 
> > 
> > We don't generally in this repo. I only added it for that one because `pip_library` is emitting a warning about it.
> 
> Hmmmm I've changed it to `python_wheel` (which seems better) but it still emits the warning. So I guess it's not `pip_library` but some property of the wheel metadata.

For posterity, this is a newer format that we'll start supporting with https://github.com/please-build/python-rules/pull/322

PR #3587, opened 2026-08-29
Fixes for puku init step & codelab

### LINKED ISSUE #3534: code lab - third party dependencies with Puku - fails on step 2
After following the [guidance](https://please.build/codelabs/puku/index.html#2) in step 1 and 2, **puku** was **failing to sync**. I'm so new to `Please` that I got `Gemini CLI` to help me remedy the issue.  Hopefully this can help improve the codelab guidance.

> I have fixed the issue where plz puku sync -w was not updating third_party/go/BUILD. 

> The primary cause was that puku (and the Go plugin) was not explicitly told where to find
  the go.mod file, which is required for synchronization. I also improved the puku alias
  configuration to ensure it runs from the repository root.

  > Changes Made:
  > 1. Exported go.mod: Created a root BUILD file to export go.mod as a Please target
      (//:gomod).
  > 2. Configured ModFile: Updated .plzconfig to set ModFile = //:gomod in the [Plugin "go"]
      section. This allows puku to correctly locate and sync dependencies.
  > 3. Improved Puku Alias: Added --wd=. to the puku alias in .plzconfig to ensure it
      executes relative to the workspace root.
  > 4. Synchronized Dependencies: Ran plz puku sync -w to populate third_party/go/BUILD with
      the go_repo() entry for github.com/google/uuid.
  > 5. Generated Source BUILD Files: Ran plz puku fmt src/hello to generate the
      src/hello/BUILD file, which now correctly references the third-party dependency.

peterebden:
Main thing is getting `plz init plugin go` to just write the filegroup and set `modfile`, which should avoid the main Puku papercut where it just silently does nothing (that behaviour also needs a fix in Puku really, it exits with 0 in too many unsuccessful cases).

Also some minor changes to the codelab as I ran through it:
 - Remove the version pin. It gets out of date and the codelab should be happy to just use the latest.
 - Don't label the puku alias as optional, because the rest of the codelab references `plz puku` so it won't work without. (We might also want `plz init plugin go` to just do this?)
 - Remove the `go install std` directive. We don't need it any more (we build it via plz on-demand)
 - Remove the initial "sync the changes" bit; it doesn't sync at that point for me. I think you need to run `plz puku fmt` first.

Fixes #3534


docs/codelabs/puku.md
```diff
@@ -210,14 +190,14 @@ plz run //src/hello
 To update a module to a specific version:
 
 ```bash
-GOTOOLCHAIN=local go get github.com/google/uuid@v1.6.0
```
toastwaffle, line 213: Presumably this was added for a reason (@scyyx5 ?); why remove it?

src/plzinit/plugin_go.go
```diff
@@ -88,14 +94,65 @@ func initGo() (map[string]string, error) {
...
+	path := filepath.Join(dir, rootBuildFileName)
+	buildFile, err := parseBuildFile(path)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse %s: %w", path, err)
+	}
+
+	// Reuse an existing filegroup if the repo already exports go.mod under another name.
+	for _, rule := range buildFile.Rules("filegroup") {
+		for _, src := range rule.AttrStrings("srcs") {
+			if src == goModFileName {
```
toastwaffle: Do we want it to be the only src in the filegroup?

toastwaffle (approved):

docs/codelabs/puku.md
peterebden: I think this prevents it from downloading a Go toolchain if the versions differ? Which they might do and we're probably better off for them to just do that. 
But it isn't applied to all the `go` invocations here either, e.g. line 172 / 160 invokes go without it.

src/plzinit/plugin_go.go
peterebden: Yeah that makes sense, have changed.

docs/codelabs/puku.md
scyyx5: Yes, this was added to prevent `go.mod` file specifies a newer version of Go than the one installed locally, which is not the case in most of the normal development workflows.

scyyx5 (approved):

PR #3591, opened 2026-09-03
make cycleDetector.stopped atomic

gzvolsky-tm:
Fixes the flaky build-darwin job and adds a regression test for the race with `go state.checkForCycles()` in `src/core/state.go`. The main goroutine can call Stop while the async check is in flight.

### DIFF
--- a/src/core/cycle_detector.go
+++ b/src/core/cycle_detector.go
@@ -3,17 +3,18 @@ package core
 import (
 	"fmt"
 	"strings"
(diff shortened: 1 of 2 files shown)


peterebden (changes requested):
I appreciate this, and it looks good, but I've addressed this in #3569 and would really like to get that reviewed and merged before making further changes to the existing setup

peterebden:
This is fixed on master now

PR #3596, opened 2026-09-08
Check target.Debug is not nil before we access it

peterebden:
Got a report of this line panicking; surely nothing has validated that the `.Debug` attribute is set on the target before it tries to dereference it.


toastwaffle (approved):

peterebden:
Right so the main way of getting there is if you do `plz debug -c opt //some:test` or something to that effect. I suppose the build def in question does not apply the debug info if it's not in a debug config.

PR #3603, opened 2026-09-15
Wrap an error & some linting

peterebden:
Was hoping I could get golangci to check for this sort of thing, but turns out it isn't that smart. Oh well.

Also reformatted BUILD files because I got a message about it.


src/parse/asp/interpreter.go
```diff
@@ -271,7 +271,7 @@ func (i *interpreter) Subinclude(pkgScope *scope, path string, label core.BuildL
 		}
 		return locals, nil
 	})
-	pkgScope.Assert(err == nil, "failed to subinclude %s: %s", label, err)
+	pkgScope.Assert(err == nil, "failed to subinclude %s: %w", label, err)
```
peterebden, line 274: This is the main thing I actually wanted. It is a bit more important for output than it used to be that we can classify errors correctly

toastwaffle (approved):

PR #3605, opened 2026-09-15
Change Graph.PackageMap to an iterator

### LINKED ISSUE #3599: Change PackageMap to return an iterator
Currently it says it returns a copy of the graph's internal map, because we didn't have a good way of exposing this earlier. It would be cleaner and faster if it exposed an iterator rather than reifying a complete copy of a map.

peterebden:
This was a bit old-school (understandably, it significantly predates iterators). No callers really want a fully reified map, so this is more efficient.

Also ended up pushing this down into cmap, and hence to a function used by the LSP which can also do it bit by bit rather than one monolithic locked call.

Resolves #3599


src/cmap/cerrmap_test.go
```diff
@@ -32,3 +33,22 @@ func TestErrWait(t *testing.T) {
...
+	m.Set(1, 2)
+	m.Set(3, 4)
+	m.SetError(5, err)
+
+	next, stop := iter.Pull2(m.Values())
+	defer stop()
+	for range 3 {
+		v, err, _ := next()
+		if v == 0 {
+			assert.Error(t, err)
```
toastwaffle: optional: you could use `assert.ErrorIs` here

src/cmap/cmap.go
```diff
@@ -85,11 +86,24 @@ func (m *Map[K, V]) Values() []V {
...
+// Items returns an iterator over each key-value pair in the map.
+// They are returned in no particular order.
+// You should not mutate the map while calling this as it may deadlock.
+func (m *Map[K, V]) Items() iter.Seq2[K, V] {
+	return func(yield func(K, V) bool) {
+		for i := range len(m.shards) {
+			shard := &m.shards[i]
+			shard.l.RLock()
+			for k, v := range shard.m {
+				if v.Wait == nil { // Only yield completed values
```
toastwaffle: Is it worth lifting this comment up to the function doc comment? Or is the fact that there's an entry with a Wait in the underlying map implementation detail?

src/query/graph.go
```diff
@@ -95,14 +96,12 @@ func (graph *JSONGraph) Subrepo(name string) *JSONGraph {
 func makeAllPackages(state *core.BuildState) <-chan JSONPackage {
 	ch := make(chan JSONPackage, 100)
 	go func() {
-		packages := state.Graph.PackageMap()
+		packages := slices.Collect(state.Graph.AllPackages())
```
toastwaffle: Why are you collecting into a slice here instead of iterating over the iterator?

toastwaffle (approved):

src/query/graph.go
peterebden: Ah nice spot. Previously this needed the whole thing to do `wg.Add(len(packages))` etc. Ended up being unnecessary once I updated it to use `wg.Go` but I didn't update this bit too.

PR #3609, opened 2026-09-17
Cut a beta release with #3569 in it

peterebden:
Arguably this could be a patch version since there isn't supposed to be anything user-visible, but I feel a bit weird cutting betas of a patch. Can be persuaded either way though.


chrisnovakovic (approved):

chrisnovakovic:
I think the release action might fail because it won't find a section heading with the corresponding version number in the ChangeLog - I might be wrong though, I haven't looked at that code in a while.

ChangeLog
```diff
@@ -1,3 +1,7 @@
+Version 17.34.0
+---------------
+    * Build parallelism changes (#3569)
```
toastwaffle, line 3: This should mention #3607 too

toastwaffle (changes requested):

ChangeLog
peterebden: Done

peterebden:
> I think the release action might fail because it won't find a section heading with the corresponding version number in the ChangeLog - I might be wrong though, I haven't looked at that code in a while.

We don't have anything in the changelog with a beta heading, and we've definitely done them before, so I think this must work somehow.
I guess if it doesn't I can figure something out. Ideally I think it should check the non-beta version, but it's possible it just doesn't generate release notes for betas.

toastwaffle (approved):

PR #3612, opened 2026-09-19
Add config property to suppress version check

### LINKED ISSUE #3600: Remote Execution API version requirements in code vs docs
https://github.com/thought-machine/please/blob/master/src/remote/remote.go#L243 checks that the RE server meets the version requirements but it seems to not match the documentation at https://please.build/remote_builds.html#gsc.tab=0

The docs say `2.1` is required but the check ensures `min <= 2.0.0 <= max` which feels mismatched but perhaps there is good historical reason.

When using plz against a recent version of buildbarn it seems the world is moving on as well.

```
Unsupported API version; we require 2.0.0 but server only supports 2.3.0 - 2.12.0
```

I am left with building a patched version of please to adjust the check with no way to disable it.

I would like a way to disable the version check (opt-in in .plzconfig would be great) so it is easier to test different remote executors.

It looks like maybe it is just buildbarn and buildfarm that have a min version of 2.3 https://github.com/bazelbuild/remote-apis#servers

peterebden:
Also bump the min required version to 2.1 (which we do very much require for non-specific output paths).

Fixes #3600


toastwaffle (approved):

src/core/config.go
```diff
@@ -575,6 +575,7 @@ type Configuration struct {
 		VerifyOutputs           bool         `help:"Whether to verify all outputs are present after a cached remote execution action. Depending on your server implementation, you may require this to ensure files are really present."`
 		UploadDirs              bool         `help:"Uploads individual directory blobs after build actions. This might not be necessary with some servers, but if you aren't sure, you should leave it on."`
 		OptionalOutputsRequired bool         `help:"Requires that any optional outputs of build actions (optional test outputs, coverage when not opted out of) are produced. By default this is a non-fatal failure, but the actions may not cache remotely."`
+		SuppressVersionCheck    bool         `help:"Suppresses the error if the server reports a version outside what we support"`
```
chrisnovakovic, line 578:
```suggestion
		SuppressVersionCheck    bool         `help:"Warns, rather than errors, if the server reports a version outside what we support"`
```

PR #3615, opened 2026-09-20
Extend cycle detection to include subincludes

peterebden:
Have wanted something like this for ages to debug the dreaded subinclude cycles. It makes all subincludes first-class nodes in the build graph (before they were somewhat oddly split between bits on the graph and on the package) so they can be walked and reported by the cycle detector.

Also fixes several minor bugs in `plz query revdeps`, `plz gc` and `plz watch` where they didn't account for nested subincludes.

Technically there is some more work done here to add the extra subinclude nodes into the graph, but the performance difference seems negligible to me.


src/cmap/cmap.go
```diff
@@ -69,6 +69,13 @@ func (m *Map[K, V]) Contains(key K) bool {
 	return m.shards[m.hasher(key)&m.mask].Contains(key)
 }
 
+// Update performs an update on the given item in-place.
+// The given function is given the old value (which will be the zero value if it isn't present) and should return the new value.
+// The function should not attempt to modify the map further.
```
toastwaffle, line 74: Is it worth stating explicitly that a map shard will be locked while the function runs?

src/core/graph.go
```diff
@@ -23,10 +21,8 @@ type BuildGraph struct {
 	packages *cmap.ErrMap[packageKey, *Package]
 	// Registered subrepos, as a map of their name to their root.
 	subrepos *cmap.Map[string, *Subrepo]
-	// Subincludes that are subincluded by other subincludes
-	subincludeSubincludes map[BuildLabel]labelSet
-	// Use a mutex as a LabelSet isn't atomic. We need to guard against inserting as well as mutating the value.
-	subincMux sync.Mutex
+	// Nested subincludes. Direct subincludes are on the package.
```
toastwaffle: This comment isn't abundantly clear to me. IIUC, keys in this map are anything that is subincluded somewhere, and the values includes everything that is subincluded as a result of subincluding that key?

Is this transitive? If `:a` subincludes `:b` subincludes `:c`, is `subincludes[":a"]` equal to `{":b"}` or `{":b", ":c"}`?

src/core/graph.go
```diff
@@ -187,37 +183,51 @@ func (graph *BuildGraph) DependentTargets(from, to BuildLabel) []BuildLabel {
...
-func (graph *BuildGraph) RegisterTransitiveSubinclude(from, to BuildLabel) {
-	graph.subincMux.Lock()
-	defer graph.subincMux.Unlock()
-
-	incs, ok := graph.subincludeSubincludes[from]
-	if !ok {
-		incs = labelSet{}
-		graph.subincludeSubincludes[from] = incs
+// SubincludeNodes returns every label that has had subincludes recorded against it.
+// These are packages as well as subinclude targets, and include packages that are still in the process of parsing.
```
toastwaffle: Is "subinclude targets" a well defined term? I presume by this you mean targets which output files which can be parsed by please, for which I'd lean on "build definitions" as a slightly common but probably equally poorly defined term?

src/core/graph.go
```diff
@@ -187,37 +183,51 @@ func (graph *BuildGraph) DependentTargets(from, to BuildLabel) []BuildLabel {
...
+func (graph *BuildGraph) SubincludeNodes() []BuildLabel {
+	ret := []BuildLabel{}
```
toastwaffle: Make this an iterator? The only place you call it you're immediately iterating over it

Makes me sad that there's nothing in stdlib to go from `iter.Iter2` to `iter.Iter`

src/core/graph_test.go
```diff
@@ -50,6 +51,74 @@ func TestSubrepo(t *testing.T) {
...
+	outer := ParseBuildLabel("//build_defs:outer", "")
+	inner := ParseBuildLabel("//build_defs:inner", "")
+	graph.AddSubinclude(pkg, outer)
+	graph.AddSubinclude(outer, inner)
+
+	// Subincludes is only the direct ones...
+	assert.Equal(t, []BuildLabel{outer}, slices.Collect(graph.Subincludes(pkg)))
+	// ...whereas AllSubincludes follows them transitively.
+	assert.ElementsMatch(t, []BuildLabel{outer, inner}, slices.Collect(graph.AllSubincludes(pkg)))
+	assert.Empty(t, slices.Collect(graph.AllSubincludes(inner)))
```
toastwaffle, line 69: It's a little odd to have this assertion, and not the full suite of all 6 possible combinations for 2 functions and 3 labels.

src/core/graph_test.go
```diff
@@ -50,6 +51,74 @@ func TestSubrepo(t *testing.T) {
...
+	b := ParseBuildLabel("//build_defs:b", "")
+	graph.AddSubinclude(a, b)
+	graph.AddSubinclude(b, a)
+
+	// A cycle here would deadlock the parser, but the graph can be walked while it's still being
+	// built (e.g. by cycle detection) so this mustn't recurse forever.
+	assert.Equal(t, []BuildLabel{b}, slices.Collect(graph.AllSubincludes(a)))
+}
+
+func TestAllSubincludesEarlyExit(t *testing.T) {
```
toastwaffle: I don't quite understand the purpose of this test? It seems to be asserting that a `break` statement works; if you were testing that the iterator were implemented correctly I'd expect at least one more assertion about something else still being healthy despite the early exit.

src/export/trimmed_exporter.go
```diff
@@ -118,7 +118,7 @@ func (e *trimmedExporter) exportSubincludes(pkg *core.Package, target core.Build
 
 	allSubincludes := usedSubincludes
 	for _, sub := range usedSubincludes {
-		for _, trans := range e.state.Graph.TransitiveSubincludes(sub) {
+		for trans := range e.state.Graph.AllSubincludes(sub) {
```
toastwaffle: nit: `trans` becomes a slightly opaque variable name now that the function being called isn't `Transitive`

toastwaffle (approved):

src/cmap/cmap.go
peterebden: Conceptually I'd prefer not to, 'shards' are supposed to be a bit of an implementation detail (although the 'shard count' thing does get exposed slightly). The third line (specifying that it should not modify the map) seems enough to tell the (currently) one consumer not to mess with it.

src/core/graph.go
peterebden: Thanks, this is a bit out-of-date (at first they were getting maintained on the package still).

It is not transitive. As a graph it's a lot easier just to store edges and let other things (like `AllSubincludes` do a transitive walk when needed.

src/core/graph.go
peterebden: Yeah I do think they have space to put some simple generic things in the `iter` package.

I can turn this into an iterator, but I didn't want to hold it  the whole time for recursive cycle detection because it will lock the underlying map for that period. I guess the interface here makes more sense (and we do already do things like that in the other functions)

PR #3616, opened 2026-09-21
Move downloading of data dependencies

peterebden:
This happens too early now in `remote.go`. Move it out of `Build()` so that function just builds and doesn't download (handling a TODO in there) and trigger the downloads externally. This actually simplifies quite a bit since we have the same "walk all inputs" in `DownloadInputsIfNeeded` and we don't need to repeat that.

Added a test case for `rex_test` since that wasn't covering this super well. It's a bit fiddly though since it's quite timing-oriented.


peterebden:
Wrong branch.

